PR #10, opened 2023-04-06
 [fea] Add statistics utilities and tests

dallasfoster:
This MR adds additional ensemble statistics useful for general and climate settings:

Calibration related statistics like the Rank Histogram and Ranked Probability Score (calibration.py)
Entropy related statistics: Entropy, Relative Entropy (KL Divergence), Normalized Entropy ('entropy.py'). The normalized entropy will become a measure for 'how confident' a model is about it's predictions (it is normalized between 0 and 1).
Extreme Forecast Index - measures how extreme a prediction is compared to a climatology.


NickGeneva (approved):
LGTM

PR #13, opened 2023-04-14
Enh Dockerfile updates

ktangsali:


akshaysubr (approved):
LGTM! Only comment I have is to move to using stable version of packages in another MR once those are ready.

ktangsali:
> LGTM! Only comment I have is to move to using stable version of packages in another MR once those are ready.

Absolutely! Will keep a track of when changes get released in stable version and update this accordingly. Thanks for the review.

PR #35, opened 2023-05-11
[IMP] code improvements for gnn layers 

stadlmax:
- improve GNN utilities
- make layers which are shared between GraphCast and MeshGraphNet generic enough to cover both (similar) models with the same set of layers 
- refactor the "MLP" layers in GraphCast to better handle the "Concat" vs. "Sum" cases of creating new edge features
- simplify Encoder/Decoder/Processor by making use of unified logic of Concat/Sum cases


mnabian (approved):
LGTM

PR #43, opened 2023-06-01
Speeding up PyTests

NickGeneva:
- Reducing iterations of K-H dataloader
- Reducing latent variable size in RNNs
- Reduction of field size in metric test


ktangsali (approved):
LGTM!

PR #65, opened 2023-07-10
Checkpoint Refactor

loliverhennigh:
# Description

This feature adds better saving and loading of models that store arguments in a json file. This allows loading the model without needing to keep track of model parameters. Along with this merge there are several related side features. from_torch allows users to bring in models from torch seamlessly assuming they have json serializable inputs. Also, we have exposed the models through endpoints following discussions about expanding usability of modulus.
Here is a list of high level features,

* Saving models now contains other information such as version and params so you can instantiate the model directly from the checkpoint from_checkpoint
* Bring in model from torch and use it seamlessly as a Modulus module, from_torch
* Register custom model dynamically. This allows loading models from checkpoints for example.
* Factory that takes a name of a Module and returns the class for it.


loliverhennigh:
This should be ready for review now. I had to mess around with my fork to change it to the public PR so hopefully nothing weird happened.

modulus/models/module.py
```diff
@@ -65,55 +105,256 @@ def debug(self):
...
+            import git
+
+            repo = git.Repo(search_parent_directories=True)
+            try:
+                version_info["git_hash"] = repo.head.object.hexsha
+            except git.InvalidGitRepositoryError:
+                version_info["git_hash"] = None
+
+        with open(directory.joinpath("version.json"), "w") as f:
+            json.dump(version_info, f)
```
NickGeneva, line 160: Does this file ever get zipped up somehow? I would like mdlus's to be a single file I can easily move around with out losing bits

modulus/models/module.py
```diff
@@ -65,55 +105,256 @@ def debug(self):
...
+
```
NickGeneva, line 161: Would we ever want a util that converts from mdlus to just a pytorch checkpoint?

NickGeneva (commented):
Initial comments

modulus/models/layers/activations.py
```diff
@@ -95,3 +95,53 @@ def __init__(self):
 
     def forward(self, x: Tensor) -> Tensor:
         return 0.5 * (x + torch.sqrt(x * x + self.b))
+
+
+# Dictionary of activation functions
+ACT2FN = {
+    "relu": nn.ReLU(),
+    "leaky_relu": nn.LeakyReLU(0.1),
```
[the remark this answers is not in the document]
loliverhennigh: Hmm, this is how they do it in HF https://github.com/huggingface/transformers/blob/main/src/transformers/activations.py#L213. Maybe we can have the leaky value as a tuple or something...

modulus/models/module.py
```diff
@@ -48,8 +73,23 @@ def __init__(self, meta: ModelMetaData = None):
...
 
-        # dummy buffer for getting where the networks device
-        self.register_buffer("device_buffer", torch.empty(0))
+    @classmethod
+    def instantiate(cls, arg_dict):
+        _mod = importlib.import_module(arg_dict["__module__"])
+        _cls_name = arg_dict["__name__"]
+
+        # Add a check if the class is one of the dynamically created ones
+        if _cls_name in cls._dynamically_created_classes:
```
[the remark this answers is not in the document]
loliverhennigh: `cls._dynamically_created_classes` starts out empty but when you add a model with either https://github.com/loliverhennigh/modulus/blob/fea-checkpoint_refactor/modulus/models/module.py#L209 or  https://github.com/loliverhennigh/modulus/blob/fea-checkpoint_refactor/modulus/models/module.py#L209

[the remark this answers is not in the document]
loliverhennigh: I am not sure if this is the best way to handle dynamically added classes. If you have an suggestions I am open on this.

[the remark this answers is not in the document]
NickGeneva: Oh, I see. So this is when you register models as a plug in. Make sense

modulus/models/module.py
```diff
@@ -65,55 +105,256 @@ def debug(self):
...
+        KeyError
+            If no model is registered under the provided name.
+        """
+
+        if name in Module._dynamically_created_classes:
+            return Module._dynamically_created_classes[name]
+
+        group = "modulus.models"
+        try:
+            entry_point = entrypoints.get_group_named(group)[name]
```
NickGeneva, line 348: Sorry Oliver, a little confused here. This is entry point in the setup.py? how would this get updated with discovered entry points from other python packages? (Say I pip install my custom NN model with a Modulus entry point)

For example FSspec has this function that will auto register entry points from existing python packages that have been installed. 

https://github.com/fsspec/filesystem_spec/blob/master/fsspec/__init__.py#L40

Did you implement something like this?

modulus/models/module.py
```diff
@@ -48,8 +73,23 @@ def __init__(self, meta: ModelMetaData = None):
...
+    @classmethod
+    def instantiate(cls, arg_dict):
+        _mod = importlib.import_module(arg_dict["__module__"])
+        _cls_name = arg_dict["__name__"]
+
+        # Add a check if the class is one of the dynamically created ones
+        if _cls_name in cls._dynamically_created_classes:
+            _cls = cls._dynamically_created_classes[_cls_name]
+        else:
+            _cls = getattr(_mod, _cls_name)
```
NickGeneva, line 85: Maybe you could you use your factory method here? I suppose _cls_name is the direct class name, not the entry point key, but consistency in how we initialize things would be nice if possible.

modulus/models/layers/activations.py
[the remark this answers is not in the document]
NickGeneva: Hmmm, yeah I was just thinking if someone wanted to try tuning things...

Not that big of a deal, since we ship complete models... users can always directly edit them for testing them. I think this is fine.

setup.py
```diff
@@ -14,4 +14,19 @@
 
 from setuptools import setup
 
-setup()
+setup(
```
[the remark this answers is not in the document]
loliverhennigh: Yes! I would really like this. I looked around a bit for something like this but let me take a look again...

modulus/models/module.py
```diff
@@ -65,55 +105,256 @@ def debug(self):
...
+        # Define an internal class as before
+        class ModulusModel(Module):
+            def __init__(self, *args, **kwargs):
+                super().__init__(meta=meta)
+                self.inner_model = torch_model_class(*args, **kwargs)
+
+            def forward(self, x):
+                return self.inner_model(x)
+
+        # Get the argument names and default values of the PyTorch model's init method
```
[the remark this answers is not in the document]
loliverhennigh: Haha, yes...

setup.py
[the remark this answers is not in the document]
NickGeneva: Now I'm also confused here, Modulus doesnt need entry points to it self right? Rather other packages should entry point into modulus... and we check to see what exists no?


So in this example here:
https://setuptools.pypa.io/en/latest/userguide/entry_point.html#entry-points-for-plugins

The timmins package does not have an entry point, it rather reads entry points fed to it via `display_eps = entry_points(group='timmins.display')`. Then the add on package `timmins-plugin-fancy` has an entry point extending the base package.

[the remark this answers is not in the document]
NickGeneva: I suppose this is kinda doubling as a known classes dictionary they way you are using it.

In fsspec this just placed inside the file system registry file:
https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/registry.html#register_implementation

loliverhennigh:
Hey @akshaysubr , @NickGeneva and I talked and I will refactor some of the registry and entrypoints stuff to be a bit more in line with fsspec. You can look at other parts of the code but I wouldn't bother seeing too much until I make these changes.

loliverhennigh:
Linking with Modulus launch PR that should be merged at the same time, https://github.com/NVIDIA/modulus-launch/pull/42

modulus/models/module.py
```diff
@@ -65,55 +105,256 @@ def debug(self):
         # TODO: set up debug log
         # fh = logging.FileHandler(f'modulus-core-{self.meta.name}.log')
 
-    def save(self, file_name: Union[str, None] = None) -> None:
+    def save(
```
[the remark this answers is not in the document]
loliverhennigh: I am open for suggestions on this, maybe `save_checkpoint`. For just saving the pytorch checkpoint now you can still do, `torch.save(model.state_dict(), PATH)` unless I broke something haha,

modulus/models/module.py
```diff
@@ -65,55 +105,256 @@ def debug(self):
...
+
+        torch.save(self.state_dict(), directory.joinpath("model.pt"))
+
+        with open(directory.joinpath("args.json"), "w") as f:
+            json.dump(self._args, f)
+            if file_name is None:
+                file_name = self.meta.name + ".pt"
+
+        # Save the modulus version and git hash (if available)
+        version_info = {"modulus_version": modulus.__version__}
```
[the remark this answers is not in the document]
loliverhennigh: Added this in now. I have a separate version for checkpoint format. Currently if it doesn't match the version it throws an error but we can expand this with if statements to handle older versions later.

modulus/models/module.py
```diff
@@ -65,55 +105,256 @@ def debug(self):
...
+        if save_git_hash:
+            import git
+
+            repo = git.Repo(search_parent_directories=True)
+            try:
+                version_info["git_hash"] = repo.head.object.hexsha
+            except git.InvalidGitRepositoryError:
+                version_info["git_hash"] = None
+
+        with open(directory.joinpath("version.json"), "w") as f:
```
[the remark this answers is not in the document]
loliverhennigh: Changed

modulus/models/module.py
loliverhennigh: I am open to this but I can't really think of any reason this would be used. We are inheriting from nn.Module already and being careful to keep all the needed nn.Module methods. Someone should be able to use a modulus.Module exactly like a pytorch model and have no issues. I can see some problems with type checking but I don't know, that kinda feels like their problem as all models will be inherited from nn.Module anyways.

NickGeneva: Yeah, its fine without. Just wanted to throw it out there and see your throughts. Not sure I exactly see the use case right now since the point of these packages is to use the Modulus utils. Resolving.

modulus/models/module.py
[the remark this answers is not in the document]
NickGeneva: I cant find my comment about this, so sorry if I'm duplicating. My main concern is if a model needs to have some sort of custom save checkpoint method (for example model parallel). There needs to be a way to save model weights consistently (like removing DDP wrappers) and also a way to overwrite with custom implementations. 

Basically users should  be able to to use `torch.save(model.state_dict(), PATH)` for  most models, but for the best method use some `module.save()` for the weight export. This also allows us to better control how exactly that state_dict is stored. (similar story with load)

loliverhennigh:
Hey @akshaysubr and @NickGeneva, This is ready for another round of revisions. I think everything here is in a pretty good place however there is one key design decision that I will explain bellow. 

So for the entrypoint exposure I followed the example here to the letter, https://amir.rachum.com/python-entry-points/. In particular I follow their final revision and expose all our models through entrypoints and treat internal and external models the same. You can see where I get the list of models here, https://github.com/loliverhennigh/modulus/blob/fea-checkpoint_refactor/modulus/models/registry.py#L20.

This goes somewhat counter to the way that fsspec does things. Instead of exposing their protocols through entrypoints and getting them from there they manually write them in a dict here https://github.com/fsspec/filesystem_spec/blob/master/fsspec/registry.py#L62C2-L62C2 . Functionally this leads to the same design and interface for users and they have good documentation on how to add protocols through entrypoints. I would say the advantage of this is they can easily add protocols from outside packages and you can see a few examples of this in the dict. If we did the same thing maybe we could bring in hugging face models in for example.

I am open to setting all this up following the fsspec example but at this point there is no benefit to doing it I think. I would say we go with the design I have in this PR and then if we want we can switch to the fsspec setup at any time. This should be easy to do given how I structured the code and will basically just be adding a dict to the `registry.py` file. This will also not effect any user interface.

Also, just to make it a bit more concrete on how a user can expose a model to modulus via entrypoints I included a little example here https://github.com/loliverhennigh/modulus/blob/fea-checkpoint_refactor/examples/ExternalPackage/example/load_model.py. Ill delete this after but it was really good to wrap my head around this stuff.

modulus/models/module.py
loliverhennigh: Fixed this

modulus/models/module.py
```diff
@@ -65,55 +106,286 @@ def debug(self):
...
+        if save_git_hash:
+            import git
+            repo = git.Repo(search_parent_directories=True)
+            try:
+                metadata_info["git_hash"] = repo.head.object.hexsha
+            except git.InvalidGitRepositoryError:
+                metadata_info["git_hash"] = None
+
+        with open(directory.joinpath("metadata.json"), "w") as f:
+            json.dump(metadata_info, f)
```
akshaysubr, line 161: Can we use fsspec abstraction here instead of just posix file writes? Something like:
```python
from modulus.utils.filesystem import _get_fs

path_str = str(directory.joinpath("metadata.json"))
fs = _get_fs(path_str)
fs.pipe(path_str, json.dumps(metadata_info))
```
https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.pipe

akshaysubr: The other option would be to write everything to a temp directory and then copy to the final destination:
```python
with tempfile.TemporaryDirectory() as tmpdirname:
    with open(os.join(tmpdirname, "metadata.json"), "w") as f:
            json.dump(metadata_info, f)
    ...
    fs.put(tmpdirname, directory, recursive=True)
```

modulus/models/module.py
```diff
@@ -65,55 +106,286 @@ def debug(self):
...
-            file_name = self.meta.name + ".pt"
 
-        file_name = Path(file_name)
-        if not file_name.exists():
-            raise IOError(f"Model checkpoint {file_name} not found")
+        directory = Module._check_checkpoint(file_name)
 
-        model_dict = torch.load(file_name, map_location=self.device)
+        model_dict = torch.load(
+            directory.joinpath("model.pt"), map_location=self.device
```
akshaysubr, line 219: Would prefer if `map_location` was exposed in the args of this method but not sure how that affects other parts of the code that use `self.device`

modulus/models/module.py
```diff
@@ -65,55 +106,286 @@ def debug(self):
...
-        """Simple utility for loading the model from checkpoint
+        # Check if the checkpoint version is compatible with the current version
+        with open(directory.joinpath("metadata.json"), "r") as f:
+            metadata_info = json.load(f)
+            if metadata_info['mdlus_file_version'] != Module.__version__:
+                raise IOError(
+                    f"Model checkpoint version {metadata_info['mdlus_file_version']} is not compatible with current version {Module.__version__}"
+                )
+
+        return directory
```
akshaysubr, line 200: Why is the return type bool but a Path object is being returned? Also, wouldn't we want all of the `raise`s to be actually `return False`? Or maybe change the return type to an enum and return either success or an error code?

modulus/models/module.py
```diff
@@ -65,55 +106,286 @@ def debug(self):
...
+
+        Raises
+        ------
+        IOError
+            If file_name provided does not exist or is not a valid checkpoint
+        """
+
+        directory = Module._check_checkpoint(file_name)
+
+        with open(directory.joinpath("args.json"), "r") as f:
```
akshaysubr, line 244: Like in the earlier comment, would be especially useful here to use a filesystem abstraction so we can support both local as well as remote filesystems or NGC registries down the line

modulus/models/module.py
```diff
@@ -65,55 +106,286 @@ def debug(self):
...
+        return list(Module._model_registry.keys())
+
+    @classmethod
+    def _clear_model_registry(cls):
+        # NOTE: This is only used for testing purposes
+        Module._model_registry = {}
+
+    @classmethod
+    def _restore_model_registry(cls):
+        # NOTE: This is only used for testing purposes
```
akshaysubr, line 386: Should name this `__restore_model_registry` with two underscores to indicate fully private use only

modulus/models/registry.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+
+    model_registry = {}
+    group = "modulus.models"
+    entrypoints = pkg_resources.iter_entry_points(group)
+    for entry_point in entrypoints:
+        model_registry[entry_point.name] = entry_point
+    return model_registry
+
+_model_registry = _construct_registry()
```
akshaysubr, line 43: Why is this needed when there is also `modulus.Module._model_registry`?

modulus/models/module.py
```diff
@@ -65,55 +106,286 @@ def debug(self):
...
+        new_class_name = f"{torch_model_class.__name__}ModulusModel"
+        ModulusModel.__name__ = new_class_name
+
+        # Add this class to the dict of models classes
+        ModulusModel.register(ModulusModel, new_class_name)
+
+        return ModulusModel
+
+    @classmethod
+    def register(cls, model, name: Union[str, None] = None):
```
akshaysubr, line 314: Would prefer the registry to be a separate singleton class rather than a class variable in `Module`. IMO it is a much clearer API to have `modulus.registry.register(ModulusModel, new_class_name)` than to do `ModulusModel.register(ModulusModel, new_class_name)`

setup.py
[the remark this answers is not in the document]
akshaysubr: Agreed, would be good to have this be dynamically generated. One way would be to have all modulus models auto registered in `_construct_registry` and entrypoints would only be used for external implementations.

akshaysubr (changes requested):
@loliverhennigh This is great work! Left some comments about suggested changes. Would also be good to change the PR title and description to something more verbose and descriptive.

akshaysubr:
>Also, just to make it a bit more concrete on how a user can expose a model to modulus via entrypoints I included a little example here https://github.com/loliverhennigh/modulus/blob/fea-checkpoint_refactor/examples/ExternalPackage/example/load_model.py. Ill delete this after but it was really good to wrap my head around this stuff.

I think it is good to keep this example. Just need to ensure it gets run every so often to ensure that it doesn't go stale.

setup.py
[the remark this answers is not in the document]
NickGeneva: @loliverhennigh we discussed removing these last week in favor of having a dictionary stored in registry.py like fsspec. Is there a reason why you kept these in setup?

modulus/models/module.py
```diff
@@ -65,55 +105,256 @@ def debug(self):
...
+        group = "modulus.models"
+        try:
+            entry_point = entrypoints.get_group_named(group)[name]
+            model_class = entry_point.load()
+            return model_class
+        except KeyError:
+            raise KeyError(f"No model is registered under the name {name}")
+
+    @classmethod
+    def _clear_dynamically_created_classes(cls):
```
NickGeneva: I feel like this should be in registry. Like registry should be a singleton that gets initialized on import

akshaysubr: Agreed, suggested the same here: https://github.com/NVIDIA/modulus/pull/65#discussion_r1266295764

modulus/models/module.py
NickGeneva: Outdated, updated in refactor. Resolving.

modulus/models/module.py
loliverhennigh: Hey @akshaysubr and @NickGeneva , Can you please look at this file and tell me if thats more what you had in mind for a singlton class model registry. https://github.com/NVIDIA/modulus/pull/65/files#diff-d0c16db3b5d360b1240b22b92e2e109e2ed837b2b9ed2ad99d02666210cc5e91.

modulus/models/module.py
akshaysubr: If we wanted to only have a single file, I would prefer a `.tar` instead of `.tar.gz` or `.zip` to avoid having to compress and decompress and having that be another dependency. Can always add compression to a tarball later if we needed that.

modulus/models/module.py
akshaysubr: @loliverhennigh Yes, this was what I had in mind. I'll add a few comments at specific lines

modulus/models/registry.py
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
akshaysubr, line 1: Can this file just be in `modulus/registry/model_registry.py` and not have the last line there?

Then we can have a `modulus/registry/__init__.py` be
```python
from model_registry import ModelRegistry
```

A user would use it as 
```python
from modulus.models import Module
from modulus.registry import ModelRegistry

class CustomModel(Module):
    ...

# Since singleton class, don't need to keep around an instantiated version of this class
ModelRegistry().register(CustomModel, "custom_model")
```

modulus/models/registry.py
```diff
@@ -0,0 +1,67 @@
...
+    def __clear_model_registry(cls):
+        # NOTE: This is only used for testing purposes
+        cls._model_registry = {}
+
+    @classmethod
+    def __restore_model_registry(cls):
+        # NOTE: This is only used for testing purposes
+        cls._model_registry = cls._construct_registry()
+
+registry = ModelRegistry()
```
akshaysubr, line 67: This isn't needed here since this is a singleton class and a user can just instantiate it whenever and have the state be carried over. See how `DistributedManager` is used here https://github.com/NVIDIA/modulus/blob/main/modulus/distributed/utils.py#L108

modulus/models/registry.py
```diff
@@ -0,0 +1,67 @@
...
+# This model registry follows conventions similar to fsspec,
+# https://github.com/fsspec/filesystem_spec/blob/master/fsspec/registry.py#L62C2-L62C2
+# Tutorial on entrypoints: https://amir.rachum.com/blog/2017/07/28/python-entry-points/
+class ModelRegistry:
+    _instance = None
+    _model_registry = {}
+
+    def __new__(cls):
+        if not cls._instance:
+            cls._instance = super(ModelRegistry, cls).__new__(cls)
```
akshaysubr, line 27: What is the benefit of this style of singleton pattern vs the Borg pattern we have for `DistributedManager`? If there is one pattern that is clearly better, we should switch both classes to use that pattern.

Some interesting discussion here: https://stackoverflow.com/questions/1318406/why-is-the-borg-pattern-better-than-the-singleton-pattern-in-python

modulus/models/registry.py
loliverhennigh: Changed it to this.

modulus/models/registry.py
loliverhennigh: No preference here so switched to borg.

modulus/models/registry.py
loliverhennigh: Sounds good

loliverhennigh:
Hey @akshaysubr and @NickGeneva , Ready for another round of reviews! I think I addressed everything. Could I get another look when you have time. Thanks!

NickGeneva (approved):

akshaysubr (approved):

PR #84, opened 2023-07-25
fix: AMP in static capture

NickGeneva:
# Modulus Pull Request

## Description

Refactoring how AMP gradient scalars are handled in the static capture.

- Splitting amp options to auto-cast and gradient scalar
- Adding support for multiple gradient scalars
- Updating class functions for checkpointing to be more PyTorchic

I tested this on FCN with multiple validation passes and then also a restart of the run 20 epochs in and the training/validation loss looks correct, not jumps:
image

Sister PR: https://github.com/NVIDIA/modulus-launch/pull/51

CLoses: https://github.com/NVIDIA/modulus/issues/76

## Checklist


modulus/utils/capture.py
```diff
@@ -32,18 +32,20 @@ class _StaticCapture(object):
...
 
-    # Grad scalar singleton use for checkpointing
-    # This limits the number of staticcapture AMP training instances to just one per program
-    scaler_dict = None
-    scaler_singleton = None
+    # Grad scaler and checkpoint singleton use for checkpoint saving and loading
+    # Since an instance of Static capture does not exist for checkpoint functions
+    # one must use class functions to access state dicts
+    _amp_scalers = {}
+    _amp_scaler_checkpoints = {}
```
akshaysubr, line 39: If the intent is for this class to be a singleton, shouldn't this be enforced through either a singleton or Borg pattern?

modulus/utils/capture.py
```diff
@@ -229,31 +230,74 @@ def _amp_forward(self, *args, **kwargs) -> Any:
...
-            Python console logger
-        """
-        if cls.scaler_dict:
+    def _init_amp_scaler(
+        self, scaler_enabled: bool, logger: Logger
+    ) -> torch.cuda.amp.GradScaler:
+        # Create gradient scaler
+        scaler = torch.cuda.amp.GradScaler(enabled=scaler_enabled)
+        # Store scaler in class variable
+        scaler_id = f"scaler_{len(_StaticCapture._amp_scalers.keys())}"
```
akshaysubr, line 239: Is this robust to permutations of the order in which models are captured in the training script? Wondering if it might be good to allow a user to pass in an optional label for the static capture

modulus/utils/capture.py
```diff
@@ -309,7 +353,7 @@ class StaticCaptureTraining(_StaticCapture):
     Note
     ----
     Presently only a single instance of training static capture with AMP can be
-    used due to a grad scalar singleton.
+    used due to a grad scaler singleton.
```
akshaysubr, line 356: Is this still true?

akshaysubr (commented):
Handling the grad scaler in this way is nice and the decorator interface at the call site is quite clean. But this also comes with the downside that we have to manage all grad scalers and do the bookkeeping vs allowing a user to do this explicitly. Not sure which workflow is better though.

modulus/utils/capture.py
NickGeneva: Yeah right now its not... I had actually put an error because its order dependent (ignore the spelling error)

https://github.com/NickGeneva/modulus/blob/fix/static_capture/modulus/utils/capture.py#L254C68-L254C76

I suppose I figured that people arent typically moving stuff around during training and if you are odds are you'll want to implement all this yourself for direct control. Do you have a specific example in mind where you would have this occur?

akshaysubr: The only example where this might be an issue is if someone trains a model, then makes some changes to the training script but still wants to restart from the same checkpoint that they happen to know should work.

If we just add a `label : Optional[str] = None` kwarg to the constructor though, a user can set that and if they don't, then they're relying on this implicit naming convention?

modulus/utils/capture.py
NickGeneva: Yeah, doc was not correct, was more so a class function. But switched to singleton approach

modulus/utils/capture.py
```diff
@@ -308,8 +367,10 @@ class StaticCaptureTraining(_StaticCapture):
 
     Note
     ----
-    Presently only a single instance of training static capture with AMP can be
-    used due to a grad scalar singleton.
+    Static captures must be checkpointed when training using the `state_dict()` if AMP
+    is being used with gradient scaler. By default, this requires static captures to be
+    instantiated in the same order as when they were checkpointed. The label parameter
+    can be used to avoid this.
```
akshaysubr, line 373: Change the last line to avoid confusion?
> The label parameter can be used to relax/circumvent this ordering requirement.

NickGeneva: Updated!

modulus/utils/capture.py
NickGeneva: Nope, thanks for the catch. Updated!

modulus/utils/capture.py
NickGeneva: Added this with a test that checks the order in-variance does indeed work (test took longer than the feature).

loliverhennigh:
Hey, I have tested this as a sanity check and all the issue seem to be resolved. The validation and restoring no longer cause issues. I took a look at the PR and it looks good to me so I will approve these.

loliverhennigh (approved):

akshaysubr (approved):
Looks good, approving.

PR #90, opened 2023-07-28
Expanding climate datapipes to include cosin zenith, land sea mask, geopotential and latlong

loliverhennigh:
# Expanding climate datapipe

## Description

Refactor ERA5 datapipe so that it contains cosin zenith, lsm, geopotential and latlong stuff. This also generalized the loader to work with HRMIP data as well. Ultimate goal is to unify all weather modeling training scripts

Closes: https://github.com/NVIDIA/modulus/issues/77


loliverhennigh:
Hey @mnabian, this is not finished yet but should unblock us for now. Basically a refactor of the current datapipe that allows us to get all needed fields to train graphcast and other models. If you could take a quick look to make sure this works for our desired problems that would be great!

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,510 @@
+# Copyright (c) 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
mnabian, line 1: Do we want to keep both `climate_hdf5.py` and the existing `era5_hdf5.py`?

examples/make_datapipe.py
```diff
@@ -0,0 +1,85 @@
+
```
mnabian, line 1: These are great examples but maybe we should move them to modulus-launch?

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,510 @@
...
+
+from dataclasses import dataclass
+from typing import Iterable, List, Union, Tuple
+from pathlib import Path
+from torch.utils.data import Dataset
+
+from modulus.datapipes.datapipe import Datapipe
+from modulus.datapipes.meta import DatapipeMetaData
+
+from zenith_angle import cos_zenith_angle
```
mnabian, line 39: `zenith_angle` currently lives in the `examples` folder and should be moved.

mnabian (changes requested):

modulus/datapipes/climate/climate_hdf5.py
loliverhennigh: This is a great question I will address it bellow with some explination.

examples/make_datapipe.py
loliverhennigh: Yeah I will remove them for now. I was just testing it.

modulus/datapipes/climate/climate_hdf5.py
loliverhennigh: Fixed this.

loliverhennigh:
Hey @mnabian and @ktangsali, This is ready for final review. I just merged the static data files needed so hopefully it will pass the CI tests. A couple of notes on how we can move forward with this in regards to getting the unified recipe together. My suggestion is to review this PR and merge is ASAP if it look good. Then I will work to get the new data loader working with @mnabian's new train script. There might (but hopefully not) be some small issues that come up along the way that will may require modifications to unified datapipe. I suggest we make them in the PRs for the unified recipe instead of delaying this PR because it will results in an a super large PR and get unmanagable. When we do the PR with the unified recipes then we can also clean stuff up like remove the old `ERA5HDF5Datapipe`. Does this sound good?

Also, should I still be setting the merge to `main` or is there a different development branch now?

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,591 @@
...
+    """
+
+    def __init__(
+        self,
+        data_dir: str,
+        stats_dir: Union[str, None] = None,
+        channels: Union[List[int], None] = None,
+        batch_size: int = 1,
+        stride: int = 1,
+        dt: float = 6,
```
dallasfoster, line 161: Should this be a float?

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,591 @@
...
+        if isinstance(device, str):
+            device = torch.device(device)
+
+        # Need a index id if cuda
+        if device.type == "cuda" and device.index == None:
+            device = torch.device("cuda:0")
+        self.device = device
+
+        # check root directory exists
+        if not self.data_dir.is_dir():
```
dallasfoster, line 222: What if the data_dir is not on some local file system? Should we use something like fsspec for generality?

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,591 @@
...
+            if self.patch_size is not None:
+                self.cropped_data_shape = [
+                    s - s % self.patch_size[i] for i, s in enumerate(self.data_shape)
+                ]
+            else:
+                self.cropped_data_shape = self.data_shape
+            self.logger.info(f"Input data shape: {self.cropped_data_shape}")
+
+            # Get total length
+            self.total_length = self.n_years * self.num_samples_per_year
```
dallasfoster, line 281: Why do we have total_length and length?

dallasfoster (approved):
Most of my comments are for future improvements. Only a small number of open questions are pertinent for this PR.

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,591 @@
...
+
+class ClimateHDF5Datapipe(Datapipe):
+    """
+    A Climate DALI data pipeline for HDF5 files. This pipeline loads data from
+    HDF5 files, which can include latitude, longitude, cosine of zenith angle,
+    geopotential, and land sea mask if specified. Additionally, it normalizes
+    the data if a statistics file is provided. The pipeline returns a dictionary
+    with the following structure:
+
+    - `state_seq`: Tensor of shape (batch_size, num_steps, num_channels, height,
```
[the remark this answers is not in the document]
loliverhennigh: This is definitely possible. I think the long term solution is to make the data pipe more modular so the user can kinda construct the datapipe with whatever data they have. We could take the torchdata approach for example. Then we can give tools like computing the cosine of the solar zenith length as a transformation in the datapipe.

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,591 @@
...
+    ├── global_means.npy
+    │
+    ├── global_stds.npy
+    ```
+    The HDF5 files should contain the following variable
+    with the corresponding name:
+    - `fields`: Tensor of shape (num_timesteps, num_channels, height, width),
+      containing climate data. The order of the channels should match the order
+      of the channels in the statistics files. The statistics files should be
+      `.npy` files with the shape (1, num_channels, 1, 1).
```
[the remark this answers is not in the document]
loliverhennigh: Yeah I really don't like this but wanted to keep it consistent. I am actually much more of a fan of normalizing inside the network. You can use batch norm with infinite moving average for example. Then we can do away with keeping track of these files in the first place.

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,591 @@
...
+    auto_device: bool = True
+    cuda_graphs: bool = True
+    # Parallel
+    ddp_sharding: bool = True
+
+
+class ClimateHDF5Datapipe(Datapipe):
+    """
+    A Climate DALI data pipeline for HDF5 files. This pipeline loads data from
+    HDF5 files, which can include latitude, longitude, cosine of zenith angle,
```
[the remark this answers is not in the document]
loliverhennigh: fixed

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,591 @@
...
+    │   ├── 1981.h5
+    │   ├── 1982.h5
+    │   ├── ...
+    │   └── 2020.h5
+    ├── stats_dir
+    │
+    ├── global_means.npy
+    │
+    ├── global_stds.npy
+    ```
```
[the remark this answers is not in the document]
loliverhennigh: So the user specifies what the `data_dir` and `stats_dir` are right now. I think this is fine for now.

modulus/datapipes/climate/climate_hdf5.py
loliverhennigh: fixed

modulus/datapipes/climate/climate_hdf5.py
loliverhennigh: Fixed this.

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,591 @@
...
+class MetaData(DatapipeMetaData):
+    name: str = "ClimateHDF5"
+    # Optimization
+    auto_device: bool = True
+    cuda_graphs: bool = True
+    # Parallel
+    ddp_sharding: bool = True
+
+
+class ClimateHDF5Datapipe(Datapipe):
```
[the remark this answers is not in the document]
loliverhennigh: This should work with `netcdf`. Do you know if `netCDF4` can open zarr arrays as well?

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,591 @@
...
+    ├── global_stds.npy
+    ```
+    The HDF5 files should contain the following variable
+    with the corresponding name:
+    - `fields`: Tensor of shape (num_timesteps, num_channels, height, width),
+      containing climate data. The order of the channels should match the order
+      of the channels in the statistics files. The statistics files should be
+      `.npy` files with the shape (1, num_channels, 1, 1).
+
+    This pipeline assumes the HDF5 files have no metadata, such as timestamps.
```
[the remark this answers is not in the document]
loliverhennigh: I wanted to do this but was a bit worried. The metadata that we need is just the time stamps. The issue I see is that I have seen people specify these timestamps in their netcdf files different which might mess up the cosine of the solar zenith angle. I don't know, I would say we can fix this in the future...

modulus/datapipes/climate/climate_hdf5.py
loliverhennigh: This is a good point and we have some file abstraction already in Modulus. Right now though this datapipe should definitely not be used for any non-local file systems. If you tried this with s3 for example it would be very slow. There are some other issue that come up with pulling data with multiple threads as well that might cause issues.

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,591 @@
...
+      width). This sequence is drawn from the HDF5 file and normalized if a
+      statistics file is provided.
+    - `timestamps`: Tensor of shape (batch_size, num_steps), containing
+      timestamps for each timestep in the sequence.
+    - `land_sea_mask`: Tensor of shape (batch_size, 1, height, width),
+      containing the land sea mask if a path to a land sea mask file is
+      provided.
+    - `geopotential`: Tensor of shape (batch_size, 1, height, width), containing
+      geopotential if a path to a geopotential file is provided.
+    - `latlon`: Tensor of shape (batch_size, 2, height, width), containing
```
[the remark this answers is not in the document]
loliverhennigh: Responded to above

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,590 @@
...
+            self.logger.info(f"Climate file found: {data_path}")
+        self.n_years = len(self.data_paths)
+        self.logger.info(f"Number of years: {self.n_years}")
+
+        # get total number of examples and image shape from the first file,
+        # assuming other files have exactly the same format.
+        self.logger.info(f"Getting file stats from {self.data_paths[0]}")
+        with h5py.File(self.data_paths[0], "r") as f:
+            # truncate the dataset to avoid out-of-range sampling
+            data_samples_per_year = f["fields"].shape[0] - self.num_steps * self.stride
```
mnabian, line 260: Shouldn't this be  `data_samples_per_year = f["fields"].shape[0] - (self.num_steps-1) * self.stride`

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,590 @@
...
+    def __init__(
+        self,
+        data_dir: str,
+        stats_dir: Union[str, None] = None,
+        channels: Union[List[int], None] = None,
+        batch_size: int = 1,
+        stride: int = 1,
+        dt: float = 6.0,
+        start_year: int = 1980,
+        num_steps: int = 2,
```
mnabian, line 163: This might create some confusion. I suggest we have the `num_steps` to represent the number of prediction steps, and add one to that for the input.

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,590 @@
...
+        AssertionError
+            If loaded numpy arrays are not of correct size
+        """
+        # If no stats dir we just skip loading the stats
+        if self.stats_dir is None:
+            self.mu = None
+            self.std = None
+            return
+        # load normalisation values
+        mean_stat_file = self.stats_dir / Path("global_means.npy")
```
mnabian, line 318: Ideally, we should not hardcode the file names.

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,590 @@
...
+        if not std_stat_file.exists():
+            raise IOError(f"Std statistics file {std_stat_file} not found")
+
+        # has shape [1, C, 1, 1]
+        self.mu = np.load(str(mean_stat_file))[:, self.channels]
+        # has shape [1, C, 1, 1]
+        self.sd = np.load(str(std_stat_file))[:, self.channels]
+
+        if not self.mu.shape == self.sd.shape == (1, len(self.channels), 1, 1):
+            raise AssertionError("Error, normalisation arrays have wrong shape")
```
mnabian, line 332: Suggestion for a more informative message:
`raise AssertionError(f"Error, normalisation arrays mean and std have shapes {self.mu.shape}, {self.sd.shape}, but expected arrays of shape {(1, len(self.channels), 1, 1)}")`

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,590 @@
...
+    def _load_geopotential(self, normalize: bool = True) -> None:
+        """Get geopotential from netCDF file."""
+        ds = nc.Dataset(self.geopotential_filename)
+        geop = np.array(ds["z"]).astype(np.float32)
+        geop = np.flip(
+            geop, axis=1
+        )  # flip latitude axis, TODO hacky fix and we should get this from the file
+        assert (
+            geop.shape[1:] == self.data_shape
+        ), "Geopotential shape does not match data shape"
```
mnabian, line 356: Again, a more informative error message can be
`f"Geopotential shape {geop.shape[1:]} does not match data shape {self.data_shape}"`

mnabian:
> Hey @mnabian and @ktangsali, This is ready for final review. I just merged the static data files needed so hopefully it will pass the CI tests. A couple of notes on how we can move forward with this in regards to getting the unified recipe together. My suggestion is to review this PR and merge is ASAP if it look good. Then I will work to get the new data loader working with @mnabian's new train script. There might (but hopefully not) be some small issues that come up along the way that will may require modifications to unified datapipe. I suggest we make them in the PRs for the unified recipe instead of delaying this PR because it will results in an a super large PR and get unmanagable. When we do the PR with the unified recipes then we can also clean stuff up like remove the old `ERA5HDF5Datapipe`. Does this sound good?
> 
> Also, should I still be setting the merge to `main` or is there a different development branch now?

That sounds good to me. And yes, please merge to `main`

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,590 @@
+# Copyright (c) 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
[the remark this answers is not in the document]
loliverhennigh: No reason for the naming really and would be open to something else. I think we should keep the current ERA5 dataloader though until at least after we make the unified recipes. This can just be a smallish PR working towards that.

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,590 @@
...
+
+        Raises
+        ------
+        IOError
+            If mean or std numpy files are not found
+        AssertionError
+            If loaded numpy arrays are not of correct size
+        """
+        # If no stats dir we just skip loading the stats
+        if self.stats_dir is None:
```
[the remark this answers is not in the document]
loliverhennigh: Added this

modulus/datapipes/climate/climate_hdf5.py
```diff
@@ -0,0 +1,590 @@
...
+        if normalize:
+            geop = (geop - geop.mean()) / geop.std()
+        self.geopotential = dali.types.Constant(geop)
+
+    def _load_latlon(self) -> None:
+        """Load latitude and longitude coordinates from data shape and compute cos/sin versions."""
+
+        # get latitudes and longitudes from data shape
+        lat = np.linspace(-90, 90, self.cropped_data_shape[0]).astype(np.float32)
+        lon = np.linspace(-180, 180, self.cropped_data_shape[1]).astype(np.float32)
```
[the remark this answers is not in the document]
loliverhennigh: Fixed this. The way you are doing it is actually correct and the way I had it would make the long off by 1 cell

modulus/datapipes/climate/climate_hdf5.py
loliverhennigh: Fixed this.

modulus/datapipes/climate/climate_hdf5.py
loliverhennigh: Fixed this.

modulus/datapipes/climate/climate_hdf5.py
loliverhennigh: So I am actually much more in favor of giving a sequence of states and having the user decide what they want to do for the input output and sequence unroll. Some of the networks such as DLWP do some strange things where it takes in `n` time steps and then predicts `m` time steps. This can get pretty confusing in the loop unroll as to predict the next step you might need to take in some of the previous input time steps and some or all of the predicted time steps. I think writing a loop unroll for these cases can get very confusing when you explicitly pass the network invar and outvar.

modulus/datapipes/climate/climate_hdf5.py
loliverhennigh: Yeah, don't really like this either. I think it is best to leave this in for now though. The way we are dealing with normalization if very prone to error and I think explicitly making the file names like this makes it a little less likely someone will swap files or something...

modulus/datapipes/climate/climate_hdf5.py
loliverhennigh: fixed this

modulus/datapipes/climate/climate_hdf5.py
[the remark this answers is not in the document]
loliverhennigh: Added this

mnabian (approved):

PR #92, opened 2023-07-29
Adding utilities to DistributedManager to create process groups

akshaysubr:
# Modulus Pull Request

## Description
- Adds the ability to create process groups and orthogonal process groups using the distributed manager.
- Adds the ability to control what distributed initialization method to use based on an environment variable.
- Reintroduce distributed AFNO implementation.

closes https://github.com/NVIDIA/modulus/issues/58

## Checklist

## Dependencies

No additional dependencies


modulus/models/afno/distributed/mappings.py
```diff
@@ -12,135 +12,136 @@
...
-# # -----------------
-# # matmul parallel
-# def copy_to_matmul_parallel_region(input_):
-#     return _CopyToMatmulParallelRegion.apply(input_)
 
+# -----------------
+# Helper functions.
+# -----------------
+# matmul parallel
+def copy_to_matmul_parallel_region(input_):
```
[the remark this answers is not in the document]
akshaysubr: Yes, I agree. I moved them to `modulus.distributed` in this follow up MR: https://github.com/NVIDIA/modulus/pull/98

modulus/distributed/utils.py
```diff
@@ -130,90 +130,91 @@ def gather_loss(loss: float, dst_rank: int = 0, mean: bool = True):
...
-#     return output
+    return output
 
 
-# def _gather(input_, dim_, group=None):
-#     """Gather tensors and concatinate along the last dimension."""
-#     # get input format
-#     input_format = get_memory_format(input_)
+def _gather(input_, dim_, group=None):
+    """Gather tensors and concatinate along the last dimension."""
```
stadlmax, line 196: - While left over from previous versions, I think we should fix the typos of `concatinate` now in several places :) 
- `_gather` here is over the specified dimension, right, and not only the last dimension?
- I'm unsure about naming these utilities and their arguments following the `_` convention. I wouldn't oppose just a collection of "public" wrappers for gather/scatter etc. including their Autograd Wrappers. 
- With the comment above in mind, this should could be something like `all_gather_wrapper` instead of `_gather`

akshaysubr: Fixed the docstring issues. I agree about renaming the utilities, but wondering if we can do this in a follow-up PR. I would like to keep this PR fairly minimal since it is just reintroducing utilities. Would be good to keep actual changes separate from this PR.

akshaysubr: Created an issue to track this: https://github.com/NVIDIA/modulus/issues/112

daviddpruitt (approved):

modulus/distributed/utils.py
stadlmax: Thanks 👍🏽

stadlmax (approved):
LGTM

PR #105, opened 2023-08-03
[FEA] distributed autograd primitives (for tensors of variable size on each rank)

stadlmax:
# Modulus Pull Request

## Description
- provides wrappers around `torch.distributed` primitives to implement `gather_v`, `scatter_v`, `all_gather_v`, `all_reduce_v`, `indexed_all_gather` primitives
- provides autograd functions for these distributed primitives
- adds multi-gpu tests for these primitives

## Checklist

## Dependencies
no new added dependencies


modulus/distributed/utils.py
```diff
@@ -218,3 +219,293 @@ def _gather(input_, dim_, group=None):
...
+                tensor_list[dst] = tensor
+            else:
+                tensor_shape[dim] = sizes[r]
+                tensor_list[r] = torch.empty(
+                    tensor_shape,
+                    device=tensor.device,
+                    dtype=tensor.dtype,
+                )
+
+                req_list[r] = dist.irecv(tensor_list[r], src=r, group=group)
```
akshaysubr, line 339: Why not just use `torch.distributed.gather` here? From the documentation, seems like it supports variable tensor sizes across processes: https://pytorch.org/docs/stable/distributed.html#torch.distributed.gather

modulus/distributed/utils.py
```diff
@@ -218,3 +219,293 @@ def _gather(input_, dim_, group=None):
...
+    scatter_list = None
+    if rank == src:
+        scatter_list = torch.split(tensor, sizes, dim=dim)
+        req_list = [None] * comm_size
+        for r in range(comm_size):
+            tensor_to_scatter_to_r = scatter_list[r]
+            if r == src:
+                output = tensor_to_scatter_to_r
+            else:
+                req_list[r] = dist.isend(tensor_to_scatter_to_r, dst=r, group=group)
```
akshaysubr, line 400: Same thing with scatter: https://pytorch.org/docs/stable/distributed.html#torch.distributed.scatter

modulus/distributed/utils.py
```diff
@@ -218,3 +219,293 @@ def _gather(input_, dim_, group=None):
...
+
+
+def indexed_all_gather_wrapper(
+    tensor: torch.Tensor,
+    scatter_indices: List[torch.Tensor],
+    recv_sizes: List[int],
+    send_sizes: List[int],
+    group: Optional[dist.ProcessGroup] = None,
+) -> torch.Tensor:
+    """implementation of the forward pass of a distributed indexed all_gather primitive"""
```
akshaysubr, line 420: Maybe add the fact that it concats along dim 0 here?

modulus/distributed/utils.py
```diff
@@ -218,3 +219,293 @@ def _gather(input_, dim_, group=None):
...
+                req_list[r].wait()
+
+    else:
+        req = dist.irecv(output, src=src, group=group)
+        req.wait()
+
+    return output
+
+
+def indexed_all_gather_wrapper(
```
akshaysubr, line 413: Why is this called `indexed_all_gather`? Isn't this like a neighbor all-to-all kinda like `MPI_Neighbor_Alltoallv`?

test/distributed/test_autograd.py
```diff
@@ -0,0 +1,247 @@
...
+
+    expected_grad = torch.ones_like(tensor) * (-1)
+    assert torch.allclose(tensor.grad, expected_grad)
+
+    del os.environ["RANK"]
+    del os.environ["WORLD_SIZE"]
+    del os.environ["MASTER_ADDR"]
+    del os.environ["MASTER_PORT"]
+
+    DistributedManager.cleanup()
```
akshaysubr, line 210: Do you think it is sufficient to have this one cleanup method or should we also have a per group cleanup like you had earlier?

akshaysubr (commented):
Thanks for the well structured PR! The split of tools within the distributed module makes a lot of sense to me and the tests look good as well. Had a few comments that I've detailed in the review.

modulus/distributed/utils.py
stadlmax: IIUC, scatter definitely says `Note that all Tensors in scatter_list must have the same size.`

modulus/distributed/utils.py
stadlmax: Never was really sure how to call this to be honest. However, I agree that following MPI in that regard more closely could make sense, I'll change it to `all_to_all_v` then.

modulus/distributed/utils.py
stadlmax: Actually this is a good point. In a few places, the documentation clearly states that tensors should have the same size while here it only specifies `appropriately-sized tensors`. I am not sure if I tried gather but I thought that I indeed saw issues when tensors don't have the same size using just plain gather.

modulus/distributed/autograd.py
```diff
@@ -0,0 +1,319 @@
...
+
+    @staticmethod
+    def backward(ctx, grad_output: torch.Tensor):
+        """backward pass of the Distributed AllGatherV primitive"""
+
+        grad_tensor = None
+        needs_grad = ctx.needs_input_grad[0]
+
+        if needs_grad:
+            grad_tensor = all_reduce_v_wrapper(
```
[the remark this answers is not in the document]
stadlmax: Totally agree, this often is a real nightmare to work with in general.

This PR originally was intended for the "tensor-parallel" setting of distributed GNN operations. In these cases, the natural backward pass of an allgather indeed should be all reduce IMHO. An example could be a tensor-parallel-style operation in one layer, followed by a multi-headed layer or an "expert-style" parallelism. You call allgather to have the same data across all ranks and then have one "expert" per rank all operating on the same data in parallel with all the other ranks. The gradient through allgather then should be an all-reduce if I am correct. 

Another example would be a loss computation based on an allgathered prediction in which you don't know which rank e.g. computes the loss. While not performant, all ranks besides one could compute a "dummy" loss of 0 without breaking the autograd idea.

If the weights operating on the allgathered inputs would be the same, each rank would perform the same computation. So this setting wouldn't make much sense to me in the first place. If, however, you later have different operations on different ranks, it still should be correct to allreduce gradients during the backward pass.

modulus/distributed/autograd.py
```diff
@@ -0,0 +1,319 @@
...
+
+
+def all_gather_v(
+    tensor: torch.Tensor,
+    sizes: List[int],
+    dim: int = 0,
+    use_fp32: bool = True,
+    group: Optional[dist.ProcessGroup] = None,
+) -> torch.Tensor:
+    """Performs a distributed all_gather of tensors of varying sizes
```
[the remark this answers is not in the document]
stadlmax: Yeah, docstrings haven't been too detailed until now. I'll try to improve this. :)

test/distributed/test_autograd.py
stadlmax: I needed my "per group" cleanup particularly because of my usage of things like "group id" etc.. I see two orthogonal things here actually. 

One thing would be to take care of the "set" environment variables in DistributedManager - which now requires these "del" statements in a lot of tests.

The other thing would be indeed a per group cleanup, i.e. basically a removal of all the things w.r.t. a certain process group - potentially including all its sub/orthogonal groups without having to reset the whole state.

modulus/distributed/utils.py
stadlmax: Actually, I still think that indexed_all_gather or something similar would be more appropriate. 

What this routine basically does is something like
```
node 0 has data elements 0, ..., 9
node 1 has data elements 10, ..., 15
etc...

node 0 wants to gather 0, 5 from 0
node 1 wants to gather 1, 4, 5 from 0
node 2 wants to gather 6, 7, 9 from 0
etc... and similarly each node gathers certain values from all the remaining ranks
```

So technically, this is not just each rank gathering a variable amount of (contiguous) data elements from each other rank but a variable amount of arbitrary (non-contiguous) elements. Does this make sense?

akshaysubr: Yes, this makes sense. But I still think all_gather is somewhat confusing here. Typically, all_gather means at the end of the call, all ranks have the same output data which is not the case here right? I agree that there should be an indexed qualifier though because it is non-contiguous elements like you said. An `indexed_all_to_all_v` perhaps?

stadlmax: That sounds better, I agree on the difference of `all_to_all_v` and `all_gather` here. 👍🏽

stadlmax:
@akshaysubr 
Following up on our conversations, the updates now
- should make the docstrings more precise and detailed
- drops `all_to_all_single` due its current experimental status to avoid potential breaking changes
- renames `indexed_all_gather` to `indexed_all_to_all_v` as suggested

I also checked again that scatter/gather indeed currently assume tensors to have the same size on each rank justifying the current implementation relying on send/recv constructs.

akshaysubr (approved):
Changes look good to me! Approving, only have a couple of minor comments.

daviddpruitt (approved):

PR #134, opened 2023-08-10
custom json encoder for omegaconf

mnabian:
# Modulus Pull Request

## Description

closes https://github.com/NVIDIA/modulus/issues/133

## Checklist

### DIFF
--- a/modulus/models/module.py
+++ b/modulus/models/module.py
@@ -21,6 +21,7 @@ import importlib
 import tempfile
 import tarfile
 import pkg_resources
+import omegaconf
 from typing import Union, List, Dict, Any
 from pathlib import Path
 import torch.nn as nn
@@ -178,7 +179,7 @@ class Module(torch.nn.Module):
             torch.save(self.state_dict(), local_path / "model.pt")
 
             with open(local_path / "args.json", "w") as f:
-                json.dump(self._args, f)
+                json.dump(self._args, f, cls=OmegaConfEncoder)
 
             # Save the modulus version and git hash (if available)
             metadata_info = {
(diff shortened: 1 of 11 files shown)


modulus/models/module.py
```diff
@@ -394,3 +395,12 @@ def num_parameters(self) -> int:
         for name, param in self.named_parameters():
             count += param.numel()
         return count
+
+
+class OmegaConfEncoder(json.JSONEncoder):
```
akshaysubr, line 400: Why have a custom JSONEncoder instead of something like this before the json.dump call?
```python
self._args = {k : (
    omegaconf.OmegaConf.to_container(v, resolve = True)
    ) if isinstance(v, (omegaconf.ListConfig, omegaconf.DictConfig)) else v}
```

akshaysubr (commented):
The implementation of the custom encoder itself looks good. But I'd like to understand where this is being used and why are `OmegaConf` objects being passed in as args to model constructors.

mnabian:
@akshaysubr please see the discussions here: https://github.com/NVIDIA/modulus/issues/133
The reason for adding this custom encoder is to be able to directly use hydra configs to specify the model arguments.

modulus/models/module.py
mnabian: In my opinion both are the same, but I think custom JSON encoder is a cleaner implementation and is consistent with this: https://github.com/python/cpython/blob/main/Lib/json/encoder.py#L161

akshaysubr: IMO, we should fix this at the call site instead of adding omegaconf as a dependency to modulus-core as mentioned in  https://github.com/NVIDIA/modulus-launch/pull/79#pullrequestreview-1579366654

Creating a custom JSON encoder for one package is clean. But I worry that we'll get into a maintenance issue if we need to start having more custom encoders which would then have to inherit from this `OmegaConfEncoder` and create a serial inheritance chain of encoders.

mnabian: I agree. Fixing this at the call site is a better approach. I will close this PR then.

PR #139, opened 2023-08-18
[FEA] distributed gnn primitives

stadlmax:
# Modulus Pull Request

## Description
- note: hold back until #105 is merged, this PR builds upon these changes
- based on distributed primitives in #105, this PR adds distributed GNN primitives using these generic routines
- add a naive graph partitioning scheme
- modifies GraphCastNet and MeshGraphNet to (more or less) allow straightforward tensor-parallel distribution of these models
- adds tests for distributed runs for these two models testing underlying GNN primitives implicitly 

## Checklist


modulus/distributed/utils.py
```diff
@@ -615,9 +615,9 @@ def indexed_all_to_all_v_wrapper_bwd(
     tensor_shape = list(tensor.shape)
 
     # scatter gradients, roles reversed compared to forward pass
-    recv_sizes = [sizes[r][rank] for r in range(comm_size)]
     recv_list = [None] * comm_size
     for r in range(comm_size):
+        recv_sizes = [sizes[i][r] for i in range(comm_size)]
```
stadlmax, line 620: bugfix

mnabian:
@stadlmax the coverage test is failing because it falls below 80%. Could you please use `# no pragma` in your code until we stand up the multi-GPU CI and coverage? Example: https://github.com/NVIDIA/modulus/blob/main/modulus/models/sfno/sfnonet.py#L104

modulus/models/graphcast/graph_cast_net.py
```diff
@@ -129,9 +129,19 @@ def __init__(
         use_cugraphops_decoder: bool = False,
         do_concat_trick: bool = False,
         recompute_activation: bool = False,
+        partition_size: int = 1,
+        partition_group_name: Optional[str] = None,
+        expect_partitioned_input: bool = False,
```
akshaysubr, line 134: When this is `True`, how would one account for that in the dataloader? Is there a mechanism for that? If not, we should think about how we can implement some such mechanisms in a future PR.

modulus/models/graphcast/graph_cast_net.py
```diff
@@ -173,35 +170,108 @@ def __init__(
         self.mesh_edata = self.mesh_graph.edata["x"]
         self.mesh_ndata = self.mesh_graph.ndata["x"]
 
-        if use_cugraphops_encoder:
-            offsets, indices, edge_ids = self.g2m_graph.adj_tensors("csc")
-            n_in_nodes, n_out_nodes = (
+        if use_cugraphops_encoder or self.is_distributed:
+            print(self.g2m_graph)
```
akshaysubr, line 174: Debug code?

modulus/models/graphcast/graph_cast_net.py
```diff
@@ -173,35 +170,108 @@ def __init__(
...
-            n_in_nodes, n_out_nodes = (
+            if self.is_distributed:
+                self.g2m_edata = self.g2m_graph.get_edge_features_in_partition(
+                    self.g2m_edata
+                )
+
+        if use_cugraphops_decoder or self.is_distributed:
+            try:
+                offsets, indices, edge_ids = self.m2g_graph.adj_tensors("csc")
+            except:
```
akshaysubr, line 202: Since this is a specific test in the try block, would be good to either catch only that specific exception so other exceptions still get raised or use something like `if hasattr(self.m2g_graph, "adj_tensors")`

modulus/models/graphcast/graph_cast_net.py
```diff
@@ -173,35 +170,108 @@ def __init__(
...
-            n_in_nodes, n_out_nodes = (
+            if self.is_distributed:
+                self.m2g_edata = self.m2g_graph.get_edge_features_in_partition(
+                    self.m2g_edata
+                )
+
+        if use_cugraphops_processor or self.is_distributed:
+            try:
+                offsets, indices, edge_ids = self.mesh_graph.adj_tensors("csc")
+            except:
```
akshaysubr, line 227: Same exception specification comment as above

modulus/models/layers/fused_silu.py
```diff
@@ -15,7 +15,11 @@
 import functools
 import torch
 from torch.autograd import Function
-from nvfuser._C import Fusion, FusionDefinition, DataType
+
+try:
+    from nvfuser._C import Fusion, FusionDefinition, DataType
+except:
+    from torch._C._nvfuser import Fusion, FusionDefinition, DataType
```
akshaysubr, line 22: This is to accommodate earlier versions of PyTorch? Would also be good to specify an `ImportError` exception here

test/models/graphcast/test_graphcast_snmg.py
```diff
@@ -0,0 +1,171 @@
...
+    del os.environ["WORLD_SIZE"]
+    del os.environ["MASTER_ADDR"]
+    del os.environ["MASTER_PORT"]
+
+    DistributedManager.cleanup()
+
+
+@pytest.mark.multigpu
+@pytest.mark.parametrize("dtype", [torch.float32, torch.float16, torch.bfloat16])
+@pytest.mark.parametrize("do_concat_trick", [False, True])
```
akshaysubr, line 154: Would it be worth to parameterize on checkpointing as well to ensure it works in conjunction with distributed graphs?

modulus/models/graphcast/graph_cast_net.py
```diff
@@ -173,35 +170,108 @@ def __init__(
...
+                partition_group_name=partition_group_name,
             )
+            if self.is_distributed:
+                self.mesh_edata = self.mesh_graph.get_edge_features_in_partition(
+                    self.mesh_edata
+                )
+                self.mesh_ndata = self.mesh_graph.get_dst_node_features_in_partition(
+                    self.mesh_ndata
+                )
+                print(self.mesh_ndata)
```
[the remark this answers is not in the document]
stadlmax: my mistake, left over from stupid printf debugging ^^

modulus/models/layers/fused_silu.py
stadlmax: agreed

modulus/models/graphcast/graph_cast_net.py
stadlmax: agreed, good suggestion

modulus/models/graphcast/graph_cast_net.py
stadlmax: yes

modulus/models/gnn_layers/distributed_graph.py
```diff
@@ -0,0 +1,357 @@
...
+        self.scatter_indices = scatter_indices
+
+        for r in range(self.partition_size):
+            assert self.sizes[self.partition_rank][r] == self.scatter_indices[r].numel()
+
+        dist.barrier(self.process_group)
+
+    def get_src_node_features_in_partition(
+        self,
+        global_node_features: torch.Tensor,
```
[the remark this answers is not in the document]
stadlmax: Technically, these things are all assumed to live on the same device as device_manager.device. If they are not, routines have optional flags to indicate whether things should be broadcast/scattered/gathered/etc. I'll add more docstrings to make this clearer.

modulus/models/gnn_layers/distributed_graph.py
```diff
@@ -0,0 +1,357 @@
...
+        self, partitioned_edge_features: torch.Tensor
+    ) -> torch.Tensor:
+        # current partitioning scheme assumes that
+        # local graph is built from partitioned IDs
+        return partitioned_edge_features
+
+    def get_global_src_node_features(
+        self,
+        partitioned_node_features: torch.Tensor,
+        get_on_all_ranks: bool = True,
```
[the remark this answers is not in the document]
stadlmax: In relation to the comment above, I think it could be more elegant to add a more general "source-device" or a "destination device". This way, the user has more control on which device certain things should live - potentially even with CPU support.

modulus/models/gnn_layers/distributed_graph.py
```diff
@@ -0,0 +1,357 @@
...
+        assert self.partition_size == partition_size, error_msg
+        self.process_group = dist_manager.group(name=graph_partition_group_name)
+
+        # this partitions offsets and indices on each rank in the same fashion
+        # it could be rewritten to do it on one rank and exchange the partitions
+        # however, as we expect the global graphs not to be too large for one CPU
+        # we do it once and then can get rid of it afterwards without going through
+        # tedious gather/scatter routines for communicating the partitions
+
+        # get distribution of destination IDs
```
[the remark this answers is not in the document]
stadlmax: Agreed. I see two steps here.
- one is simply factoring things out, I'll add this to this PR which makes addressing a few other comments easier, too
- a follow-up would be thinking about how to generalize this to other strategies, too

modulus/models/graphcast/graph_cast_net.py
stadlmax: Unfortunately, the dataloader side hasn't really been thought about. The main use-case for this would e.g. be multi-step roll-out, i.e. when the output is partitioned, the next-steps can `expect_partitioned_input` which for now would require re-setting this flag or doing a manual partitioning before the rollout

modulus/models/graphcast/graph_cast_net.py
stadlmax: agreed

modulus/models/gnn_layers/distributed_graph.py
```diff
@@ -0,0 +1,357 @@
...
+        self.num_dst_nodes_in_each_partition = [None] * partition_size
+
+        # mapping of local IDs to their corresponding global IDs
+        self.partitioned_indices_to_global = None
+        self.partitioned_src_node_ids_to_global = None
+        self.partitioned_dst_node_ids_to_global = None
+
+        # utility variables for torch.distributed
+        dist_manager = DistributedManager()
+        self.device_id = dist_manager.device
```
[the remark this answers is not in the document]
stadlmax: agreed

modulus/models/gnn_layers/graph.py
```diff
@@ -0,0 +1,365 @@
...
+        to its rank in the process group across which the graph is partitioned,
+        get the global edge features either on all group ranks or on group rank 0.
+        """
+        if self.is_distributed:
+            return self.dist_graph.get_global_edge_features(
+                local_efeat, get_on_all_ranks
+            )
+        return local_efeat
+
+    def to(self, *args: Any, **kwargs: Any) -> "CuGraphCSC":
```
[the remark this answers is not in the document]
stadlmax: This is something we have in many places. IIUC, this is about properly handling forward references as we can't use the class itself (here CuGraphCSC) as type when defining the class or methods.

modulus/models/gnn_layers/distributed_graph.py
```diff
@@ -0,0 +1,357 @@
...
+                    src_offsets_in_partition[rank], src_offsets_in_partition[rank + 1]
+                )
+                self.partitioned_dst_node_ids_to_global = range(
+                    dst_offsets_in_partition[rank], dst_offsets_in_partition[rank + 1]
+                )
+                self.partitioned_indices_to_global = range(
+                    global_offsets[offset_start], global_offsets[offset_end]
+                )
+
+                self.local_offsets = offsets.detach().clone().to(device=self.device_id)
```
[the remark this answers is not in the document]
stadlmax: I think this might be a leftover of an intermediate version which used things differently and thus this assignment needed a detach/clone as offsets was just a view of a larger global buffer. IIUC, this shouldn't be the case anymore and indeed could go away.

modulus/models/graphcast/graph_cast_net.py
```diff
@@ -129,9 +129,19 @@ def __init__(
         use_cugraphops_decoder: bool = False,
         do_concat_trick: bool = False,
         recompute_activation: bool = False,
+        partition_size: int = 1,
+        partition_group_name: Optional[str] = None,
```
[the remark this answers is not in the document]
stadlmax: Currently, this is somewhat intentional actually. If no group is specified, it is just assumed that the partitioning (or in this case model parallelism) is across the global process group. 
But definitely open to suggestions here.

modulus/models/graphcast/graph_cast_net.py
```diff
@@ -129,9 +129,19 @@ def __init__(
         use_cugraphops_decoder: bool = False,
         do_concat_trick: bool = False,
         recompute_activation: bool = False,
+        partition_size: int = 1,
```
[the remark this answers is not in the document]
stadlmax: definitely, yes

modulus/models/gnn_layers/graph.py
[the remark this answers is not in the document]
akshaysubr: Wow, this led me down a fascinating rabbit hole and I learnt something new today! Apparently you can do C++ template style generic types in python and still have annotations and static type checking for it!

WRT this specific comment, I found the documentation for this [here](https://mypy.readthedocs.io/en/stable/generics.html#generic-methods-and-generic-self) and it seems like using `"CuGraphCSC"` would be an explicit annotation but you could also use `TypeVar` so child classes of this can also be typed appropriately. I'll leave it up to you if you want to make this change though it seems somewhat unnecessary here.

modulus/models/graphcast/graph_cast_net.py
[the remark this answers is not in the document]
akshaysubr: Ah, okay. I would actually prefer if the default was to not use the global process group since the default `partition_size` is also 1. So the current defaults would work for a single GPU run, but not for non-distributed-graph training using DDP?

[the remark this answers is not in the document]
akshaysubr: Basically, what I am getting at is that it would be good to not require the user create a single GPU process group if they just want DDP and no model parallelism.

modulus/models/graphcast/graph_cast_net.py
akshaysubr: Okay, that makes sense. Just opened https://github.com/NVIDIA/modulus/issues/151 to track this dataloader issue.

modulus/models/gnn_layers/graph.py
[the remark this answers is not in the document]
stadlmax: Ah wow, yeah, the generic type indeed would be better in this case - despite not really necessary.

modulus/models/graphcast/graph_cast_net.py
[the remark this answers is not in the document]
stadlmax: Ah, now I understand. If implemented correctly, the logic should always be
- if partition_size <= 1, the code should run in the "single-gpu" path, i.e. the process group isn't actually needed
- if partition_size > 1, the graph is assumed to be partitioned across multiple GPUs and the passed process group will be used. If passed as None, the default process group will be used. However, there are usually checks in place to assert partition_size == group_size.

I'll add a comment about the validity of process groups though and the usage of the default process group otherwise.

modulus/models/graphcast/graph_cast_net.py
stadlmax: Thanks, definitely makes sense 👍🏽

modulus/models/gnn_layers/graph.py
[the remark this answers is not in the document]
stadlmax: Actually, it is much simpler to just use `typing.Self` as these things indeed just have a `return self`.

mnabian:
@stadlmax could you please black format your code? I recommend using pre-commit

stadlmax:
> @stadlmax could you please black format your code? I recommend using pre-commit

I usually do so, apparently I forgot it this time. Will do :)

modulus/models/gnn_layers/graph.py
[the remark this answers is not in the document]
akshaysubr: I'm not sure about using `typing.Self` since it is only supported in python 3.11 onwards according to [PEP 673](https://peps.python.org/pep-0673/). Is there a way to backport it to 3.9/3.10 which is what the NGC pytorch containers use?

modulus/distributed/autograd.py
```diff
@@ -50,7 +50,7 @@ def forward(
         dim: int = 0,
         use_fp32: bool = True,
         group: Optional[dist.ProcessGroup] = None,
-    ) -> torch.Tensor:
+    ) -> torch.Tensor: # pragma: no cover
```
akshaysubr: I understand why the pragma is needed here. I created a new issue to try and fix this. Can you add any other details I might have missed to this issue? https://github.com/NVIDIA/modulus/issues/165

modulus/models/gnn_layers/distributed_graph.py
```diff
@@ -27,6 +27,197 @@
...
 )
 
+@dataclass
+class GraphPartition:
+    partition_size: int
+    partition_rank: int
+    device: torch.device
+    # data structures for local graph
+    # of this current partition rank
+    local_offsets: torch.Tensor = None
```
akshaysubr, line 42: Should this be `Optional[torch.Tensor]` if default is `None`?

akshaysubr: I think generally, the defaults for many of this dataclass elements don't match the type annotations. This would cause lots of issues once we add static type checking https://github.com/NVIDIA/modulus/issues/121

modulus/models/gnn_layers/distributed_graph.py
```diff
@@ -35,206 +226,88 @@ def __init__(
         global_indices: torch.Tensor,
         partition_size: int,
         graph_partition_group_name: str,
-    ):
-        """Utility Class representing a distributed graph. Based on a CSC
-        structure represented by corresponding offsets and indices buffers,
-        the global graph is partioned into smaller local graphs.
+        graph_partition: Optional[GraphPartition] = None
```
akshaysubr: Really like this approach!

test/models/graphcast/utils.py
```diff
@@ -36,3 +37,13 @@ def get_icosphere_path():
     script_path = os.path.abspath(__file__)
     icosphere_path = os.path.join(os.path.dirname(script_path), "icospheres.json")
     return icosphere_path
+
+
+def compare_quantiles(t: torch.Tensor, ref: torch.Tensor, quantiles: List[float], tolerances: List[float]):
```
akshaysubr: This is a nice utility! Would be nice if you could add some docstrings so other tests can also use this.

test/models/meshgraphnet/test_meshgraphnet_snmg.py
```diff
@@ -125,13 +129,18 @@ def run_test_distributed_meshgraphnet(rank, world_size, dtype):
     loss.backward()
 
     out_multi_gpu = model_multi_gpu(nfeat_multi_gpu, efeat_multi_gpu, graph_multi_gpu)
-    loss = out_multi_gpu.sum()
+    # PyTorch unfortunately averages across all process groups within DistributedDataParallel
+    # by default, for tensor-parallel applications like this one, potential solutions
+    # are either custom gradient hooks (the best solution for actual training workloads)
+    # or multiplying by world_size to cancel out the normalization. As this is just a simple
+    # test, we do the easier thing here as we only compare the weight gradients.
+    loss = out_multi_gpu.sum() * world_size
```
akshaysubr, line 137: Is this not an artifact of the fact that you're using DDP with `process_group=manager.group("graph_partition")` and also creating the distributed graph using `partition_group_name="graph_partition"`? Ideally, wouldn't it be that you set up DDP with `process_group=manager.group("orthogonal_graph_partition")`?

akshaysubr (approved):
@stadlmax Thanks for addressing all the earlier comments. Adding some more relatively minor comments. Should be good to merge after these are addressed.

mnabian (approved):

modulus/distributed/autograd.py
stadlmax: No, seems good to me, thanks for opening the issue 👍🏽

modulus/models/gnn_layers/distributed_graph.py
stadlmax: I am actually not sure about this dataclass here.
What I want here is
- a container for the related variables defining a graph partition
- without the need to specify all of these values when initializing the object
- hence, a dataclass with mostly defaults

For tensors, you can't really define a proper default here as things like size, device, and especially dtype are only known at a later stage. I'll go with Optional for now, so technically this is not really accurate

test/models/graphcast/utils.py
stadlmax: It also would be great to have a place for these, actually.

test/models/meshgraphnet/test_meshgraphnet_snmg.py
stadlmax: IIUC, no. I would use the orthogonal process group if the the weights in each process group in `graph_partition` are not shared, right? Then you just would accumulate gradients for the same partition rank across different data-parallel groups.
Here, the model is shared and in a rather tensor-parallel setting. So the main artifact should stem from the fact that DDP by default averages the gradients based on world_size. My solution in actual code would be a custom gradient hook - though I wanted to keep the tests simpler and not deal with that.

akshaysubr: When the weights are shared, I was envisioning that you would instead do the all-reduce within the process group as a no-op autograd function that performs an all-reduce in the backward pass. But what you're saying is a valid approach too. This is another aspect to model parallel training that to me seems challenging to generalize. Would be good to think about this in more detail after getting this PR merged in.

PR #153, opened 2023-09-12
Add a way to configure the required process group structure for a model

akshaysubr:
# Modulus Pull Request

## Description

This PR adds a `ProcessGroupConfig` class to define the configuration of a model's parallel process group structure as a tree. Each node of the tree is of type `ProcessGroupNode`.                          

Once the process group config structure (i.e, the tree structure) is set, it is sufficient to set only the sizes for each leaf process group. Then, the size of every parent group can be automatically computed as the product reduction of the sub-tree of that parent group node.

The intent is to use this config structure as follows (which will require some more additions to `DistributedManager` and changes to distributed model implementations):

```python
class MyModel(Module):
    def __init__(self):
        pass

    @staticmethod
    def get_model_parallel_info(self) -> ProcessGroupConfig:
        mp = ProcessGroupNode("model_parallel", orthogonal_group="data_parallel")
        config = ProcessGroupConfig(mp)
        config.add_node(ProcessGroupNode("spatial_parallel"), parent=mp)
        config.add_node(ProcessGroupNode("channel_parallel"), parent="model_parallel")
        return config                                                                          


pg_config = MyModel.get_model_parallel_info()
group_sizes = {"channel_parallel": 4, "spatial_parallel": 2}
pg_config.set_leaf_group_sizes(group_sizes)

print(pg_config.leaf_groups())  # ['spatial_parallel', 'channel_parallel']
print(pg_config.get_node("model_parallel").size)  # 8

DistributedManager.create_process_groups(pg_config, topology_mapping)
manager = DistributedManager()

model = MyModel()
model = DDP(model, process_group=manager.group("data_parallel"))
```

Closes: https://github.com/NVIDIA/modulus/issues/142

## Checklist

## Dependencies
This PR adds [treelib](https://treelib.readthedocs.io/en/latest/) as a dependency to manage the process group config

### DIFF
--- a/modulus/distributed/config.py
+++ b/modulus/distributed/config.py
@@ -1,0 +1,259 @@
+# Copyright (c) 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from treelib import Tree
+from typing import Dict, List, Optional, Union
+
+
+class ProcessGroupNode:
+    """
+    Class to store the attributes of a distributed process group
+
+    Attributes
+    ----------
+    name : str
+        Name of the process group
+    size : Optional[int]
+        Optional, size of the process group
+    orthogonal_group : Optional[str]
+        Optional, name of an orthogonal process group to create
+    """
+
+    def __init__(
+        self,
+        name: str,
+        size: Optional[int] = None,
+        orthogonal_group: Optional[str] = None,
+    ):
+        """
+        Constructor for the ProcessGroupNode class
+
+        Parameters
+        ----------
+        name : str
+            Name of the process group
+        size : Optional[int]
+            Optional, size of the process group
+        orthogonal_group : Optional[str]
+            Optional, name of an orthogonal process group to create
+        """
+        self.name = name
+        self.size = size
+        self.orthogonal_group = orthogonal_group
+
+    def __str__(self):
+        """
+        String representation of the process group node
+
+        Returns
+        -------
+        str
+            String representation of the process group node
+        """
+        return (
+            "ProcessGroupNode("
+            f"name={self.name}, "
+            f"size={self.size}, "
+            f"orthogonal_group={self.orthogonal_group})"
+        )
+
+    def __repr__(self):
+        """
+        String representation of the process group node
+
+        Returns
+        -------
+        str
+            String representation of the process group node
+        """
+        return self.__str__()
+
+
+class ProcessGroupConfig:
+    """
+    Class to define the configuration of a model's parallel process group structure as a
+    tree. Each node of the tree is of type `ProcessGroupNode`.
+
+    Once the process group config structure (i.e, the tree structure) is set, it is
+    sufficient to set only the sizes for each leaf process group. Then, the size of
+    every parent group can be automatically computed as the product reduction of the
+    sub-tree of that parent group node.
+
+    Examples
+    --------
+    >>> from modulus.distributed.config import ProcessGroupNode, ProcessGroupConfig
+    >>> mp = ProcessGroupNode("model_parallel", orthogonal_group="data_parallel")
+    >>>
+    >>> # Create model parallel group with data parallel as the orthogonal group
+    >>> mp = ProcessGroupNode("model_parallel", orthogonal_group="data_parallel")
+    >>>
+    >>> # Create the process group config with the highest level process group
+    >>> config = ProcessGroupConfig(mp)
+    >>>
+    >>> # Create spatial and channel parallel sub-groups
+    >>> config.add_node(ProcessGroupNode("spatial_parallel"), parent=mp)
+    >>> config.add_node(ProcessGroupNode("channel_parallel"), parent="model_parallel")
+    >>>
+    >>> pg_config.leaf_groups()
+    ['spatial_parallel', 'channel_parallel']
+    >>>
+    >>> # Set leaf group sizes
+    >>> group_sizes = {"channel_parallel": 3, "spatial_parallel": 2}
+    >>> pg_config.set_leaf_group_sizes(group_sizes)  # Update all parent group sizes too
+    >>> pg_config.get_node("model_parallel").size
+    6
+    """
+
+    def __init__(self, node: ProcessGroupNode):
+        """
+        Constructor to the ProcessGroupConfig class
+
+        Parameters
+        ----------
+        node : ProcessGroupNode
+            Root node of the tree, typically would be 'model_parallel'
+            Note, it is generally recommended to always set the orthogonal_group for
+            the 'model_parallel' group to be 'data_parallel' to aid with distributed
+            data parallel training
+        """
+        self.root = node
+        self.root_id = node.name
+        self.tree = Tree()
+        self.tree.create_node(node.name, node.name, data=node)
+
+    def add_node(self, node: ProcessGroupNode, parent=Union[str, ProcessGroupNode]):
+        """
+        Add a node to the process group config
+
+        Parameters
+        ----------
+        node : ProcessGroupNode
+            The new node to be added to the config
+        parent : Union[str, ProcessGroupNode]
+            Parent node of the node to be added. Should already be in the config.
+            If str, it is the name of the parent node. Otherwise, the parent
+            ProcessGroupNode itself.
+        """
+        if isinstance(parent, ProcessGroupNode):
+            parent = parent.name
+        self.tree.create_node(node.name, node.name, data=node, parent=parent)
+
+    def get_node(self, name: str) -> ProcessGroupNode:
+        """
+        Method to get the node given the name of the node
+
+        Parameters
+        ----------
+        name : str
+            Name of the node to retrieve
+
+        Returns
+        -------
+        ProcessGroupNode
+            Node with the given name from the config
+        """
+        return self.tree.get_node(name).data
+
+    def update_parent_sizes(self, verbose: bool = False) -> int:
+        """
+        Method to update parent node sizes after setting the sizes for each leaf node
+
+        Parameters
+        ----------
+        verbose : bool
+            If True, print a message each time a parent node size was updated
+
+        Returns
+        -------
+        int
+            Size of the root node
+        """
+        return tree_product_reduction(self.tree, self.root_id, verbose=verbose)
+
+    def leaf_groups(self) -> List[str]:
+        """
+        Get a list of all leaf group names
+
+        Returns
+        -------
+        List[str]
+            List of all leaf node names
+        """
+        # return find_leaf_nodes(self.tree, self.root_id)
+        return [n.identifier for n in self.tree.leaves()]
+
+    def set_leaf_group_sizes(
+        self, group_sizes: Dict[str, int], update_parent_sizes: bool = True
+    ):
+        """
+        Set process group sizes for all leaf groups
+
+        Parameters
+        ----------
+        group_sizes : Dict[str, int]
+            Dictionary with a mapping of each leaf group name to its size
+        update_parent_sizes : bool
+            Update all parent group sizes based on the leaf group if True
+            If False, only set the leaf group sizes.
+        """
+        for id, size in group_sizes.items():
+            assert self.tree.contains(
+                id
+            ), f"Process group {id} is not in this process group config"
+            node = self.tree.get_node(id)
+            assert node.is_leaf(), f"Process group {id} is not a leaf group"
+            node.data.size = size
+
+        if update_parent_sizes:
+            self.update_parent_sizes()
+
+
+def tree_product_reduction(tree, node_id, verbose=False):
+    """
+    Function to traverse a tree and compute the product reduction of
+    the sub-tree for each node starting from `node_id`
+    """
+    children = tree.children(node_id)
+    node = tree.get_node(node_id)
+    if not children:
+        assert node.data.size is not None, "Leaf nodes should have a valid size set"
+        return node.data.size
+
+    product = 1
+
+    for child in children:
+        product *= tree_product_reduction(tree, child.identifier)
+
+    if node.data.size != product:
+        if verbose:
+            print(
+                "Updating size of node "
+                f"{node.data.name} from {node.data.size} to {product}"
+            )
+        node.data.size = product
+
+    return product
+
+
+def find_leaf_nodes(tree, node_id):
+    """
+    Function to find all leaf node identifiers of `tree` starting from `node_id`
+    """
+    if not tree.children(node_id):
+        return [node_id]
+    else:
+        leaf_nodes = []
+        for child_id in tree.children(node_id):
+            leaf_nodes.extend(find_leaf_nodes(tree, child_id.identifier))
+        return leaf_nodes
--- a/modulus/distributed/process_group.py
+++ b/modulus/distributed/process_group.py
@@ -25,7 +25,7 @@ class ProcessGroupNode:
     name : str
         Name of the process group
     size : Optional[int]
-        Optional, size of the process group
+        Optional, number of processes in the process group
     orthogonal_group : Optional[str]
         Optional, name of an orthogonal process group to create
     """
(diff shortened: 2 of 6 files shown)


modulus/distributed/config.py
dallasfoster, line 1: Is there a particular reason why `config.py` was chosen over something more informative about the contents of the file?

modulus/distributed/config.py
```diff
@@ -0,0 +1,259 @@
...
+        orthogonal_group: Optional[str] = None,
+    ):
+        """
+        Constructor for the ProcessGroupNode class
+
+        Parameters
+        ----------
+        name : str
+            Name of the process group
+        size : Optional[int]
```
dallasfoster, line 46: It is not clear to me exactly what size is meant to represent. Is it the number of gpus, the expected number of nodes in the tree?

modulus/distributed/config.py
```diff
@@ -0,0 +1,259 @@
...
+
+    Once the process group config structure (i.e, the tree structure) is set, it is
+    sufficient to set only the sizes for each leaf process group. Then, the size of
+    every parent group can be automatically computed as the product reduction of the
+    sub-tree of that parent group node.
+
+    Examples
+    --------
+    >>> from modulus.distributed.config import ProcessGroupNode, ProcessGroupConfig
+    >>> mp = ProcessGroupNode("model_parallel", orthogonal_group="data_parallel")
```
dallasfoster, line 96: typo?

modulus/distributed/config.py
```diff
@@ -0,0 +1,259 @@
...
+    >>>
+    >>> # Create model parallel group with data parallel as the orthogonal group
+    >>> mp = ProcessGroupNode("model_parallel", orthogonal_group="data_parallel")
```
dallasfoster, line 99: Can only the root of the tree have an `orthogonal_group` or can any of the leafs also have their own orthogonal group? What would be the use case for this?

modulus/distributed/config.py
```diff
@@ -0,0 +1,259 @@
...
+    >>>
+    >>> # Create the process group config with the highest level process group
+    >>> config = ProcessGroupConfig(mp)
+    >>>
+    >>> # Create spatial and channel parallel sub-groups
+    >>> config.add_node(ProcessGroupNode("spatial_parallel"), parent=mp)
```
dallasfoster, line 105: Can we add nodes to the `orthogonal_group`?

modulus/distributed/config.py
```diff
@@ -0,0 +1,259 @@
...
+    >>> config.add_node(ProcessGroupNode("channel_parallel"), parent="model_parallel")
+    >>>
+    >>> pg_config.leaf_groups()
+    ['spatial_parallel', 'channel_parallel']
+    >>>
+    >>> # Set leaf group sizes
+    >>> group_sizes = {"channel_parallel": 3, "spatial_parallel": 2}
+    >>> pg_config.set_leaf_group_sizes(group_sizes)  # Update all parent group sizes too
+    >>> pg_config.get_node("model_parallel").size
```
dallasfoster, line 114: Is `size` informative enough of an attribute name?

dallasfoster (commented):
The only thing that is not quite clear is what the purpose and algebra of the `orthogonal_group` is. There doesn't appear to be anything in this MR that sets relationships or interactions with any orthogonal group. Can I have more than 1 orthogonal group?

modulus/distributed/config.py
akshaysubr: This file has utilities to describe the configuration of process groups, so I thought it made sense to have it be in `modulus.distributed.config`. Open to suggestions for other/better names.

modulus/distributed/config.py
akshaysubr: No, this is correct. What seems wrong here?

dallasfoster: It seemed like you repeated this line 96 at line 99.

modulus/distributed/config.py
dallasfoster: I would have thought that `modulus.distributed.process_group` might be more apt. My instinct is that `modulus.distributed.config` would configure the distributed manager instead of the process group (although I realize that the process group is a component of configuring the manager).

akshaysubr:
@dallasfoster The main purpose of having an orthogonal group is to perform communication/reductions along the orthogonal dimension of a process group. For example, if you have 32 GPUs and a model parallel size of 8, then you have 4 sets of these model parallel groups that you do data parallel training over. Now when you want to reduce the gradients, you would want to only do an `all_reduce` along all GPUs with constant model parallel rank (8 independent all reductions across 4 GPUs each). This is basically doing that `all_reduce` along the orthogonal process group to the `model_parallel` process group, which we're calling the `data_parallel` group. That's what is in the last line of the code snippet in the PR description:
```python
model = DDP(model, process_group=manager.group("data_parallel"))
```

Hope this helps explain why we need orthogonal process groups. There might be other instances where a model needs to perform reductions in a dimension that is orthogonal to a specific group. I don't know of any specific use case for this though, so there is less of a justification for having orthogonal groups for non-root groups maybe.

modulus/distributed/config.py
akshaysubr: Is there any ambiguity in what `ProcessGroupNode.size` would mean?

modulus/distributed/config.py
akshaysubr: In the current design, no. Is there a case where this might be needed? We can structure the tree as a binary tree with the right node always being the orthogonal of the left but that has less than ideal usability and is also invariant under certain permutations making it a bit harder to deal with from both a user and library perspective.

modulus/distributed/config.py
akshaysubr: This is a single process group node, so size refers to just the number of processes in this process group. I'll make that more explicit in the docstring.

modulus/distributed/config.py
akshaysubr: Oh, you're right! Good catch, I'll remove this

modulus/distributed/config.py
NickGeneva: +1 to Dallas' idea, don't have to change the file name but can we add some of these classes to the __init__.py so they can get imported right next to the Distributed manager?

(maybe just call the file groups.py or something? idk)

modulus/distributed/config.py
akshaysubr: Any node can have an orthogonal group. I tried to outline a use case for this in this comment: https://github.com/NVIDIA/modulus/pull/153#issuecomment-1718405515

pyproject.toml
```diff
@@ -23,7 +23,8 @@ dependencies = [
     "torch>=1.12",
     "xarray>=2023.1.0",
     "s3fs>=2023.5.0",
-    "scikit-learn>=1.0.2"
+    "scikit-learn>=1.0.2",
+    "treelib>=1.2.5"
```
akshaysubr, line 27: Flagging the addition of a new dependency here.

modulus/distributed/config.py
akshaysubr: Good point about adding to `__init__.py`. `modulus.distributed.process_group` sounds good to me. I'll change the name to that.

pyproject.toml
NickGeneva: Apache 2.0 license

modulus/distributed/config.py
```diff
@@ -0,0 +1,259 @@
...
+
+        Returns
+        -------
+        str
+            String representation of the process group node
+        """
+        return self.__str__()
+
+
+class ProcessGroupConfig:
```
NickGeneva, line 83: I really like this Akshay, I'm mainly curious of how the distributed manager will know what groups to stick on a internode vs intranode?

For example in the example you gave theres three groups ("channel_parallel", "spatial_parallel", and "data_parallel"). How would we know to stick the first two on the same node?

modulus/distributed/config.py
NickGeneva: Is there another use for the orthogonal group other than DDP?

modulus/distributed/config.py
NickGeneva, line 83: Another just thought experiment, there's nothing else that would be needed for pipeline parallelism as far the the group structure. Its just how the model is getting executed and making it aware of what micro batch its processing correct?

modulus/distributed/config.py
```diff
@@ -0,0 +1,259 @@
...
+        name : str
+            Name of the process group
+        size : Optional[int]
+            Optional, size of the process group
+        orthogonal_group : Optional[str]
+            Optional, name of an orthogonal process group to create
+        """
+        self.name = name
+        self.size = size
+        self.orthogonal_group = orthogonal_group
```
NickGeneva, line 53: Any instance where you would want more than one ortho group? (cant think of a case but figured id ask)

modulus/distributed/config.py
```diff
@@ -0,0 +1,259 @@
...
+            Note, it is generally recommended to always set the orthogonal_group for
+            the 'model_parallel' group to be 'data_parallel' to aid with distributed
+            data parallel training
+        """
+        self.root = node
+        self.root_id = node.name
+        self.tree = Tree()
+        self.tree.create_node(node.name, node.name, data=node)
+
+    def add_node(self, node: ProcessGroupNode, parent=Union[str, ProcessGroupNode]):
```
NickGeneva, line 135: What happens if the user gives a parent node id that isnt in the tree?

modulus/distributed/config.py
```diff
@@ -0,0 +1,259 @@
...
+    def leaf_groups(self) -> List[str]:
+        """
+        Get a list of all leaf group names
+
+        Returns
+        -------
+        List[str]
+            List of all leaf node names
+        """
+        # return find_leaf_nodes(self.tree, self.root_id)
```
NickGeneva, line 193: left over or intentional?

modulus/distributed/config.py
```diff
@@ -0,0 +1,259 @@
...
+            ), f"Process group {id} is not in this process group config"
+            node = self.tree.get_node(id)
+            assert node.is_leaf(), f"Process group {id} is not a leaf group"
+            node.data.size = size
+
+        if update_parent_sizes:
+            self.update_parent_sizes()
+
+
+def tree_product_reduction(tree, node_id, verbose=False):
```
NickGeneva, line 222: this feels like an answer to a leetcode question. Have you considered dynamic programming here? Think of the efficiency! :point_up::nerd_face: (kidding)

modulus/distributed/config.py
```diff
@@ -0,0 +1,259 @@
...
+            print(
+                "Updating size of node "
+                f"{node.data.name} from {node.data.size} to {product}"
+            )
+        node.data.size = product
+
+    return product
+
+
+def find_leaf_nodes(tree, node_id):
```
NickGeneva, line 249: Is tree a `ProcessGroupConfig`? Not sure I understand this vs `def leaf_groups(self) -> List[str]:` on line 184.

dallasfoster:
> The main purpose of having an orthogonal group is to perform communication/reductions along the orthogonal dimension of a process group.

I understand the intent of the so-called "orthogonal process group", I suppose my concern/confusion is around the definition, rules, and algebra of this group as set forth in this MR because it doesn't appear to have any. I would expect some behavior or rules enforced to ensure the "orthogonality" of the passed process group to `ProcessGroupNode` but instead it just accepts a `str`, which is not entirely meaningful. Will these rules be enforced by the distributed manager? If so I would like just a sketch or outline of what they would look like.

modulus/distributed/config.py
dallasfoster: I just think the current documentation is a tautology and a user would at least be more helped by a better definition or explanation.

akshaysubr:
@dallasfoster https://github.com/NVIDIA/modulus/pull/188 adds the `DistributedManager` utility that creates a process group and it's orthogonal process group. Hopefully this makes things more clear.

There is one issue though that an orthogonal process group is orthogonal in the scope of the entire `group.WORLD` right now, not within the scope of the parent group. I'm not sure yet which one is preferable. The only real use case for using an orthogonal process group currently is DDP and both are equivalent for that.

modulus/distributed/config.py
akshaysubr: There can only be one orthogonal group, so no

modulus/distributed/config.py
akshaysubr: Currently there isn't such a way. In your example, you'd have a `model_parallel` group with `data_parallel` being the orthogonal group. `model_parallel` would have `channel_parallel` and `spatial_parallel` child groups. When we create the process groups, we first create the `model_parallel` group which would use a contiguous set of ranks. So if the total `model_parallel` group size was 8, that would be on a single node. See https://github.com/NVIDIA/modulus/pull/188 for the specifics of how this works.

But I want to emphasize that this is just the logical group config. It is up to the `DistributedManager` to actually instantiate this group config and that's where this kind of topology optimization needs to come in. Right now it is this hard coded topology mapping, but there isn't a reason we cannot add the ability for a user to specify a custom topology mapping when passing this config to the `DistributedManager`

akshaysubr: Yeah, the group structure required for pipeline parallelism is expressible through this group config. The simplest pipeline parallelism really only needs a single `model_parallel` group. See the example here: https://pytorch.org/docs/stable/pipeline.html#pipe-apis-in-pytorch

modulus/distributed/config.py
akshaysubr: Currently no. Just DDP. But internally, https://github.com/NVIDIA/modulus/pull/188 uses orthogonal groups to properly create all child groups of a parent process group.

modulus/distributed/config.py
akshaysubr: @dallasfoster https://github.com/NVIDIA/modulus/pull/188/files#diff-953c2b18aa629b13fc90a0c3df4fd3adfcbcf7407651517ce6844668f430a5f7R524 might help explain how we might reformulate this process group config as a binary tree with the right node always being the orthogonal of the corresponding left node.

modulus/distributed/config.py
akshaysubr: It'll raise an exception in `treelib`

modulus/distributed/config.py
akshaysubr: Yeah, this is not needed. I implemented this first and then realized treelib had a method to get the leaves natively so I switched to that but forgot to remove this.

modulus/distributed/config.py
akshaysubr: This is infact dynamic programming. It's recursive though which I don't like but I can't be bothered to refactor this into an iterative version 😝

modulus/distributed/config.py
akshaysubr: Made this change

modulus/distributed/config.py
akshaysubr: Added more explicit docs for the size attribute being the number of processes in this process group

modulus/distributed/config.py
akshaysubr: Removed

modulus/distributed/config.py
akshaysubr: Added better docs for the size attribute

modulus/distributed/config.py
akshaysubr: This was not intentional. Removed

modulus/distributed/config.py
akshaysubr: Removed

modulus/distributed/process_group.py
```diff
@@ -0,0 +1,244 @@
...
+    >>> config.add_node(ProcessGroupNode("spatial_parallel"), parent=mp)
+    >>> config.add_node(ProcessGroupNode("channel_parallel"), parent="model_parallel")
+    >>>
+    >>> pg_config.leaf_groups()
+    ['spatial_parallel', 'channel_parallel']
+    >>>
+    >>> # Set leaf group sizes
+    >>> group_sizes = {"channel_parallel": 3, "spatial_parallel": 2}
+    >>> pg_config.set_leaf_group_sizes(group_sizes)  # Update all parent group sizes too
+    >>> pg_config.get_node("model_parallel").size
```
dallasfoster, line 113: Same typos as in the other PR. I assume you'll fix both in merge.

dallasfoster (approved):
Small comment, otherwise looks good to me given #188.

modulus/distributed/config.py
azrael417: I think the MR looks good but Nick is right: there needs to be some notion of what is close and what is not. So basically an order of initialization. So for example the leaf node list should be ordered, then it should be clear.

stadlmax:
What's the current status?

akshaysubr:
Since this PR and https://github.com/NVIDIA/modulus/pull/188 are so tightly coupled, consolidating these changes in https://github.com/NVIDIA/modulus/pull/188. Closing this PR.

PR #171, opened 2023-09-21
Add ruff with flake8, bandit, and isort rules

### LINKED ISSUE #118: 🚀[FEA]: Add flake8 to pre-commit hooks

New Feature

Medium

Add better style guide enforcement using [flake8](https://flake8.pycqa.org/en/latest/)

[ruff](https://github.com/astral-sh/ruff)

### LINKED ISSUE #119: 🚀[FEA]: Add isort to pre-commit hooks

New Feature

Critical (currently preventing usage)

Use [isort](https://pycqa.github.io/isort/) to sort imports

[ruff](https://beta.ruff.rs/docs/)

### LINKED ISSUE #120: 🚀[FEA]: Add bandit to pre-commit hooks

New Feature

Medium

Add [bandit](https://github.com/PyCQA/bandit) hook for security scanning

[flake8-bandit](https://github.com/tylerwince/flake8-bandit)

[ruff](https://beta.ruff.rs/docs/rules/#flake8-bandit-s)

### LINKED ISSUE #123: 🚀[FEA]: Investigate using ruff to replace flake8, bandit, isort, etc

New Feature

Medium

[ruff](https://github.com/astral-sh/ruff) can do almost everything that flake8, bandit, isort, etc. can do and is supposed to be a much faster alternative as well. Moving to using ruff would mean we only need ruff in addition to black for formatting and mypy for type checking simplifying our pre-commit hooks while also making CI faster.

dallasfoster:
# Modulus Pull Request

## Description
Adds ruff to the pre-commit hooks with rules that include flake8 linting style, bandit security checks, and isort fixes. See `ruff.toml` for the configuration.

There are a number of choices that are open for debate:
1. Selection of ["E", "F", "S", "I"] rules
2. isort rules are fixable.
3. Do not enforce line length violations (assumes we are keeping black), this is temporary
5. import violations do not apply to `__init__.py` files or to `docs`.
6. Assertion violations do not apply to `test`.

(Have to list like this for Github to recognize it closes all of these)
Closes #118 
Closes #119 
Closes #120 
Closes #123

## Checklist


Makefile
```diff
@@ -21,6 +21,9 @@ interrogate:
 lint:
 	pre-commit run markdownlint -a
 
+ruff:
```
NickGeneva, line 24: Can we call ruff a linter? The make file options correspond to CI stages. We can make a ruff stage is we want but then all repos using the CI pipline would need to add ruff in their make file.

ruff.toml
```diff
@@ -0,0 +1,17 @@
+# Enable flake8-bandit (`S`) and isort (`I`) rules.
+select = ["E", "F", "S", "I"]
```
NickGeneva, line 2: Can you quicky just doc that "E" is flake8's error and "F" is pyflakes

test/deploy/test_onnx_fft.py
```diff
@@ -299,27 +302,6 @@ def forward(self, x):
     assert torch.allclose(output.imag, output_onnx_imag, rtol, atol)
 
 
-@check_ort_version()
```
NickGeneva, line 302: Why did this get removed?

NickGeneva (changes requested):

Makefile
NickGeneva: * If you think ruff really needs to be a seperate stage, then we totally can. Just a little more work (although I would like to make the stages more generic such as black -> format and interogate -> doctest  or something

dallasfoster: I was wondering about this. I think we can call ruff a linter and add it to the CI stage.

test/deploy/test_onnx_fft.py
dallasfoster: This is a duplicate to the test starting on L256.

ruff.toml
dallasfoster: Added.

test/deploy/test_onnx_fft.py
NickGeneva: Ah, good catch!

NickGeneva (approved):
If the pipeline can clear with ruff, looks good to me. :)

akshaysubr (changes requested):

ruff.toml
```diff
@@ -0,0 +1,17 @@
+# Enable flake8/pycodestyle (`E`), Pyflakes (`F`), flake8-bandit (`S`), and isort (`I`) rules.
```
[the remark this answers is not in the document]
NickGeneva: Forgot about this, good call

ruff.toml
```diff
@@ -0,0 +1,17 @@
...
+select = ["E", "F", "S", "I"]
```
akshaysubr, line 2: We should also add `"PERF"` maybe

ruff.toml
```diff
@@ -0,0 +1,17 @@
...
+fixable = ["I"]
+
+# Never enforce `E402`, `E501` (line length violations),
```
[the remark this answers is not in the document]
dallasfoster: So the main issue is that black does not apply line-length fixes to comments while flake8/ruff will flake comments that exceed this length. 

The thought is also that black will fix the line length violations anyways since it will be run before flake8/ruff.

ruff.toml
```diff
@@ -0,0 +1,17 @@
...
+# and `S311` (random number generators)
+ignore = ["E402", "E501", "S311"]
```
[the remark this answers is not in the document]
dallasfoster: In principle, yes. Inclusion of this error throws flags on a few files (pix2pix.py, super_res_net.py) that have  `# ignore_header_test` flags at the top of the file. We can individually ignore this rule for those files but it seems sort of hacky. @NickGeneva ?

ktangsali:
Does this need an addition to `pyproject.toml` as well? Especially to the `[dev]` section. Basically let's add any required packages there too to enable non pre-commit workflows. (The `[dev]` section will show up after you rebase :))

ruff.toml
dallasfoster: Added!

ruff.toml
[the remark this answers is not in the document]
NickGeneva: I know theres an import mid file for distributed comms too that would error with 402... case by case ignore is fine when its just a few outliers.

Wouldn't any try except statements trigger this or no?
E.g. given that we have so many opt deps, if we want to include more meaningful error messages for people.

If it does for this, then I would suggest ignoring the issue entirely

ruff.toml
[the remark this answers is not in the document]
NickGeneva: If its just comments that will get picked up in addition then I think this could be okay :thinking: ? It would be nice to have them both compatible.

Granted imo line length is an opinion for the formatter and should be left to the formatter.

But Akshay has some good points about development environments.

dallasfoster:
Need to add changes to the contribution guidelines.

ruff.toml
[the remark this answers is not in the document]
akshaysubr: Decided that it would be best to leave these checks out of this PR but add them back in as part of a new PR. Created an issue to track this: https://github.com/NVIDIA/modulus/issues/172

akshaysubr (approved):

ruff.toml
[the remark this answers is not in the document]
dallasfoster: So I lifted this ignore and put in a few file `ruff: noqa: E402` for the handful of places that had issues. There are a number of `test/` files that follow this pattern:
```python
script_path = os.path.abspath(__file__)
sys.path.append(os.path.join(os.path.dirname(script_path), ".."))

import common
```

Any thoughts?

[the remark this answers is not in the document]
akshaysubr: My initial impression is that those tests can just be one directory up and have all the utilities for the tests inside model specific directory. If that's not acceptable, then the imports should use relative imports instead of changing the python path: https://docs.python.org/2/whatsnew/2.5.html#pep-328-absolute-and-relative-imports

Here's a test with relative imports:
```zsh
❯ tree
.
├── common
│   ├── __init__.py
│   └── utils.py
├── __init__.py
└── test_sfno
    ├── __init__.py
    └── test_sfno.py
```

Contents of `test_sfno.py`
```python
from ..common import utils


def test_value():
    print(f"value in module common = {utils.value}")
    assert utils.value == 1
```

pytest output:
```zsh
❯ pytest
==================================================================================================================================================== test session starts ====================================================================================================================================================
platform linux -- Python 3.10.12, pytest-7.4.2, pluggy-1.3.0
rootdir: /home/akshays/Codes/legate/import_test
plugins: anyio-3.7.1
collected 1 item                                                                                                                                                                                                                                                                                                            

test_sfno/test_sfno.py .                                                                                                                                                                                                                                                                                              [100%]

===================================================================================================================================================== 1 passed in 0.01s =====================================================================================================================================================
```

pyproject.toml
```diff
@@ -51,10 +51,30 @@ dev = [
     "black==22.10.0",
     "interrogate==1.5.0",
     "coverage==6.5.0",
+    "ruff==0.0.290",
```
akshaysubr, line 54: Why the strict versioning here? Can this not be `>=`?

pyproject.toml
```diff
@@ -51,10 +51,30 @@ dev = [
...
+# isort (`I`), and performance 'PERF' rules.
+select = ["E", "F", "S", "I", "PERF"]
+fixable = ["I"]
+
+# Never enforce `E402`, `E501` (line length violations),
+# and `S311` (random number generators)
+ignore = ["E501", "S311"]
+
+[tool.ruff.per-file-ignores]
+# Ignore `F401` (import violations) in all `__init__.py` files, and in `docs/*.py`.
```
akshaysubr, line 74: These comments are extremely useful! Thanks!

test/ci_tests/header_check.py
```diff
@@ -43,9 +41,7 @@ def get_top_comments(_data):
         elif "from" in line:
             break
 
-    comments = []
-    for line in lines_to_extract:
-        comments.append(_data[line])
+    comments = [_data[line] for line in lines_to_extract]
```
akshaysubr, line 44: Was this to fix a `PERF` error?

akshaysubr (approved):

pyproject.toml
NickGeneva: Does ruff have backwards compatibility?

Its could also a similar issue with black in the sense, different versions may have slight differences but both pass which can cause odd git history

test/ci_tests/header_check.py
dallasfoster: Yes, one of the `PERF` errors is about list comprehension.

pyproject.toml
dallasfoster: Good point, although I don't know how stable code is with `0.0.~`. Changed.

ruff.toml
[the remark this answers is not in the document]
dallasfoster: Turns out there is not a clean way to do this for now since the @akshaysubr 's above solution introduces an `E999` flag with the linter. I tried out another solution to no avail. For now, I will keep the present solution.

NickGeneva (approved):

PR #174, opened 2023-09-28
ARM container numcodecs fix

ktangsali:
# Modulus Pull Request

## Description
Workaround fix for #175 

## Checklist


Dockerfile
```diff
@@ -38,20 +38,36 @@ RUN pip install "tensorly>=0.8.1" https://github.com/tensorly/torch/archive/715a
...
+        pip install --force-reinstall /modulus/deps/numcodecs-0.11.0-cp310-cp310-linux_aarch64.whl; \
+    else \
+        echo "Numcodecs wheel for $TARGETPLATFORM is not present, attempting to build from pip, but might fail" && \
+	pip install numcodecs; \
+    fi
+
 # install vtk and pyvista
-RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
-	echo "Installing vtk and pyvista for: $TARGETPLATFORM" && \
+RUN if [ "$TARGETPLATFORM" = "linux/arm64" ] && [ -e "/modulus/deps/vtk-9.2.6.dev0-cp310-cp310-linux_aarch64.whl" ]; then \
```
akshaysubr, line 55: Where does one get this wheel from and can this path be a dockerfile arg that can be overridden?

akshaysubr (approved):
This PR looks good overall. But maybe we should think about how we can make Dockerfile changes like this and test container builds it in an environment that mimics an end user's and doesn't have all these custom wheels available.

ktangsali:
> This PR looks good overall. But maybe we should think about how we can make Dockerfile changes like this and test container builds it in an environment that mimics an end user's and doesn't have all these custom wheels available.

Thanks @akshaysubr , you are right. I have done such testing for this and the previous PR. i.e. these builds work, even when these pre-built wheels are not available as it just reverts to source build in that case. The motivation to add these checks for pre-built wheels was to speed up internal builds. We need to do them nightly and having things build from source every time slows things down. 

That being said, you brought up an interesting point. I have been discussing this with @NickGeneva recently. We need to have a robust source build instructions that mimic the docker build behavior. A lot of projects have a third_party folder to track such dependencies and we should consider adding such to Modulus. Prepared this issue to track this further: https://github.com/NVIDIA/modulus/issues/176

Dockerfile
ktangsali: Referring to my comment here: https://github.com/NVIDIA/modulus/pull/174#issuecomment-1739820835, just want to highlight that these pre-built wheels are not a necessity. In case they are unavailable, the build will attempt to install from source. I have tested this on an arm and x86 system. The motivation to add these pre-built wheels is primarily to speed-up internal builds.

NickGeneva (approved):

PR #187, opened 2023-10-06
Restore contents that were previously in LFS

ktangsali:
# Modulus Pull Request

## Description

This PR restores the files from Git LFS. Git LFS makes cloning and working with repos difficult. After this PR, an attempt should be made to limit the sizes of the files that are being committed to this repository to keep the repo size manageable. This is also now added as a pre-commit hook/CI check. To allow for current files, the limit has been set to 5 mB, but an attempt should be made to reduce this limit further. 

## Checklist

- [] An [issue](https://github.com/NVIDIA/modulus/issues) is linked to this pull request.

Closes: https://github.com/NVIDIA/modulus/issues/131


Makefile
```diff
@@ -20,7 +20,8 @@ interrogate:
 
 lint:
 	pre-commit run markdownlint -a && \
-	pre-commit run ruff -a
```
dallasfoster, line 23: Not sure there is a better place, but I figure I will open the discussion: Is `lint` the right place to put the large file check?

dallasfoster (approved):

Makefile
ktangsali: Created one here: https://github.com/NVIDIA/modulus/issues/190

NickGeneva: I agree lint is an odd stage for this but there's not a better spot at this moment... updating stages breaks all repos so this will have to be a coordinated effort to update when ever that happens

NickGeneva (commented):
Is the plan to eventually scrub the repo of some of these files? I can imagine we can easily update the CI unit tests to use the data repo eventually.

NickGeneva (approved):

.pre-commit-config.yaml
```diff
@@ -49,3 +49,9 @@ repos:
     hooks:
     - id: ruff
       args: [--fix]
+
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.4.0
+    hooks:
+    - id: check-added-large-files
+      args: [--maxkb=5000]
```
NickGeneva, line 57: Whats the lowest this can go right now?

ktangsali: Maybe like 3-4 MB? Currently also wanted to keep a nice round number.

ktangsali:
> Is the plan to eventually scrub the repo of some of these files? I can imagine we can easily update the CI unit tests to use the data repo eventually.

I certainly think we should aim to move in that direction.

Issue #191, opened 2023-10-11
🚀[FEA]: Modulus Launch and Core integration


ktangsali:
### Is this a new feature, an improvement, or a change to existing functionality?

Change

### How would you describe the priority of this feature request

Medium

### Please provide a clear description of problem you would like to solve.

This is a proposal to bring the utilities from Modulus Launch into Modulus core and deprecate a separate pip wheel and dockerfile for the Modulus-Launch repo. The utilites in Modulus launch have matured enough to be widely applicable across different data-driven examples and can be integrated into the core Modulus package. The additional dependencies of Modulus Launch will be handled via the route of optional dependencies, i.e. with a `[launch]` option in the `.toml` file. From a user standpoint, the APIs/import statements are not affected as namespace packaging is adopted across the Modulus repos.

NickGeneva:
Done in linked PR. Closing

PR #192, opened 2023-10-11
Launch core integration part1

ktangsali:
# Modulus Pull Request

## Description
Brings the utilities from Launch preserving history. Also makes relevant changes to the Pyproject.toml and Dockerfile. 

Addresses #191 

## Checklist


Dockerfile
```diff
@@ -29,6 +29,8 @@ ENV _CUDA_COMPAT_TIMEOUT=90
 
 # Install other dependencies
 RUN pip install "h5py>=3.7.0" "mpi4py>=3.1.4" "netcdf4>=1.6.3" "ruamel.yaml>=0.17.22" "scikit-learn>=1.0.2" 
+RUN pip install "hydra-core>=1.2.0" "termcolor>=2.1.1" "wandb>=0.13.7" "mlflow>=2.1.1" "pydantic>=1.10.2" "imageio>=2.28.1" "moviepy>=1.0.3" "tqdm>=4.60.0"
```
akshaysubr, line 32: Why here instead of in `pyproject.toml`?

pyproject.toml
```diff
@@ -30,6 +30,17 @@ classifiers = [
...
+    "termcolor>=2.1.1",
+    "wandb>=0.13.7",
+    "mlflow>=2.1.1",
+    "pydantic>=1.10.2",
+    "imageio>=2.28.1",
+    "moviepy>=1.0.3",
+    "tqdm>=4.60.0",
+]
+
 all = [
```
akshaysubr, line 44: Should ideally use self-referential dependencies here instead of re-listing everything:
https://github.com/pypa/hatch/blob/master/docs/config/dependency.md#self-referential

akshaysubr (changes requested):

Dockerfile
ktangsali: In the builder stage, we want to only install the dependencies and not the source (because of the CI image referencing the builder). Unfortunately, this means that this install will be repeated in toml and dockerfile. Ideally, pip should support a `--only-deps` option, but that is still being considered (Ref: https://www.github.com/pypa/pip/issues/11440). The alternatives AFAIK are writing a separate requirements.txt either ourselves or through some pip-tools. I don't think it is a very good alternative, so I have been wanting to revisit this once the PyPi team implemented that feature. That being said, I feel like at the cost of having to maintain dependencies at two places, it is still worth doing it this way, because shipping docker image places a lot of constraints on what software can go in due to license issues etc. Consider tensorly-torch for example, that package's latest wheel is not compliant with our container requirements. We have a patch pushed, but the wheel is not public yet, so we have to manually install it inside our docker container following the process currently implemented.

pyproject.toml
ktangsali: Great idea! Implemented it

Dockerfile
akshaysubr: The `--only-deps` option is nice. I agree, can wait for that to remove this duplication. But maybe we should add a test for a source only install in a vanilla cuda container in the interim?

pyproject.toml
```diff
@@ -43,6 +54,7 @@ all = [
     "pyvista>=0.40.1",
     "torch-harmonics>=0.6.2",
     "onnx>=1.14.0",
+    "nvidia-modulus[launch]"
```
akshaysubr, line 57: What are the deps above this? Those are docker deps? Also, why no [dev] in [all]?

akshaysubr (approved):
Approving. Added another minor comment about dependency management.

Dockerfile
ktangsali: Yes certainly! I think we should make this part of the QA - i.e. test the pip/source installs in a conda env or in a vanilla cuda container as you suggest. Thanks @akshaysubr, marking this as resolved now.

pyproject.toml
ktangsali: Thanks! added [dev] to [all]. Also, yes the other dependencies are to get the full functionality of the Core (basically docker deps). We added the instructions for them here: https://docs.nvidia.com/deeplearning/modulus/getting-started/index.html (under the pip section)

NickGeneva (approved):

PR #194, opened 2023-10-17
Add routines to compute TISR

nbren12:
# Modulus Pull Request

## Description

Google graph-cast is trained with TOA insolation data from ERA5, but it is not convenient for inference to rely on downloading this data. For example, where should we get this data for real-time forecasts? There is no ERA5 for the future!

This function adds a routine `modulus.utils.zenith_angle.toa_incident_solar_radiation_accumulated` to help compute this variable. It is not identical to the ERA5 output, but the error is quite small  (0.1%). I'm not sure exactly how IFS computes it. See these plots.
tisr

line

Note the implementation is more complex than cos zenith since the irradiance varies over the year by about 6% (more radiation when the earth is closer to the sun in its elliptic orbit). Also TISR is accumulated over the last hour rather than an instantaneous value.

I also fixed some time zone handling in cos zenith

## Checklist


test/utils/test_zenith_angle.py
```diff
@@ -61,3 +66,118 @@ def test_zenith_angle_array():
...
+
+
+def test_toa():
+    t = datetime(2000, 7, 1, 12, 0, 0, tzinfo=utc).timestamp()
+    lat, lon = 0.0, 0.0
+    ans = toa_incident_solar_radiation_accumulated(t, lat, lon)
+    assert ans >= 0
+
+
+def test_tisr_matches_cds():
```
nbren12, line 93: This was helpful for validating against IFS, but I can remove it from the PR if we are comfortable with the skill.

modulus/utils/zenith_angle.py
```diff
@@ -101,6 +105,174 @@ def cos_zenith_angle_from_timestamp(
...
+    """
+    seconds_per_solar_day = 86400
+    mean_tropical_year = mean_tropical_year * seconds_per_solar_day
+
+    year_2000_equinox = datetime.datetime(2000, 3, 20, 7, 35, tzinfo=pytz.utc)
+
+    # from appendix of Berger 1978
+    lam = (t - year_2000_equinox.timestamp()) % mean_tropical_year
+    lam = lam / mean_tropical_year * 2 * np.pi
+    v = lam - perihelion_longitude * np.pi / 180
```
akshaysubr, line 156: This is the mean anomaly, isn't it? But the next line computes `rho` (which ideally should be renamed to `rho_inv`) but the formula should have the true anomaly which is very slightly different from the mean anomaly for small eccentricities like it is for Earth. See section 6.8 here: https://link.springer.com/book/10.1007/978-3-662-53045-0

Standard way is to go from the mean anomaly to the eccentric anomaly and then to the true anomaly, but this doesn't have a closed form solution so would need a nonlinear solver.

akshaysubr (changes requested):
There seems to be an error in computing the irradiance arising from the fact that the true anomaly does not change linearly with time. See comment below for more details.

modulus/utils/zenith_angle.py
nbren12: Thanks. It would be nice to get the plot to line up exactly. Sec 3 of [Berger](https://journals.ametsoc.org/view/journals/atsc/35/12/1520-0469_1978_035_2362_ltvodi_2_0_co_2.xml?tab_body=pdf) also describes some iterative procedure, but I skimmed over that part.

nbren12:
I improved the calculation somewhat with @akshaysubr's guidance. Still not a perfect match, but the line plots overlay a bit better.

line
tisr

nbren12:
ready for re-review. Would someone be so kind as to run the CI?

pyproject.toml
```diff
@@ -21,6 +21,7 @@ dependencies = [
     "nvidia_dali_cuda110>=1.16.0",
     "setuptools>=67.6.0",
     "certifi>=2023.7.22",
+    "pytz>=2023.3",
```
akshaysubr, line 24: MIT license, so should be good: https://pythonhosted.org/pytz/#license

test/utils/test_zenith_angle.py
```diff
@@ -61,3 +66,122 @@ def test_zenith_angle_array():
...
+
+            if i == 0:
+
+                a.set_title("ERA5")
+                b.set_title("Computed")
+                c.set_title("Difference")
+                fig.suptitle("Solar irradiance (W/m^2)")
+
+    plot_comparison(ds.sel(time="2018"))
+    plt.savefig("tisr.png")
```
akshaysubr, line 163: Would be good to convert this into a test that checks for accuracy up to a certain threshold. Can keep the plotting code here for future use, but branch it off using a verbosity flag. Something like https://github.com/NVIDIA/modulus/blob/main/test/distributed/test_distributed_fft.py#L79

test/utils/test_zenith_angle.py
```diff
@@ -61,3 +66,122 @@ def test_zenith_angle_array():
...
+
+
+@pytest.skip(
+    "This test requires CDS API and manual review of the tisr.png "
+    "and line .png. Consider running if modifying the tisr code."
+)
+def test_tisr_matches_cds():
+    import os
+
+    import cdsapi
```
akshaysubr, line 100: We probably don't want to add the cdsapi as a dependency here? Would be good if there is a way around it. Maybe use the same mechanism as in https://github.com/NVIDIA/modulus/blob/main/test/datapipes/test_climate_hdf5.py?

akshaysubr: Actually, I just checked and the cdsapi is apache 2.0, so we should be okay to add it as a `[dev]` optional dependency.

Tagging @NickGeneva to comment about any security vulnerabilities with the cdsapi.

test/utils/test_zenith_angle.py
```diff
@@ -61,3 +66,122 @@ def test_zenith_angle_array():
...
+    import matplotlib.pyplot as plt
```
akshaysubr, line 101: matplotlib also not in `[dev]` dependencies currently.

modulus/utils/zenith_angle.py
```diff
@@ -101,6 +105,197 @@ def cos_zenith_angle_from_timestamp(
...
+    # E - e sin(E) = M
+    def f(E):
+        return E - e * np.sin(E) - M
+
+    def fp(E):
+        return 1 - e * np.cos(E)
+
+    # newton iterations
+    # initial guess
+    E = M / (1 - e)
```
akshaysubr, line 177: Optional, but might be an enhancement:

This initial condition is a good one in the limit `E -> 0`, but a better asymptotic expansion is to do `(E-M) -> 0`. If you do that, the first order solution ends up being

```math
E = M + \frac{e \sin M}{1 - e \cos M}
```

And it is satisfying that in the limit that `M -> 0`, you recover the initial guess that you have 😄

akshaysubr (commented):
The update looks good to me and seems like the error reduced to half of what it was before. The only other concern I had was with some of the packages required for the test not being reflected in `pyproject.toml`.

test/utils/test_zenith_angle.py
nbren12: this is why i skipped the test.

test/utils/test_zenith_angle.py
nbren12: this is why i skipped the test.

nbren12:
@akshaysubr let me know what I should do with the test deps. I skipped the test as a workaround. If this is not good enough, I would be inclined to delete the test rather then touch the project-wide deps.

test/utils/test_zenith_angle.py
nbren12: removed the test.

test/utils/test_zenith_angle.py
nbren12: removed the test.

nbren12:
@akshaysubr updated the PR. hopefully the tests pass now.

akshaysubr:
@nbren12 Deleting the test is good for now. Don't think we want to keep this PR waiting to figure out the right solution for that. For now, it would help to add a `# pragma: no cover` comment to exclude untested code from the coverage report to get CI to pass. See https://github.com/NVIDIA/modulus/blob/main/modulus/distributed/autograd.py#L53

Can you also open an issue to track the testing blockers?

akshaysubr (approved):

NickGeneva (dismissed):

nbren12:
@akshaysubr There is another test which ensures the output is positive. This will at least satisfy the checks.

modulus/utils/zenith_angle.py
```diff
@@ -101,6 +105,197 @@ def cos_zenith_angle_from_timestamp(
...
+    mean_tropical_year=365.2422,
+    newton_iterations: int = 3,
+):
+    """The flux of solar energy in W/m2 towards Earth
+
+    The default orbital parameters are set to 2000 values.
+    Over the period of 1900-2100 this will result in an error of at most 0.02%,
+    so can be neglected for many applications.
+
+    Args:
```
NickGeneva, line 122: These doc strings are great!

Can the sections quickly get switched to Numpy styling?

```
Parameters
---------

Notes
-----
```

NickGeneva (changes requested):

NickGeneva (approved):
Approving so youre not blocked if you update the doc string styling.

modulus/utils/zenith_angle.py
nbren12: done

nbren12:
@NickGeneva @akshaysubr I updated the docstrings. should be ready to go.

nbren12:
looks like the tests finally ran.

PR #216, opened 2023-11-02
Adding NGC model file support to filesystem

NickGeneva:
# Modulus Pull Request

## Description

Adding NGC model file protocol for filesystem in Modulus. Allows users to fetch / unzip files on NGC model registry both public and private.

Users can now use `ngc://org/team/model/version/file` to fetch files from NGC. Also works with private registries if users have set `API_KEY` to the environment variable. If its a zip file, it will automatically unzip its contents. This can be very useful for shipping resources for examples and such.

Related to: https://github.com/NVIDIA/earth2mip/issues/57

Closes: https://github.com/NVIDIA/modulus/issues/217

## Checklist

### DIFF
--- a/modulus/utils/filesystem.py
+++ b/modulus/utils/filesystem.py
@@ -102,7 +102,7 @@ def _download_ngc_model_file(path: str, out_path: str, timeout: int = 300) -> st
         total_size_in_bytes = int(r.headers.get("content-length", 0))
         chunk_size = 1024  # 1 kb
         progress_bar = tqdm(total=total_size_in_bytes, unit="iB", unit_scale=True)
-        progress_bar.set_description(f"Downloading {filename}")
+        progress_bar.set_description(f"NGC file {filename}")
         with open(local_url, "wb") as f:
             for chunk in r.iter_content(chunk_size=chunk_size):
                 progress_bar.update(len(chunk))
--- a/modulus/utils/filesystem.py
+++ b/modulus/utils/filesystem.py
@@ -102,7 +102,7 @@ def _download_ngc_model_file(path: str, out_path: str, timeout: int = 300) -> st
         total_size_in_bytes = int(r.headers.get("content-length", 0))
         chunk_size = 1024  # 1 kb
         progress_bar = tqdm(total=total_size_in_bytes, unit="iB", unit_scale=True)
-        progress_bar.set_description("Downloading NGC model file")
+        progress_bar.set_description(f"Downloading {filename}")
         with open(local_url, "wb") as f:
             for chunk in r.iter_content(chunk_size=chunk_size):
                 progress_bar.update(len(chunk))
--- a/modulus/utils/filesystem.py
+++ b/modulus/utils/filesystem.py
@@ -16,6 +16,7 @@ import hashlib
 import json
 import logging
 import os
+import re
 import urllib.request
 import zipfile
 
@@ -53,7 +54,7 @@ def _download_ngc_model_file(path: str, out_path: str, timeout: int = 300) -> st
 
     Args:
         path (str): NGC model file path of form:
-            `ngc://models/<org_id/team_id/model_id>/<version>/<path/in/repo>`
+            `ngc://models/<org_id/team_id/model_id>@<version>/<path/in/repo>`
         out_path (str): Output path to save file / folder as
         timeout (int): Time out of requests, default 5 minutes
 
@@ -65,18 +66,16 @@ def _download_ngc_model_file(path: str, out_path: str, timeout: int = 300) -> st
     """
     # Strip ngc model url prefix
     suffix = "ngc://models/"
-    if not path.startswith(suffix):
+    # The regex check
+    pattern = re.compile(f"{suffix}[\w-]+/[\w-]+/[\w-]+@[A-Za-z0-9.]+/[\w/]+")
+    if not pattern.match(path):
         raise ValueError(
-            "Invalid URL, should be of form ngc://models/<repo_id>/<version>/<path/in/repo>"
+            "Invalid URL, should be of form ngc://models/<org_id/team_id/model_id>@<version>/<path/in/repo>"
         )
+
     path = path.replace(suffix, "")
-
-    if len(path.split("/", 4)) != 5:
-        raise ValueError(
-            "Invalid URL, should be of form ngc://models/<org_id/team_id/model_id>/<version>/<path/in/repo>"
-        )
-
-    (org, team, model, version, filename) = path.split("/", 4)
+    (org, team, model_version, filename) = path.split("/", 4)
+    (model, version) = model_version.split("@", 1)
     token = ""
     # If API key environment variable
     if "NGC_API_KEY" in os.environ:
--- a/modulus/utils/filesystem.py
+++ b/modulus/utils/filesystem.py
@@ -102,7 +102,7 @@ def _download_ngc_model_file(path: str, out_path: str, timeout: int = 300) -> st
         total_size_in_bytes = int(r.headers.get("content-length", 0))
         chunk_size = 1024  # 1 kb
         progress_bar = tqdm(total=total_size_in_bytes, unit="iB", unit_scale=True)
-        progress_bar.set_description(f"NGC file {filename}")
+        progress_bar.set_description(f"Fetching {filename}")
         with open(local_url, "wb") as f:
             for chunk in r.iter_content(chunk_size=chunk_size):
                 progress_bar.update(len(chunk))
--- a/modulus/utils/filesystem.py
+++ b/modulus/utils/filesystem.py
@@ -23,6 +23,7 @@ import fsspec
 import fsspec.implementations.cached
 import requests
 import s3fs
+from tqdm import tqdm
 
 logger = logging.getLogger(__name__)
 
@@ -61,15 +62,22 @@ def _download_ngc_model_file(path: str, out_path: str, timeout: int = 300) -> st
     Returns:
         str: output file / folder path
     """
-    if len(path.split("/")) != 5 or not path.startswith("ngc://"):
+    # Strip ngc url prefix
+    if not path.startswith("ngc://"):
+        raise ValueError(
+            "Invalid URL, should be of form ngc://org/team/model/version/filename"
+        )
+    path = path[6:]
+
+    if len(path.split("/")) != 5:
         raise ValueError(
             "Invalid URL, should be of form ngc://org/team/model/version/filename"
         )
 
-    (org, team, model, version, filename) = path.lstrip("ngc://").split("/")
+    (org, team, model, version, filename) = path.split("/")
     token = ""
     # If API key environment variable
-    if os.environ["API_KEY"]:
+    if "API_KEY" in os.environ:
         try:
             session = requests.Session()
             session.auth = ("$oauthtoken", os.environ["API_KEY"])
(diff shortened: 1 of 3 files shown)


modulus/utils/filesystem.py
```diff
@@ -43,11 +46,88 @@ def _get_fs(path):
...
 
 
-def _download_cached(path: str, recursive: bool = False) -> str:
+def _download_ngc_model_file(path: str, out_path: str, timeout: int = 300) -> str:
+    """Pulls files from model registry on NGC. Supports private registries when NGC
+    API key is set the the API_KEY enviroment variable. If download file is a zip folder
+    it will get unzipped.
+
+    Args:
+        path (str): NGC model file path of form `ngc://org/team/model/version/filename`
```
akshaysubr, line 55: This is nice! An enhancement might be to align this string format with the huggingface filesystem abstraction so it is more familiar to users.
https://huggingface.co/docs/huggingface_hub/main/en/guides/hf_file_system

modulus/utils/filesystem.py
```diff
@@ -43,11 +46,88 @@ def _get_fs(path):
...
+
+    # Download file, apparently the URL for private registries is different than the public?
+    if len(token) > 0:
+        file_url = f"https://api.ngc.nvidia.com/v2/org/{org}/team/{team}/models/{model}/versions/{version}/files/{filename}"
+    else:
+        file_url = f"https://api.ngc.nvidia.com/v2/models/{org}/{team}/{model}/versions/{version}/files/{filename}"
+    local_url = f"{LOCAL_CACHE}/{filename}"
+    headers = {"Authorization": f"Bearer {token}", "Content-Type": "application/json"}
+    # Streaming here for larger files
+    with requests.get(file_url, headers=headers, stream=True, timeout=timeout) as r:
```
akshaysubr, line 100: Not for this PR, but it would be great to implement an fsspec filesystem as a child class of the [HTTPFileSystem](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.implementations.http.HTTPFileSystem) which will then allow us to use fsspec [URL chaining](https://filesystem-spec.readthedocs.io/en/latest/features.html#url-chaining). So, what we could do is to take `"ngc://org/team/model/version/filename"` and translate that to `"filecache::zip::https://api.ngc.nvidia.com/v2/models/{org}/{team}/{model}/versions/{version}/files/{filename}"` and then not have to do the unzip and caching manually 😄

modulus/utils/filesystem.py
```diff
@@ -43,11 +46,88 @@ def _get_fs(path):
...
+            r.raise_for_status()
+            token = json.loads(r.content)["token"]
+        except requests.exceptions.RequestException:
+            logger.warning("Failed to get JWT using the API set in API_KEY")
+
+    # Download file, apparently the URL for private registries is different than the public?
+    if len(token) > 0:
+        file_url = f"https://api.ngc.nvidia.com/v2/org/{org}/team/{team}/models/{model}/versions/{version}/files/{filename}"
+    else:
+        file_url = f"https://api.ngc.nvidia.com/v2/models/{org}/{team}/{model}/versions/{version}/files/{filename}"
```
akshaysubr, line 96: This only works for ngc models right? In that case, maybe the path format should specify that somehow. Maybe use `ngcmodel://...` instead of `ngc://...`? That would allow us to expand to resources or datasets down the line.

modulus/utils/filesystem.py
```diff
@@ -43,11 +46,88 @@ def _get_fs(path):
         return fsspec.filesystem("file")
 
 
-def _download_cached(path: str, recursive: bool = False) -> str:
+def _download_ngc_model_file(path: str, out_path: str, timeout: int = 300) -> str:
```
ktangsali, line 49: This is great! Now that we have a few ways to handle packages, I think we should better document them. Can we add something to the docs of the `Package` class to specify what all data sources are supported?

modulus/utils/filesystem.py
```diff
@@ -43,11 +46,88 @@ def _get_fs(path):
...
+            "Invalid URL, should be of form ngc://org/team/model/version/filename"
+        )
+
+    (org, team, model, version, filename) = path.split("/")
+    token = ""
+    # If API key environment variable
+    if "API_KEY" in os.environ:
+        try:
+            session = requests.Session()
+            session.auth = ("$oauthtoken", os.environ["API_KEY"])
```
akshaysubr, line 83: Generic env var name. Should specialize for NGC. Maybe NGC_API_KEY?

modulus/utils/filesystem.py
```diff
@@ -43,11 +46,88 @@ def _get_fs(path):
...
+            headers = {"Accept": "application/json"}
+            authn_url = f"https://authn.nvidia.com/token?service=ngc&scope=group/ngc:{org}&group/ngc:{org}/{team}"
+            r = session.get(authn_url, headers=headers, timeout=5)
+            r.raise_for_status()
+            token = json.loads(r.content)["token"]
+        except requests.exceptions.RequestException:
+            logger.warning("Failed to get JWT using the API set in API_KEY")
```
akshaysubr, line 90: `set in API_KEY` -> `set in the API_KEY environment variable`?

akshaysubr: Also, good practice to re-raise the exception.
```python
try:
    ...
except requests.exceptions.RequestException:
    logger.warning("Failed to get JWT using the API set in API_KEY")
    raise  # Will re-raise the exception with the same stack trace
```

akshaysubr (changes requested):
Very nice utility that's super useful. Had only a few QOL improvement suggestions.

test/utils/test_filesystem.py
```diff
@@ -53,3 +56,29 @@ def test_http_package():
 
     known_checksum = "e075b2836d03f7971f754354807dcdca51a7875c8297cb161557946736d1f7fc"
     assert calculate_checksum(path) == known_checksum
+
+
+@pytest.mark.skip("Skipping because slow, need better test solution")
```
ktangsali, line 61: Thanks for adding this, but not sure if we should skip. If looking for a smaller file, this [MGN checkpoint](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/modulus/models/modulus_ns_meshgraphnet) is ~50% of the DLWP checkpoint size and might be worth a try.

modulus/utils/filesystem.py
NickGeneva: Going off your other suggestion how about `ngc://model/....`? Then a dataset can be `ngc://datasets/....` like HF

(So following `hf://[<repo_type_prefix>]<repo_id>[@<revision>]/<path/in/repo>`)

modulus/utils/filesystem.py
```diff
@@ -43,11 +46,88 @@ def _get_fs(path):
...
+
+    # Download file, apparently the URL for private registries is different than the public?
```
ktangsali, line 92: Has this been confirmed? Ref: https://docs.nvidia.com/ngc/gpu-cloud/ngc-catalog-user-guide/index.html#download-models-via-wget-authenticated-access

ktangsali (approved):

modulus/utils/filesystem.py
NickGeneva: Yeah, I agree this would be really nice to do. Trouble is the NGC org -> team concept 

Also seems HF isnt fully consistent between datasets and models. I like the idea of trying to follow:
```
hf://[<repo_type_prefix>]<repo_id>[@<revision>]/<path/in/repo>
```
(repo_id is essentially org/team/name of registry)


So I'm thinking:

<repo_id> = `org/team/name of repo`
@<revision> = version
<repo_type_prefix> = 'models' or 'datasets'

modulus/utils/filesystem.py
NickGeneva: Yeah its dumb, the one used for authentication for private does not work without a token. So for public model you need to use the other URL. Don't know why, but way it is :shrug:

test/utils/test_filesystem.py
NickGeneva: Yeah I'm waiting on the diagnostic models to get uploaded, I can use climatenet checkpoint for this test which is like 2Mb download.

Will have a follow up PR when they are live.

modulus/utils/filesystem.py
NickGeneva: Sounds good, I had originally left this to get passed so if someone was actually trying to access a public file it would still make the attempt despite the auth failure.

But I agree that it should just fail here if someone has the environment variable set so its transparent.

modulus/utils/filesystem.py
NickGeneva: Agreeded

NickGeneva: (Blame NGC cli docs for this name not me :P )

modulus/utils/filesystem.py
```diff
@@ -71,6 +151,9 @@ def _download_cached(path: str, recursive: bool = False) -> str:
         if path.startswith("s3://"):
             fs = _get_fs(path)
             fs.get(path, cache_path, recursive=recursive)
+        elif path.startswith("ngc://"):
+            path = _download_ngc_model_file(path, cache_path)
```
loliverhennigh, line 155: Maybe add timeout input here and have timeout be given to `_download_cached`. Not a big deal though.

loliverhennigh (approved):

modulus/utils/filesystem.py
akshaysubr: Yeah, I like that. `<repo_type_prefix>` should probably have a `/` at the end though.

So in the test here, we would have something like this?
```
ngc://models/nvidia/modulus/modulus_dlwp_cubesphere@v0.2/dlwp_cubesphere.zip
```

modulus/utils/filesystem.py
NickGeneva: Updated!

modulus/utils/filesystem.py
NickGeneva: Updated!

modulus/utils/filesystem.py
NickGeneva: Updated

modulus/utils/filesystem.py
NickGeneva: Considered this, but then you would then have to realistically also control the fsspec timeout too and even potentially pass it up to the Package class.

Anyway, I just left it as a controllable input now so we can do this in the future if really needed. :+1:

modulus/utils/filesystem.py
NickGeneva: Implemented, clean up the valid url check too with some chatgpt regex and added unit test to check invalid urls.

loliverhennigh (approved):

modulus/utils/filesystem.py
NickGeneva: Added doc string to package.

modulus/utils/filesystem.py
NickGeneva: Once we really start relying on NGC registry for resources I'm all for this.

akshaysubr: Let me make an issue for this. I'd love to implement this when there is some time. It can help with a bunch of use cases potentially!

test/utils/test_filesystem.py
```diff
@@ -53,3 +56,63 @@ def test_http_package():
...
+    known_checksum = "d2a84f4b8b650937ec8f73cd8be2c74add5a911ba64df27458ed8229da804a26"
+    assert calculate_checksum(path) == known_checksum
+
+
+def test_ngc_model_file_invalid():
+    test_url = "ngc://models/nvidia/modulus/modulus_dlwp_cubesphere/v0.2"
+    package = filesystem.Package(test_url, seperator="/")
+    try:
+        package.get("dlwp_cubesphere.zip")
+        raise Exception("Processed invalid model url")
```
akshaysubr, line 92: This is better done with [`pytest.raises`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.raises)

akshaysubr (approved):

test/utils/test_filesystem.py
NickGeneva: Done

PR #252, opened 2023-11-23
ClimateDatapipe with extended functionality for climate datasets

jleinonen:
# Modulus Pull Request

## Description
Adds new datapipe `modulus.datapipes.climate.ClimateDatapipe`, based on `modulus.experimental.datapipes.climate.ClimateHDF5Datapipe`, which has been moved, renamed and updated with changes.

**New features**:
- Debugged, cleaned up and tested the experimental datapipe.
- Support for reading data from multiple sources at the same time. This is usually needed for training diagnostic models, for instance. To enable this, each source is configured by creating a `modulus.datapipes.climate.ClimateDataSourceSpec` object. The list of the sources is then passed to the `ClimateDatapipe` constructor.
- Support for NetCDF4 formatted datasets.
- Support for outputting:
    - Cosine of solar zenith angle
    - Latitude/longitude and their sin/cos 
    - Surface geopotential (if supplied with NetCDF file containing it)
    - Land-sea mask (if supplied with NetCDF file containing it)

**To be resolved with devs before merge**:
- Test coverage; the existing test file has been modified to support the new way of creating the datapipe but I haven't been able to test in the exact environment used by the devs. A dataset metadata file might have to be added to the test environment.
- Update documentation: docstrings have been updated but documentation should be checked for completeness.
- Check that naming of new components is appropriate.

## Checklist

## Dependencies

No new dependencies needed.


modulus/datapipes/climate/climate.py
```diff
@@ -0,0 +1,829 @@
...
+    variables: Union[List[str], None], optional for HDF5 files, mandatory for NetCDF4 files
+        List of named variables to load. Variables will be read in the order specified
+        by this parameter. Must be used for NetCDF4 files. Supported for HDF5 files
+        in which case it will override `channels`.
+    use_cos_zenith: bool, optional
+        If True, the cosine zenith angles corresponding to the coordinates of this
+        data source will be produced, default False
+    num_steps : int, optional
+        Number of timesteps to return, by default 1
+    num_steps : int, optional
```
YichengDWu, line 102: This parameter is repeated here.

modulus/datapipes/climate/climate.py
```diff
@@ -0,0 +1,829 @@
...
+      provided.
+    - `geopotential`: Tensor of shape (batch_size, 1, height, width), containing
+      geopotential if a path to a geopotential file is provided.
+    - `latlon`: Tensor of shape (batch_size, 2, height, width), containing
+      latitude and longitude meshgrid if specified.
+    - `cos_latlon`: Tensor of shape (batch_size, 3, height, width), containing
+      `[cos(lat), sin(lon), cos(lon)]` if specified. This is required by many
+      neural climate models.
+    - `cos_zenith-{name}`: Tensors of shape (batch_size, num_steps, 1, height, width),
+      containing the cosine of the solar zenith angle if specified.
```
dallasfoster, line 334: I don't think I support hard-coding these names into the datapipe because it restricts/confines exactly what inputs users need to use with this datapipe. If I wanted to bring in another variable then do I have to make my own datapipe, even if everything else is the same? I think we should think of a solution that adds variables in a more flexible way.

modulus/datapipes/climate/climate.py
```diff
@@ -0,0 +1,829 @@
...
+    start_year : int, optional
+        Start year of dataset, by default 1980
+    lsm_filename : str, optional
+        Path to land sea mask file, by default None
+    lsm_variable : str, optional
+        Variable to load from land sea mask file, by default "LSM"
+    geopotential_filename : str, optional
+        Path to geopotential file, by default None
+    geopotential_variable : str, optional
+        Variable to load from geopotential file, by default "Z"
```
dallasfoster, line 370: Could a dictionary-based strategy be used here instead?

modulus/datapipes/climate/climate.py
```diff
@@ -0,0 +1,829 @@
...
+            sd = np.array([norm[var]["std"] for var in self.variables])
+            self.sd = sd.reshape((1, len(sd), 1, 1))
+        else:
+            raise ValueError(("Invalid statistics file specification"))
+
+        if not self.mu.shape == self.sd.shape == (1, len(self.channels), 1, 1):
+            raise ValueError("Error, normalisation arrays have wrong shape")
+
+
+class ClimateDatapipe(Datapipe):
```
dallasfoster, line 307: I have a hard time accepting the term 'ClimateDatapipe' when most of the inputs here are not climate related. 'ClimateDatapipe' also seems like a very generic term but this class is quite restrictive.

modulus/datapipes/climate/climate.py
```diff
@@ -0,0 +1,829 @@
...
+            self._load_geopotential()
+        if self.use_latlon or any(spec.use_cos_zenith for spec in sources):
+            self._load_latlon()
+
+        # Create pipeline
+        self.pipe = self._create_pipeline()
+
+    def _read_invariant_autoflip(self, filename, var, normalize=False):
+        """Get invariant from netCDF file."""
+        with nc.Dataset(filename) as ds:
```
dallasfoster, line 483: Can we replace netcdf with the more flexible xarray load? That way we can move towards supporting other file formats.

dallasfoster (commented):
This is an initial review because there are some larger concerns about the datapipes.

modulus/datapipes/climate/climate.py
jleinonen: Yes, these were ported over from the `ClimateHDF5Datapipe` in `experimental`. Having a more flexible system, where more auxiliary variables can be added by the user, would be nice. At the same time I think it's good to offer some of the most common ones out of the box.

modulus/datapipes/climate/climate.py
jleinonen: Yes, these can be in a dictionary. Would it work to have the auxiliary variables configured with a dict mapping variable names to callables that provide them?

modulus/datapipes/climate/climate.py
jleinonen: I'm perfectly fine with changing the name, it does seem a bit awkward. I made it up simply by taking the existing `ClimateHDF5Datapipe` and dropping the `HDF5` since it was no longer HDF5-specific.

modulus/datapipes/climate/climate.py
jleinonen: Yes we can certainly change these to xarray if preferred.

jleinonen: Changed to xarray in 45661fe98135162d952302ce9e7deea2b1da70ae

loliverhennigh:
Hey @jleinonen, I looked over the code and don't have any strong comments other then what Dallas said before. Have there been any changes here you would like to get in? If not we can merge if for the release I think.

modulus/datapipes/climate/climate.py
jleinonen: A dict-based interface to adding auxiliary variables is now implemented in e3a916f6a3b9bb4984f13b67fc53a41fe884624b.

modulus/datapipes/climate/climate.py
jleinonen: A dict-based interface to adding auxiliary variables is now implemented in e3a916f6a3b9bb4984f13b67fc53a41fe884624b.

loliverhennigh:
Hey @jleinonen, This looks fantastic! I really like the aux functions. This gets rid of a lot of the previous configs. We could also get rid of the cos zenith with the same function. I would be ok just filing a separate issue for this where we put together a numpy function for this.

modulus/datapipes/climate/utils/invariant.py
```diff
@@ -0,0 +1,119 @@
...
+        if normalize:
+            self.data = (self.data - self.data.mean()) / self.data.std()
+        
+        self.interp_method = interp_method
+
+    def __call__(self, latlon: np.ndarray):
+        (lat, lon) = latlon
+        lat = xr.DataArray(lat, dims=["latitude", "longitude"])
+        lon = xr.DataArray(lon, dims=["latitude", "longitude"])
+        return self.data.interp(
```
loliverhennigh, line 117: Unrelated but if we are using this interpolation for the CoreDiff it will be quite slow. Xarray I have found has problems. Probably not a big deal but something to keep an eye on.

modulus/datapipes/climate/climate.py
```diff
@@ -0,0 +1,790 @@
...
+        timestamps = np.array(
+            [
+                (start_time + timedelta(hours=i * self.stride * self.dt)).timestamp()
+                for i in range(self.num_steps)
+            ]
+        )
+
+        # outputs from auxiliary sources
+        aux_outputs = (
+            callback(timestamps, self.latlon)
```
loliverhennigh, line 731: I'm a little confused here. It makes sense to give arbitrary functions of time and lat lon but when I look at the invariant functions they are only of lat lon and not with time. Am I missing something?

modulus/datapipes/climate/climate.py
```diff
@@ -0,0 +1,790 @@
...
+    file_type: str
+        Type of files to read, supported values are "hdf5" (default) and "netcdf4"
+    stats_files: Union[Mapping[str, str], None], optional
+        Numpy files to data statistics for normalization. Supports either a channels
+        format, in which case the dict should contain the keys "mean" and "std", or a
+        named-variable format, in which case the dict should contain the key "norm" .
+        If None, no normalization will be used, by default None
+    metadata_path: Union[Mapping[str, str], None], optional for NetCDF, required for HDF5
+        Path to the metadata JSON file for the dataset (usually called data.json).
+    channels : Union[List[int], None], optional
```
loliverhennigh, line 92: You could imagine removing `channels` altogether and using `variables` for both where you give a dict of key and list of channel pairs. For example, the default would be `{'field': [0, 1, 3}`. Then you would basically combine the variables and channels. If None is given it will just take all and act as if you are just giving the name of the variables. What do you think?

loliverhennigh:
Hey @jleinonen and @dallasfoster, Wanted to propose something. My understanding is that this datapipe is being used for the corediff currently. Its fairly well generalized but we can take it a bit further I think. My proposal is to merge this now following review and I will make a PR to add zarr support and make it compatible with my unified recipe. That way we can be in a place where we have all the datapipes unified. Looking through the code I think this is possible and I will outline the key points,

* Remove cos zenith and make it an aux function (easy)
* Add Zarr support (easy)
* Follow my suggestion about removing the channels and variables params and just have one variables param as a dict of names and channels. You can also make it a bit more friendly for different dimentional datasets such as healpix.
* Find a way to handle multi-files such as `1980.h5`, `1981.h5`, ... and one file with all the data like I have set up now. I have an idea for this but need to think about it.

What do yall think?

modulus/datapipes/climate/utils/invariant.py
jleinonen: Yeah here it's only called once so not a big deal.

modulus/datapipes/climate/climate.py
jleinonen: Maybe it's a bit awkward, but the invariants that don't depend on time (like land-sea mask - hopefully sea level rise won't be so much that we have to reconsider this!) are set in the `invariants` argument of the datapipe, and these only accept the `latlon` argument. Meanwhile the time-dependent ones are set in `aux_variables` of the ClimateDataSourceSpec. This is because multiple time steps can be returned if `num_steps > 1`, in which case we also return the aux_variables for each time step.

modulus/datapipes/climate/climate.py
jleinonen: I'm not totally sure how you intend the dict to work - why would a variable name map to multiple channels? But I do agree it's better to specify the variables by name than by channel number. The channels option is there more for backward compatibility. Maybe we can deal with this later?

jleinonen:
> Hey @jleinonen and @dallasfoster, Wanted to propose something. My understanding is that this datapipe is being used for the corediff currently. Its fairly well generalized but we can take it a bit further I think. My proposal is to merge this now following review and I will make a PR to add zarr support and make it compatible with my unified recipe. That way we can be in a place where we have all the datapipes unified. Looking through the code I think this is possible and I will outline the key points,
> 
> * Remove cos zenith and make it an aux function (easy)
> * Add Zarr support (easy)
> * Follow my suggestion about removing the channels and variables params and just have one variables param as a dict of names and channels. You can also make it a bit more friendly for different dimentional datasets such as healpix.
> * Find a way to handle multi-files such as `1980.h5`, `1981.h5`, ... and one file with all the data like I have set up now. I have an idea for this but need to think about it.
> 
> What do yall think?

Currently we are not using this datapipe for CorrDiff. It has its own dataloaders and the datasets are in Zarr format. So I agree it would be nice to add Zarr support. But before using it for CorrDiff we would have to consider if we want to generalize it to data in multiple resolutions, which is needed for CorrDiff. Adding that option might make this datapipe too difficult to maintain. _Maybe_ it's better to not have all that complexity in a single datapipe.

The cos zenith could easily be an aux function. But I couldn't figure out how to fit it to the workflow using the current DALI implementation, so I left it as a separate option for now. We could implement it on the GPU using PyTorch tensors instead.

Implementing support for files with all the data shouldn't be too difficult. You mostly need to change the calculation of `year_idx` and `in_idx` in `ClimateDaliExternalSource.__call__`.

modulus/datapipes/climate/climate.py
loliverhennigh: Oh I see. Makes sense

modulus/datapipes/climate/climate.py
loliverhennigh: Yeah, Ill think through this more and see if I have a solution. Lets not worry about it now.

loliverhennigh:
Ok, Yeah I saw we merge this now then and then I make a few modifications so it can be used with the unified recipe as well.

modulus/datapipes/climate/utils/zenith_angle.py
dallasfoster, line 1: It makes me uncomfortable that there is so much redundant code here that is also in `modulus/utils/zenith_angle.py`. Can any of that code be used here?

modulus/datapipes/climate/utils/invariant.py
```diff
@@ -0,0 +1,119 @@
...
+
+    # does longitude wrap around the globe?
+    lon_wraparound = ((bounds[1][0] % 360) == (bounds[1][1] % 360))
+    if lon_wraparound:
+        # treat differently from lat due to wrap-around
+        lon = np.linspace(*bounds[1], shape[1]+1, dtype=np.float32)[:-1]
+    else:
+        lon = np.linspace(*bounds[1], shape[1], dtype=np.float32)
+
+    return np.meshgrid(lat, lon, indexing='ij')
```
dallasfoster, line 28: Either make this code private `i.e. _latlon_grid` or add additional docstring commenting on the assumptions that this function makes about the regularity of the grid (or both).

modulus/datapipes/climate/utils/invariant.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class Invariant(ABC):
+    @abstractmethod
+    def __call__(self, latlon: np.ndarray):
+        pass
```
dallasfoster, line 34: I would appreciate more docstring on this class and what it is intended for.

modulus/datapipes/climate/utils/invariant.py
```diff
@@ -0,0 +1,119 @@
...
+        interp_method: str, optional
+            Any argument accepted by xarray.DataArray.interp.
+            Default 'linear'.
+        """
+
+        with xr.open_dataset(filename) as ds:
+            self.data = ds[var_name].astype(np.float32)
+            self.lat = ds["latitude"].to_numpy().astype(np.float32)
+            self.lon = ds["longitude"].to_numpy().astype(np.float32)
+
```
dallasfoster, line 104: So what if I build a NetCDF4 file with 'lat' and 'lon' instead of 'latitude' and 'longitude'?

dallasfoster: What if I have more than 1 `var_name` that I want to load?

dallasfoster (changes requested):
I agree with the discussion above about merging soon but I signaled a few minor details for changes.

modulus/datapipes/climate/utils/invariant.py
jleinonen: > So what if I build a NetCDF4 file with 'lat' and 'lon' instead of 'latitude' and 'longitude'?
Sure we could have the names configurable but is it worth the added complexity? After the all the user can always rename the NetCDF variables, or build their own function/class to load from whatever kind of file they want.

dallasfoster: I suppose not, this function is well documented with variable assumptions in any case.

dallasfoster: I think my question about the single `var_name` remains: What is the decision for only allowing a single `var_name` and is it conceivable that someone would want to load additional invariants?

modulus/datapipes/climate/utils/invariant.py
jleinonen: I'll add a docstring (and there should be a type hint for the return value as well) to this tomorrow. It's just there to specify the interface that the `Invariant` subclasses are supposed to follow.

test/datapipes/test_climate.py
```diff
@@ -46,34 +54,55 @@ def geopotential_filename():
...
-        patch_size=8,
-        num_samples_per_year=None,
-        batch_size=1,
+        stats_files=stats_files,
+        metadata_path=metadata_path,
+        **spec_kwargs
+    )
+    datapipe = ClimateDatapipe(
+        [spec],
         lsm_filename=lsm_filename,
```
NickGeneva, line 102: lsm_filename does not exist, same with geo potential, need to update all tests here.

NickGeneva:
Closing since this was rebased into a separate PR

PR #256, opened 2023-11-29
Add git hash to the docker container

ktangsali:
# Modulus Pull Request

## Description
Adds git hash to the docker container. If the build arg is not provided, it will default to unknown. 

Sample output:

```
docker run --rm --shm-size=1g --ulimit memlock=-1 --ulimit stack=67108864 --runtime nvidia -v ${PWD}:/workspace -it modulus:deploy /bin/bash -c 'echo $MODULUS_GIT_HASH'
...
<header content>
...
9fc5d8b
```

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,9 +60,10 @@ else
     $(error Unknown CPU architecture ${ARCH} detected)
 endif
 
+MODULUS_GIT_HASH = $(shell git rev-parse --short HEAD)
 
 container-deploy:
-	docker build -t modulus:deploy --build-arg TARGETPLATFORM=${TARGETPLATFORM} --target deploy -f Dockerfile .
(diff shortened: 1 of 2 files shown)


mnabian (approved):

akshaysubr (approved):

Makefile
```diff
@@ -60,9 +60,10 @@ else
     $(error Unknown CPU architecture ${ARCH} detected)
 endif
 
+MODULUS_GIT_HASH = $(shell git rev-parse --short HEAD)
```
akshaysubr, line 63: Do we want the `--short` here? Why not keep the full hash?

ktangsali: I think it was more from presentation standpoint. The full hash is 40 characters long, which I feel might be a bit too much. The short hash should still be resolvable with fallbacks present: https://stackoverflow.com/a/56012434

Alternative would be to make the ENV variable full hash and then truncate it only when printing in the header. I guess that can be done easily. What do you think is preferable?

ktangsali: Actually short will also automatically try to ensure that uniqueness is maintained by automatically increasing the length if required. Resolving this.

PR #269, opened 2023-12-06
Parallel mlflow

MarsuPila:
# Modulus Pull Request

## Description
Only the root process logs to MLFlow. LaunchLogger gathers all losses, so global variables are logged in MLFlow. Additionally, a method for logging mlflow artifacts was added to the LaunchLogger. The two darcy flow examples have been updated accordingly, the nested FNO training is now parallelised.

## Checklist


modulus/launch/logging/mlflow.py
```diff
@@ -82,16 +81,19 @@ def initialize_mlflow(
...
     """
     dist = DistributedManager()
-    if DistributedManager.is_initialized() and dist.distributed:
-        group_name = create_ddp_group_tag(run_name)
-        run_name = f"{run_name}-Process_{dist.rank}"
+    root = True
+    if dist.is_initialized():
+        root = dist.rank == 0
+
+    if not root:  # only one process should be logging to mlflow
```
NickGeneva, line 88: Can just do instead of this if/else:
```
if dist.rank != 0:
    return
```

The is initialized doesn't make full sense there because 83 initializes it regardless.

modulus/launch/logging/launch.py
```diff
@@ -324,6 +323,14 @@ def _log_backends(
             metric_dict[step[0]] = step[1]
             wandb.log(metric_dict)
 
+    def log_mlflow_figure(self, figure, artifact_file: str = "artifact"):
```
NickGeneva, line 326: Can you type hint `figure`

NickGeneva, line 326: lets change this to `log_figure`... launch logger should not have specific back end functions, rather general ones that then have the logic to figure out how to handle specific backends.

examples/cfd/darcy_nested_fnos/train_nested_darcy.py
```diff
@@ -225,7 +279,7 @@ def TrainModel(cfg: DictConfig, base: SetUpInfrastructure, loaded_epoch: int) ->
                         epoch,
                     )
                     total_loss += loss * batch["darcy"].shape[0] / len(base.valid_set)
-                log.log_epoch({"Validation error": total_loss})
+                    log.log_minibatch({"Validation error": total_loss})
```
NickGeneva, line 282: Why log minibatch here?

NickGeneva:
Please add a line in the controbution.md file about updating the MLFlow logger to only operate on proc 0. Thanks!

examples/cfd/darcy_nested_fnos/train_nested_darcy.py
MarsuPila: was experimenting, reverted to `log_epoch`

modulus/launch/logging/launch.py
NickGeneva: Great!

examples/cfd/darcy_nested_fnos/utils.py
```diff
@@ -224,8 +217,9 @@ def compare(
             fig.colorbar(im[ii], ax=ax[ii], location="bottom", fraction=0.046, pad=0.04)
             ax[ii].set_title(self.headers[ii])
 
-        mlflow.log_figure(fig, f"val_step_{step}.png")
-        fig.savefig(os.path.join(self.out_dir, f"validation_step_{step}.png"))
+        LaunchLogger("valid").log_figure(
```
NickGeneva, line 220: You should pass the logger in as a parameter to this function not instantiate.

This will reset the epoch number so when it logs in console, it will read epoch 1 for all validation.

modulus/launch/logging/launch.py
```diff
@@ -324,6 +325,51 @@ def _log_backends(
             metric_dict[step[0]] = step[1]
             wandb.log(metric_dict)
 
+    def log_figure(
+        self,
+        figure,
```
NickGeneva, line 330: FYI removed type hint so matplotlib is not a requirement

NickGeneva (approved):
Approving to unblock, but please fix bug in FNO exmaple first. Works otherwise when I tested it!

modulus/launch/logging/launch.py
Alexey-Kamenev: Instead of removing, you could do something like `figure : "Figure"` though this may or may not upset type checker.

ktangsali (approved):

examples/cfd/darcy_nested_fnos/utils.py
MarsuPila: done

PR #275, opened 2023-12-08
Physics Informed examples: Stokes Flow

ktangsali:
# Modulus Pull Request

## Description
Adds Stokes flow example using GNNs with Physics-informed inference.

## Checklist


mnabian:
I reviewed `pi_inference.py` and it looks good to me. I have already reviewed the other scripts a while ago. It would be good to add some comments or documentation of how the Modulus Graph is used here.

mnabian (approved):

examples/cfd/stokes_mgn/pi_inference.py
```diff
@@ -0,0 +1,525 @@
...
+            loss_u_in,
+            loss_v_in,
+            loss_u_noslip,
+            loss_v_noslip,
+            loss_mom_u,
+            loss_mom_v,
+            loss_cont,
+        )
+
+    def train(self):
```
akshaysubr, line 333: Why is the class called Inferencer when it is also used to train?

examples/cfd/stokes_mgn/pi_inference.py
```diff
@@ -0,0 +1,525 @@
...
+            loss_p,
+            loss_u_in,
+            loss_v_in,
+            loss_u_noslip,
+            loss_v_noslip,
+            loss_mom_u,
+            loss_mom_v,
+            loss_cont,
+        ) = self.loss()
+        loss = (
```
akshaysubr, line 346: Add a comment here for what this weighting is doing

examples/cfd/stokes_mgn/pi_inference.py
```diff
@@ -0,0 +1,525 @@
...
+            results_int = self.graph.forward({"x": x_int, "y": y_int})
+            pred_u, pred_v, pred_p = (
+                results_int["u"],
+                results_int["v"],
+                results_int["p"],
+            )
+            pred_u = pred_u.detach().cpu().numpy()
+            pred_v = pred_v.detach().cpu().numpy()
+            pred_p = pred_p.detach().cpu().numpy()
+            error_u = np.linalg.norm(self.ref_u - pred_u) / np.linalg.norm(self.ref_u)
```
akshaysubr, line 388: Can't these all be done on the GPU using torch.norm and only moving the final error values to the CPU?

examples/cfd/stokes_mgn/pi_inference.py
```diff
@@ -0,0 +1,525 @@
...
+        x_proj = torch.cat([torch.sin(x_proj), torch.cos(x_proj)], dim=-1)
+
+        # Pass through layers
+        out = self.layers(x_proj)
+        return out
+
+
+class MdlsSymDNN(Arch):
+    """
+    Wrapper model to convert PyTorch model to Modulus-Sym model.
```
akshaysubr, line 160: Add some more docs here with links to what a Key is, and how this wrapper does the mapping

examples/cfd/stokes_mgn/pi_inference.py
```diff
@@ -0,0 +1,525 @@
...
+    ):
+        super(MdlsSymDNN, self).__init__(
+            input_keys=input_keys,
+            output_keys=output_keys,
+        )
+
+        self.mdls_model = DNN(layers, fourier_features)
+
+    def forward(self, dict_tensor: Dict[str, torch.Tensor]):
+        x = self.concat_input(
```
akshaysubr, line 178: Add comment here to explain where this concat_input comes from

examples/cfd/stokes_mgn/pi_inference.py
```diff
@@ -0,0 +1,525 @@
...
+            dict_tensor,
+            self.input_key_dict,
+            detach_dict=self.detach_key_dict,
+            dim=-1,
+        )
+        out = self.mdls_model(x)
+        return self.split_output(out, self.output_key_dict, dim=1)
```
akshaysubr, line 185: Add comment here explaining where split_output comes from and why it is needed and how it maps to output_keys

examples/cfd/stokes_mgn/pi_inference.py
akshaysubr: Also how it maps to input_keys

examples/cfd/stokes_mgn/pi_inference.py
```diff
@@ -0,0 +1,525 @@
...
+                Key("momentum_x"),
+                Key("momentum_y"),
+            ],
+            func_arch=False,
+        ).to(self.device)
+
+        self.optimizer = torch.optim.Adam(self.model.parameters(), lr=0.001)
+
+    def parabolic_inflow(self, y, U_max=0.3):
+        u = 4 * U_max * y * (0.4 - y) / (0.4**2)
```
akshaysubr, line 277: Where is this 0.4 coming from? Maybe make it a variable so it has a descriptive name?

examples/cfd/stokes_mgn/pi_inference.py
```diff
@@ -0,0 +1,525 @@
...
+        coords,
+        coords_inflow,
+        coords_noslip,
+        nu,
+        ref_u,
+        ref_v,
+        ref_p,
+    )
+
+    logger.info("Physics-informed inference started...")
```
akshaysubr, line 455: Infernce? The code below calls `train`?

examples/cfd/stokes_mgn/pi_inference.py
akshaysubr: Also, the readme says 0.41, not 0.4

examples/cfd/stokes_mgn/pi_inference.py
```diff
@@ -0,0 +1,525 @@
...
+            self.input_key_dict,
+            detach_dict=self.detach_key_dict,
+            dim=-1,
+        )
+        out = self.mdls_model(x)
+        return self.split_output(out, self.output_key_dict, dim=1)
+
+
+class PhysicsInformedInferencer:
+    """Class to define all the inference utils."""
```
akshaysubr, line 189: Needs more docs on what all these gnn_u, etc mean

akshaysubr (changes requested):
The implementation of the physics informed model using sym utilities itself looks very good, using the `Graph` abstraction and all the derivative utilities.

It might just be that I don't have the needed context here, but my main comment about this example is that it's not clear what the physics informed part is doing. The naming of some of the utilities doesn't help there either since they have an inference suffix, but they're actually training a new model from the looks of it. The README doesn't help clarify this.

examples/cfd/stokes_mgn/pi_inference.py
ktangsali: I can understand how this can be confusing. I have added more docs to the readme and the script itself to hopefully make it clearer. The inferencer class here is basically encapsulating the utils for physics-informed fine-tuning and the subsequent inference.

examples/cfd/stokes_mgn/pi_inference.py
ktangsali: Thanks, added.

examples/cfd/stokes_mgn/pi_inference.py
ktangsali: Thanks, added!

examples/cfd/stokes_mgn/pi_inference.py
ktangsali: Thanks, added!

examples/cfd/stokes_mgn/pi_inference.py
ktangsali: Added!

examples/cfd/stokes_mgn/pi_inference.py
ktangsali: Thanks, please refer my response here: https://github.com/NVIDIA/modulus/pull/275#discussion_r1433020288

I have added more context to hopefully make it clearer.

examples/cfd/stokes_mgn/pi_inference.py
ktangsali: Thanks added!

examples/cfd/stokes_mgn/pi_inference.py
ktangsali: Good catch, fixed!

examples/cfd/stokes_mgn/pi_inference.py
ktangsali: Certainly possible! Just updated it.

ktangsali:
> The implementation of the physics informed model using sym utilities itself looks very good, using the `Graph` abstraction and all the derivative utilities.
> 
> It might just be that I don't have the needed context here, but my main comment about this example is that it's not clear what the physics informed part is doing. The naming of some of the utilities doesn't help there either since they have an inference suffix, but they're actually training a new model from the looks of it. The README doesn't help clarify this.

Thanks Akshay, I have added more info to the README and the docs to hopefully provide more context.

examples/cfd/stokes_mgn/pi_inference.py
akshaysubr: Then why not call it fine-tuning instead of inference?

examples/cfd/stokes_mgn/pi_inference.py
```diff
@@ -0,0 +1,561 @@
...
+            torch.tensor(coords_inflow, requires_grad=True).float().to(self.device)
+        )
+        self.coords_noslip = (
+            torch.tensor(coords_noslip, requires_grad=True).float().to(self.device)
+        )
+
+        self.model = DNN(
+            layers=[2, 128, 128, 128, 128, 3],
+            fourier_features=64,
+        ).to("cuda")
```
akshaysubr, line 253: Should be `self.device` here?

examples/cfd/stokes_mgn/pi_inference.py
```diff
@@ -0,0 +1,561 @@
...
+                Key("v"),
+                Key("p"),
+                Key("continuity"),
+                Key("momentum_x"),
+                Key("momentum_y"),
+            ],
+            func_arch=False,
+        ).to(self.device)
+
+        self.optimizer = torch.optim.Adam(self.model.parameters(), lr=0.001)
```
akshaysubr, line 296: Can we use add `fused=True if torch.cuda.is_available() else False`?

examples/cfd/stokes_mgn/pi_inference.py
```diff
@@ -0,0 +1,561 @@
...
+            results_int["momentum_y"],
+            results_int["continuity"],
+        )
+        pred_u, pred_v, pred_p = results_int["u"], results_int["v"], results_int["p"]
+
+        u_in, v_in = self.parabolic_inflow(self.coords_inflow[:, 1:2])
+
+        # Compute losses
+        # data loss
+        loss_u = torch.mean((self.gnn_u - pred_u) ** 2)
```
akshaysubr, line 325: I don't understand this. This doesn't look like fine-tuning to me. It looks like a regression problem where you're training a PINN to satisfy the GNN outputs rather than the reference data. Why not train directly on the reference data and why is the GNN useful here?

examples/cfd/stokes_mgn/pi_inference.py
ktangsali: Thanks done!

examples/cfd/stokes_mgn/pi_inference.py
ktangsali: Done!

examples/cfd/stokes_mgn/pi_inference.py
ktangsali: We cannot train on reference data as in the inference mode, ground truth is typically not available. Training on the GNN predictions along with the physics losses allow the predictions to be more accurate (This is made clear by the figure in the README, basically in the event where the GNN prediction is not accurate enough, the physics loss can help to guide the predictions to a more realistic one). I am also adding @mnabian to add more context if required as he was the primary author of this example.

examples/cfd/stokes_mgn/pi_inference.py
ktangsali: I am honestly fine calling it either way. In my opinion, both are apt as the fine-tuning step is almost part of the inference process. Adding @mnabian to the thread to check his opinion as well.

mnabian: I agree with Akshay that fine-tuning is a better naming option than inference.

ktangsali: Thanks folks! Updated it to use fine-tuning terminology.

examples/cfd/stokes_mgn/pi_inference.py
ktangsali: Resolving this as this is also being discussed here: https://github.com/NVIDIA/modulus/pull/275#discussion_r1432058257

akshaysubr (approved):

PR #291, opened 2024-01-08
Fea unified recipe

loliverhennigh:
# Modulus Pull Request

Unified recipe for training global weather models on ERA5 data. This features,

* Model switching with configs
* Data curation via ETL pipeline from ARCO ERA5
* 

## Description

Unified recipe for training global weather models on ERA5 data. This features,

* Model switching with configs
* Data curation via ETL pipeline from ARCO ERA. This allows things such as regrinding.

## Checklist

## Dependencies

None


loliverhennigh:
Hello @dallasfoster, wanted to give a proposal for changing the dataset following our offline discussion.

Currently the ETL scripts generates a zarr array with the following variables,
`predicted: (nr_steps, channels, lat, lon)`
`unpredicted: (nr_steps, channels, lat, lon)`
Your comments are that we should make the ETL and datapipe both be agnostic to the predicted and unpredicted variable sets and only the training script should care about this split. What do you think about having the ETL script save every single pressure and single level variable as a new variable in the zarr array. For example,
`pressure_100: (nr_steps, lat, lon)`
`sea_surface_temp: (nr_steps, lat, lon)`
etc...
Then the current zarr datapipe stays the same and the train script does the splitting to predicted and unpredicted. I don't think the performance hit in splitting up the variables will be significant as well but would want to test to be sure. If you did this setup adding new variables to the dataset can be done without needed to create a new dataset.

An alternative idea is to save all the pressure levels of a single field to an variable in the zarr dataset. Basically just sub-sample the pressure levels by a desired amount. This might get rid of any performance issue however when adding new pressure levels you may need to recreate the dataset. It might also complicate either the datapipe or the training recipe.

dallasfoster:
>  What do you think about having the ETL script save every single pressure and single level variable as a new variable in the zarr array. For example, `pressure_100: (nr_steps, lat, lon)` `sea_surface_temp: (nr_steps, lat, lon)` etc... Then the current zarr datapipe stays the same and the train script does the splitting to predicted and unpredicted. I don't think the performance hit in splitting up the variables will be significant as well but would want to test to be sure. If you did this setup adding new variables to the dataset can be done without needed to create a new dataset.

I think this would be the best solution, since as you pointed out, it would be easy to add additional channels later after the initial construction of the zarr dataset.

akshaysubr:
> What do you think about having the ETL script save every single pressure and single level variable as a new variable in the zarr array.

@loliverhennigh Can we not just merge the `predicted` and `unpredicted` variables into one?
```
data: (nr_steps, channels, lat, lon)
```
The training recipe can just look at the `channels` dimension metadata and index based on that to select the predicted and unpredicted portions right? The benefit of this is that the dataloader can use advanced indexing and if used with a store that supports `getitems` like the S3 store or the GDS store, it can asynchronously fetch all of that data getting around the zarr synchronous API.

examples/weather/unified_recipe/fetch_data.py
```diff
@@ -0,0 +1,91 @@
...
+    # Resolve config so that all values are concrete
+    OmegaConf.resolve(cfg)
+
+    # Get transform function
+    if cfg.transform.name == "downsample":
+        transform = lambda x: downsample_transform(x, downsample_factor=cfg.transform.downsample_factor)
+    elif cfg.transform.name == "trim_lat720":
+        transform = trim_lat720_transform
+    else:
+        raise NotImplementedError("Transform not implemented")
```
dallasfoster, line 52: Is this a sustainable way to include transforms? What if this becomes an endless list of if statements?

examples/weather/unified_recipe/seq_zarr_datapipe.py
```diff
@@ -0,0 +1,253 @@
...
+            batch_size=self.batch_size,
+            num_threads=2,
+            prefetch_queue_depth=2,
+            py_num_workers=1,
+            device_id=self.device.index,
+            py_start_method="spawn",
+        )
+
+        with pipe:
+            # ERA5 source
```
dallasfoster, line 126: Remove comment.

examples/weather/unified_recipe/train.py
```diff
@@ -0,0 +1,358 @@
...
+    OmegaConf.resolve(cfg)
+
+    # Initialize distributed environment for training
+    DistributedManager.initialize()
+    dist = DistributedManager()
+
+    # Initialize loggers
+    initialize_mlflow(
+        experiment_name="Modulus-Launch-Dev",
+        experiment_desc="Modulus launch development",
```
dallasfoster, line 70: Should these be config inputs?

examples/weather/unified_recipe/train.py
```diff
@@ -0,0 +1,358 @@
...
+
+    # Normalizer (TODO: Maybe wrap this into model)
+    predicted_batch_norm = nn.BatchNorm2d(cfg.dataset.nr_predicted_variables, momentum=None, affine=False).to(dist.device)
+    unpredicted_batch_norm = nn.BatchNorm2d(cfg.dataset.nr_unpredicted_variables, momentum=None, affine=False).to(dist.device)
+    def normalize_variables(variables, batch_norm):
+        shape = variables.shape
+        variables = variables.flatten(0, 1)
+        variables = batch_norm(variables)
+        variables = variables.view(shape)
+        return variables
```
dallasfoster, line 154: Should this be in the unified recipe? Why or why not?

examples/weather/unified_recipe/train.py
```diff
@@ -0,0 +1,358 @@
...
+
+                    # Switch back to train mode
+                    model.train()
+
+            # Sync after each epoch
+            if dist.world_size > 1:
+                torch.distributed.barrier()
+
+            # Save checkpoint
+            if (epoch % 5 == 0 or epoch == 1) and dist.rank == 0:
```
dallasfoster, line 342: Can this be configurable?

examples/weather/unified_recipe/train.py
```diff
@@ -0,0 +1,358 @@
...
+from modulus.launch.utils import load_checkpoint, save_checkpoint
+from modulus.metrics.general.mse import batch_normalized_mse
+
+from seq_zarr_datapipe import SeqZarrDatapipe
+
+
+
+@hydra.main(version_base="1.2", config_path="conf", config_name="config")
+def main(cfg: DictConfig) -> None:
+
```
dallasfoster, line 59: Can we get more documentation on exactly what this function does, the logic/order of operations for users. 

For example: 
Why do we batch normalize?
What are stages?
Why do we do static capturing?

What options are configurable and what aren't?

modulus/metrics/general/mse.py
```diff
@@ -61,3 +61,32 @@ def rmse(pred: Tensor, target: Tensor, dim: int = None) -> Union[Tensor, float]:
         Root mean squared error value(s)
     """
     return torch.sqrt(mse(pred, target, dim=dim))
+
+
+def batch_normalized_mse(pred: Tensor, target: Tensor) -> Union[Tensor, float]:
```
NickGeneva, line 66: Unit test or remove!

NickGeneva (approved):

examples/weather/unified_recipe/train.py
loliverhennigh: I am strongly in favor of these being in the unified recipe although there might be a better way to add them. I am adding some parts now where it saves the mean and std from these normalizes to a npy file that will be compatible with downstream tasks.

examples/weather/unified_recipe/train.py
loliverhennigh: Ill worry about this in a future PR for this.

examples/weather/unified_recipe/train.py
dallasfoster: We would expect that the mean and std computed from these `BatchNorm2d` operators will tend toward a sample mean and std, but what is the effect on the initial iterations (especially if you are limited to batch size of 1 like graphcast)?

loliverhennigh:
Hey @NickGeneva and @dallasfoster, If possible I would like to get this merged now and leave some of the other comments for future PRs. The main comments are regarding the dataset and how we create and store the data in zarr. I think I will look at this later as I see if I can reuse the datapipes we have in modulus now. I tried to address the other comments and resolve those threads.

dallasfoster (approved):

PR #296, opened 2024-01-12
Fea argconf model instantiation

loliverhennigh:
# Modulus Pull Request
Allow for arbitrary kwargs to models.

## Description
This fixes and issue where we can have arbitrary arguments to instantiate a model. The following is an example of this,

```python
from dataclasses import dataclass
import modulus
import torch.nn as nn

class UNet(modulus.Module):
    def __init__(self, in_channels=1, out_channels=1, **other_args):
        super(UNet, self).__init__(meta=None)

        self.enc1 = self.conv_block(in_channels, 64)
        self.enc2 = self.conv_block(64, 128)

        self.dec1 = self.upconv_block(128, 64)
        self.final = nn.Conv2d(64, out_channels, kernel_size=1)

        self.my_args = other_args
        print(self.my_args)

    def conv_block(self, in_channels, out_channels):
        return nn.Sequential(
            nn.Conv2d(in_channels, out_channels, 3, padding=1),
            nn.ReLU(inplace=True),
            nn.MaxPool2d(2)
        )

    def upconv_block(self, in_channels, out_channels):
        return nn.Sequential(
            nn.ConvTranspose2d(in_channels, out_channels, 2, stride=2),
            nn.Conv2d(out_channels, out_channels, 3, padding=1),
            nn.ReLU(inplace=True)
        )

    def forward(self, x):
        x1 = self.enc1(x)
        x2 = self.enc2(x1)
        x = self.dec1(x2)
        return self.final(x)

if __name__ == "__main__":
    model = UNet(1, boop=2, out_channels=3, beep=4)
    model.save("unet.mdlus")
    new_model = modulus.Module.from_checkpoint("unet.mdlus")
    assert model.my_args == new_model.my_args, f"{model.my_args} != {new_model.my_args}"
```

Turns out this is just a 1 line change and I think we just implemented it wrong in the first place.

Closes https://github.com/NVIDIA/modulus/issues/287

## Checklist


modulus/models/module.py
```diff
@@ -65,7 +65,7 @@ def __new__(cls, *args, **kwargs):
         out._args = {
             "__name__": cls.__name__,
             "__module__": cls.__module__,
-            "__args__": {k: v for k, v in bound_args.arguments.items()},
+            "__args__": {k: v for k, v in bound_args.kwargs.items()},
```
akshaysubr, line 68: I'm not sure this is the right solution here. This won't cover any positional arguments, will it?

The root cause for this issue makes sense since the docs say that `arguments` will only have explicitly bound arguments: https://docs.python.org/3/library/inspect.html#inspect.BoundArguments.arguments

But `kwargs` only has the keyword args and `args` has positional arguments. The current solution will skip any positional args.

akshaysubr (changes requested):
The root cause for the issue makes sense to me. However, I had a comment on the proposed solution.

Can you also add some tests to ensure that these cases are tested by any solution in this PR? One for an unbound kwarg and one testing positional args?

loliverhennigh:
https://github.com/NVIDIA/modulus/issues/287

modulus/models/module.py
loliverhennigh: The positional arguments get bound to the kwargs from the above two lines. Let me take a look though.

loliverhennigh: Oh weird,
```python
        print(bound_args.kwargs)
        bound_args.arguments.pop("self", None)
        print(bound_args.kwargs)
        exit()
```
This gives,
```
{'boop': 2, 'beep': 4}
{'in_channels': 1, 'out_channels': 3, 'boop': 2, 'beep': 4}
```
popping `self` causes kwargs to get all the arguments? What is going on haha.

loliverhennigh: Dang, I am not sure a good way to fix this...

loliverhennigh:
This should be ready for review and merge now @akshaysubr and @mnabian. I think the current solution is much less hacky and makes sense.

modulus/models/module.py
```diff
@@ -57,17 +57,32 @@ class Module(torch.nn.Module):
...
+        # Get args and kwargs (excluding self and unroll kwargs)
+        instantiate_args = {}
+        for param, (k, v) in zip(sig.parameters.values(), bound_args.arguments.items()):
+            if k != "self":  # Skip self
+                if param.kind == param.VAR_KEYWORD:
+                    instantiate_args.update(v)
+                else:
+                    instantiate_args[k] = v
+            else:
+                continue
```
akshaysubr, line 79: Maybe refactor this to be a bit more readable?
```python
if k == "self":
    continue

if param.kind == param.VAR_KEYWORD:
    instantiate_args.update(v)
else:
    instantiate_args[k] = v
```

modulus/models/module.py
```diff
@@ -57,17 +57,32 @@ class Module(torch.nn.Module):
...
         )  # Add None to account for self
         bound_args.apply_defaults()
-        bound_args.arguments.pop("self", None)
 
+        # Get args and kwargs (excluding self and unroll kwargs)
+        instantiate_args = {}
+        for param, (k, v) in zip(sig.parameters.values(), bound_args.arguments.items()):
+            if k != "self":  # Skip self
+                if param.kind == param.VAR_KEYWORD:
+                    instantiate_args.update(v)
```
akshaysubr, line 75: This is a nice solution! There is no need for a recursive approach here, is there? Like if this dict itself had a dict entry, etc?

akshaysubr (approved):
This new solution looks much cleaner than the earlier approach. Only had a couple of small comments.

modulus/models/module.py
loliverhennigh: Apologies but I don't think I understand your comment. Was the issue with the if statements? Not sure I see how this is a recursive approach.

akshaysubr: I was just asking if there was a need for a recursive approach here. Sounds like no.

mnabian:
Follow up issue to be fixed by myself: https://github.com/NVIDIA/modulus/issues/367

mnabian (approved):

PR #333, opened 2024-02-07
Add SFNO tests and Makani optional dependency.

Alexey-Kamenev:
# Modulus Pull Request

## Description
1. Added optional Makani dependency required for SFNO models.
2. Added basic tests to verify that Makani SFNO can be used in Modulus.
3. Minor reformat of Dockerfile: removed trailing spaces, replaced tabs with spaces.

## Checklist


pyproject.toml
```diff
@@ -54,6 +54,13 @@ dev = [
     "ruff==0.0.290",
 ]
 
+makani = [
+    "makani @ git+https://github.com/NVIDIA/modulus-makani.git@v0.1.0",
+    "torch_harmonics @ git+https://github.com/NVIDIA/torch-harmonics.git@v0.6.5",
```
ktangsali, line 59: Oh! I am not sure if this is going to break our pypi uploads... In the past we have had issues with such git links. Refer: https://github.com/pypa/twine/issues/726 https://github.com/NVIDIA/modulus/pull/64

ktangsali: Is there a possibility of the torch_harmonics and makani team releasing pypi wheels?

Dockerfile
ktangsali, line 1: Just curious, did you use any automated linter or did you have to do this manually?

pyproject.toml
Alexey-Kamenev: That's a very good point! I saw the [Attention](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#direct-url-dependencies) notice in the docs but mistakenly thought it's only related to mandatory dependencies, not optional.
Let me discuss it with Makani folks.
For now, here is what we can do:
1. Comment out (with TODO comment) these direct URL dependencies.
2. Install them in Dockerfile CI stage (so the tests could run)

Once Makani is in pypi we can update the code accordingly.

Thought?

Dockerfile
Alexey-Kamenev: No magic here, just VS Code :)
* `trimTrailingWhitespace` to trim trailing whitespaces automatically on save.
* "Convert indentation to spaces" command - click on "Spaces: 4" in the status bar (or via Ctrl+Shift+P).

pyproject.toml
bonevbs: torch-harmonics can be installed through pypi already. Are we sure that this is breaking?

makani requires some workarounds at the moment to fix some issues with complex numbers in PyTorch. This requires overwriting some files in the torch installation and as long as this is the case, I'd prefer having it as an "experimental" repository that needs to be installed through git.

test/models/test_sfno.py
```diff
@@ -0,0 +1,80 @@
...
+    torch.manual_seed(0)
+
+    # Construct SFNO models.
+    model_1 = _create_model().to(device)
+    model_2 = _create_model().to(device)
+
+    bsize = 2
+    invar = torch.randn(bsize, INP_CHANS, *IN_OUT_SHAPE).to(device)
+
+    assert common.validate_checkpoint(model_1, model_2, (invar,))
```
bonevbs, line 80: we could add some more tests here if you like. It would maybe make sense to test forward and backward pass of the model. You could also just take the unittest `test_models.py` from makani and adapt it.

Alexey-Kamenev: Yes, this is a good idea. For now I just want to make sure that we can instantiate and use SFNO model using Modulus and that checkpoints can be loaded (so we can avoid accidentally breaking it). But we also do not want to duplicate a code from Makani - Modulus assumes SFNO model itself in Makani is already sufficiently tested.

pyproject.toml
ktangsali: Yes, I am reasonably confident that having the git links in the optional dependency will also be flagged. The solution suggested by @Alexey-Kamenev sounds like a good alternative (installing it via git inside the Dockerfile). Additionally, we can also add appropriate docs to the READMEs of Makani/Modulus (wherever appropriate) to mention this step to the users.

Dockerfile
ktangsali: Fantastic! Thanks for sharing your setup, I will configure my VSCode accordingly now :)

ktangsali (approved):

pyproject.toml
bonevbs: That sounds like a reasonable workaround.

loliverhennigh (approved):

Issue #312, opened 2024-01-23
🐛[BUG]: ERA5 DALI dataloader has a potential for deadlock with certain combinations of batch_size and world_size


ktangsali:
### Version

0.4.0

### On which installation method(s) does this occur?

Docker, Pip, Source

### Describe the issue

The ERA5 dataloader [here](https://github.com/NVIDIA/modulus/blob/main/modulus/datapipes/climate/era5_hdf5.py) can run into deadlock some particular combinations of `batch_size`, `world_size` and `num_samples_per_year`. For example, the current [AFNO weather example](https://github.com/NVIDIA/modulus/tree/main/examples/weather/fcn_afno) runs into a deadlock on 32 GPUs. A potential fix can be to ensure that the `num_samples_per_year` is an exact multiple of the `world_size`.

PR #337, opened 2024-02-08
FCN example deadlock fixes

ktangsali:
# Modulus Pull Request

## Description
Currently the FCN example sometimes runs into a deadlock on multi-node systems. This PR tries to fix that by making the `samples_per_year` an exact multiple of the `world_size`. It also makes the `num_workers` for the dataloaders configurable in the training recipe. 

Closes: #312 

## Checklist


modulus/datapipes/climate/era5_hdf5.py
```diff
@@ -157,7 +157,9 @@ def parse_dataset_files(self) -> None:
         self.logger.info(f"Getting file stats from {self.data_paths[0]}")
         with h5py.File(self.data_paths[0], "r") as f:
             # truncate the dataset to avoid out-of-range sampling
-            data_samples_per_year = f["fields"].shape[0] - self.num_steps * self.stride
+            data_samples_per_year = (
```
Alexey-Kamenev, line 160: So the idea here is that each rank should have the same number of samples? Maybe add a comment to make it more clear?

Alexey-Kamenev (approved):

modulus/datapipes/climate/era5_hdf5.py
ktangsali: That is correct! Good point about adding a comment. Let me do that

loliverhennigh (approved):

stefan-maxar:
Perhaps I am missing something, but I think this logic might still deadlock since `data_samples_per_year` is only used if `self.num_samples_per_year` is `None` (line 175 in era5_hdf5.py) and given that `num_samples_per_year_train` is currently defined in the FCN config YAML? 

I think it the same logic as provided to `data_samples_per_year` can be applied to the config-defined `num_samples_per_year_train`...but am not quite sure. That being said, considering the yearly HDF5 files have a full sample of 1460 (365 days * 4 samples per day), perhaps there is a better way to do this such that all samples can be used and/or a fixed max number of samples based on world_size.

PR #352, opened 2024-02-20
Fea unified recipe model card

loliverhennigh:
# Modulus Pull Request

closes https://github.com/NVIDIA/modulus/issues/343

Add model card saving utility to unified recipe

## Checklist


examples/weather/unified_recipe/save_model_card.py
```diff
@@ -0,0 +1,82 @@
...
+    unpredicted_variables,
+    save_path,
+    readme=None,
+    ):
+    """
+    Save model card for model global weather model.
+
+    The model card will be a directory with the following files:
+    - model.mdlus: The model file.
+    - variables.json: A json file with the predicted and unpredicted variables.
```
dallasfoster, line 32: I think it would be preferable to have a `metadata.json` file that contains the variable information but also the grid information (usually a list of lat/lon but we should think on this), explicit dimension information ( e.g. ['batch', 'variable', 'lat' 'lon'] or ['time', 'variable', 'face', 'lat', 'lon'] in the DLWP case), and model time step / history information.

loliverhennigh: I added the entire `cfg` as `metadata.json`. This will have all needed information. It gets a bit complicated with the regridding stuff so seemed like this was the best option. The configs have all the information and we can write a decoding to get lat and lon arrays if needed.

examples/weather/unified_recipe/train.py
```diff
@@ -51,9 +51,10 @@
 from modulus.utils import StaticCaptureEvaluateNoGrad, StaticCaptureTraining
 
 from seq_zarr_datapipe import SeqZarrDatapipe
+from model_packages import save_inference_model_package
```
dallasfoster, line 54: Where is this function located?

examples/weather/unified_recipe/save_model_card.py
dallasfoster: I am not sure this is sufficient. I might be mistaken but it is not clear to me that the domain information can be gleaned from the `cfg` alone. The `lat`/`lon` information seems to come from the dataset but isn't implicit in the config. I think we should do a little bit more here.

dallasfoster (changes requested):

examples/weather/unified_recipe/train.py
loliverhennigh: whoops, forgot to commit.

examples/weather/unified_recipe/save_model_card.py
loliverhennigh: I disagree, you can get the `lat` `lon` knowing that the data comes from ARCO and all the transformation
 configs are present. Here is a config for reference [metadata.json](https://github.com/NVIDIA/modulus/files/14503210/metadata.json). At least the point being that just from this config file I would be able to get a `lat` `lon` grid. I think you can argue we should compute this ourselves from the dataset and save it to the json file. The problem is though that this introduces a lot of variations in what the `lat` and `lon` should actually be for different griddings. The `face` `lat`, `lon` for healpix might work but I am not sure what this information would even be used for. For the healpix you really need the transformation configs if you want to from lat lon to healpix and vice versa. Ideally we would have regrdding in Modulus maybe and then reference that but thats kind of an open problem I would say.

dallasfoster: > you can get the lat lon knowing that the data comes from ARCO and all the transformation
configs are present.

How can this be made programmatic? There is nothing in the config that makes it so that I can assume the data came from ARCO and I don't think that even should be an assumption.

The `metadata.json` that you provided only contains the following information regarding the domain: `chunking`, `transform`, and `input_shape`. None of those actually specify what the actual dimensions are. I think at minimum we need to know the exact values of the domain for the model (`['lat', 'lon']` or `['face', 'lat', 'lon']`)

I think the issue is that this "model card" should have all of the relevant information needed to be run in the inference package. If you believe that the calculation for the domain dimensions should be handled elsewhere then where? Why not here? The dimensions of the data (both domain and non-domain dimensions) are part of the model provenance and should be known by (or calculable by) the training script.

loliverhennigh: Yeah, at the very least I suppose I can give the `lat` `lon` on ARCO with the JSON. This PR kinda feels like building half a bridge though. I think tomorrow I am going to try and see how much work it will be to get a model from this actually working in the inference service. If we managed to get like a PR for both parts then that would clear things up.

NickGeneva: Just weighing in here a bit late but some items to think about:

- We've been in the regridding discussion before, we determined its complicated to implement generally, lots of nuance and decided to move past it for now. I suggest we still consider the same. I sounds like a lot of work Oliver.
- Oliver you're right that one could back calc the lat lon etc from the data set + transformation. However, this is a very round about way to do this which would be someones immediate question about the inputs. I shouldnt have to know a model was trained on ARCO to figure out its a 0.25 deg model.
- Data set information is important to catalog however for the Model Card ++
- The coordinate data does not have to be in a JSON, its array data. Package it up in a npz if you're not a fan of 1440 lines of floats in your ASCII file.
- In onnx its required you provide the shape and the name of the input/output tensors. This is the minimum. We have more info needed because we need to know the geo-physical info of what the inputs are.
- Lets start simple, consider our own use case now for guidance, do whats needed there, and then expand as customers need more.

PR #356, opened 2024-02-22
Diagnostic model example

jleinonen:
# Modulus Pull Request

## Description
Adds an example of training diagnostic models (models that predict an additional variable from the atmospheric state). Also includes scripts for basic validation and for exporting models in an Earth2MIP-compatible format.

This PR also fixes a couple of bugs in `modulus.datapipes.climate.ClimateDatapipe` that were found when testing it with a new dataset.

For further information about how the example works, refer to the `README.md` included in this PR.

## Checklist

## Dependencies

No new dependencies needed.

### DIFF
--- a/examples/weather/diagnostic/config/diagnostic_precip.yaml
+++ b/examples/weather/diagnostic/config/diagnostic_precip.yaml
@@ -1,0 +1,52 @@
+model:
+  model_type: "afno"
+  model_name: "diagnostic-precip-afno"
+  inp_shape: [720, 1440]
+  in_channels: 27 # TODO: this could be automated
+  out_channels: 1
+
+sources:
+  state_params:
+    # ERA5 atmospheric state data
+    data_dir: "/data/era5-73"
+    metadata_path: "/data/era5-73/data.json"
+    name: era5
+    # Inputs following Pathak et al. (2022)
+    variables: [
+      "u10m", "v10m", "t2m", "sp", "msl", "tcwv",
+      "u500", "u850", "u1000", "v500", "v850", "v1000",
+      "z50", "z500", "z850", "z1000", "t500", "t850",
+      "r500", "r850"
+    ]
+    use_cos_zenith: True
+    stats_files:
+      mean: "/data/era5-73/stats/global_means.npy"
+      std: "/data/era5-73/stats/global_stds.npy"
+  diag_params:
+    # precipitation data
+    data_dir: "/data/precip"
+    name: precip
+    file_type: netcdf4
+    variables: ["tp06"]
+
+datapipe:
+  geopotential_filename: "/data/era5-73/invariants/orography.nc"
+  lsm_filename: "/data/era5-73/invariants/land_sea_mask.nc"
+  use_latlon: True
+  num_samples_per_year_train: 1456
+  num_samples_per_year_valid: 64
+  batch_size_train: 4
+  crop_window: [[0, 720], [0,1440]]
+
+training:
+  max_epoch: 50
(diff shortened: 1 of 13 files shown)


examples/weather/diagnostic/config/diagnostic_precip.yaml
```diff
@@ -0,0 +1,68 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+model:
+  model_type: "afno"
+  model_name: "diagnostic-precip-afno"
+  inp_shape: [720, 1440]
+  in_channels: 27 # TODO: this could be automated
```
loliverhennigh, line 21: You can use `eval` in yaml files for stuff like this. For example, https://github.com/loliverhennigh/modulus/blob/fea-unified_recipe_model_card/examples/weather/unified_recipe/conf/dataset/7_var.yaml#L25.

examples/weather/diagnostic/diagnostic/export.py
```diff
@@ -0,0 +1,114 @@
...
+        model,
+        in_channel_names,
+        out_channel_names,
+        grid_shape=grid_shape,
+        input_stats={"mean": in_source.mu, "std": in_source.sd},
+        output_stats={"mean": out_source.mu, "std": out_source.sd},
+    )
+
+
+def _create_e2mip_package(
```
loliverhennigh, line 80: We should put a method like this in modulus source. I think @daviddpruitt is using it too. Not for this PR though.

examples/weather/diagnostic/diagnostic/models.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def setup_model(
+    model_type: str = "afno", model_name: Union[str, None] = None, **model_cfg
+) -> Module:
+    """Setup model from config dict."""
+    model_kwargs = default_model_params[model_type].copy()
+    model_kwargs.update(model_cfg)
+
+    if model_type == "afno":
```
loliverhennigh, line 40: We can actually use the `instantiate` Module method. There is an example of it here, https://github.com/loliverhennigh/modulus/blob/fea-unified_recipe_model_card/examples/weather/unified_recipe/train.py#L99. This will make instantiating the model automatic from configs.

examples/weather/diagnostic/config/diagnostic_precip.yaml
jleinonen: I think it's a bit complex to do with eval because in addition to the number of variables in the data generator, `in_channels` also depends on how many auxiliaries/invariants you add. It's better to implement it using a function in the Python code that allows removing this input parameter.

examples/weather/diagnostic/diagnostic/export.py
jleinonen: Yeah, happy to coordinate this with you. We'll also need to add some code to Earth-2 MIP that can recognize the auxiliary/invariant inputs. Marking this as resolved since no changes are suggested.

examples/weather/diagnostic/diagnostic/models.py
jleinonen: Great suggestion, changed in [2e2f4b7](https://github.com/NVIDIA/modulus/pull/356/commits/2e2f4b7f46350d2db3bc801e09ca2befa9a1878c). This way you can use any compatible model without having to add it to the list.

NickGeneva (approved):

loliverhennigh (approved):

PR #366, opened 2024-02-27
Add more config options to FCN AFNO example.

Alexey-Kamenev:
# Modulus Pull Request

## Description

Added more config options to the example to make it more convenient for the end users.

## Checklist


examples/weather/fcn_afno/conf/config.yaml
```diff
@@ -20,12 +20,26 @@ hydra:
   run:
     dir: ./outputs/
 
+train_dir: ./dataset/train
```
ktangsali, line 23: Can we keep the defaults to `/data/*`? This example has been integrated with a few workflows internally, like perf testing, qa, etc. and I think this change will break that

examples/weather/fcn_afno/conf/config.yaml
```diff
@@ -20,12 +20,26 @@ hydra:
...
+validation_dir: ./dataset/val
```
ktangsali, line 24: also, for internal datasets, this will have to be `/data/test/`

examples/weather/fcn_afno/train_era5.py
```diff
@@ -130,23 +131,23 @@ def main(cfg: DictConfig) -> None:
...
         validation_datapipe = ERA5HDF5Datapipe(
-            data_dir="/data/test/",
-            stats_dir="/data/stats/",
-            channels=[i for i in range(20)],
-            num_steps=8,
-            num_samples_per_year=4,
-            batch_size=1,
+            data_dir=to_absolute_path(cfg.validation_dir),
+            stats_dir=to_absolute_path(cfg.stats_dir),
+            channels=list(range(cfg.num_channels)),
```
ktangsali, line 136: If we are configuring this, is it possible to make this a list of channels instead of number or channels in the yaml?

ktangsali (approved):
Thanks! Looks good, I added a few comments regarding defaults, mostly to see if we can prevent the current workflows breaking.

examples/weather/fcn_afno/conf/config.yaml
Alexey-Kamenev: Great point - I forgot about our internal QA workflows!

examples/weather/fcn_afno/conf/config.yaml
Alexey-Kamenev: Fixed.

examples/weather/fcn_afno/train_era5.py
Alexey-Kamenev: Good point - done.

loliverhennigh (approved):
Lgtm

PR #381, opened 2024-03-04
Fix corr-diff conditional scaling

### LINKED ISSUE #229: 🐛[BUG]: Resdiff conditional input scaling
### Version

n/a

_No response_

### Describe the issue

I know resdiff isn't in the code base yet, but wanted to file this since I'll be gone.

Not a bug per-se, but @MortezaMardani, Arash, Jaideep and I have discussed an issue with how the conditional inputs are scaled in the EDMPreconditioner:
```
        #SR: concatenate input channels
        if img_lr is None:
            x = x
        else:
            x = torch.cat((x, img_lr), dim=1)

        x = x.to(torch.float32)
        sigma = sigma.to(torch.float32).reshape(-1, 1, 1, 1)
        class_labels = None if self.label_dim == 0 else torch.zeros([1, self.label_dim], device=x.device) if class_labels is None else class_labels.to(torch.float32).reshape(-1, self.label_dim)
        dtype = torch.float16 if (self.use_fp16 and not force_fp32 and x.device.type == 'cuda') else torch.float32

        c_skip = self.sigma_data ** 2 / (sigma ** 2 + self.sigma_data ** 2)
        c_skip = 0.0*c_skip
        c_out = sigma * self.sigma_data / (sigma ** 2 + self.sigma_data ** 2).sqrt()
        c_out = torch.ones_like(c_out)
        c_in = 1 / (self.sigma_data ** 2 + sigma ** 2).sqrt()
        c_in = torch.ones_like(c_in)
        c_noise = sigma.log() / 4
        c_noise = 0.0*c_noise
        
        F_x = self.model((c_in * x).to(dtype), c_noise.flatten(), class_labels=class_labels, **model_kwargs)
```

Specifically, for large noise c_in is very small. So both the conditional input "img_lr" and the latents are scaled down. I think we agreed the conditional inputs should probably not scaled down like this (i.e. the class labels are not).

The code would be modified like this:
```
        x = x.to(torch.float32)
        sigma = sigma.to(torch.float32).reshape(-1, 1, 1, 1)
        class_labels = (
            None
            if self.label_dim == 0
            else torch.zeros([1, self.label_dim], device=x.device)
            if class_labels is None
            else class_labels.to(torch.float32).reshape(-1, self.label_dim)
        )
        dtype = (
            torch.float16
            if (self.use_fp16 and not force_fp32 and x.device.type == "cuda")
            else torch.float32
        )

        c_skip = self.sigma_data**2 / (sigma**2 + self.sigma_data**2)
        c_out = sigma * self.sigma_data / (sigma**2 + self.sigma_data**2).sqrt()
        c_in = 1 / (self.sigma_data**2 + sigma**2).sqrt()
        c_noise = sigma.log() / 4

        arg = c_in * x

        if img_lr is not None:
            arg = torch.cat([arg, img_lr], dim=1)

        F_x = self.model(
            arg.to(dtype),
            c_noise.flatten(),
            class_labels=class_labels,
            **model_kwargs,
        )
```

cc @mnabian 

_No response_

### Relevant log output

_No response_

### Environment details

_No response_

nbren12:
# Modulus Pull Request

## Description

This PR fixes the scaling for conditional inputs in corrdiff. Previously the conditional inputs were concatenated with the noisy target BEFORE scaling. This means that for large noise the conditional inputs will also be scaled down to near 0. This is undesirable since the large-noise limits corresponds to deterministic prediction---we expect the conditional inputs to matter most in this case! This PR fixes this by scaling the noise image BEFORE concatenating it with the conditioning.

I wrote this PR with unit testing in mind, so I wrote a modular class `_ConditionalPrecond` which is testable in isolation. I added a public API `EDMPrecondSRV2` to provide a similar interface as `EDMPrecondSR` that also builds the UNet.

Closes #229.

## Checklist

### DIFF
--- a/modulus/models/diffusion/preconditioning.py
+++ b/modulus/models/diffusion/preconditioning.py
@@ -896,18 +896,9 @@ class _ConditionalPrecond(torch.nn.Module):
         return torch.as_tensor(sigma)
 
 
-def EDMPrecondSRV2(
-    img_resolution,
-    img_in_channels,
-    img_out_channels,
-    label_dim=0,
-    use_fp16=False,
-    sigma_min=0.0,
-    sigma_max=float("inf"),
-    sigma_data=0.5,
-    model_type="DhariwalUNet",
-    **model_kwargs,
-) -> _ConditionalPrecond:
+class EDMPrecondSRV2(_ConditionalPrecond, Module):
+    # note this order of inheritance is necessesary  since Module.__init__ calls
+    # super().__init__
     """EDM Preconditioner with appropriate handling of conditional inputs via concatenation
 
     This helper is provided to have a similar interface to EDMPrecondSR,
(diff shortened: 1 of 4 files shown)


nbren12:
BTW, I've used this code preconditioner in some other projects with good results.

MortezaMardani:
makes sense to me.

MortezaMardani (approved):

yairchn (approved):
LGTM!

modulus/models/diffusion/preconditioning.py
```diff
@@ -806,3 +814,154 @@ def round_sigma(sigma: Union[float, List, torch.Tensor]):
         See EDMPrecond.round_sigma
         """
         return EDMPrecond.round_sigma(sigma)
+
+
+class _ConditionalPrecond(torch.nn.Module):
```
ktangsali, line 819: Can this subclass from `modulus.Module`? Similar to: https://github.com/NVIDIA/modulus/blob/main/modulus/models/diffusion/preconditioning.py#L240

test/models/diffusion/test_preconditioning.py
```diff
@@ -0,0 +1,55 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+
+from modulus.models.diffusion.preconditioning import EDMPrecondSRV2, _ConditionalPrecond
+
+
+def test__ConditionalPrecond():
```
ktangsali, line 22: nit: Just one underscore should be sufficient I think.

ktangsali (approved):

modulus/models/diffusion/preconditioning.py
nbren12: I'm not sure.  These docs claim that all arguments of `__init__` need to be json serializable, but `model` is a Module: https://github.com/NVIDIA/modulus/blob/cea2e6074ba679ed18dfda1d5235191c059f2d9a/modulus/models/module.py#L42C1-L45C23

Any suggestions?

nbren12: I added a workaround.

ktangsali: Thanks! Looks good to me :)

PR #385, opened 2024-03-08
Fea unified recipe sfno conf

loliverhennigh:
# Modulus Pull Request

Adding SFNO conf to unified recipe following current Makani configs.

## Checklist


examples/weather/unified_recipe/README.md
```diff
@@ -84,3 +84,14 @@ training, run
 
 If running inside a docker container, you may need to include the
 `--allow-run-as-root` in the multi-GPU run command.
+
+## SFNO Training
+
+One of the showcased models available in the configs is [Spherical Fourier Neural Operators:
+Learning Stable Dynamics on the Sphere](https://arxiv.org/pdf/2306.03838.pdf).
+
+The config file can be modified to train the SFNO model by uncommenting all SFNO configs. Following the prior dataset fetching and
```
ktangsali, line 93: "Following the prior dataset fetching and" - this is sounding incomplete sentence... Am I interpreting this correctly?

ktangsali (approved):

examples/weather/unified_recipe/train.py
```diff
@@ -51,9 +51,10 @@
 from modulus.utils import StaticCaptureEvaluateNoGrad, StaticCaptureTraining
 
 from seq_zarr_datapipe import SeqZarrDatapipe
+from model_packages import save_inference_model_package
 
 
-def batch_normalized_mse(pred: Tensor, target: Tensor) -> Union[Tensor, float]:
```
Alexey-Kamenev, line 56: 1. My understanding is that `batch_normalized_mse` expects `Tensor`s as input, so why remove the type hint?
2. The function returns the result of `torch.mean`, which is actually a `Tensor`, not `float`, so the type hint is incorrect. If you absolutely must return `float` you should use `.item()` but note that it will cause a sync.

examples/weather/unified_recipe/train.py
```diff
@@ -432,6 +433,20 @@ def train_step_forward(
...
         # Finish training
         if dist.rank == 0:
+            # Save model card
+            logger.info("Saving model card")
+            save_inference_model_package(
+                model,
+                cfg,
+                predicted_variable_normalizer=predicted_batch_norm,
+                unpredicted_variable_normalizer=unpredicted_batch_norm,
+                latitude=zarr.open(cfg.dataset.dataset_filename, mode="r")["latitude"],
```
Alexey-Kamenev, line 443: Is `zarr.open` fast? If not, maybe call it once and store the result?
```py
with zarr.open(cfg.dataset.dataset_filename, mode="r") as zarr_store:
...
    latitude=zarr_store["latitude"]
```

Alexey-Kamenev (approved):
LGTM

examples/weather/unified_recipe/train.py
loliverhennigh: Its actually extremely fast so I got a bit lazy doing it when I might not need to.

examples/weather/unified_recipe/README.md
loliverhennigh: Whoops, Finished the configs with info on install Makani. Tested it out and it runs. I would really have liked to make SFNO the default but I decided to keep a tiny AFNO model for now.

examples/weather/unified_recipe/train.py
loliverhennigh: Fixed this

PR #386, opened 2024-03-11
Added Physics-Informed Magnetohydrodynamics Example

fresleven:
# Modulus Pull Request

## Checklist


fresleven:
I also changed the SWE Nonlinear loss function that I previously put since I noticed that there was a slight problem with the indexing last time (it should've gone from [1:-1] for all values since we ignore the boundaries, but there were some values where it went from [0:-2] (i.e. for `2 * nt - 2`. Thus, I made it simpler by just indexing at the end.

fresleven:
Hi, I just wanted to know when you would be able to review my code?

loliverhennigh:
Hey @fresleven, I have been looking at the code and it looks good but would like to try it out. Do you have a dataset already made you could send? Even if its a smaller subset. I was running into issues installing Dedalus.

fresleven:
Hi, I just made a google drive with 100 options:
https://drive.google.com/drive/folders/10xeaTHRMLKnkPW5uXXEbBDktPgdQ5qW2?usp=sharing

you may need to change the config files based on how much training/testing files there are

fresleven:
Also, please let me know what issues you were having with Dedalus and I can try and see to replicate/fix them soon. I was using conda, and that worked for me.

fresleven:
also, Professor Huerta notified me that the deadline is by 5pm EST today. I know it's quite short, but hopefully could you complete the review by then?

ktangsali (approved):
Looks good. I have a few comments that can potentially be addressed in a separate PR:

1. We should investigate how to offer the model architecture code inside `tfno` more generally via `modulus.models`. 
2. Provide the small dataset you have as maybe the default config (we can host the dataset via NGC if you are interested [Modulus Resources](https://catalog.ngc.nvidia.com/resources?filters=&orderBy=nameASC&query=Modulus)) and then have it auto-download as part of the training. That way the startup time for the users is very quick without requriring the download of dedalus library. 

Thanks!

PR #387, opened 2024-03-11
Improve initial data loading in Ahmed body example.

Alexey-Kamenev:
# Modulus Pull Request

## Description

Improvements to Ahmed body dataset and example:

* improved initial dataset loading time.
* minor refactoring to the dataset
* added an option to control checkpoint frequency.

## Checklist

### DIFF
--- a/modulus/datapipes/gnn/ahmed_body_dataset.py
+++ b/modulus/datapipes/gnn/ahmed_body_dataset.py
@@ -18,7 +18,7 @@ import concurrent.futures as cf
 import os
 import re
 from dataclasses import dataclass
-from typing import Any, Dict, Iterable, List, Tuple, Union
+from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
 
 import numpy as np
 import torch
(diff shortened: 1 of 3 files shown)


loliverhennigh (approved):
Need to look more at the ahmeb body example but in general this looks good to me.

modulus/datapipes/gnn/ahmed_body_dataset.py
```diff
@@ -191,83 +208,46 @@ def __init__(
...
+        def get_num_workers():
+            # Make sure we don't oversubscribe CPUs on a node.
+            # TODO(akamenev): this should be in DistributedManager.
+            local_node_size = max(
+                int(os.environ.get("OMPI_COMM_WORLD_LOCAL_SIZE", 1)), 1
+            )
+            num_workers = len(os.sched_getaffinity(0)) // local_node_size
+            return max(num_workers - 1, 1)
+
+        num_workers = get_num_workers()
```
ktangsali, line 227: Does it make sense to make this configurable? And then maybe one can use `dist.world_size` for that parameter?

modulus/datapipes/gnn/ahmed_body_dataset.py
```diff
@@ -92,7 +108,7 @@ def __init__(
         data_dir: str,
         split: str = "train",
         num_samples: int = 10,
-        invar_keys: List[str] = [
+        invar_keys: Iterable[str] = (
```
ktangsali, line 111: Does this change the docstrings here: https://github.com/NVIDIA/modulus/pull/387/files#diff-1af1f52943601c7e922fb5b8b6fde91bfa932071db5abcdcfdc8c37e759ca7adR88 ? Or do we want to keep it the same indicating `List[str]` as the preferable option?

modulus/datapipes/gnn/ahmed_body_dataset.py
ktangsali: Or maybe give an option to overwrite, but if not specified, it uses this implementation. Not sure if I am overthinking this...

ktangsali (approved):
Looks good! The speed improvement is awesome. Just left a few minor comments.

modulus/datapipes/gnn/ahmed_body_dataset.py
Alexey-Kamenev: Good point - fixed the docstring. The code works with any iterable equally well, so there is no need to use more specific type.

modulus/datapipes/gnn/ahmed_body_dataset.py
Alexey-Kamenev: Good suggestion - made it configurable, with default being auto-select.
Note that `world_size` would not work as expected when the number of nodes > 1 (`world_size` would be the total world size, but we need the local node size), so that's why I have to use `OMPI_COMM_WORLD_LOCAL_SIZE`.

PR #400, opened 2024-03-22
add example which trains a distributed GraphCast model on shallow-water-equations data

stadlmax:
# Modulus Pull Request

## Description
- add example which trains a GraphCast model on a shallow-water-equations dataset
- setup example in such a way to showcase "tensor-parallel" training of a GNN model
- improvements and bugfixes in used distributed communication primitives

## Checklist


modulus/distributed/manager.py
```diff
@@ -359,24 +359,24 @@ def setup(
             else:
                 manager._local_rank = local_rank
 
+        manager._device = torch.device(
+            f"cuda:{manager.local_rank}" if torch.cuda.is_available() else "cpu"
+        )
+
+        if manager._distributed:
             # Setup distributed process group
             dist.init_process_group(
```
stadlmax, line 368: there is an option to set `device_id` for `init_process_group`

> a single, specific device to “bind” this process to, allowing for backend-specific optimizations. Currently this has two effects, only under NCCL: the communicator is immediately formed (calling ncclCommInit* immediately rather than the normal lazy call) and sub-groups will use ncclCommSplit when possible to avoid unnecessary overhead of group creation. If you want to know NCCL initialization error early, you can also use this field.

modulus/distributed/manager.py
```diff
@@ -359,24 +359,24 @@ def setup(
...
                 backend, rank=manager.rank, world_size=manager.world_size
             )
 
-        manager._device = torch.device(
-            f"cuda:{manager.local_rank}" if torch.cuda.is_available() else "cpu"
-        )
-        # Needed for cuda graphs
         if torch.cuda.is_available():
```
stadlmax, line 372: this has been somewhat weird so far anyways
- we sometimes guard with `torch.cuda.is_available` but not always
- we call both `torch.cuda.device` and `torch.cuda.set_device`

modulus/distributed/utils.py
```diff
@@ -400,36 +403,48 @@ def gather_v_wrapper(
...
 
-        for r in range(comm_size):
-            if r != dst:
-                req_list[r].wait()
+    dist.all_to_all(x_recv, x_send, group=group)
 
-    else:
-        req = dist.isend(tensor, dst=dst, group=group)
-        req.wait()
+    # TODO: clean gather/scatter and some examples up
```
stadlmax, line 434: discussion item for the future

modulus/distributed/utils.py
```diff
@@ -472,41 +487,36 @@ def scatter_v_wrapper(
...
+        if r == src:
+            tensor_shape[dim] = sizes[rank]
+        else:
+            tensor_shape[dim] = 0
+        x_recv[r] = torch.empty(tensor_shape, device=tensor.device, dtype=tensor.dtype)
 
-    else:
-        req = dist.irecv(output, src=src, group=group)
-        req.wait()
+    dist.all_to_all(x_recv, x_send, group=group)
```
stadlmax, line 517: we can use all_to_all (natively supporting all_to_all_v) directly to avoid manual recv/send pairs which are also a bit more prone to hangs

modulus/models/gnn_layers/distributed_graph.py
```diff
@@ -82,12 +82,19 @@ class GraphPartition:
     num_local_dst_nodes: int = -1
     # number of edges in local graph on each `partition_rank`
     num_local_indices: int = -1
-    # mapping from local to global ID space (source node IDs)
-    partitioned_src_node_ids_to_global: Optional[torch.Tensor] = None
+    # mapping from global to local ID space (source node IDs)
```
stadlmax, line 85: few small changes and fixes in this file

modulus/utils/graphcast/graph_utils.py
```diff
@@ -74,13 +78,17 @@ def create_heterograph(
...
+        number of nodes for some node types, see dgl.heterograph for more information
 
     Returns
     -------
     DGLGraph
         The dgl Graph.
     """
-    graph = dgl.heterograph({labels: ("coo", (src, dst))}, idtype=dtype)
+    graph = dgl.heterograph(
+        {labels: ("coo", (src, dst))}, num_nodes_dict=num_nodes_dict, idtype=dtype
```
stadlmax, line 90: avoid issue with some of the processor graphs at higher resolution erroring if max(idx) in src/dst does not match the specified number of nodes

test/distributed/test_autograd.py
```diff
@@ -30,6 +30,7 @@
 
 def run_test_scatter_v(rank, world_size):
     os.environ["RANK"] = f"{rank}"
+    os.environ["LOCAL_RANK"] = f"{rank}"
```
stadlmax, line 33: not sure if necessary, but tried a few things after observing several different hangs for local tests

test/models/test_distributed_graph.py
```diff
@@ -0,0 +1,366 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+import pytest
+import torch
+
+from modulus.distributed import DistributedManager
```
stadlmax, line 22: add further unit tests for basic functionality of partitioning, i.e. to test for consistency, e.g. distribute+gather should match initial tensor

examples/cfd/swe_distributed_gnn/pde_dataset.py
```diff
@@ -0,0 +1,146 @@
+# coding=utf-8
```
mnabian, line 1: Need to include `# ignore_header_test` to pass the license test

examples/cfd/swe_distributed_gnn/shallow_water_equations.py
```diff
@@ -0,0 +1,463 @@
+# coding=utf-8
```
mnabian, line 1: Need to include # ignore_header_test to pass the license test

examples/cfd/swe_distributed_gnn/train.py
```diff
@@ -0,0 +1,640 @@
+# coding=utf-8
```
mnabian, line 1: Need to include # ignore_header_test to pass the license test.
Also, would be great to mention that this example was adopted from the torch harmonics package, to give credit to the developers.

examples/cfd/swe_distributed_gnn/train.py
```diff
@@ -0,0 +1,640 @@
...
+            os.path.join(output_dir, run_suffix, "output_data", "metrics.json"), "w"
+        ) as f:
+            json.dump(metrics, f)
+
+    if DistributedManager().distributed:
+        DistributedManager.cleanup()
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
```
mnabian, line 612: Need to use hydra configs instead of argparse. Please let me know if I can help with that.

modulus/models/graphcast/graph_cast_net.py
```diff
@@ -43,6 +43,53 @@
...
+    """Utility Function to get separation intervals for lat-lon
+    grid for partition_sizes of interest.
+
+    Parameters
+    ----------
+    partition_size : int
+        size of graph partition, currently supported sizes: {1, 2, 4, 8, 16}
+    """
+
+    if partition_size == 1:
```
mnabian, line 56: I wish we could make this to work with any `partition_size`.

mnabian:
Tagging @akshaysubr to review the changes to the distributed manager.

examples/cfd/swe_distributed_gnn/train.py
stadlmax: done

examples/cfd/swe_distributed_gnn/train.py
stadlmax: done

examples/cfd/swe_distributed_gnn/shallow_water_equations.py
stadlmax: done

examples/cfd/swe_distributed_gnn/pde_dataset.py
stadlmax: done

modulus/models/graphcast/graph_cast_net.py
stadlmax: I updated the logic to make it more general. Not it should support any `partition_size` but defaults back to just splitting the domain into `partition_size` along the longitude values. 
We should have somewhat balanced distribution schemes for `partition_size in {2, 4, 6, 8, 9, 12, 16, 20, 24}` which likely will cover most of the relevant sizes for now anyways.

modulus/distributed/utils.py
```diff
@@ -400,36 +403,48 @@ def gather_v_wrapper(
...
+            x_send[r] = torch.empty(
+                tensor_shape,
+                dtype=tensor.dtype,
+                device=tensor.device,
+            )
 
-        for r in range(comm_size):
-            if r != dst:
-                req_list[r].wait()
+    dist.all_to_all(x_recv, x_send, group=group)
```
akshaysubr, line 432: Can this be simplified by just using [`dist.all_to_all_single`](https://pytorch.org/docs/stable/distributed.html#torch.distributed.all_to_all_single)?

modulus/distributed/utils.py
akshaysubr: This is a nice improvement. Like in the previous comment, wondering if we can also avoid the explicit `torch.split` above and the for loop over all processes and just use `dist.all_to_all_single`

modulus/models/graphcast/graph_cast_net.py
```diff
@@ -42,6 +43,119 @@
...
+    def _divide(num_lat_chunks: int, num_lon_chunks: int):
+        # divide lat-lon grid into equally-sizes chunks along both latitude and longitude
+        if (num_lon_chunks * num_lat_chunks) != partition_size:
+            raise ValueError(
+                "Can't divide lat-lon grid into grid {num_lat_chunks} x {num_lon_chunks} chunks for partition_size={partition_size}."
+            )
+        # divide latitutude into num_lat_chunks of size 180 / num_lat_chunks
+        # divide longitude into chunks of size 360 / (partition_size / num_lat_chunks)
+        lat_bin_width = 180.0 / num_lat_chunks
+        lon_bin_width = 360.0 / num_lon_chunks
```
akshaysubr, line 68: This is a good solution for the inputs which are on a lat-lon grid, but is this ideal for the icosahedral grid as well once you have `num_lat_chunks > 2`? Since the poles will have much fewer points in the icosahedral grid right? One easy solution for this would be to add a `use_cosine_lat_scaling=False` option in this method and partition based on `cos(lat)` in `[-1,1]` which would be a more even split for the icosahedral grid.

modulus/models/graphcast/graph_cast_net.py
akshaysubr: Can we also have a default where we just take the two closest factors (closest to `sqrt(partition_size)`)?

modulus/models/graphcast/graph_cast_net.py
```diff
@@ -114,16 +228,31 @@ class GraphCastNet(Module):
         passing no process group name leads to a parallelism across the default
         process group. Otherwise, the group size of a process group is expected
         to match partition_size.
-    expect_partitioned_input : bool, default=False,
+    use_lat_lon_partitioning : bool, default=False
```
akshaysubr, line 231: Not super important for GraphCast, since the graph structure is fixed here, but for MeshGraphNet, it might be good to figure out a way to expose the `GraphPartition` object or something similar so one can pass in a custom partition.

test/distributed/test_autograd.py
akshaysubr: That's weird. The default is to set the local rank to `rank % gpus_per_node` if this env var is not defined. If that's failing, maybe there is some other issue in the setup process that's worth looking into?

test/models/graphcast/test_graphcast_snmg.py
```diff
@@ -24,20 +24,23 @@
...
-from torch.nn.parallel import DistributedDataParallel
 
-from modulus.distributed import DistributedManager
+from modulus.distributed import DistributedManager, mark_module_as_shared
 from modulus.models.graphcast.graph_cast_net import GraphCastNet
 
 icosphere_path = get_icosphere_path()
 
 
+torch.backends.cuda.matmul.allow_tf32 = False
```
akshaysubr, line 34: Why is this needed here?

test/models/meshgraphnet/test_meshgraphnet_snmg.py
```diff
@@ -186,22 +198,20 @@ def run_test_distributed_meshgraphnet(rank, world_size, dtype, partition_scheme)
     loss.backward()
 
     out_multi_gpu = model_multi_gpu(nfeat_multi_gpu, efeat_multi_gpu, graph_multi_gpu)
```
akshaysubr, line 200: Can you help me understand how this multi gpu model is working? Looks like I'm missing something about how the partitioned graph gets set in the meshgraphnet model. I don't think I see any attributes of `model_multi_gpu` being set after construction and the constructor doesn't take in any graph data right?

modulus/distributed/utils.py
stadlmax: I thought about that, too. `all_to_all_single`, however, is marked as "experimental and subject to change" at the moment. Hence, I would go for the more stable version for now.

examples/cfd/swe_distributed_gnn/train.py
```diff
@@ -0,0 +1,580 @@
...
+import matplotlib.pyplot as plt
+
+from pde_dataset import PdeDataset
+
+from modulus.distributed import DistributedManager, mark_module_as_shared
+from modulus.models.graphcast.graph_cast_net import GraphCastNet
+
+
+logger = logging.getLogger(__name__)
+OmegaConf.register_new_resolver("rand_uuid", lambda: int(uuid.uuid4()))
```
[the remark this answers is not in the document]
stadlmax: Yes, good point. Main advantage of such a solution is just to be "safe" in cases where you start multiple runs at the same time on a larger cluster. Just using time sometimes doesn't suffice.

modulus/models/graphcast/graph_cast_net.py
stadlmax: MeshGraphNet is a different case anyways. There, the "graph" is only part of the "forward" signature. To run the model in a "graph-parallel" fashion, the user has to define a `GraphPartition` beforehand to setup the `DistributedGraph` properly and then pass this graph structure into the model.

With the current GraphCast design, the graph structure inherently is tied to the model. E.g. the "mesh" itself is often only created within the constructor, i.e. there is no way for the user to define a `GraphPartition` on that graph beforehand anyways.

test/models/graphcast/test_graphcast_snmg.py
stadlmax: I sometimes use this to get tighter bounds on checking the numerics of different implementations. TF32 actually can have quite large errors - especially in a model with multiple layers. 
I have to admit that I am not consistent with this, but especially when starting to develop things, one can identify actual implementation errors quicker than having to figure out whether minor differences come from numerics or bugs.

test/models/meshgraphnet/test_meshgraphnet_snmg.py
stadlmax: You are right. MeshGraphNet in its current design takes in the graph in the forward method. `model_multi_gpu` thus is more or less just the same model as in the single-gpu case on each rank with the only difference that you call `mark_module_as_shared` to register the necessary backward hooks.

The other difference to the single-gpu case is that you specify this "graph" object explicitly with a `graph_partition` and the process group.

```    
graph_multi_gpu = CuGraphCSC(
        offsets.to(manager.device),
        indices.to(manager.device),
        num_nodes,
        num_nodes,
        partition_size=world_size,
        partition_group_name="graph_partition",
        graph_partition=graph_partition,
    )
```

This example actually existed before this PR. I just improved a few minor things here and there after discovering initial bugs in the partitioning logic.

The distributed GraphCastNet on its own actually also doesn't have too many differences from its single-gpu model. As this model "own" more information (e.g. about creating the graph structures in the constructor, converting the layout of input/outputs within the forward pass, etc..), you have to move more of these setups into the model.

examples/cfd/swe_distributed_gnn/train.py
```diff
@@ -0,0 +1,580 @@
...
+        "world_size": world_size,
+        **OmegaConf.to_container(cfg, resolve=True),
+    }
+
+    if dist_manager.rank == 0:
+        logger.info(f"starting to run ...")
+
+    mesh_base_path = os.path.dirname(os.path.realpath(__file__))
+    meshgraph_path = os.path.join(
+        mesh_base_path, f"./icospheres_{cfg.data.mesh_size}.json"
```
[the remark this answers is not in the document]
stadlmax: good point, mohammad just gave me these, we might need a way to figure these things out
note, however, that with the recent GraphCast improvements, these meshes are created on the fly anyways, i.e. this path doesn't exist anymore anyways

modulus/models/graphcast/graph_cast_net.py
stadlmax: In this way, you are correct. We might have to explore these things much more anyways. The thing is, as of now, `num_lat_chunks` only occurs for `partition_size > 8` anyways in which case the number of points per partition for GraphCast probably will be too small anyways compared to necessary communication within the processor across multiple GPU nodes. In this case, we actually might even look at letting some GPUs sit idle and use a smaller partition for the latent mesh itself.

The consideration here is also a different one. The "edges" between the grid and the mesh, mainly are based on e.g. nearest-neighbor relations of grid points on the lat-lon grid and the processor mesh.
By using the same partitioning throughout, you can avoid the processor nodes having to "gather" grid nodes from other partitions, as most of the mesh-nodes' neighbors are within the same lat-lon chunks.

modulus/distributed/manager.py
stadlmax: Probably not, the way the documentation is written, I actually think that this should be the default anyways.

test/distributed/test_autograd.py
stadlmax: Personally, not really sure about that. I guess this is just left over from debugging something else. I'll remove these.

modulus/models/graphcast/graph_cast_net.py
stadlmax: sure, this is actually the better solution for the "generic" case anyways and should even cover all the special cases

examples/cfd/swe_distributed_gnn/pde_dataset.py
```diff
@@ -0,0 +1,147 @@
...
+    def __getitem__(self, index):
+        if self.rank == 0:
+            with torch.inference_mode():
+                with torch.no_grad():
+                    inp, tar = self._get_sample()
+
+                    if self.normalize:
+                        inp = (inp - self.inp_mean) / torch.sqrt(self.inp_var)
+                        tar = (tar - self.inp_mean) / torch.sqrt(self.inp_var)
+                        inp = inp.clone()
```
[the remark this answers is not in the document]
stadlmax: actually, i don't really know, i'll remove them

examples/cfd/swe_distributed_gnn/pde_dataset.py
```diff
@@ -0,0 +1,147 @@
...
+        self.ictype = ictype
+
+    def set_num_examples(self, num_examples=32):
+        self.num_examples = num_examples
+
+    def _get_sample(self):
+        if self.ictype == "random":
+            inp = self.solver.random_initial_condition(mach=0.2)
+        elif self.ictype == "galewsky":
+            inp = self.solver.galewsky_initial_condition()
```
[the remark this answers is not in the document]
stadlmax: I would assume so. Note that most parts of this example have been provided by Mohammad and are more or less taken from the FNO example in torch-harmonics. (this includes e.g. also the naming of files etc.)
Should we modify the example then to make it more reasonable? We can always remove things like e.g. the galewsky initial condition and streamline other things, too.
I mainly focused on "distributing" the given example.

examples/cfd/swe_distributed_gnn/train.py
```diff
@@ -0,0 +1,580 @@
...
+        loss_fn = l2loss
+    elif cfg.model.loss_fn == "l2sphere":
+        loss_fn = l2loss_sphere
+    else:
+        raise ValueError(
+            f"loss_fn={cfg.model.loss_fn} not supported, expected 'l2' or 'l2sphere'."
+        )
+
+    losses = np.zeros(nics)
+    fno_times = np.zeros(nics)
```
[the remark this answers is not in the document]
stadlmax: left over from the initial torch-harmonics example which used fno

examples/cfd/swe_distributed_gnn/train.py
```diff
@@ -0,0 +1,580 @@
...
+            with amp.autocast(enabled=cfg.model.enable_amp):
+                prd = model(inp)
+                for _ in range(cfg.data.num_future):
+                    prd = model(prd)
+                loss = loss_fn(prd, tar, solver=solver, relative=False)
+
+            acc_loss += loss.item() * inp.size(0)
+            optimizer.zero_grad(set_to_none=True)
+            if dist_manager.rank == 0:
+                gscaler.scale(loss).backward()
```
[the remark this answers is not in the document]
stadlmax: As the output is "accumulated" onto rank0, the loss also is only valid on rank0.
If we implement the gather/scatter in a way that ranks where the input is not valid, you actually even only could call the backward on the only rank on which the output is not `None`

examples/cfd/swe_distributed_gnn/train.py
```diff
@@ -0,0 +1,580 @@
...
+    torch.cuda.manual_seed(cfg.seed)
+
+    # initialize DistributedManager and define process group
+    # across which graph is partitioned and model is parallel across
+    # for now, based
+    DistributedManager.initialize()
+    if DistributedManager().distributed:
+        graph_partition_pg_name = "graph_partition"
+        world_size = torch.distributed.get_world_size()
+        DistributedManager.create_process_subgroup(
```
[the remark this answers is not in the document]
stadlmax: The current example e.g. doesn't use DDP, i.e. only model-parallelism. Technically, we don't even need to create the process group for the graph partition. This is rather just to show that one might want to use a dedicated process group for that.
Your call on whether we want to show this in this simplified fashion or by using `ProcessGroupConfig`.

examples/cfd/swe_distributed_gnn/train.py
```diff
@@ -0,0 +1,580 @@
...
+        meshgraph_path=meshgraph_path,
+        static_dataset_path=None,
+        input_res=input_res,
+        input_dim_grid_nodes=3,
+        input_dim_mesh_nodes=3,
+        input_dim_edges=4,
+        output_dim_grid_nodes=3,
+        processor_layers=cfg.model.processor_layers,
+        hidden_dim=cfg.model.hidden_dim,
+        partition_size=dist_manager.group_size(graph_partition_pg_name),
```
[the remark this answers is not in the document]
stadlmax: Yes and no. In a way, yes, group_size always should match the size of the partition group. As far as I can recall, I check for these things in the later distributed primitives.
However, it is more robust do not query the size from the process group as this would enforce the user to define the process group beforehand. In the current implementation, the user has a bit more freedom w.r.t. ordering things in his setup before the main training loop.

examples/cfd/swe_distributed_gnn/train.py
```diff
@@ -0,0 +1,580 @@
...
+    ).to(device=dist_manager.device)
+
+    # since model is "tensor-parallel" in graph-partition
+    # mark model as "shared" which sets gradient hooks and
+    # aggregates gradients in the backward pass accordingly
+    if (
+        dist_manager.is_initialized()
+        and dist_manager.group_size(graph_partition_pg_name) > 1
+    ):
+        mark_module_as_shared(model, graph_partition_pg_name)
```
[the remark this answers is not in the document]
stadlmax: Technically, yes. 
I nevertheless would move this outside of the model to e.g. allow it being part of a larger Wrapper Model for instance. In this case, only a single `mark_module_as_shared` on the outer model should suffice.

examples/cfd/swe_distributed_gnn/train.py
```diff
@@ -0,0 +1,580 @@
...
+            current_lr = optimizer.param_groups[0]["lr"]
+            metrics["epochs"].append(epoch)
+            metrics["current_lr"].append(current_lr)
+            metrics["epoch_time"].append(epoch_time)
+            metrics["acc_loss"].append(acc_loss)
+            metrics["val_loss"].append(valid_loss)
+
+        scheduler.step()
+
+    train_time = time.time() - train_start
```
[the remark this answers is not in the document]
stadlmax: The main saves checkpoints after the training loop.
As the model is the same on each device, it is actually rather simple. Rank0 saves the state_dict, every rank can load this state_dict.
What we don't consider at the moment, is the use of Adam in a setting with shared weights, but this is a different story.
Do we want to complicate the example to also allow storing intermediate training checkpoints?

modulus/distributed/utils.py
stadlmax: see comment above w.r.t. all_to_all_single being marked as experimental and subject to change at the moment

modulus/distributed/autograd.py
```diff
@@ -66,17 +66,16 @@ def forward(
...
-
-        if needs_grad:
-            grad_tensor = all_gather_v_bwd_wrapper(
-                grad_output,
-                ctx.sizes,
-                dim=ctx.dim,
-                use_fp32=ctx.use_fp32,
-                group=ctx.group,
-            )
+        grad_tensor = all_gather_v_bwd_wrapper(
```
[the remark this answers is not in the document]
stadlmax: Technically, you could have a scenario
- 4 ranks with tensors, only two of them require gradients
- AllGatherV (similar also just GatherV)
- then backward pass
- ranks that have received information typically don't know if their original "senders" require a gradient or not
- all_gather_v_bwd_wrapper, however, is a collective operation, for now it is simpler to let all ranks participate
- finally, only return the gradient on the ranks that actually need it

Does this make sense? I am personally not sure how relevant this is in practice though.

modulus/distributed/utils.py
stadlmax: closing this with opening https://github.com/NVIDIA/modulus/issues/520

examples/cfd/swe_distributed_gnn/train.py
```diff
@@ -0,0 +1,580 @@
...
+import torch
+from tqdm.auto import tqdm
+from tqdm.contrib.logging import logging_redirect_tqdm
+from torch.utils.data import DataLoader
+from torch.cuda import amp
+
+import numpy as np
+import matplotlib.pyplot as plt
+
+from pde_dataset import PdeDataset
```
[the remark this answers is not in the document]
stadlmax: taken from the torch-harmonics example
- technically, it originally was intended to be somewhat general, however, only the shallow-water-equation is implemented anyways
@mnabian what do you think?

modulus/distributed/manager.py
stadlmax: checked it again, it is only available from PyTorch 2.3.0 onwards, if possible, the code now set this id

test/distributed/test_autograd.py
stadlmax: checked these again, the issue was with LOCAL_RANK being set but being empty
in this case, `int(os.environ.get(...)` throws a type error and `initialize_env` is aborted, the manager tries unsuccessfully to `initialize_slurm` or `initialize_open_mpi`. As these are not even tried (because the "guarding" environment variables don't exist), `initialize` just returns without setting up anything. The first NCCL primitive then errors out with the default process group not having been initialized.

```        
if initialization_method is None:
            try:
                DistributedManager.initialize_env()
            except TypeError:
                if "SLURM_PROCID" in os.environ:
                    DistributedManager.initialize_slurm(port)
                elif "OMPI_COMM_WORLD_RANK" in os.environ:
                    DistributedManager.initialize_open_mpi(addr, port)
```

I have removed all LOCAL_RANK occurrences in all tests but also updated the DistributedManager to cover this case in a more robust fashion. An empty LOCAL_RANK now doesn't lead to things erroring out but to assuming the default of `rank % device_count`.

modulus/models/graphcast/graph_cast_net.py
stadlmax: done

examples/cfd/swe_distributed_gnn/train.py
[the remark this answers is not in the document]
stadlmax: Actually, this was probably me trying to improve things without testing. Revisited this again, this leads to hangs. However, this then also shows that having empty "dummy" tensors is important to e.g. ensure that all ranks call `.backward`.

examples/cfd/swe_distributed_gnn/train.py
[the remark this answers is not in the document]
stadlmax: updated the PR to the more recent main branch, as these have been removed, there shouldn't now also be any notion of `meshgraph_path` in the example

modulus/distributed/utils.py
akshaysubr: That's the case for `all_to_all` too though: https://pytorch.org/docs/stable/distributed.html#torch.distributed.all_to_all

test/distributed/test_autograd.py
akshaysubr: Ah, it's about `LOCAL_RANK` being defined but empty. That makes sense.

> DistributedManager to cover this case in a more robust fashion

Yeah, I agree this is needed. Can you ensure that the changes you are introducing are not already covered in #475 or causes any conflicts with that PR?

examples/cfd/swe_distributed_gnn/pde_dataset.py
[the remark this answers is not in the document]
akshaysubr: I think it is okay to leave it as is. Maybe worth adding a small comment here just to clarify this is only an example and not a fully fleshed out recipe?

mnabian (approved):

examples/cfd/swe_distributed_gnn/train.py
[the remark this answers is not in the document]
akshaysubr: I think it would be good to use `ProcessGroupConfig` to define the graph partitions in the model implementation as well as move the mark as shared into the model (though it's trivial in this case). That way, we can also use this model and example as a template for other users that want to build model parallel architectures. Thoughts?

examples/cfd/swe_distributed_gnn/train.py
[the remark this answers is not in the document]
akshaysubr: Probably not needed to add checkpointing support to this PR. But I feel like that is another challenge with a model parallel recipe that users will encounter and will have questions about.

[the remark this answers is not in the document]
akshaysubr: Let's open a separate issue to add checkpointing to this example and update the modulus checkpointing utilities to support this use case.

examples/cfd/swe_distributed_gnn/train.py
```diff
@@ -0,0 +1,580 @@
...
+    if dist_manager.rank == 0:
+        logger.info(f"number of trainable params: {num_params}")
+    metrics["num_params"] = num_params
+
+    if cfg.load_checkpoint:
+        if cfg.checkpoint_dir is None:
+            raise ValueError(
+                "load_checkpoint=True but cfg.checkpoint_dir is not set, abort."
+            )
+        model.load_state_dict(
```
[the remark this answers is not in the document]
akshaysubr: It is for sure, but I wanted to add that so the modulus checkpointing utilities properly support model parallelism. But let's not worry about that in this PR and add it in a separate PR

modulus/distributed/autograd.py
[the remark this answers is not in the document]
akshaysubr: That makes sense.

modulus/models/graphcast/graph_cast_net.py
akshaysubr: Yeah, makes sense that this is only needed for `partition_size > 8`. Let's add a comment here so we know what to do if this is needed down the line?

modulus/models/graphcast/graph_cast_net.py
akshaysubr: That makes sense and why I said maybe not super important for GraphCast. I guess for now, having booleans to control the partitioning scheme is fine.

modulus/distributed/utils.py
stadlmax: True, realized that just now. 
`all_to_all_single` however still comes with the restriction of splits only working alongside dim0. An explicit split currently would be more flexible for now. Hence, I would vote for keeping the current approach as is.

test/distributed/test_autograd.py
stadlmax: oh wow, this PR is still open? 
indeed, it introduces a "warn" in the "else"-branch which later leads to an exception afterwards
let's merge this PR first and i'll resolve the conflicts

examples/cfd/swe_distributed_gnn/pde_dataset.py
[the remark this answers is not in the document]
stadlmax: Looked a bit into this, this "static" initial condition is one that is often used as a reference of some kind for testing things. I added a comment that this can be used for valuation of trained models for instance.

examples/cfd/swe_distributed_gnn/train.py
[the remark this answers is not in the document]
stadlmax: slightly cleaned up things and renamed it to ShallowWaterPDEDataset, it implemented only this PDE anyways

modulus/distributed/utils.py
akshaysubr: Makes sense. We can always update this to `all_to_all_single` down the line if we feel like it might help with perf.

akshaysubr (approved):

PR #401, opened 2024-03-25
Patch-based CorrDiff Integration - 2

tge25:
# Modulus Pull Request

## Description
Add the option to incorporate the regression's output as an additional conditioning of the diffusion model.
Allow arbitrary length of sinusoid positional embedding. 
Enable learnable positional embedding.
Move positional embedding to network architecture
Update the config file for patch-based training and generation.

## Checklist


examples/generative/corrdiff/conf/config_generate.yaml
```diff
@@ -89,6 +90,7 @@ dataset:
   gridtype: "sinusoidal"
```
nbren12, line 90: It's unclear what gridtypes are supported.

examples/generative/corrdiff/datasets/img_utils.py
```diff
@@ -17,6 +17,7 @@
 
 import numpy as np
 import torch
+import warnings
 
 
 def reshape_fields(
```
nbren12, line 23: It's a bit confusing to have the position encoding in a function called "reshape fields". Perhaps this functions does too much and should be split into two.

examples/generative/corrdiff/generate.py
```diff
@@ -226,15 +234,18 @@ def generate_fn(image_lr):
                 with nvtx.annotate("regression_model", color="yellow"):
                     image_reg = generate(
                         net=net_reg,
-                        img_lr=image_lr_patch,
+                        img_lr=image_lr_patch[:, 0:16],
```
nbren12, line 237: Magic number. Should this be removed?

examples/generative/corrdiff/train.py
```diff
@@ -265,6 +269,8 @@ def main(cfg: DictConfig) -> None:
     )
 
     (img_shape_y, img_shape_x) = dataset.image_shape()
+    img_shape_x = dataset_cfg["img_shape_x"]
+    img_shape_y = dataset_cfg["img_shape_y"]
```
nbren12, line 273: What is wrong with "image_shape"? Should it be fixed?

modulus/metrics/diffusion/loss.py
```diff
@@ -484,12 +490,20 @@ def __call__(self, net, img_clean, img_lr, labels=None, augment_pipe=None):
...
+            patched_train = True
+        else:
+            patched_train = False
         rnd_normal = torch.randn([img_clean.shape[0], 1, 1, 1], device=img_clean.device)
         sigma = (rnd_normal * self.P_std + self.P_mean).exp()
         weight = (sigma**2 + self.sigma_data**2) / (sigma * self.sigma_data) ** 2
 
         # augment for conditional generaiton
-        img_tot = torch.cat((img_clean, img_lr), dim=1)
+        img_tot = torch.cat((img_clean, img_lr[:, :16]), dim=1)
```
nbren12, line 506: magic number? What is :16?

modulus/metrics/diffusion/loss.py
```diff
@@ -526,7 +561,7 @@ def __call__(self, net, img_clean, img_lr, labels=None, augment_pipe=None):
 
             # global interpolation
             input_interp = torch.nn.functional.interpolate(
-                img_lr[:, 0 : c_in - 4],
+                img_lr[:, 0:12],
```
nbren12, line 564: more magic numbers? What is 12?

modulus/metrics/diffusion/loss.py
```diff
@@ -445,6 +445,9 @@ def __init__(
         P_mean: float = 0.0,
         P_std: float = 1.2,
         sigma_data: float = 0.5,
+        pos_embed: int = 4,
+        hr_mean_conditioning: bool = False,
+        gridtype: str = "sinusoidal",
```
nbren12, line 450: Why does the loss function need to know about this? Seems these things could all be encapsulated in the network.

nbren12 (changes requested):
Thanks for adding this important feature into main @tge25.

I'm finding the flow of the configuration a bit hard to follow since there are a lot of new conditional statements added. It also seems like there is too much coupling in the code since you needed to add new if statements in the loss, training_loop and datasets.

 A feature like positional encoding feature should only need to touch one place in the code ideally. I think this will be less tightly coupled, if we move more of the position encoding logic into the architecture.

Also, doesn't seem like there any new unit tests of the changes to modulus/

examples/generative/corrdiff/conf/config_generate.yaml
```diff
@@ -56,8 +56,9 @@ use_torch_compile: True
 ## Data options
 patch_shape_x: 448
 patch_shape_y: 448
-overlap_pix: 4 
-boundary_pix: 2
+overlap_pix: 0
```
mnabian, line 59: Can we add a short description for each option?

mnabian:
@tge25 CI failed, please black format your code

modulus/models/diffusion/song_unet.py
```diff
@@ -169,6 +171,17 @@ def __init__(
             )
 
         super().__init__(meta=MetaData())
+        if gridtype[:9] == "learnable":
+            pos_dim = int(gridtype[10:])
```
mnabian, line 175: Same here

modulus/models/diffusion/song_unet.py
```diff
@@ -149,6 +149,8 @@ def __init__(
         encoder_type: str = "standard",
         decoder_type: str = "standard",
         resample_filter: List[int] = [1, 1],
+        gridtype: str = "sinusoidal",
+        task: str = "regression",
```
mnabian, line 153: We should keep these network architectures as generic as possible. Instead of passing the task arg, consider directly passing the pos_embd.

examples/generative/corrdiff/training/training_loop.py
```diff
@@ -242,12 +256,16 @@ def training_loop(
...
+            if patch_shape_x != img_shape_x:
+                g["lr"] *= 0.8 ** ((cur_nimg - lr_rampup_kimg * 1000) // 4e6)
             wb.log({"lr": g["lr"]}, step=cur_nimg)
         for param in net.parameters():
             if param.grad is not None:
                 torch.nan_to_num(
                     param.grad, nan=0, posinf=1e5, neginf=-1e5, out=param.grad
                 )
+        if patch_shape_x != img_shape_x:
+            grad_norm = torch.nn.utils.clip_grad_norm_(net.parameters(), 1e5)
```
mnabian, line 268: Can we make the grad norm threshold configurable?

examples/generative/corrdiff/training/training_loop.py
```diff
@@ -242,12 +256,16 @@ def training_loop(
             g["lr"] = optimizer_kwargs["lr"] * min(
                 cur_nimg / max(lr_rampup_kimg * 1000, 1e-8), 1
             )  # TODO better handling (potential bug)
+            if patch_shape_x != img_shape_x:
+                g["lr"] *= 0.8 ** ((cur_nimg - lr_rampup_kimg * 1000) // 4e6)
```
mnabian, line 260: Where is this 0.8 coming from? Is that empirical? Can we make it configurable?

modulus/models/diffusion/song_unet.py
```diff
@@ -169,6 +171,17 @@ def __init__(
             )
 
         super().__init__(meta=MetaData())
+        if gridtype[:9] == "learnable":
+            pos_dim = int(gridtype[10:])
+            print("learnable %s positional embedding" % gridtype[10:])
```
mnabian, line 176: These type of print statements should not be in the network architecture. Consider adding that to the training script instead.

examples/generative/corrdiff/train.py
tge25: Nothing wrong. I forget that img_shape has been defined. Deleted 272 and 273 and resolving this conversation. Thanks

examples/generative/corrdiff/conf/config_generate.yaml
tge25: Added supported gridtypes

examples/generative/corrdiff/generate.py
tge25: That's the tricky part. Thanks for bringing it up. The pretrained regression model takes 12 channels + 4 positional embeddings as input. If a different positional embedding is utilized in the diffusion model, we'll need to retain both positional embeddings in the dataset and select them as required for regression and diffusion. Hard-coding isn't an elegant solution, but the diffusion training code lacks information regarding the positional embedding of the regression. Should we consider making it configurable?

nbren12: >  If a different positional embedding is utilized in the diffusion model, we'll need to retain both positional embeddings in the dataset and select them as required for regression and diffusion.

will this be fixed by moving the position encoding into the network rather than the dataset? See also https://github.com/NVIDIA/modulus/pull/401#discussion_r1537920232

tge25:
> Thanks for adding this important feature into main @tge25.
> 
> I'm finding the flow of the configuration a bit hard to follow since there are a lot of new conditional statements added. It also seems like there is too much coupling in the code since you needed to add new if statements in the loss, training_loop and datasets.
> 
> A feature like positional encoding feature should only need to touch one place in the code ideally. I think this will be less tightly coupled, if we move more of the position encoding logic into the architecture.
> 
> Also, doesn't seem like there any new unit tests of the changes to modulus/



> Thanks for adding this important feature into main @tge25.
> 
> I'm finding the flow of the configuration a bit hard to follow since there are a lot of new conditional statements added. It also seems like there is too much coupling in the code since you needed to add new if statements in the loss, training_loop and datasets.
> 
> A feature like positional encoding feature should only need to touch one place in the code ideally. I think this will be less tightly coupled, if we move more of the position encoding logic into the architecture.
> 
> Also, doesn't seem like there any new unit tests of the changes to modulus/

Thanks @nbren12 for all the comments. I have cleaned up the code and moved the positional embedding to network architecture. Some unit tests have also been added.

examples/generative/corrdiff/training/training_loop.py
```diff
@@ -242,12 +259,18 @@ def training_loop(
...
             )  # TODO better handling (potential bug)
+            if patch_shape_x != img_shape_x:
+                g["lr"] *= lr_decay ** ((cur_nimg - lr_rampup_kimg * 1000) // 5e6)
             wb.log({"lr": g["lr"]}, step=cur_nimg)
         for param in net.parameters():
             if param.grad is not None:
                 torch.nan_to_num(
                     param.grad, nan=0, posinf=1e5, neginf=-1e5, out=param.grad
                 )
+        if patch_shape_x != img_shape_x:
```
mnabian, line 270: What is the rationale behind supporting gradient clipping only for patching? I think we should support gradient clipping even without patching. Something like this:

modulus/models/diffusion/song_unet.py
```diff
@@ -364,3 +370,57 @@ def forward(self, x, noise_labels, class_labels, augment_labels=None):
...
+            grid_list = []
+            grid_x, grid_y = np.meshgrid(
+                np.linspace(0, 2 * np.pi, self.img_resolution),
+                np.linspace(0, 2 * np.pi, self.img_resolution),
+            )
+            for freq in freq_bands:
+                for p_fn in [np.sin, np.cos]:
+                    grid_list.append(p_fn(grid_x * freq))
+                    grid_list.append(p_fn(grid_y * freq))
+                print(freq, (grid_x * freq)[12, 12], p_fn(grid_x * freq)[12, 12])
```
mnabian, line 421: Can we remove this print statement?

test/models/diffusion/test_song_unet.py
```diff
@@ -133,6 +144,41 @@ def test_song_unet_constructor(device):
     except ValueError:
         pass
 
+    try:
```
nbren12, line 147: This try-except formulation is tricky to follow. Better to use https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.raises

```
with pytest.raises(Exception):
    model = ...
```

nbren12 (changes requested):
Thanks for the changes. 

I had some nits about the tests.

I terms of required changes, I think the positional embedding can be encapsulated better within the architecture.

modulus/models/diffusion/song_unet.py
mnabian, line 1: Hi @tge25 , this changes the default behavior of `SongUNet`. I think we should keep this network architecture generic, as it should correspond to the model proposed in Song et al. paper. Another PR for turbulence super resolution is also using this architecture, and your changes will most likely break it: https://github.com/NVIDIA/modulus/pull/398. Can we instead make make another class that inherits from this class and overwrites the `forward` method, something like class `SongUNetPosEmbd(SongUNet)`?

test/models/diffusion/test_song_unet.py
```diff
@@ -133,6 +144,41 @@ def test_song_unet_constructor(device):
...
+        model = UNet(
+            img_resolution=img_resolution,
+            in_channels=in_channels,
+            out_channels=out_channels,
+            gridtype=None,
+        ).to(device)
```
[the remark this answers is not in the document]
tge25: Other test codes utilize the decorator @pytest.mark.parametrize("device", ["cuda:0", "cpu"]) to conduct tests on both CPU and GPU. Shall we retain it for the sake of consistency?

modulus/metrics/diffusion/loss.py
```diff
@@ -406,6 +406,11 @@ def __call__(self, net, img_clean, img_lr, labels=None, augment_pipe=None):
         y = y_tot[:, : img_clean.shape[1], :, :]
         y_lr = y_tot[:, img_clean.shape[1] :, :, :]
 
+        # add positional embedding
+        pos_embd = net.model.pos_embd.expand(img_lr.shape[0], -1, -1, -1).to(
+            device=img_clean.device
+        )
+        y_lr = torch.cat((y_lr, pos_embd), dim=1)
```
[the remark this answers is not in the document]
tge25: Thanks for this comment. I have added a new submodule and moved the evaluation of positional embedding to the forward pass, so resolving this issue

examples/generative/corrdiff/generate.py
```diff
@@ -182,7 +186,7 @@ def main(cfg: DictConfig) -> None:
     # load regression network, move to device, change precision
     if not diffusion_only:
         logger0.info(f'Loading network from "{reg_ckpt_filename}"...')
-        net_reg = Module.from_checkpoint(reg_ckpt_filename)
+        net_reg = Module_reg.from_checkpoint(reg_ckpt_filename)
```
mnabian, line 189: I think `Module_reg` is not defined

test/models/diffusion/test_song_unet.py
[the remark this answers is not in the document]
nbren12: This is just a nit. but I would assume those tests are actually passing some data to the unet. 

By including .to, you are now testing two things. 1) whether the constructors fails as you desire and 2) whether .to fails. This makes the test more ambiguous. For example (1) could actually fail to raise an error, but by using the wrong device (2) would raise an error and the overall check would pass. This is confusing. 

 It's best to test the thing you care about...which is (1). I think we  trust .to to work.

examples/generative/corrdiff/training/training_loop.py
```diff
@@ -264,12 +283,17 @@ def training_loop(
             g["lr"] = optimizer_kwargs["lr"] * min(
                 cur_nimg / max(lr_rampup_kimg * 1000, 1e-8), 1
             )  # TODO better handling (potential bug)
+            if patch_shape_x != img_shape_x:
```
mnabian, line 286: Decaying learning rate should be independent of whether using patching or not.

examples/generative/corrdiff/generate.py
tge25: Ah, forgot to change it back. Resolved

examples/generative/corrdiff/training/training_loop.py
```diff
@@ -264,12 +283,17 @@ def training_loop(
...
+                g["lr"] *= lr_decay ** ((cur_nimg - lr_rampup_kimg * 1000) // 5e6)
             wb.log({"lr": g["lr"]}, step=cur_nimg)
         for param in net.parameters():
             if param.grad is not None:
                 torch.nan_to_num(
                     param.grad, nan=0, posinf=1e5, neginf=-1e5, out=param.grad
                 )
+        grad_norm = torch.nn.utils.clip_grad_norm_(
```
mnabian, line 294: Note that this is changing the default behavior. Previously grad norm was not used but now you are using a non-zero threshold even without patching.

test/models/diffusion/test_song_unet_pos_embd.py
```diff
@@ -0,0 +1,231 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2024 NVIDIA CORPORATION & AFFILIATES.
```
nbren12, line 1: needs test where global_index is passed.

test/models/diffusion/test_song_unet.py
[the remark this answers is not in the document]
tge25: Yes, that makes sense. Removed ".to(device)" for test_fails_if_grid_is_invalid

modulus/models/diffusion/song_unet.py
```diff
@@ -364,3 +364,264 @@ def forward(self, x, noise_labels, class_labels, augment_labels=None):
...
+
+    def positionalEmbedding(self):
+        if self.N_grid_channels == 0:
+            return None
+        elif self.gridtype == "learnable":
+            grid = torch.nn.Parameter(
+                torch.randn(
+                    self.N_grid_channels, self.img_resolution, self.img_resolution
+                )
+            )
```
nbren12, line 584: Thinking forward to the ICON hpx work. It would be more flexible to have the position embedding be a 2D array with shape (channels, n_global_grid), and `global_index` be a (batch, patch, patch) array of indices into this array.

I worry we'll need to have another backwards incompatible change to support the HPX work without this change. Is it hard?

examples/generative/corrdiff/training/training_loop.py
tge25: I think we had this conversion during the previous round of review: whether or not we should support gradient clipping and learning rating scheduling for non patch-based training. By default, the code does not use. Shall we set a sufficiently large value for gradient clipping threshold in the config file when is not used in order to keep it consistent?

modulus/metrics/diffusion/loss.py
```diff
@@ -546,6 +560,14 @@ def __call__(self, net, img_clean, img_lr, labels=None, augment_pipe=None):
...
             )
+            global_index = torch.zeros(
+                b * self.patch_num,
+                2,
+                self.patch_shape_x,
+                self.patch_shape_y,
+                dtype=torch.int,
+                device=img_clean.device,
+            )
             for i in range(self.patch_num):
```
nbren12, line 571: **not for this PR since this code already exists**

Eventually, the loss should not know anything about patches and just treat them like batches (hey this rhymes). This can be achieved by moving the patching  logic either to the dataloader or training_loop.py. global_index can then be passed to the loss object.  Let's open an issue for this refactor.

examples/generative/corrdiff/datasets/cwb.py
```diff
@@ -472,34 +447,34 @@ def __len__(self):
...
         """Get longitude values from the dataset."""
         lon = self._dataset.longitude()
-        return lon if self.train else lon[..., : self.crop_size_y, : self.crop_size_x]
+        return lon if self.train else lon[..., : self.img_shape_y, : self.img_shape_x]
 
     def latitude(self):
         """Get latitude values from the dataset."""
         lat = self._dataset.latitude()
-        return lat if self.train else lat[..., : self.crop_size_y, : self.crop_size_x]
+        return lat if self.train else lat[..., : self.img_shape_y, : self.img_shape_x]
```
nbren12, line 455: **For later**

Not new code, but it is odd for a few reasons
1. different answers based on training vs not. I understand why, but to me lat is lat.
2. the crops always start at (0,0) is the also true in the __getitem__ below?

Overall, for the case where lat/lon may depend on patching/cropping etc, I think it's easier if these coordinate values are returned alongside the data or if you modify these coordinates values to take global_index as an argument.

```
def latitude(self, global_index):
    ...
```

examples/generative/corrdiff/training/training_loop.py
tge25: changed

examples/generative/corrdiff/training/training_loop.py
nbren12: I'm okay with some backwards breaking changes to the training.

nbren12 (changes requested):
@tge25 Thanks for implementing the changes with the position embedding.

Loss.py is a bit messier now, but I think there is a clear path to refactor the patching logic into the dataset or training_loop.py where IMO it is a much better fit.

I did suggest some changes that will enable future support for unstructured/global grids. Would be nice to avoid more backwards incompatible changes to the networks.

examples/generative/corrdiff/training/training_loop.py
tge25: Never mind. I have set the default config to not using graident clipping

modulus/metrics/diffusion/loss.py
tge25: Agreed. However, it could be tricky since the patching operation should be done after the regression inference, while the regression inference is inside the loss function

test/models/diffusion/test_song_unet_pos_embd.py
tge25: added tests

modulus/models/diffusion/song_unet.py
nbren12: agreed to defer this to future work.

examples/generative/corrdiff/datasets/cwb.py
tge25: In the previous version, starting location is defined by a random number generator if crop size is smaller than image shape, which seems to be a prototype of patch-based training. In the future we could consider move the patching operation back to dataloader and let the loader return coordinate values. The patching operation for regression's output could be performed based on returned global indices. I have added this to issue #448 as a reminder. Resolving this conversation for now.

mnabian (approved):

nbren12 (approved):
approving too. thanks! 62 comments!

Issue #412, opened 2024-04-04
🐛[BUG]: accumulated loss should be divided by accumulation steps to get the mean loss for wandb report 


yairchn:
### Version

0.5.0

### On which installation method(s) does this occur?

Pip

### Describe the issue

In the training loop, an accumulated loss is computing in an additive manner [here](https://github.com/NVIDIA/modulus/blob/9b0b7cc71fb280c028e09ee56726c0caa387c63b/examples/generative/corrdiff/training/training_loop.py#L236) across all steps in the `num_accumulation_rounds`. This lose should be divided by the value of `num_accumulation_rounds` to get the mean rather than the sum of loss to be recorded to wandb:
`loss_accum += loss/num_accumulation_rounds`

ktangsali:
Closing as fixed.

PR #413, opened 2024-04-04
Add online validation loss with wandb 

yairchn:
# Modulus Pull Request

## Description
This PR adds a validation loss computation to training loop. The new loss is recorded both in the report stats and to wandb. While working on this PR I oped an issue for wandb loss saving #412 which this PR handles as well. 

Also closes https://github.com/NVIDIA/modulus/issues/416

## Checklist

- [?] New or existing tests cover these changes.


examples/generative/corrdiff/training/training_loop.py
```diff
@@ -231,11 +237,11 @@ def training_loop(
                     labels=labels,
                     augment_pipe=augment_pipe,
                 )
-                training_stats.report("Loss/loss", loss)
+                training_stats.report("Loss/training loss", loss)
```
yairchn, line 240: this name change will break some people loss plotting script, but I believe we should make this change to avoid ambiguity

examples/generative/corrdiff/train.py
```diff
@@ -84,6 +85,7 @@ def main(cfg: DictConfig) -> None:
     wandb_mode = getattr(cfg, "wandb_mode", "disabled")
     tick = getattr(cfg, "tick", 1)
     dump = getattr(cfg, "dump", 500)
+    validation_dump = getattr(cfg, "validation_dump", 500)
```
mnabian, line 88: Can we make this configurable by adding a `validation_dump` item to  `config_train_base.yaml`?

examples/generative/corrdiff/training/training_loop.py
```diff
@@ -71,7 +73,11 @@ def training_loop(
     patch_shape_y=448,
     patch_num=1,
     wandb_mode="disabled",
+    wandb_project="Modulus-Generative",
```
mnabian, line 76: why are we adding the wandb configs here? If you want to make these configurable, you'd also need to add them to the hydra config file and parse them.

examples/generative/corrdiff/training/training_loop.py
```diff
@@ -249,6 +255,29 @@ def training_loop(
                     param.grad, nan=0, posinf=1e5, neginf=-1e5, out=param.grad
                 )
         optimizer.step()
+        valid_loss_accum = 0
+        num_validation_steps  = 10
```
mnabian, line 259: Can we make this configurable?

examples/generative/corrdiff/training/training_loop.py
nbren12: maybe keep the name the same, and but use a new name for the validation loss.

mnabian (approved):

examples/generative/corrdiff/train.py
```diff
@@ -260,10 +271,20 @@ def main(cfg: DictConfig) -> None:
...
+    # create a training dataset
     (dataset, dataset_iterator) = init_dataset_from_config(
         dataset_cfg, data_loader_kwargs, batch_size=batch_size_gpu, seed=seed
     )
-
+    
+    # create a valiadtion dataset
+    validation_dataset_cfg = copy.deepcopy(dataset_cfg)
+    validation_dataset_cfg['all_times'] = False
+    validation_dataset_cfg['train'] = False
```
jleinonen, line 282: This will only work with dataloaders that have the parameters `all_times` and `train` (CWB will work but any future dataloaders might be configured differently).

yairchn: @jleinonen I have modified the use of this call by moving the sprint to a wrapper function inside dataset, where in train.py the default is false and the behavior should ignore the validation test split if none is provided in the yaml

modulus/launch/logging/wandb.py
```diff
@@ -94,7 +94,7 @@ def initialize_wandb(
         wandb_name = f"{name}_{time_string}"
 
     if not os.path.exists(wandb_dir):
-        os.makedirs(wandb_dir)
+        os.makedirs(wandb_dir, exist_ok=True)
```
yairchn, line 97: oddly I encountered race conditions here without this flag `exist_ok=True`

yairchn:
I am testing this PR in a small training in an interactive session to this [WANDB project ](https://wandb.ai/nv-research-climate/diffusion/runs/zmcm69ku?nw=nwuseryairchn)

mnabian:
Reviewed the most recent changes. LGTM

mnabian:
@yairchn black test failed

jleinonen:
LGTM - left some optional improvements but these can be deferred to later PRs too.

examples/generative/corrdiff/datasets/dataset.py
```diff
@@ -28,6 +28,45 @@
...
+    Returns:
+    - Tuple[base.DownscalingDataset, Iterable, Optional[base.DownscalingDataset], Optional[Iterable]]: A tuple containing the training dataset and iterator, and optionally the validation dataset and iterator if train_test_split is True.
+    """
+
+    config = copy.deepcopy(dataset_cfg)
+    config.pop('train_test_split', None)
+    (dataset, dataset_iter) = init_dataset_from_config(
+        config, dataloader_cfg, batch_size=batch_size, seed=seed
+    )
+    if train_test_split:
```
daviddpruitt, line 57: Instead of having a seperate flag for splits can we check if valid_years has values? This could also potentially push the logic all into the dataset.

modulus/launch/logging/wandb.py
daviddpruitt: I've seen this as well. There should be a check to make sure only one process is in charge of initialization. I don't know where the responsibility should be for that though.

yairchn: @daviddpruitt  its a bit tricky as the validation set mightn't be specific years, people might have a random sample of the data s validation and the exact details would be in the dataloader. I think I would leave the current flag but we can change this in the future

examples/generative/corrdiff/datasets/dataset.py
yairchn: @daviddpruitt  its a bit tricky as the validation set mightn't be specific years, people might have a random sample of the data s validation and the exact details would be in the dataloader. I think I would leave the current flag but we can change this in the future

jleinonen (changes requested):

modulus/launch/logging/wandb.py
mnabian: Thanks for checking with me, I leave that up to you and Jussi to decide

jleinonen (approved):

PR #423, opened 2024-04-08
Add AMP support to CorrDiff

jleinonen:
# Modulus Pull Request

## Description
Add Automatic Mixed Precision (AMP) support to CorrDiff training:
- Add an `fp_mode` config parameter that can be one of `{'fp32', 'fp16', 'amp'}`. This supersedes the `fp16` config parameter, thought the latter is still supported if `fp_mode` is not present. `fp_mode == 'fp32'` and `fp_mode == 'fp16'` correspond to `fp16 == False` and `fp16 == True` respectively, while `fp_mode == 'amp'` activates AMP.
- Execute training forward pass in a `torch.autocast` environment that is enabled if `fp_mode == 'amp'`.
- Disable certain datatype checks in U-Nets if AMP is enabled (these would otherwise cause an error).

## Checklist


examples/generative/corrdiff/training/training_loop.py
```diff
@@ -224,13 +225,16 @@ def training_loop(
...
 
-                loss = loss_fn(
-                    net=ddp,
-                    img_clean=img_clean,
-                    img_lr=img_lr,
-                    labels=labels,
-                    augment_pipe=augment_pipe,
-                )
+                with torch.autocast(
+                    "cuda", dtype=torch.bfloat16, enabled=(fp_mode == "amp")
```
mnabian, line 229: Does it make sense to make the `dtype` configurable, to support `fp16` as well?

examples/generative/corrdiff/train.py
```diff
@@ -246,7 +253,7 @@ def main(cfg: DictConfig) -> None:
     logger0.info(f"Preconditioning & loss:  {precond}")
     logger0.info(f"Number of GPUs:          {dist.world_size}")
     logger0.info(f"Batch size:              {c.batch_size_global}")
-    logger0.info(f"Mixed-precision:         {c.network_kwargs.use_fp16}")
```
mnabian, line 249: Should we also remove it from here? https://github.com/NVIDIA/modulus/blob/b10fe30c8bb18aade11650817731ece8fe95ec48/examples/generative/corrdiff/train.py#L218

mnabian (approved):
LGTM. Left a couple of minor comments. Thanks for the PR!

mnabian:
Closes https://github.com/NVIDIA/modulus/issues/424

examples/generative/corrdiff/train.py
jleinonen: (Later we can discuss whether it makes sense to keep the existing FP16 mode which does quite a bit of manual configuration inside the U-Nets. Removing it could simplify the U-Net code nicely. But I wouldn't try to remove it in this PR as this requires feedback from other devs).

examples/generative/corrdiff/training/training_loop.py
jleinonen: It's easy to add technically (maybe using `fp_mode == 'amp-fp16'` and `'fp_mode' == 'amp-bf16'`). However I haven't tested AMP with FP16, while I have fully trained CorrDiff with BF16 so I'm confident it'll work ok. Is it better to add it now anyway?

mnabian: No we can keep the code as is. But would be nice to clarify in the README or even as a comment [here](https://github.com/NVIDIA/modulus/blob/b10fe30c8bb18aade11650817731ece8fe95ec48/examples/generative/corrdiff/conf/config_train_base.yaml#L44) that the AMP implementation uses `bfloat16`

examples/generative/corrdiff/conf/config_train_base.yaml
```diff
@@ -40,8 +40,9 @@ augment: 0.0
   # Augment probability
 
 ## Performance options
-fp16: false
-  # Enable mixed-precision training
+fp_mode: fp32
```
akshaysubr, line 43: Would prefer renaming this to `tensor_core_mode` or `matmul_fp_mode` since this option should not affect group norm and other precision sensitive non-matmul layers.

examples/generative/corrdiff/training/training_loop.py
akshaysubr, line 229: Why always enable `bfloat16`? This should be configurable ideally.

akshaysubr (changes requested):

examples/generative/corrdiff/conf/config_train_base.yaml
akshaysubr: I think if we don't want to specifically indicate tensor cores or matmuls, we can maybe rename to `fp_compute_mode`? That would cover TF32 as well. If that's not acceptible, `fp_optimizations` is okay.

examples/generative/corrdiff/training/training_loop.py
jleinonen: Implemented in https://github.com/NVIDIA/modulus/pull/423/commits/de4f91833da024d35098009f09cca8a1808a2784.

examples/generative/corrdiff/conf/config_train_base.yaml
jleinonen: Hmm, I think `fp_compute_mode` would be too general in the same way `fp_mode` is. Renamed to `fp_optimizations` in https://github.com/NVIDIA/modulus/pull/423/commits/de4f91833da024d35098009f09cca8a1808a2784.

PR #456, opened 2024-04-18
config updates for clearity and better defaults

yairchn:
# Modulus Pull Request

## Description
two config updates that I think are useful:
1. the name of the seed parameter in generation is not clearly linked to what is does, I changes the text next to it 
2. the dump parameter training was set to a value unsuitable for a 4h job, I reduced it and updated the text

## Checklist

### DIFF
--- a/examples/generative/corrdiff/conf/config_generate.yaml
+++ b/examples/generative/corrdiff/conf/config_generate.yaml
@@ -27,7 +27,7 @@ reg_ckpt_filename: "./checkpoints/regression.mdlus"
 image_outdir: "image_outdir"
   # Where to save the output images
 seeds: "0-63"
-  # Random seeds used for generation
+  # Number of samples on the ensemble dimension
 class_idx: null  
   # Class label. Null is random
 num_steps: 8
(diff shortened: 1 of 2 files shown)


examples/generative/corrdiff/conf/config_generate.yaml
```diff
@@ -27,7 +27,7 @@ reg_ckpt_filename: "./checkpoints/regression.mdlus"
 image_outdir: "image_outdir"
   # Where to save the output images
 seeds: "0-63"
-  # Random seeds used for generation
+  # Number of samples on the ensemble dimension
```
mnabian, line 30: This is technically not the number of samples, but the random seeds that are used for each ensemble

examples/generative/corrdiff/conf/config_train_base.yaml
```diff
@@ -64,8 +64,8 @@ checkpoint_dir: checkpoints
   # Where to save the checkpoints
 tick: 1
   # How often to print progress
-dump: 500
-  # How often to dump state
+dump: 50
+  # How often to write a ckpt (diff needs more than reg)
```
mnabian, line 68: "diff needs more than reg" I'm not sure if I understand this

examples/generative/corrdiff/conf/config_generate.yaml
yairchn: @mnabian 
if the user wants to run a generation with 100 members, that parameter controls that number right? there is nowhere else one can control the number of members.

mnabian: That is correct

PR #471, opened 2024-04-24
0.6.0 docker patch

ktangsali:
# Modulus Pull Request

## Description
With different base containers, we have different requirements for the wheels. This PR allows a mechanism to control this using a `build-arg`.

Also tested the build using just the defaults (in the absence of pre-built wheels)

Closes: https://github.com/NVIDIA/modulus/issues/453

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,26 +38,32 @@ COPY . /modulus/
 
 # Install Numcodecs (This needs a separate install because Numcodecs ARM pip install has issues)
 # A fix is being added here: https://github.com/zarr-developers/numcodecs/pull/315 but the public release is not ready yet.
+ARG NUMCODECS_ARM64_WHEEL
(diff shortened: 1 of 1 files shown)


NickGeneva (approved):

akshaysubr (approved):
This looks good to me. Just to confirm, this PR should now allow local docker builds without having the cached wheels right?

ktangsali:
> This looks good to me. Just to confirm, this PR should now allow local docker builds without having the cached wheels right?

Yes, that is correct. There is one caveat, if the base container is changed such that the fallbacks don't work (say has unsupported CUDA version, etc.), then that might require some additional handling, but other than that, it should work fine. If you run into issues, please let me know. I tested this dockerfile without any cached wheels (just defaults, i.e. `make container-deploy` and that worked.

PR #484, opened 2024-05-05
Fixed indexing error for modulus loss function

fresleven:
# Modulus Pull Request

## Description

I forgot to include a commit when pushing the MHD example, which was also supposed to fixe the SWE PINO indexing incorrectly. This is the updated version

https://github.com/NVIDIA/modulus/pull/338#issuecomment-2094897687

## Checklist


NickGeneva:
Instead cropping we should consider padding @ktangsali 

We should pad the repo with a reflection boundary so we can compute edge derivatives. Makes the domain periodic iirc so the calculation is better.

See: https://github.com/NVIDIA/modulus-sym/blob/main/examples/darcy/darcy_PINO.py#L95

NickGeneva:
Merged here: https://github.com/NVIDIA/modulus/pull/545

Thanks for the patch.

PR #510, opened 2024-05-21
GraphCast improvements - Part I

mnabian:
# Modulus Pull Request

## Description

Closes https://github.com/NVIDIA/modulus/issues/506, https://github.com/NVIDIA/modulus/issues/505, https://github.com/NVIDIA/modulus/issues/486, https://github.com/NVIDIA/modulus/issues/508, https://github.com/NVIDIA/modulus/issues/509, https://github.com/NVIDIA/modulus/issues/511, https://github.com/NVIDIA/modulus/issues/516, https://github.com/NVIDIA/modulus/issues/517

## Checklist


stadlmax:
@mnabian 
Since you are revisiting GraphCast now, adding a few comments

- Can we add the option to use `transformer_engine.LayerNorm`? In AIFS benchmarks, we just could get a 1.3x end-to-end improvement from doing so since the PyTorch implementation is rather bad for the sizes we encounter in these workloads.
- Can you check whether the current combination of `MeshGraphNodeBlock` and `MeshGraphEdgeBlock` actually matches the paper (https://github.com/NVIDIA/modulus/blob/main/modulus/models/graphcast/graph_cast_processor.py#L97-L98) I created a schematic of the GraphCast architecture for some Arch folks last week, and I think the order of residuals over the edges does not match the paper here. I might have made a mistake when trying to use shared primitives here the last time. The issue here is that in MeshGraphNet, EdgeBlock already applies the "residual" on the edge features, while the NodeBlock would expect then the features including the residual connection prior to message-passing while in GraphCast, all residual connections are only applied after both the updated edge and node features are computed (at least according to the paper).
- What would you think of splitting the GraphCastNet into a GraphCastNetERA5 and a GraphCastNet model? The current issue I see with GraphCastNet is that it is very specific to the nature of the ERA5 dataset (e.g. when it comes to preparing the input and output to switch between the `HxW` layout and the typical "serial" graph layout. GraphCastNet then could be a rather data-agnostic model defining the operations on (g2m_graph, mesh_graph, m2g_graph), while GraphCastNetERA5 defines the things somewhat specific to the workload like checkpointing, input/output conversions, etc.. In the longer term, I think it really could make sense to try to make things a bit more modular. In particular, this also includes things like "history" or the actual "prediction" mode, i.e. whether GraphCastNetERA5 predicts `y_t = f(x_t-1)` or `y_t = x_t - 1 + f(x_t-1)`. It could make sense if the "backbone" is agnostic to these things while having a specialized prediction wrapper.

mnabian:
> @mnabian Since you are revisiting GraphCast now, adding a few comments
> 
> * Can we add the option to use `transformer_engine.LayerNorm`? In AIFS benchmarks, we just could get a 1.3x end-to-end improvement from doing so since the PyTorch implementation is rather bad for the sizes we encounter in these workloads.
> * Can you check whether the current combination of `MeshGraphNodeBlock` and `MeshGraphEdgeBlock` actually matches the paper (https://github.com/NVIDIA/modulus/blob/main/modulus/models/graphcast/graph_cast_processor.py#L97-L98) I created a schematic of the GraphCast architecture for some Arch folks last week, and I think the order of residuals over the edges does not match the paper here. I might have made a mistake when trying to use shared primitives here the last time. The issue here is that in MeshGraphNet, EdgeBlock already applies the "residual" on the edge features, while the NodeBlock would expect then the features including the residual connection prior to message-passing while in GraphCast, all residual connections are only applied after both the updated edge and node features are computed (at least according to the paper).
> * What would you think of splitting the GraphCastNet into a GraphCastNetERA5 and a GraphCastNet model? The current issue I see with GraphCastNet is that it is very specific to the nature of the ERA5 dataset (e.g. when it comes to preparing the input and output to switch between the `HxW` layout and the typical "serial" graph layout. GraphCastNet then could be a rather data-agnostic model defining the operations on (g2m_graph, mesh_graph, m2g_graph), while GraphCastNetERA5 defines the things somewhat specific to the workload like checkpointing, input/output conversions, etc.. In the longer term, I think it really could make sense to try to make things a bit more modular. In particular, this also includes things like "history" or the actual "prediction" mode, i.e. whether GraphCastNetERA5 predicts `y_t = f(x_t-1)` or `y_t = x_t - 1 + f(x_t-1)`. It could make sense if the "backbone" is agnostic to these things while having a specialized prediction wrapper.

Thanks @stadlmax , I'll add your comments to my [epic](https://github.com/NVIDIA/modulus/issues/499) and consider them all.

mnabian:
Note to myself: API updates breaks GraphCast tests. Need to update them all.

mnabian:
@stadlmax as far as I remember, we were using fused layernorm and that gave us nice speedup (although I can't find it in the most recent code)... Did you also compare  `transformer_engine.LayerNorm` with fused layernorm?

stadlmax:
> @stadlmax as far as I remember, we were using fused layernorm and that gave us nice speedup (although I can't find it in the most recent code)... Did you also compare `transformer_engine.LayerNorm` with fused layernorm?

Yes, for AIFS, I found TE > APEX > PyTorch throughout a bunch of usual sizes AIFS had in their RFI benchmark. Especially the backward kernels in TE are much better for our cases. (reported numbers are runtimes, lower is better)

num_channels = 256
| layer_norm_impl    |   1626240 x 256 |   327660 x 256 |   40962 x 256 |   542080 x 256 |   814540 x 256 |
|:-------------------|----------------:|---------------:|--------------:|---------------:|---------------:|
| apex               |         9.75127 |       2.03821  |      0.371149 |       3.32072  |        4.9402  |
| pytorch            |        10.752   |       4.17265  |      0.957743 |       3.63721  |       10.2774  |
| transformer_engine |         2.59236 |       0.580879 |      0.801795 |       0.916124 |        1.33596 |

num_channels = 384
| layer_norm_impl    |   1626240 x 384 |   327660 x 384 |   40962 x 384 |   542080 x 384 |   814540 x 384 |
|:-------------------|----------------:|---------------:|--------------:|---------------:|---------------:|
| apex               |        11.2164  |       2.3109   |      0.359366 |        3.79922 |        5.64847 |
| pytorch            |        11.8419  |       4.33466  |      0.583828 |        3.99414 |       10.6802  |
| transformer_engine |         3.98762 |       0.849599 |      0.396184 |        1.38306 |        2.022   |

num_channels = 512
| layer_norm_impl    |   1626240 x 512 |   327660 x 512 |   40962 x 512 |   542080 x 512 |   814540 x 512 |
|:-------------------|----------------:|---------------:|--------------:|---------------:|---------------:|
| apex               |        12.1739  |        2.50785 |      0.37578  |        4.11927 |        6.14573 |
| pytorch            |        12.7752  |        4.5477  |      0.615464 |        4.30874 |       11.2191  |
| transformer_engine |         4.90182 |        1.04243 |      0.391352 |        1.6877  |        2.4967  |

mnabian:
> > @stadlmax as far as I remember, we were using fused layernorm and that gave us nice speedup (although I can't find it in the most recent code)... Did you also compare `transformer_engine.LayerNorm` with fused layernorm?
> 
> Yes, for AIFS, I found TE > APEX > PyTorch throughout a bunch of usual sizes AIFS had in their RFI benchmark. Especially the backward kernels in TE are much better for our cases. (reported numbers are runtimes, lower is better)
> 
> num_channels = 256
> 
> layer_norm_impl	1626240 x 256	327660 x 256	40962 x 256	542080 x 256	814540 x 256
> apex	9.75127	2.03821	0.371149	3.32072	4.9402
> pytorch	10.752	4.17265	0.957743	3.63721	10.2774
> transformer_engine	2.59236	0.580879	0.801795	0.916124	1.33596
> num_channels = 384
> 
> layer_norm_impl	1626240 x 384	327660 x 384	40962 x 384	542080 x 384	814540 x 384
> apex	11.2164	2.3109	0.359366	3.79922	5.64847
> pytorch	11.8419	4.33466	0.583828	3.99414	10.6802
> transformer_engine	3.98762	0.849599	0.396184	1.38306	2.022
> num_channels = 512
> 
> layer_norm_impl	1626240 x 512	327660 x 512	40962 x 512	542080 x 512	814540 x 512
> apex	12.1739	2.50785	0.37578	4.11927	6.14573
> pytorch	12.7752	4.5477	0.615464	4.30874	11.2191
> transformer_engine	4.90182	1.04243	0.391352	1.6877	2.4967

This is great comparison, thanks! I'll switch to te then. Do we have any reason to still keep fused layernorm from apex, or we should just remove it?

stadlmax:
> This is great comparison, thanks! I'll switch to te then. Do we have any reason to still keep fused layernorm from apex, or we should just remove it?

I guess, no, not really. TE also should be decently covered when it comes to development specifically for Blackwell and beyond. I know a few POCs that try to optimize The LN in TE even further.
If we are based on the DLFW containers, TE also should come pre-installed.

mnabian:
@stadlmax added support for TE layernorm.

mnabian:
> Note to myself: API updates breaks GraphCast tests. Need to update them all.

Done

modulus/utils/graphcast/graph.py
```diff
@@ -132,53 +118,31 @@ def create_g2m_graph(self, verbose: bool = True) -> Tensor:
...
-            self.icospheres["order_" + str(self.max_order) + "_faces"][:, 2]
-        ]
-        edge_len_3 = np.max(get_edge_len(edge_src, edge_dst))
-        edge_len = max([edge_len_1, edge_len_2, edge_len_3])
+
+        max_edge_len = max_edge_length(
+            self.finest_mesh_vertices, self.finest_mesh_src, self.finest_mesh_dst
+        )
 
         # create the grid2mesh bipartite graph
```
stadlmax, line 126: since we are revisiting this: we might also explore to generalize the grid2mesh / mesh2grid routines a bit
AIFS e.g. uses the same definition in both "mapping" layers leading to the same number of edges in both graphs while GraphCast for some reason uses a non-symmetrical approach
we basically should have
- nearest neighbor grid-> mesh (and optional filtering based on max_edge_len)
- centroid based
- and then it's up to the configuration to specify these options
Would this make sense?

modulus/utils/graphcast/graph.py
```diff
@@ -40,47 +44,33 @@ class Graph:
...
-        self.max_order = (
-            len([key for key in self.icospheres.keys() if "faces" in key]) - 2
-        )
 
         # flatten lat/lon gird
         self.lat_lon_grid_flat = lat_lon_grid.permute(2, 0, 1).view(2, -1).permute(1, 0)
 
+        # create the multi-mesh
+        _meshes = get_hierarchy_of_triangular_meshes_for_sphere(splits=multimesh_level)
+        merged_mesh = merge_meshes(_meshes)
```
stadlmax, line 66: should be add a config-option specifying whether we want to merge meshes?
e.g. GenCast has
> Unlike the multimesh used in GraphCast, the mesh in GenCast is a 6-times refined icosahedral mesh as
defined in Lam et al. (2023), with 41,162 nodes and 246,960 edge

which on the other hand only works due to the use of self-attention on a 32-hop neighborhood
> The 4-head self-attention mechanism in each block is such that each
node in the mesh attends to itself and to all other nodes which are within its 32-hop neighbourhood
on the mesh

on a second thought, the latter would require no "merging" step but adding these k-hops neighborhoods explicitly anyways, so probably a bit more complicated anyways

feel free to ignore this comment or add further comments :D

examples/weather/graphcast/conf/config.yaml
```diff
@@ -32,18 +35,18 @@ checkpoint_encoder_finetune: True
...
 dataset_path: "/data"
-static_dataset_path: "datasets/static"
-latlon_res: (721, 1440)
-num_samples_per_year_train: 1448
+static_dataset_path: null
+latlon_res: [721, 1440]  
+num_samples_per_year_train: 1408
 num_workers: 0  # 8
-num_channels: 3  # 34
+num_channels: 474
```
stadlmax, line 49: since we are revisiting this: I feel like we have been a bit inconsistent when it comes to "num_channels". In GraphCast, we technically have `num_channels = num_data_channels * (1 + history_len) + static_channels`. (and the number of static channels technically also depends on the length of history)

For now, the model neither has been fully agnostic to how the input is actually prepared nor fully aware of it. We included the option for a "static" dataset in GraphCast itself but would have assumed the history to be taken care of outside of the model. 

Would it make sense to move the static dataset outside of GraphCast, too, and then just have "num_channels", i.e. have an agnostic model when it comes to input preparation? 

Or should be add a "prepare_input" method which applies the history and the static dataset and let the model compute "num_channels" based on "num_data_channels", "num_static_channels", and "history_len"?

dallasfoster: I vote in favor of moving the static dataset outside of GraphCast. If anything, I could imagine a wrapper class that does ERA5 specific utilities like appending types of static data.

modulus/models/gnn_layers/mesh_graph_mlp.py
```diff
@@ -26,11 +26,11 @@
...
 
 try:
-    from apex.normalization import FusedLayerNorm
+    from transformer_engine import pytorch as te
 
-    apex_imported = True
+    te_imported = True
 except ImportError:
-    apex_imported = False
+    te_imported = False
```
dallasfoster, line 33: If we are going to add `transformer_engine` as an optional library then I think it should be included in the toml. Is `transformer_engine` included in the pytorch base container that we use?

modulus/models/graphcast/graph_cast_net.py
```diff
@@ -66,11 +66,10 @@ class GraphCastNet(Module):
 
     Parameters
     ----------
-    meshgraph_path : str
-        Path to the meshgraph file. If not provided, the meshgraph will be created
-        using PyMesh.
     static_dataset_path : str
         Path to the static dataset file.
```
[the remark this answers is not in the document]
mnabian: Yes, this was on my to-do list for the Part II PR, but I can take care of it in this PR. We've been going back and forth about this issue for a while. Initially, the reason was to ease integration with E2MIP and avoid a wrapper class, but I agree it should change now.

modulus/models/gnn_layers/mesh_graph_mlp.py
mnabian: Yes, it is included in the pytorch base image > 23.10

modulus/datapipes/climate/era5_hdf5.py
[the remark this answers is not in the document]
mnabian: As part of this PR, I have enabled GraphCast to support any input resolution. I have also included the configs for training a 1-degree model that replicates the `graphcast_small` model released by Google. My goal is to ultimately train a 1-degree GraphCast model and compare it with Google's released model. We can either resize the inputs in the dataloader or prepare another dataset at the desired resolution. The latter approach is what we did for DLWP. I chose the former for improved usability, although it has its own disadvantages. Overall, I think it would be useful to have this feature in the dataloader to add flexibility in the size of the inputs.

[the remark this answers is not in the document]
dallasfoster: Related: how does the input resolution interact with the multigrid level? It seems like there is some implicit relationship that is not being fully captured. What happens if you specify a 1-degree model and a multimesh level of 6?

modulus/models/graphcast/graph_cast_net.py
```diff
@@ -183,13 +183,8 @@ def __init__(
...
 
         # construct the graph
-        try:
-            self.graph = Graph(meshgraph_path, self.lat_lon_grid)
-        except FileNotFoundError:
-            raise FileNotFoundError(
-                "The icospheres_path is corrupted. "
-                "Tried using pymesh to generate the graph but could not find pymesh"
-            )
+        self.graph = Graph(self.lat_lon_grid, multimesh_level)
```
dallasfoster, line 186: I also wanted to call out, in L176, the definition of `longitudes` is not consistent with the ERA5 dataset, which I believe is typically `linspace(0, 360, 1440, endpoint = False)`. I don't think this actually makes a different in the behavior of the model, however.

mnabian: You are right, but I wanted to remain consistent with the paper. See equation A.4 on page 25: https://arxiv.org/pdf/2212.12794v1

modulus/datapipes/climate/era5_hdf5.py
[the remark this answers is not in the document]
stadlmax: The multimesh is more or less a latent graph and a multimesh of level 6 e.g. consists of around 40,000 points. The grid2mesh and mesh2grid graphs map the input and/or output grid onto that latent representation. This is e.g. done via finding nearest neighbors, i.e. each grid point is e.g. connected to the 4-closed mesh points.

When going to higher resolutions, it is likely that you also would like to increase the resolution of the latent graph while significantly lower resolutions could even require lower ones. Technically, however, the multimesh level then is just a hyperparameter.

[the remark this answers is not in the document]
mnabian: There is a bipartite graph (grid2mesh) with single message passing that maps from the latlon grid to multimesh. An edge is added if the distance between the mesh node and the grid node is smaller or equal than 0.6 times the length of the edges in mesh highest level. 
So if you specify a 1-degree model and a multimesh level of 6, your grid2mesh graph will change accordingly.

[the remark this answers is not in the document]
dallasfoster: Thank you for the clarification.

dallasfoster (approved):

modulus/utils/graphcast/graph.py
mnabian: How about we add this comment to our [epic](https://github.com/NVIDIA/modulus/issues/499) and revisit it if/when we decided to extend to GenCast?

modulus/utils/graphcast/graph.py
mnabian: I have doubts about this. GraphCast's approach for constructing grid2mesh and mesh2grid edges are significantly different, leading to 3x more edges in mesh2grid compared to grid2mesh. Although I agree that the underlying reason for such design is unclear. Since our immediate goal is to compare directly with GraphCast, I suggest we stick to their design, and consider/revisit this in future enhancements post comparison. Would yo agree?

modulus/utils/graphcast/graph.py
stadlmax: makes perfect sense, yes

examples/weather/graphcast/conf/config.yaml
mnabian: Done, please take a look.

modulus/models/graphcast/graph_cast_net.py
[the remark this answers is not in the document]
mnabian: Done

modulus/utils/graphcast/graph.py
stadlmax: Fully agree with you regarding GraphCast and how they use things. My point was rather potentially future proofing things. For now, GraphCast definitely will just use the mappings as defined in the original version for sure. 
But good point, let's revisit things later.

examples/weather/graphcast/conf/config_small.yaml
```diff
@@ -0,0 +1,73 @@
...
+checkpoint_processor_finetune: False
+checkpoint_decoder_finetune: False
+concat_trick: False
+cugraphops_encoder: False
+cugraphops_processor: False
+cugraphops_decoder: False
+recompute_activation: False
+wb_mode: "online"
+synthetic_dataset: false
+dataset_path: "/data/era5_73var"  #"/code/datasets/era5_73var"
```
stadlmax, line 43: nitpick: comments

stadlmax:
Thanks for addressing the feedback, looks good to me.

stadlmax (approved):

examples/weather/graphcast/conf/config_small.yaml
mnabian: Thanks, fixed it

PR #544, opened 2024-06-05
AeroGraphNet example

Alexey-Kamenev:
# Modulus Pull Request

## Description

This example demonstrates how to train MeshGraphNet model on two automotive aerodynamics datasets, Ahmed Body and DrivAerNet.

The example uses [Hydra experiments](https://hydra.cc/docs/patterns/configuring_experiments/) as a way to specify experiment configuration.
For example, to train default MGN on DrivAerNet dataset, the following command can be used:
```
python train.py +experiment=drivaernet/mgn data.data_dir=/data/DrivAerNet/
```
See the example README for more information.

The example also contains visualization of the meshes which allows to better interpret the training progress.

## Checklist


modulus/datapipes/gnn/drivaernet_dataset.py
```diff
@@ -0,0 +1,391 @@
...
+            # Extract connectivity information from the mesh.
+            # Traversal API is faster comparing to iterating over mesh.cell.
+            polys = mesh.GetPolys()
+            if polys is None:
+                raise ValueError("Failed to get polygons from the mesh.")
+
+            polys.InitTraversal()
+
+            edge_list = []
+            for _ in range(polys.GetNumberOfCells()):
```
ktangsali, line 280: I think this piece of code does not capture all the edges. (Can we verify the edges generated from this matches with what VTK graph has natively?). For the Stokes dataset, I had to change the following to (I am yet to submit a PR for that):

```
id_list = vtk.vtkIdList()
for _ in range(polys.GetNumberOfCells()):
    polys.GetNextCell(id_list)
    num_ids = id_list.GetNumberOfIds()
    for j in range(num_ids):
        edge_list.append((id_list.GetId(j), id_list.GetId(j+1) % num_ids))
```

The `% num_ids` was required to close the loop and get all the connections / edges of a triangle. 

cc. @mnabian

ktangsali: image

ktangsali: Probably does not matter for graph training, but can become important if we expect the graph to be used in further downstream tasks such as computing gradients for which having accurate connectivity information is important.

examples/cfd/ahmed_body_mgn/README.md
ktangsali, line 1: I see that this directory is getting a new name. Can we update the entry in the docs too - that way the docs will reflect the change. https://github.com/Alexey-Kamenev/modulus/blob/aero-graph-net/docs/index.rst?plain=1#L43

examples/cfd/aero_graph_net/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+Output of the model are:
+
+- Surface pressure
+- Wall shear stresses
+- Drag coefficient - optional, computed using pressure and shear stress outputs.
+
+![Comparison between the AeroGraphNet prediction and the
+ground truth for surface pressure, wall shear stresses, and the drag coefficient for one
+of the samples from the test dataset.](../../../docs/img/ahmed_body_results.png)
```
ktangsali, line 81: Would be very cool if we could add some images like this for the drivernet dataset too :)

examples/cfd/aero_graph_net/inference.py
```diff
@@ -0,0 +1,181 @@
...
+                assert (
+                    len(batch["name"]) == 1
+                ), "Only batch size 1 is currently supported."
+                pv_graph.save(self.output_dir / (batch["name"][0] + ".vtp"))
+
+
+@hydra.main(version_base="1.3", config_path="conf", config_name="config")
+def main(cfg: DictConfig) -> None:
+    # initialize distributed manager
+    DistributedManager.initialize()
```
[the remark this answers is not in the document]
ktangsali: After this: https://github.com/NVIDIA/modulus/pull/549, you can most likely remove...

modulus/datapipes/gnn/drivaernet_dataset.py
ktangsali: Ref: https://github.com/NVIDIA/modulus/pull/549

examples/cfd/aero_graph_net/train.py
```diff
@@ -0,0 +1,262 @@
...
+    # initialize distributed manager
+    DistributedManager.initialize()
+    dist = DistributedManager()
+
+    init_python_logging(cfg)
+    logger.info(f"Config summary:\n{OmegaConf.to_yaml(cfg, sort_keys=True)}")
+
+    torch.set_float32_matmul_precision("high")
+    torch.backends.cudnn.benchmark = True
+    torch.backends.cudnn.allow_tf32 = True
```
mnabian, line 208: I'm curious if you have made any accuracy comparisons between the models trained with and without TF32?

examples/cfd/aero_graph_net/utils.py
```diff
@@ -14,7 +14,21 @@
...
 
+from typing import Any, Mapping, Optional
+
 import torch
+from torch import Tensor
+
+from dgl import DGLGraph
+
+
+class RRMSELoss(torch.nn.Module):
```
mnabian, line 25: Can this be used instead? https://github.com/NVIDIA/modulus/blob/main/modulus/metrics/general/mse.py#L47

mnabian (approved):

modulus/datapipes/gnn/drivaernet_dataset.py
mnabian: Also this PR proposes another fix: https://github.com/NVIDIA/modulus/pull/561

examples/cfd/aero_graph_net/conf/config.yaml
```diff
@@ -0,0 +1,96 @@
...
+hydra:
+  run:
+    dir: ${output}
+  output_subdir: hydra  # Default is .hydra which causes files not being uploaded in W&B.
+
+# Main output directory.
+output: outputs/${now:%Y-%m-%d}/${now:%H-%M-%S}
+
+# The directory to search for checkpoints to continue training.
+resume_dir: ${output}
```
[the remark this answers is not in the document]
Alexey-Kamenev: That's correct: the new output dir is created on each run, which is most likely what you want anyway - as experiment settings can change (e.g. learning rate or other hyperparameters).
By default, `resume_dir` will point to the `output` and will not actually do anything as there will be no checkpoints in the newly created dir. However, if user overrides output and points to existing dir with checkpoints, then the latest checkpoint should be loaded just fine (similar to previous behavior). The only complication in re-using the dir is log and other config merging. That's why the recommended way would be to use separate `output` dirs for each experiment but point `resume_dir` to a dir with checkpoints.

examples/cfd/aero_graph_net/inference.py
```diff
@@ -0,0 +1,181 @@
...
+            to_absolute_path(cfg.resume_dir),
+            models=self.model,
+            device=self.device,
+        )
+
+        # instantiate losses.
+        logger.info("Creating the losses...")
+        self.loss = instantiate(cfg.loss)
+
+    @torch.no_grad()
```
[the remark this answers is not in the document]
Alexey-Kamenev: I tend to prefer using `no_grad` instead of `inference_mode` as the latter had a known issue in the past when it was giving an error (`no_grad` worked in that case). As I understand, that bug has been fixed, but my habit stayed :)
Will change it.

examples/cfd/aero_graph_net/inference.py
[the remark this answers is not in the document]
Alexey-Kamenev: `initialize()` is still needed (dist manager is also used in logger init), but I updated the code to use the device from the manager on line 88.

examples/cfd/aero_graph_net/loggers.py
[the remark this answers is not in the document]
Alexey-Kamenev: That's a great question which I wanted to discuss as well. My thoughts are:
1. There is no need for `PythonLogger` and `RankZeroLoggingWrapper` - all of that can be done using standard Python logging mechanisms. This is what I do in AGN example. Adding unnecessary wrappers only complicates things.
2. The reason for my `ExperimentLogger` is that Modulus' metric logging is embedded into the `LaunchLogger`. The main problem is that `LaunchLogger` does *a lot* of things, most importantly hiding the training loop, which is confusing to most of the users (from my experience), that expect to see a standard training loop. Also, `LaunchLogger` class often mixes a lot of things in a single function, including calling to different logger types (e.g. MLFlow and WB) in a single function (see `__enter__` and `__exit__` as an example). As a result, adding support for new logger type (e.g. TensorBoard) is not trivial.

In my case, using `CompositeLogger` provides simple abstraction and implementation that can easily support different logger types without any changes to the main code or other loggers or anything else.

Happy to discuss it further, if needed.

examples/cfd/aero_graph_net/train.py
```diff
@@ -0,0 +1,262 @@
...
+        self.model = instantiate(cfg.model)
+
+        if cfg.compile.enabled:
+            self.model = torch.compile(self.model, **cfg.compile.args).to(
+                self.dist.device
+            )
+        else:
+            self.model = self.model.to(self.dist.device)
+
+        # distributed data parallel for multi-node training
```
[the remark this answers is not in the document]
Alexey-Kamenev: Done.

examples/cfd/aero_graph_net/conf/config.yaml
```diff
@@ -0,0 +1,96 @@
...
+test:
+  batch_size: 1
+  dataloader:
+    shuffle: false
+    num_workers: 1
+    pin_memory: true
+    drop_last: false
+
+compile:
+  enabled: false
```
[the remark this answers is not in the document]
Alexey-Kamenev: Unfortunately, I've tried that and MGN still does not work with any backend, including inductor, due lack of support in DGL. I hope DGL will support it some day.

examples/cfd/aero_graph_net/conf/config.yaml
```diff
@@ -0,0 +1,96 @@
...
+    _target_: torch.cuda.amp.GradScaler
+    enabled: ${..enabled}
+
+loggers:
+  wandb:
+    _target_: loggers.WandBLogger
+    project: car-cfd # aero-graph-net
+    entity: modulus
+    name: agn
+    group:
```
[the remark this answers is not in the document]
Alexey-Kamenev: Yes, it will be None value and, in case of W&B, the group will simply be empty.

examples/cfd/aero_graph_net/train.py
Alexey-Kamenev: No, not yet, but it would an interesting experiment.

modulus/datapipes/gnn/drivaernet_dataset.py
Alexey-Kamenev: @ktangsali @mnabian I like the proposed solution better than `% num_ids` - it makes it clearer of what's going on, so I've incorporated it into my code.

examples/cfd/aero_graph_net/loggers.py
```diff
@@ -0,0 +1,108 @@
...
+    pylog_cfg.rank = DistributedManager().rank
+    # Enable logging only on rank 0, if requested.
+    if pylog_cfg.rank0_only and pylog_cfg.rank != 0:
+        pylog_cfg.handlers = {}
+        pylog_cfg.loggers.agnet.handlers = []
+    # Configure logging.
+    logging.config.dictConfig(OmegaConf.to_container(pylog_cfg, resolve=True))
+
+
+def rank0(func):
```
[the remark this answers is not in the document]
Alexey-Kamenev: @ktangsali has the same question, I [responded](https://github.com/NVIDIA/modulus/pull/544#discussion_r1653601627) to it earlier in the review. Happy to have a separate discussion among 3 of us.

examples/cfd/aero_graph_net/utils.py
Alexey-Kamenev: Hmm, but that one is a different function, right? That is, Ahmed body (and subsequently my code), uses RRMSE (**Relative** Root Mean Squared Error) loss, while the function you mentioned is a simple RMSE.
Now, I don't know why RRMSE was used in the original code, @ktangsali any comments?

examples/cfd/ahmed_body_mgn/README.md
Alexey-Kamenev: Good point - updated.

examples/cfd/aero_graph_net/loggers.py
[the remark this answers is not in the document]
ktangsali: I think I understand your reasoning. I will resolve this in that case to unblock. I think we can discuss the logging issue you mentioned in a different PR / Epic as some of the concerns with deviation from a standard training loop are very valid. 

Also tagging @NickGeneva for visibility.

examples/cfd/aero_graph_net/loggers.py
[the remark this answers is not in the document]
ktangsali: Thanks, I resolved my comment as Alexey's concern is valid. Looks like this might need a separate discussion

modulus/datapipes/gnn/drivaernet_dataset.py
ktangsali: I am happy as long as all the edges are included :)

examples/cfd/aero_graph_net/conf/config.yaml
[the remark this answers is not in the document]
ktangsali: Thanks! This is making sense to me

examples/cfd/aero_graph_net/inference.py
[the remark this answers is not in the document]
ktangsali: I see, and this is the logger you have defined in this PR correct?

examples/cfd/aero_graph_net/README.md
Alexey-Kamenev: Done - added an image, updated the README.

ktangsali: Thanks! Looks great!

examples/cfd/aero_graph_net/utils.py
ktangsali: Great question - yeah I agree, the rmse from metrics might not be useful here. As to Alexey's question why RRMSE was used, I believe @mnabian might know best as he trained the ahmed body example initially.

ktangsali (approved):
Looks good!

examples/cfd/aero_graph_net/inference.py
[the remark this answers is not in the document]
Alexey-Kamenev: Yep - it's defined [here](https://github.com/NVIDIA/modulus/pull/544/files#diff-56b78f47e6e2657ca66fbed56b5c8a9ec69719f668eb7622c8855e119fa19143R30). Maybe instead of using `DistributedManager` inside of this function, I can pass rank as an argument to make it more clear?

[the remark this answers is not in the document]
ktangsali: I think that would be great! Might make it modular too. :)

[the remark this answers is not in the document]
Alexey-Kamenev: Done.

PR #547, opened 2024-06-11
ModAFNO for interpolation

jleinonen:
# Modulus Pull Request

## Description
This adds a Modulated AFNO (ModAFNO) layer that adds a modulation input to the AFNO architecture. This can be used to build tunable AFNO models and works, for example, to build a weather interpolation model.

## Checklist

## Dependencies
No new dependencies needed


modulus/models/afno/modafno.py
```diff
@@ -0,0 +1,649 @@
...
+
+        sequence = [nn.Linear(in_features, hidden_features), activation_fn()]
+        for _ in range(hidden_layers):
+            sequence += [nn.Linear(hidden_features, hidden_features), activation_fn()]
+        sequence.append(nn.Linear(hidden_features, out_features * 2))
+        self.net = nn.Sequential(*sequence)
+
+    def forward(self, x: Tensor):
+        (scale, shift) = torch.chunk(self.net(x), 2, dim=1)
+        return (1 + scale, shift)
```
NickGeneva, line 73: Why a plus 1 here?

modulus/models/afno/modafno.py
```diff
@@ -0,0 +1,649 @@
...
+    def __init__(
+        self,
+        hidden_size: int,
+        mod_features: int,
+        num_blocks: int = 8,
+        sparsity_threshold: float = 0.01,
+        hard_thresholding_fraction: float = 1,
+        hidden_size_factor: int = 1,
+        scale_shift_kwargs: Union[dict, None] = None,
+        scale_shift_mode: str = "complex",
```
NickGeneva, line 167: type `Literal["complex", "real"]` seems more appropriate here

modulus/models/afno/modafno.py
```diff
@@ -0,0 +1,649 @@
...
+                self.w1[1],
+            )
+            + self.b1[1]
+        )
+
+        # scale-shift operation
+        (scale, shift) = self.scale_shift(mod_embed)
+        scale = scale.view(*scale_shift_shape)
+        shift = shift.view(*scale_shift_shape)
+        if self.scale_shift_mode == "real":
```
NickGeneva, line 257: In the constructor there is presently no check if `scale_shift_mode` is real or complex.

I would suggest either line 178 gets changed to through an error if not one of these, or line 260 gets changed to an else for consistency.

modulus/models/afno/modafno.py
```diff
@@ -0,0 +1,649 @@
...
+        if scale_shift_kwargs is None:
+            scale_shift_kwargs = {}
+        self.scale_shift = ScaleShiftMlp(
+            mod_features, latent_features, **scale_shift_kwargs
+        )
+
+    def forward(self, x: Tensor, mod_embed: Tensor) -> Tensor:
+        (scale, shift) = self.scale_shift(mod_embed)
+
+        scale_shift_shape = (scale.shape[0],) + (1,) * (x.ndim - 2) + (scale.shape[1],)
```
NickGeneva, line 121: Is this ever a variable dim or do we always know the shape will be something like (B, 1, 1, N)

modulus/models/afno/modafno.py
```diff
@@ -0,0 +1,649 @@
...
+        self.scale_shift = ScaleShiftMlp(
+            mod_features,
+            self.num_blocks
+            * self.block_size
+            * self.hidden_size_factor
+            * self.channel_mul,
+            **scale_shift_kwargs,
+        )
+
+    def forward(self, x: Tensor, mod_embed: Tensor) -> Tensor:
```
NickGeneva, line 190: May consider adding an assert check here to check consistency between x and mod_embed

modulus/models/afno/modafno.py
```diff
@@ -0,0 +1,649 @@
...
+        double_skip: bool = True,
+        sparsity_threshold: float = 0.01,
+        hard_thresholding_fraction: float = 1.0,
+        modulate_filter: bool = True,
+        modulate_mlp: bool = True,
+        scale_shift_mode: str = "real",
+    ):
+        super().__init__()
+        self.norm1 = norm_layer(embed_dim)
+        if modulate_filter:
```
NickGeneva, line 384: Would users realistically want to shut this off? Would this just become AFNO with out these? I wonder if this option is really needed or can just be considered part of the arch.

modulus/models/afno/modafno.py
```diff
@@ -0,0 +1,649 @@
...
+                drop=drop,
+            )
+        self.double_skip = double_skip
+        self.modulate_filter = modulate_filter
+        self.modulate_mlp = modulate_mlp
+
+    def forward(self, x: Tensor, mod_embed: Tensor) -> Tensor:
+        residual = x
+        x = self.norm1(x)
+        x = self.filter(x, mod_embed) if self.modulate_filter else self.filter(x)
```
NickGeneva, line 424: Could so some thing like

filder_model = AFNOMlp()
self.filer = lambda x, mode_embed : filder_model(x)

to avoid having an if statement in your forward.

modulus/models/afno/modafno.py
```diff
@@ -0,0 +1,649 @@
...
+
+        if self.double_skip:
+            x = x + residual
+            residual = x
+
+        x = self.norm2(x)
+        x = self.mlp(x, mod_embed) if self.modulate_mlp else self.mlp(x)
```
NickGeneva, line 431: Similarly, not a big fan of conditionals in the forward.

modulus/models/afno/modafno.py
```diff
@@ -0,0 +1,649 @@
...
+        x = x + mod_additive
+
+        x = x.reshape(B, self.h, self.w, self.embed_dim)
+        for blk in self.blocks:
+            x = blk(x, mod_embed=mod_embed)
+
+        return x
+
+    def forward(self, x: Tensor, mod: Tensor) -> Tensor:
+        """The full ModAFNO model logic."""
```
NickGeneva, line 637: maybe assert check should be here

modulus/models/afno/modafno.py
```diff
@@ -0,0 +1,649 @@
...
+
+    def forward_features(self, x: Tensor, mod: Tensor) -> Tensor:
+        """Forward pass of core ModAFNO"""
+        B = x.shape[0]
+        x = self.patch_embed(x)
+        x = x + self.pos_embed
+        x = self.pos_drop(x)
+
+        mod_embed = self.mod_embed_net(mod)
+        mod_additive = self.mod_additive_proj(mod_embed).unsqueeze(dim=(1))
```
NickGeneva, line 627: Just Confirm that dim=1 is always valid and that mod will never be higher dimensional needing something like dim=-1c

modulus/models/afno/modafno.py
```diff
@@ -0,0 +1,649 @@
...
+        Ratio of layer MLP latent variable size to input feature size, by default 4.0
+    drop_rate : float, optional
+        Drop out rate in layer MLPs, by default 0.0
+    num_blocks : int, optional
+        Number of blocks in the block-diag frequency weight matrices, by default 16
+    sparsity_threshold : float, optional
+        Sparsity threshold (softshrink) of spectral features, by default 0.01
+    hard_thresholding_fraction : float, optional
+        Threshold for limiting number of modes used [0,1], by default 1
+    modulate_filter: bool, optional
```
NickGeneva, line 482: Nit: add `, by default True` and order these the same as the parameters of the constructor.

modulus/models/afno/modembed.py
```diff
@@ -0,0 +1,118 @@
...
+    def __init__(self, num_channels: int):
+        super().__init__()
+        self.num_channels = num_channels
+
+        freqs = torch.pi * torch.arange(
+            start=1, end=self.num_channels // 2 + 1, dtype=torch.float32
+        )
+        self.register_buffer("freqs", freqs)
+
+    def forward(self, x: Tensor):
```
NickGeneva, line 42: -> Tensor

modulus/models/afno/modembed.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+
+    def __init__(self, num_channels: int):
+        super().__init__()
+        self.num_channels = num_channels
+        ind = torch.arange(num_channels)
+        ind = ind.view(1, len(ind))
+        self.register_buffer("indices", ind)
+
+    def forward(self, t: Tensor):
```
NickGeneva, line 65: Nit: return type hint -> Tensor

modulus/models/afno/modembed.py
```diff
@@ -0,0 +1,118 @@
...
+            self.sinusoid_embed = PositionalEmbedding(dim)
+
+        self.dim = dim
+
+        blocks = []
+        for _ in range(depth):
+            blocks.extend([nn.Linear(dim, dim), activation_fn()])
+        self.mlp = nn.Sequential(*blocks)
+
+    def forward(self, t: Tensor):
```
NickGeneva, line 111: Nit: return type hint -> Tensor

modulus/models/afno/modembed.py
```diff
@@ -0,0 +1,118 @@
...
+        dim: int = 64,
+        depth: int = 1,
+        activation_fn: Type[nn.Module] = nn.GELU,
+        method: str = "sinusoidal",
+    ):
+        super().__init__()
+        self.max_time = max_time
+        self.method = method
+        if method == "onehot":
+            self.onehot_embed = OneHotEmbedding(dim)
```
NickGeneva, line 100: Can you make this and the following line `self.emb = ...` and get rid of the conditional in the forward?

modulus/models/afno/modembed.py
```diff
@@ -0,0 +1,118 @@
...
+        elif method == "sinusoidal":
+            self.sinusoid_embed = PositionalEmbedding(dim)
```
NickGeneva, line 102: Nit: add
```
else:
    raise ValueError(f"Embedding {method} not supported")
```

NickGeneva:
Can you add the model to the docs rst file:

https://github.com/NVIDIA/modulus/blob/main/docs/api/modulus.models.rst

modulus/models/afno/modafno.py
jleinonen: The net gives pretty small values after initialization so with the +1 we get scale close to 1 and shift close to 0 in the beginning of training. This turned out to lead to more stable training.

modulus/models/afno/modafno.py
jleinonen: Great, will update - my knowledge of type annotations is rather incomplete.

modulus/models/afno/modembed.py
jleinonen: Good idea, but wouldn't this break compatibility with existing checkpoints (since AFAIK layer names need to match those in the checkpoint file)?

modulus/models/afno/modembed.py
jleinonen: Fixed in https://github.com/NVIDIA/modulus/pull/547/commits/f634dfdcaae1cc73e0e67568843c7fd7fd5d1d09.

modulus/models/afno/modembed.py
jleinonen: Fixed in https://github.com/NVIDIA/modulus/pull/547/commits/f634dfdcaae1cc73e0e67568843c7fd7fd5d1d09.

modulus/models/afno/modembed.py
jleinonen: Fixed in https://github.com/NVIDIA/modulus/pull/547/commits/f634dfdcaae1cc73e0e67568843c7fd7fd5d1d09.

modulus/models/afno/modembed.py
jleinonen: Added in https://github.com/NVIDIA/modulus/pull/547/commits/0b75c9b0348c95a3096ead5437d2222dc18f4418.

modulus/models/afno/modafno.py
jleinonen: Fixed in https://github.com/NVIDIA/modulus/pull/547/commits/0b75c9b0348c95a3096ead5437d2222dc18f4418.

modulus/models/afno/modafno.py
jleinonen: Done in https://github.com/NVIDIA/modulus/pull/547/commits/8e24b676645436ffe014e2a584a4c2404f429ec5

modulus/models/afno/modafno.py
jleinonen: This would be an assert for the tensor channel dimension? In any case, I tried to add it and ruff complains:
```modulus/models/afno/modafno.py:191:9: S101 Use of `assert` detected```

File: modulus/models/afno/modafno.py (hunk 4 above)
Comment (jleinonen) [reply modulus/models/afno/modafno.py:121]: Maybe what I had in mind when I wrote this was that (Mod)AFNO can be extended to 1D/3D quite easily and then this would still work... but for the current implementation `x.ndim ==4` always.

File: modulus/models/afno/modafno.py (hunk 9 above)
Comment (jleinonen) [reply modulus/models/afno/modafno.py:639]: This would have the same issue as https://github.com/NVIDIA/modulus/pull/547#discussion_r1646497439?

**Comment by jleinonen** [comment]
> Can you add the model to the docs rst file:
> 
> https://github.com/NVIDIA/modulus/blob/main/docs/api/modulus.models.rst

Added in https://github.com/NVIDIA/modulus/pull/547/commits/5f46f434155c6cf2925eb5e8f0fb4b8cd89f8655, @NickGeneva could you check that this makes sense? Also, shouldn't `AdaptiveFourierNeuralOperator` in that file be just `AFNO`?

File: modulus/models/afno/modafno.py (hunk 7 above)
Comment (jleinonen) [reply modulus/models/afno/modafno.py]: Changed in https://github.com/NVIDIA/modulus/pull/547/commits/447c62e650e79dfc0e4342a93b1bd7780067661d, I implemented this a bit differently to avoid overwriting `self.filter` so existing checkpoints will work. Maybe check if this makes sense to you?

File: modulus/models/afno/modafno.py (hunk 8 above)
Comment (jleinonen) [reply modulus/models/afno/modafno.py]: Changed in https://github.com/NVIDIA/modulus/pull/547/commits/447c62e650e79dfc0e4342a93b1bd7780067661d as with the comment above.

File: modulus/models/afno/modafno.py (hunk 3 above)
Comment (jleinonen) [reply modulus/models/afno/modafno.py:259]: Added error check to constructor in https://github.com/NVIDIA/modulus/pull/547/commits/4750b4d689943825b371709428cd1ae51289de59.

File: modulus/models/afno/modafno.py (hunk 10 above)
Comment (jleinonen) [reply modulus/models/afno/modafno.py:635]: Yes, here `dim==1` should be always valid. The shape of `x` here is `(B, P, C)` where the `P` contains all the AFNO patches flattened into one dimension. So we create a singleton dimension in `mod_additive` to make its shape `(B, 1, C)` to make all patches get the same modulation. The flattening to a `(B, P, C)` shape should always happen (even if this happens to be turned in to a 3D ModAFNO later).

File: modulus/models/afno/modafno.py (hunk 6 above)
Comment (jleinonen) [reply modulus/models/afno/modafno.py:386]: When developing, I was testing this with either `modulate_mlp` and `modulate_filter` turned off. It turns out to work best with both, but it was surprisingly good with only `modulate_mlp` active. Anyway, this should be handled without conditionals in the forward now.

File: modulus/models/afno/modafno.py (hunk 4 above)
Comment (NickGeneva) [reply modulus/models/afno/modafno.py:121]: That makes sense. Was just tracing dims from the model.

File: modulus/models/afno/modafno.py (hunk 5 above)
Comment (NickGeneva) [reply modulus/models/afno/modafno.py:192]: Ah yeah I forgot about the asset checks, ruff will want you to have a if statement and then raise the error yourself.

I suppose the idea is to make sure that the user's x and mod_embed are of the correct sizes and give a easy to understand error if it isnt. Could help make it easier to use for people. Just my opinion though,Thoughts?

File: modulus/models/afno/modembed.py (hunk 15 above)
Comment (NickGeneva) [reply modulus/models/afno/modembed.py:100]: Yeah this would require some manual stitching of the previous checkpoint.

I'd be for making the update, I think the fix for the checkpoint should be a rename of the weight key.
```python
checkpoint = torch.load(load_name)
for key, value in checkpoint['model'].items():
    if key == "modafno.something.something.onehot_embed":
           checkpoint['model']['"modafno.something.something.embed"] = value
```

I think this would be alright (I had to do this before for I think the original AFNO checkpoint :P ). But naturally we need to make sure we have an original implementation to make sure we outputs are exact. Can help if needed.

modulus/models/afno/modafno.py
NickGeneva: Makes sense, just curious :+1:

docs/api/modulus.models.rst
```diff
@@ -41,6 +41,9 @@ Currently available models include:
    * - AdaptiveFourierNeuralOperator
```
NickGeneva, line 41: Yes, please update this to AFNO

NickGeneva:
Over all I think edits look good, confirm that the outputs are consistent with existing checkpoint and I think we can merge

NickGeneva:
Oh also please add a bullet in the change log about adding this arch.

NickGeneva (approved):
Approving to unblock

docs/api/modulus.models.rst
jleinonen: Fixed in [f849071](https://github.com/NVIDIA/modulus/pull/547/commits/f849071daffda7c4c8c055e007e59c42d9c21ed5).

jleinonen:
@NickGeneva I have confirmed that the original checkpoints loads fine into the Modulus version of ModAFNO in this PR, no changes required.

jleinonen:
@NickGeneva could you also nominate another reviewer, we'll need 2 approvals to merge.

ktangsali (approved):
Looks good! Thanks for also adding the appropriate changes to the docs - appreciate that!

PR #583, opened 2024-07-11
[FEA] Graph Transformer Processor for GraphCast

mnabian:
# Modulus Pull Request

## Description

Closes https://github.com/NVIDIA/modulus/issues/572, https://github.com/NVIDIA/modulus/issues/570, https://github.com/NVIDIA/modulus/issues/571

## Checklist


modulus/models/graphcast/graph_cast_processor.py
```diff
@@ -177,3 +179,50 @@ def forward(
...
         return efeat, nfeat
+
+
+class GraphCastProcessorGraphTransformer(nn.Module):
+    """Processor block used in GenCast operating on a latent space
+    represented by hierarchy of icosahedral meshes.
+
+    Parameters
+    ----------
+    """
```
mnabian, line 190: Complete the docstrings

modulus/models/graphcast/graph_cast_processor.py
```diff
@@ -177,3 +179,50 @@ def forward(
...
+        input_dim_nodes: int = 512,
+        hidden_dim: int = 512,
+    ):
+        super().__init__()
+
+        layers = [
+            te.pytorch.TransformerLayer(
+                hidden_size=input_dim_nodes,
+                ffn_hidden_size=hidden_dim,
+                num_attention_heads=4,
```
mnabian, line 205: Make this configurable

modulus/models/graphcast/graph_cast_processor.py
```diff
@@ -177,3 +179,50 @@ def forward(
...
+
+    def forward(
+        self,
+        nfeat: Tensor,
+    ) -> Tensor:
+        nfeat = nfeat.unsqueeze(1)
+        for module in self.processor_layers:
+            nfeat = module(
+                nfeat,
+                attention_mask=self.attn_mask.to(nfeat.device),
```
mnabian, line 224: Note: Not sure how efficiently TE implements masking. Need to do some benchmarking.

modulus/models/graphcast/graph_cast_processor.py
```diff
@@ -177,3 +179,50 @@ def forward(
...
+
+        self.processor_layers = nn.ModuleList(layers)
+        self.hidden_dim = hidden_dim
+        self.attn_mask = torch.tensor(attn_mask, dtype=torch.bool)
+
+    def forward(
+        self,
+        nfeat: Tensor,
+    ) -> Tensor:
+        nfeat = nfeat.unsqueeze(1)
```
mnabian, line 220: Note: This follows the default "sbhd” layout.

modulus/utils/graphcast/graph.py
```diff
@@ -102,9 +125,20 @@ def create_mesh_graph(self, verbose: bool = True) -> Tensor:
         # ensure fields set to dtype to avoid later conversions
         mesh_graph.ndata["x"] = mesh_graph.ndata["x"].to(dtype=self.dtype)
         mesh_graph.edata["x"] = mesh_graph.edata["x"].to(dtype=self.dtype)
+        if self.khop_neighbors > 0:
+            # Make a graph whose edges connect the k-hop neighbors of the original graph.
+            mesh_graph = dgl.khop_graph(
```
mnabian, line 130: Note: This is not very efficient and we need to also look for alternative approaches. Takes a long time to compute the khop graph of a level 6 mesh.

mnabian:
Need to update the changelog

modulus/utils/graphcast/graph.py
stadlmax: My bad, I forgot that khop_graph doesn't "filter out" duplicate edges between nodes but actually just aggregates all different k-hop graphs. This is not what we need and leads to a lot of unnecessary edges. I can imagine this being slow. 
Also: the way through networkx is also likely not ideal.

DGL also has `khop_adj` which is actually what we want, i.e. it returns the adjancency matrix of A^k. Can you try
something like `sum([dgl.khop_adj(g, k=k) for k in range(1, kmax+1)]).bool().float()` which directly should give you a binary adjacency matrix?

stadlmax: (you later have to invert the binary matrix and scale it accordingly for a valid mask)

examples/weather/graphcast/train_graphcast.py
```diff
@@ -25,11 +25,18 @@
 import torch.cuda.profiler as profiler
 from torch.optim.lr_scheduler import SequentialLR, LinearLR, CosineAnnealingLR, LambdaLR
 
+import torch._dynamo
+
+torch._dynamo.config.suppress_errors = True
```
stadlmax, line 30: What do we need this for?

modulus/utils/graphcast/graph.py
stadlmax: Looked a bit into the internal of khop_adj, by avoiding the backend-agnostic nature of DGL, some unnecessary conversions between torch, numpy, and scipy-sparse, it actually is somewhat easy to implement a standalone function relying only on the SparseTensor of PyTorch itself.

```
def khop_adj_all_k(g, kmax):
    assert g.is_homogeneous, \
        'only homogeneous graph is supported'
    with torch.no_grad():
        adj = g.adj_external(transpose=True, scipy_fmt=None)
        adj_k = adj.clone()
        adj_all = adj.clone()
        for _ in range(2, kmax + 1):
            adj_k = adj_k @ adj
            adj_all += adj_k

    return adj_all.to_dense()
```

This should return the union of all k-hop subgraphs for k in (1, ..., kmax) and surely runs a lot faster.

stadlmax (changes requested):

examples/weather/graphcast/train_graphcast.py
mnabian: Without this, torch dynamo gives an error on the machine I'm using for debugging. I'll put a note to check later if this can be removed.

modulus/models/graphcast/graph_cast_processor.py
mnabian: My understanding is that the reshape is done automatically inside the self-attention module. I get a "GEMM not possible" error if I reshape manually.

modulus/models/graphcast/graph_cast_processor.py
mnabian: Done, but I'm not sure if it's a good idea to keep the mask in the state_dict as the size of it can be huge.

stadlmax: What I meant is call `register_buffer` with the mask tensor and set `persistent=False`. This way, it is not part of the state_dict when storing the model and always created when the module is instantiated but comes with the advantage of `to` calls being propagated. 
But actually, does TE have the requirement of keeping the mask in fp32? Otherwise, don't do that because then the dtype might be affected by setting the dtype of the model.

stadlmax: Then register_buffer should work as only floating point buffers are changed when e.g. calling `model.to(dtype=torch.bfloat16)`, we sometimes use this behavior as a workaround to avoid buffers being casted by bitcasting them to int and later bitcasting back. 

I prefer the buffer approach if applicable. Thus, e.g. `model.to(device="cpu")` would move this large mask back to the CPU and not leave a "ghost" buffer back on the device.

stadlmax (approved):
Thanks for addressing the feedback. Great to see this addition in the first place 👍🏽

modulus/models/graphcast/graph_cast_net.py
```diff
@@ -132,8 +135,13 @@ class GraphCastNet(Module):
...
     ----------
-    multimesh_level: int, optional
-        Level of the multi-mesh, by default 6
+    mesh_level: int, optional
+        Level of the latent mesh, by default 6
+    multimesh: bool, optional
+        If the latent mesh is a multimesh, by default True
+        If True, the latent mesh includes the nodes corresponding
+        to the specified `mesh_level`and incorporates the edges from
+        all mesh levels ranging from level 0 up to and including `mesh_level`.
```
dallasfoster, line 144: I know this was asked above but it appears to me that renaming `multimesh_level` -> `mesh_level` would indeed break existing `GraphCastNet` checkpoints, no?

modulus/utils/graphcast/graph.py
```diff
@@ -41,36 +41,73 @@
...
     """
 
     def __init__(
-        self, lat_lon_grid: Tensor, multimesh_level=6, dtype=torch.float
+        self,
+        lat_lon_grid: Tensor,
+        mesh_level: int = 6,
+        multimesh: bool = True,
+        khop_neighbors: int = 0,
+        dtype=torch.float,
```
dallasfoster, line 73: Again, the name change of `multimesh_level` here would also be breaking?

modulus/models/graphcast/graph_cast_net.py
dallasfoster, line 1: I guess what I really don't understand the motivation for is the requirement that the same module, `GraphCastNet` must be modified to support both the message passing and transformer paradigms. I understand that there would be duplicated code but wouldn't it be preferable to create a separate model called `GenCastNet`?

test/models/graphcast/test_graphcast.py
dallasfoster, line 1: It doesn't appear that any of these test the transformer component?

dallasfoster (commented):
Just a few questions here but I think the more concerning piece are the unit tests.

test/models/graphcast/test_graphcast.py
mnabian: Added unit tests, please take a look

modulus/models/graphcast/graph_cast_net.py
mnabian: Well, both options are possible, but I personally preferred to include it in the GraphCast implementation to avoid duplicating a large chunk of code

modulus/utils/graphcast/graph.py
mnabian: This one won't break the checkpoint, but fixed it for the `GraphCastNet` module

modulus/models/graphcast/graph_cast_net.py
mnabian: Thanks for your comment. I added backward compatibility, and a deprecation warning.

dallasfoster (approved):

PR #625, opened 2024-07-26
Fix non-convergence error caused by the incorrect regex.

Alexey-Kamenev:
# Modulus Pull Request

## Description

Previous version of the code contains incorrect regex that can capture patters which do not belong to the file names.
Intended regex usage is to extract id from files like `./dataset/case10.vtp`. 
However, the regex would also capture (incorrectly) things like `./dataset1/case10.vtp` (note the `1`) - in this case, the regex match would be simply `1`! This led to extreme case when all samples were the same.

This PR fixes the problem by re-writing the file enumeration code using `pathlib` making it more robust and much simpler.
The only difference is resulting sort order which does not matter anyway.

## Checklist


mnabian (approved):
LGTM

ktangsali (approved):
LGTM!

PR #645, opened 2024-08-16
[FEA] Extending the checkpointing utility to store metadata

mnabian:
# Modulus Pull Request

## Description

Often times it is required to store some metadata with the checkpoints. This PR extends the checkpointing utility to enable that.

## Checklist


daviddpruitt (approved):

modulus/launch/utils/checkpoint.py
```diff
@@ -277,8 +282,9 @@ def load_checkpoint(
     scheduler: Union[scheduler, None] = None,
     scaler: Union[scaler, None] = None,
     epoch: Union[int, None] = None,
+    return_metadata: bool = False,
```
NickGeneva, line 285: Instead of returning meta data, can you provide a dictionary object that gets modified? This way we avoid this boolean flag and its more consistent with the other parameters:

meta_data: Union[dict, None] = None,

Then:
```
if meta_data:
  for key, value in checkpoint_dict.get("metadata"):
     meta_data[key] = value
```

NickGeneva: Return type then doesnt need to change as well

NickGeneva (approved):
Approving to unblock, see suggested edit

modulus/launch/utils/checkpoint.py
```diff
@@ -185,6 +185,7 @@ def save_checkpoint(
     scheduler: Union[scheduler, None] = None,
     scaler: Union[scaler, None] = None,
     epoch: Union[int, None] = None,
+    metadata: Union[Dict[str, Any], None] = None,
```
Alexey-Kamenev, line 188: Using `Optional` would be IMO more clear and shorter:
```py
 metadata: Optional[dict[str, Any]] = None
```
Same in other places.

modulus/launch/utils/checkpoint.py
Alexey-Kamenev: +1 to @NickGeneva suggestion. In this case, users won't have to deal with different return types that lead to code like this:
```pyt
r = save_checkpoint(..)
if isinstance(r, tuple):
  # handled tuple
elif ...
```

mnabian: Sounds good. Changed it according to the suggestion

modulus/launch/utils/checkpoint.py
mnabian: Fixed it

NickGeneva (approved):

PR #590, opened 2024-07-12
multi-level gnn

ZijieH:
multi-level gnn

# Modulus Pull Request

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/modulus/blob/main/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [ ] The [CHANGELOG.md](https://github.com/NVIDIA/modulus/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/modulus/issues) is linked to this pull request.

PR #646, opened 2024-08-16
Bistride Multiscale MeshGraphNet

Alexey-Kamenev:
# Modulus Pull Request

## Description

This PR contains an example of Bistride Multiscale MeshGraphNet model, described in the [corresponding paper](https://arxiv.org/pdf/2210.02573). This is a Modulus compatible version of the earlier experimental PR #590.

The multi-scale feature is implemented as a wrapper over existing dataloader and model (MGN) and does not require any modifications to the original dataloader or MGN.

When trained using provided recipe (experiment), the following results can be obtained:

image

See the example README.md for details.

## Checklist


modulus/models/gnn_layers/bsms.py
```diff
@@ -0,0 +1,582 @@
...
+            raise ValueError(f"Only implemented for dim 2 and 3, got {x.shape}")
+
+        if len(pos.shape) == 3:
+            pi, pj = pos[:, i], pos[:, j]
+        elif len(pos.shape) == 2:
+            pi, pj = pos[i], pos[j]
+        else:
+            raise ValueError(f"Only implemented for dim 2 and 3, got {x.shape}")
+
+        # NOTE for zijie
```
ktangsali, line 391: Was this left intentionally?

modulus/models/meshgraphnet/bsms_mgn.py
```diff
@@ -0,0 +1,163 @@
...
+        Hidden layer size for the node feature decoder, by default 128
+    num_layers_node_decoder : Union[int, None], optional
+        Number of MLP layers for the node feature decoder, by default 2.
+        If None is provided, the MLP will collapse to a Identity function, i.e. no decoder
+    aggregation: str, optional
+        Message aggregation type, by default "sum"
+    do_conat_trick: : bool, default=False
+        Whether to replace concat+MLP with MLP+idx+sum
+    num_processor_checkpoint_segments: int, optional
+        Number of processor segments for gradient checkpointing, by default 0 (checkpointing disabled)
```
ktangsali, line 85: @mnabian @Alexey-Kamenev would be nice to add some recommendation here as to what values one can use. When to use this (memory saving) etc. For this architecture and vanilla MGN too. Especially since we use this flag quite often during scaling. Any thoughts?

examples/cfd/aero_graph_net/conf/data/bsms_ahmed.yaml
```diff
@@ -0,0 +1,37 @@
+# @package _global_
```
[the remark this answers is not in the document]
Alexey-Kamenev: Good catch - this was from my previous experiments with configs. This statement tells Hydra that whatever is defined in the current config file, should be mapped directly to global package. The applicable use is in experiment configs where you would want to map, say, `data` defaults directly to `/data/` and not `/experiment/data`.
I've also removed this statement from couple of other places.

examples/cfd/aero_graph_net/conf/experiment/ahmed/mgn.yaml
```diff
@@ -28,11 +28,11 @@ data:
   train:
     name: ahmed_body_train
     split: train
-    num_samples: 683
+    num_samples: 408
```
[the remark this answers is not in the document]
Alexey-Kamenev: Yep, the converged dataset is now 408-50-50 train-val-test split.

modulus/models/gnn_layers/bsms.py
Alexey-Kamenev: Oops, yes, thank you - I've removed that note comment and cleaned up the comment itself.

modulus/models/meshgraphnet/bsms_mgn.py
Alexey-Kamenev: Yes, that's a good point. At least, outline the general approach, such as "if your model does not fit into GPU memory, then start with the smallest factor of number of processor layers. For example, if the number of layers is 15, then you can start with 3 and increase in case the model still does not fit."

Not sure what is the best place to add it though - is it docstring or maybe docs?

modulus/models/meshgraphnet/bsms_mgn.py
ktangsali, line 1: Looks great! My only suggestion is, can you add this to the docs here: https://github.com/NVIDIA/modulus/blob/main/docs/api/modulus.models.rst?plain=1#L428

so that it renders here: https://docs.nvidia.com/deeplearning/modulus/modulus-core/api/modulus.models.html#module-modulus.models.meshgraphnet.meshgraphnet

modulus/models/meshgraphnet/bsms_mgn.py
ktangsali: Yeah, maybe for now you can add it to the Modulus model docs here: https://docs.nvidia.com/deeplearning/modulus/modulus-core/api/modulus.models.html

Or just have it in the docstrings. Eventually we need to add a Perf tuning section to this repo similar to what we have for Sym. But that can be done in a separate PR :)

ktangsali (approved):
LGTM! Only left a few minor doc related comments.

modulus/models/meshgraphnet/bsms_mgn.py
Alexey-Kamenev: Done.

PR #658, opened 2024-08-28
[FEA] CAE metrics

ktangsali:
# Modulus Pull Request

## Description
Adding useful metrics for CAE domain.

## Checklist

### DIFF
--- a/modulus/metrics/cae/cfd.py
+++ b/modulus/metrics/cae/cfd.py
@@ -1,0 +1,275 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2024 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import List, Tuple, Union
+
+import numpy as np
+import pyvista as pv
+import shapely
+import torch
+from numpy.fft import fft, fftfreq
+
+
+def compute_frontal_area(mesh: pv.PolyData, direction: str = "x"):
+    """
+    Compute frontal area of a given mesh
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -168,6 +168,9 @@ RUN pip install --no-cache-dir "numpy-stl" "scikit-image>=0.24.0"
 # Install sparse-dot-mkl
 RUN pip install --no-cache-dir "sparse-dot-mkl"
 
+# Install shapely
+RUN pip install --no-cache-dir "shapely"
```
mnabian, line 172: Please add this dependency to changelog

modulus/metrics/cae/cfd.py
```diff
@@ -0,0 +1,285 @@
...
+    R = R / torch.mean(trace_strain2**1.5, dim=-1, keepdim=True)
+
+    return P, Q, R
+
+
+def compute_tke_spectrum(
+    field: np.ndarray, length: float = None
+) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+    """
+    Compute the turbulent kinetic energy spectrum
```
mnabian, line 223: For some of these more advanced metrics, it would be helpful to provide a link for the users to read more about it

modulus/metrics/cae/integral.py
```diff
@@ -0,0 +1,130 @@
...
+        integral = np.sum(np.sum(field * tangents, axis=1) * lengths)
+    else:
+        # Scalar quantity
+        integral = np.sum(field * lengths)
+
+    return integral
+
+
+def surface_integral(
+    mesh: pv.PolyData,
```
mnabian, line 69: Shouldn't we also support `pv.UnstructuredGrid`?

mnabian (approved):

Dockerfile
ktangsali: It's only in the CI image, so I don't think we need to add

modulus/metrics/cae/integral.py
ktangsali: Good point, but probably we don't want to do that for this particular utility whose scope is to only compute surface integrals.

From pyvista's docs [here](https://docs.pyvista.org/api/core/), they mention:
> We generally use [pyvista.PolyData](https://docs.pyvista.org/api/core/_autosummary/pyvista.polydata#pyvista.PolyData) to construct scattered points and closed/open surfaces (non-volumetric datasets). 

Since the `pv.UnstructuredGrid` is a more general dataset format that allows 3d grids, we can potentially write a separate utility extracts such surfaces from the 3d mesh and then calls this utility for the integrals.

modulus/metrics/cae/cfd.py
ktangsali: Added

RishikeshRanade (approved):

PR #664, opened 2024-09-03
[FEA] Physics informed examples using the `PhysicsInformer` utility

ktangsali:
# Modulus Pull Request

## Description
Updates some of the key examples to use the newer `PhysicsInformer` utility. 

## Checklist


examples/cfd/ldc_pinns/config.yaml
```diff
@@ -0,0 +1,33 @@
...
+    in_channels: 2
+    dimension: 2
+    latent_channels: 32
+    fno_layers: 4
+    fno_modes: 12
+    padding: 9
+
+scheduler:
+  initial_lr: 1.E-3
+  decay_rate: .9995
```
RishikeshRanade, line 33: Can the type of gradient calculation used be part of the config?

examples/cfd/stokes_mgn/pi_fine_tuning_gnn.py
```diff
@@ -0,0 +1,504 @@
...
+        v = torch.zeros_like(y)
+        return u, v
+
+    def loss(self):
+        out = self.model(
+            self.dgl_graph.ndata["x"], self.dgl_graph.edata["x"], self.dgl_graph
+        )
+
+        # inflow points
+        mask_inflow = (
```
RishikeshRanade, line 204: Is this the masking because we don't calculate the gradients on boundary elements/voxels/nodes? Should this be taken care of when the gradient is returned by the PhysicsInformer utility instead of the user having to code it up?

RishikeshRanade (approved):

examples/cfd/darcy_physics_informed/conf/config_deeponet.yaml
```diff
@@ -24,7 +24,7 @@ start_lr: 0.001
 gamma: 0.99948708
 max_epochs: 50
 
-phy_wt: 0.1
+physics_wt: 0.1
```
Alexey-Kamenev, line 27: I wouldn't shorten `weight` either :)

examples/cfd/darcy_physics_informed/darcy_physics_informed_fno.py
```diff
@@ -110,6 +107,14 @@ def main(cfg: DictConfig):
         padding=cfg.model.fno.padding,
     ).to(device)
 
+    phy_informer = PhysicsInformer(
+        required_outputs=["diffusion_u"],
+        equations=darcy,
+        grad_method="finite_difference",
+        device=device,
+        fd_dx=1 / 240,
```
Alexey-Kamenev, line 115: Is this a known constant? Maybe name it or add comment?

Alexey-Kamenev (approved):
LGTM!

examples/cfd/ldc_pinns/config.yaml
ktangsali: It is possible. But since this is an introductory example, I decided not to do that. The idea was to provide a minimal code with a focus on the utilities. For FD and Least Squares, the dataloader pipeline will have to be different and enabling that via config can become a bit verbose.

examples/cfd/stokes_mgn/pi_fine_tuning_gnn.py
ktangsali: So actually the `PhysicsInformer` utility will compute residuals on any points that the user passes. Since here, the points passed include the boundary points, the residuals computed on those points are not required as we need to apply BCs on them instead. Hence the masking was added. Potentially we have a higher level of abstraction at a later point?

examples/cfd/darcy_physics_informed/darcy_physics_informed_fno.py
ktangsali: Done :)

PR #784, opened 2025-02-06
Enable Domain Parallelism with ShardTensor

coreyjadams:
# Modulus Pull Request

## Description
This PR adds new capabilities to Modulus:
- `ShardTensor` is an extension to pytorch `DTensor` that enables uneven sharding of tensors across DeviceMesh objects.  While some logical sharding constraints remain, this allows more dynamic and flexible operation on distributed input data, especially in cases where the input data shape and output data shape differ.
- `ShardTensor` also enables an ecosystem of operation extensions.  Two major ones are included in this PR: convolutions (1D/2D/3D) and neighborhood attention.  When the right components of modulus are imported, these operations (when performed on sharded tensors) will automatically compute halo regions and perform data transfers to enable results consistent with single device outputs.
  - For small data, this is not useful, but for extremely large data this is a powerful way to scale training on large inputs.
- The documentation for Modulus now includes an API reference for `ShardTensor`, as well as an example of integrating multiple levels of parallelism by combining shard tensor and pytorch `FSDP`.

## Checklist

## Dependencies

Adds a dependency on `wrapt` for monkey-patching operations on sharded inputs..


docs/tutorials/fsdp_and_shard_tensor.rst
```diff
@@ -0,0 +1,230 @@
+ShardTensor and FSDP Tutorial
```
pzharrington, line 1: Would maybe put "Domain Decomposition" in the title here to highlight this is not just some other way of doing FSDP

docs/api/modulus.distributed.shardtensor.rst
```diff
@@ -0,0 +1,117 @@
...
+
+ShardTensor helps address this by:
+- Distributing the input data across multiple devices
+- Performing operations on smaller local portions
+- Coordinating the necessary communication between devices in the forward and backward passes
+
+ShardTensor is built as an extension of PyTorch's DTensor, and gains substantial functionality by leveraging the utilities already implemented in the PyTorch distributed package.  However, some operations on sharded input data are not trivial to implement correctly, nor relevant to the model sharding problem.  In Modulus, we have implemented parallelized versions of several key operations, including (so far):
+
+- Convolution (1D, 2D, 3D)
+- Neighborhood Attention (2D)
```
pzharrington, line 107: Whether it's here or elsewhere, probably good to keep an official list of ops that are supported, and maybe link to them

docs/tutorials/fsdp_and_shard_tensor.rst
```diff
@@ -0,0 +1,230 @@
...
+.. code-block:: bash
+
+    torchrun --nproc_per_node=4 train_cnn.py
+
+This will train the model using both data parallelism (FSDP) and spatial decomposition (ShardTensor) across 4 GPUs in a 2x2 configuration.
+
+Key Points
+---------
+
+1. The device mesh is split into two dimensions: one for data parallelism (FSDP) and one for spatial decomposition (ShardTensor)
```
pzharrington, line 216: Can you show these lines as well to make the example more complete?

modulus/distributed/__init__.py
```diff
@@ -27,3 +27,8 @@
     reduce_loss,
     unmark_module_as_shared,
 )
+
+from .shard_tensor import ShardTensor, scatter_tensor
+
+# Load and register custom ops:
+from .custom_ops import *
```
pzharrington, line 34: Should probably try and avoid the wildcard imports, if it's not too much trouble

modulus/distributed/shard_tensor.py
```diff
@@ -0,0 +1,564 @@
...
+)
+
+from modulus.distributed._shard_redistribute import (
+    ShardRedistribute
+)
+from torch.distributed.device_mesh import _mesh_resources, DeviceMesh
+
+
+from modulus.distributed import DistributedManager
+class _ToTorchTensor(torch.autograd.Function):
```
pzharrington, line 51: Formatting (add newline between import and class def). Did linting not catch this?

pzharrington:
\blossom-ci

modulus/distributed/shard_utils/natten_patches.py
```diff
@@ -0,0 +1,368 @@
...
+    Raises:
+        MissingShardPatch: If kernel configuration is not supported for sharding
+            - Even kernel sizes not supported
+            - Dilation != 1 not supported
+    """
+    # Currently, reject even kernel_sizes and dilation != 1:
+    if kernel_size % 2 == 0:
+        raise MissingShardPatch("Neighborhood Attention is not implemented for even kernels")
+    if dilation != 1:
+        raise MissingShardPatch("Neighborhood Attention is not implemented for dilation != 1")
```
pzharrington, line 37: Is this a NATTEN limitation or our own?

pzharrington:
Overall, I think this is looking good, nice work! I started with the documentation and then focused on unit tests to see overall functionality, as well as changes to the `DistributedManager` to see device mesh functionality and the main changes to what existed previously. Also looked at the `ShardTensor` definition, halo collectives and conv/natten patches, but didn’t spend much time on the other backend stuff.

Aside from the minor comments added, my main flag is to make the unit testing more complete, but I don’t think that should necessarily block merging. In particular I think for ops that we support (conv or nat, currently), we should add unit tests for correctness compared to a non-sharded baseline for forward and backward passes (subject to within some numerical tolerance, esp. in context of the neighborhood attention numerics we discovered).

modulus/distributed/__init__.py
coreyjadams: Good idea, resolved.

modulus/distributed/shard_utils/natten_patches.py
coreyjadams: Initially, both were my limitations and focusing on the case actually in production.  It's feasible to add either, its all about getting the halo size correct and subsequent trimming correct.

Are there use cases where these are needed?  After this release I want to prioritize the list of operations to support, including cases like these.

coreyjadams:
Thanks for the review @pzharrington!  I agree with you on the testing.  Here's my thoughts:
- Modulus should support unit tests of the basic ShardTensor functionality, "baked in".  Most of those are there, but I have locally some tests in development regarding gradient propagation through sharded tensors.  I would like to get them in but didn't want to hold the review.
- Tests on numerical accuracy are probably too much for CI/CD and unit testing.  I am working on exactly the tools you highlighted, but it's currently manually run and analyzed.  It aims to support numerical checking (and performance benchmarking!) of all the operations we patch like this in modulus, as well as extending to more comprehensive layers and even full models.  I'll get the repo up on gitlab to iron out the kinks and keep the numerical checking untied from the modulus release.  FYI, apart from the issues we found in na2d for long sequence lengths, all patched operations are passing numerical checks.

modulus/distributed/shard_utils/natten_patches.py
pzharrington: Nice. I don't think it's immediately needed but dilation would be useful to add for NATTEN as their computer vision papers show that being a nice way to increase receptive fields without adding too much computational overhead

docs/tutorials/fsdp_and_shard_tensor.rst
```diff
@@ -0,0 +1,231 @@
...
+            
+        def forward(self, x):
+            # This is automatically parallel:
+            x = self.conv(x)
+            x = self.relu(x)
+            # This operation reduces on the parallel dimension.
+            # This will leave x as a Partial placement, meaning
+            # it isn't really sharded anymore but the results on the domain
+            # pieces haven't been computed yet.
+            x = self.pool(x)
```
akshaysubr, line 78: How does this work with DTensor? I don't see any pooling ops in https://github.com/pytorch/pytorch/tree/937b41e3b55e9dc682ac2027b5af17b1d7d75489/torch/distributed/tensor/_ops

docs/tutorials/fsdp_and_shard_tensor.rst
```diff
@@ -0,0 +1,231 @@
...
+            x = torch.flatten(x, 1)
+            x = self.fc(x)
```
akshaysubr, line 80: This is automatically changed to all local ops by `distribute_model`?

docs/tutorials/fsdp_and_shard_tensor.rst
```diff
@@ -0,0 +1,231 @@
...
+
+    def train_epoch(model, optimizer, criterion):
+        model.train()
+        
+        for i in range(10):  # 10 training steps
+            # Get sharded data
+            inputs, targets = create_sample_data()
+            
+            # Forward pass
+            outputs = model(inputs)
```
akshaysubr, line 162: Presumably the outputs are replicated across the spatial mesh right? It's not clear to me where/how that happens.

docs/api/modulus.distributed.shardtensor.rst
```diff
@@ -0,0 +1,117 @@
...
+    from modulus.distributed import DistributedManager
+    from modulus.distributed.shard_tensor import ShardTensor, scatter_tensor
+
+    def main():
+        # Initialize distributed environment
+        DistributedManager.initialize()
+        dist = DistributedManager()
+
+        # Create a 1D device mesh
+        mesh = DeviceMesh(device_type="cuda", mesh=[0, 1, 2, 3])
```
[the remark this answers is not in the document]
coreyjadams: Sure, I can update this script.  I've updated `DistributedManager` in this script to enable DeviceMesh creation too.

docs/tutorials/fsdp_and_shard_tensor.rst
coreyjadams: Under the hood in the pooling layer, it's doing an `aten.mean` operation.  That gets intercepted and redispatched by `DTensor`, which is the fallback dispatch for `ShardTensor`'s dispatch.  The mean operation is handled in upstream pytorch here:
https://github.com/pytorch/pytorch/blob/937b41e3b55e9dc682ac2027b5af17b1d7d75489/torch/distributed/tensor/_ops/_math_ops.py#L306-L355

In general, most operations can fall back to DTensor successfully.  There is a small hiccup here in this operation (doesn't show up in this example, of course) actually:  If this tensor was sharded asymmetrically, the mean would do local averaging followed by a global average when the output is needed.  The global average needs to be weighted by the shape of the un-reduced tensor, but currently isn't.  It's actually an expected-fail in the tests in this branch, for exactly this reason, but I don't think it should block progress since it's a relatively rare path to go down.  It's on my agenda to fix but likely requires sub-classing the `Partial` placement in DTensor to catch and handle.

docs/tutorials/fsdp_and_shard_tensor.rst
coreyjadams: No, it changes all the underlying parameters in the layers from `torch.Tensor` to `torch.distributed.tensor.DTensor`.  `DTensor` comes with it's own dispatch, and when you call the thing returned by `distribute_module` with a DTensor or ShardTensor it will make a decision of how to process based on a) the operation being dispatched and b) the input tensor(s). 

When you do the `pool` operation in this example, the pooling actually takes place over the sharded dimension.  So, the output `x` on line 78 is a ShardTensor/DTensor with `Replicate()` placement instead of `Shard(2)` placement (which was the height dimension).

Subsequent operations will see that the weights in `fc` are `Replicate` sharding, and the inputs to `fc` are `Replicate` sharding, so it just has to use an all-local operation and wrap the output in a `Replicate`'d DTensor.  ShardTensor then intercepts that and keeps the domain data as `ShardTensor`.

In the backwards pass, DTensor (and hence ShardTensor) should automatically un-replicate the gradients and shard them appropriately.  So the input gradients to the upstream layers here (conv, relu) will receive shard tensors of the correct size.

docs/tutorials/fsdp_and_shard_tensor.rst
coreyjadams: Here, yes, they are replicated.  In this example it happens specifically in the `pool` layer.

In general, it need not be replicated - depending on your model and loss function you may have different needs.  For this classification example, where the labels are small and replicated, it makes sense to replicate them.  If we didn't have the `pool` layer the loss computation would call `outputs = outputs.full_tensor()` which explicitly (and differentiably) summons the outputs.

On the other hand, for something like a segmentation loss where the output size == input size == label size, the computation of the loss may actually make more sense to perform as a distributed operation.  In that case, a per-pixel softmax cross entropy between the output and the label would produce itself a sharded tensor.  When you sum (or average) across the image shape, you'll be left with scalars on each rank in the `Partial` placement - meaning they are reduced on the local device but not the global device.

Most of this is transparently handled by the DTensor dispatch + operation rules.  Combined with the generic way to redistribute tensors (from dtensor) + the reimplementation for asymmetrically sharded tensors here, most single-tensor operations and straightforward operations (like `a + b` for two DTensors) can work out of the box and under the hood.

modulus/distributed/shard_utils/conv_patches.py
```diff
@@ -0,0 +1,495 @@
...
+
+    
+
+
+@wrapt.patch_function_wrapper('torch.nn.functional', 'conv1d')
+def conv1d_wrapper(wrapped, instance, args, kwargs):
+    
+    return generic_conv_nd_wrapper(wrapped, instance, args, kwargs)    
+    
+@wrapt.patch_function_wrapper('torch.nn.functional', 'conv2d')
```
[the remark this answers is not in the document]
coreyjadams: This is the feedback I have been expecting!  :)

I hear your point.  I'll outline *why* I did it this way and not the otherway, maybe we'll find a path that meets all needs.  Sorry it took awhile to figure out, I can (if we keep it) do a write up of how exactly this is working.

I wanted to meet these user-facing requirements in the implementation:
- Easy to use interface.  Ideally, the "model code" should remain unchanged or as close to unchanged as possible to enable domain parallelism.
  - We could, maybe, have a big registry of "nn.Conv2d -> modulus.distributed.nn.Conv2d" that gets used when `distribute_module` is applied as a partition_fn.  Then, users would have to apply our partition function + any other partition functions needed.
  - The spirit of the partition function, though, appears to be to enable model parallelism (not domain parallelism as I am doing here).  `distribute_module` is operating on the model parameters and not how those parameters are interacting with the input data.
- Don't break un-sharded codes
  - I wanted the operation to get called correctly regardless of whether the input is sharded or not.  Using a module replacement technique, it likely comes down to the same steps, but every module we built to do this would essentially be a custom layer.
  - I know you could do this with a module mapping/replacement too.
- Support `torch.nn.functional` interface
  - What happens if you have a custom layer that stores a convolutional weight, bias and then calls an operation with `torch.nn.functional.conv2d`?  This would need to be specifically enabled.

In the end, I went with this approach since it was most flexible to the user.  The monkey patching will pick the right operation based on the input types, and in fact I built the sharded ops on even lower level operations (`aten.convolution.default`) for example so it's a one-op-fits all to support convolutions.

The best of both worlds, after writing this up, is likely to use this functional-level mapping and a registry for dispatch within `ShardTensor` itself.  I think that's possible; it definitely won't be ready for this release.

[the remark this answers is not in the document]
coreyjadams: Also, can you elaborate on how this breaks the save / from_checkpoint interface?  I think it might actually be `distribute_module` that might break these, but in general this dispatch/dynamic interception of layers shouldn't break the way models are stored and loaded.

The one exception is if users need to shard their models parameters for domain sharding.  This can occur if you're adding position embeddings or have other trainable parameters with shapes dependent on data shape.  Then, we'd need to update save / from_checkpoint to ensure:
- Models coalesce full parameters before saving.
- Sharding / placements are (optionally) restored on loading.

docs/tutorials/fsdp_and_shard_tensor.rst
akshaysubr: Yeah, that makes sense that `fc` weights are replicated since that's the default with `distribute_model`.

This actually brings up a good point. Ideally, we'd want to parallelize the `fc` in this example as well since it's a lot of redundant compute right? How would one do that? Through a custom `partition_fn` in `distribute_model`? That would work in this case. How would it work in a larger model with many different `Linear` layers which all might need different handling in `partition_fn`?

docs/tutorials/fsdp_and_shard_tensor.rst
akshaysubr: Got it, traced that down to this piece of code that tells you that it will be a `Partial` placement: https://github.com/pytorch/pytorch/blob/937b41e3b55e9dc682ac2027b5af17b1d7d75489/torch/distributed/tensor/_ops/_math_ops.py#L248-L251.

When does this `Partial` placement get resolved? In the `fc` I assume? Is that basically calling a `redistribute` to `Replicated` and that triggers an allreduce? If that `fc` was rowwise (or colwise depending on convention) sharded, then it would `redistribute` to `Shard(0)` and that would trigger a reduce-scatter instead of an allreduce?

coreyjadams: I'm not 100% sure when it resolves, actually.  But what I *think* is happening is:
- Ops have a set of rules defined, for input placements, to determine output placements.
- If the placements aren't going to match after computation, a redistribute is triggered.  https://github.com/pytorch/pytorch/blob/937b41e3b55e9dc682ac2027b5af17b1d7d75489/torch/distributed/tensor/_dispatch.py#L177-L185

I want to look into this more: the details of when and why are important when parallelism is being composed.

docs/tutorials/fsdp_and_shard_tensor.rst
coreyjadams: This model is so small that the `fc` is probably irrelevant in terms of compute, but yes, I get your point.  I think weigh partitioning, in almost every case, should be handled with existing FSDP tools.  That ought to allow most parallelizations to work out of the box, though it needs to be fully checked and confirmed.

This is on my to-do list for the next release, 3D parallelism over batching, domain sharding, and model parallelism.  It's not ready in this release though.

ktangsali:
Reminder to change the target to [`0.10.0-rc`](https://github.com/NVIDIA/modulus/tree/0.10.0-rc) branch before merging.

docs/api/modulus.distributed.shardtensor.rst
```diff
@@ -0,0 +1,117 @@
+
+Modulus Shard Tensor
+===========
+
+In scientific AI applications, the parallelization techniques to enable state of the art 
+models are different from those used in training large language models.  Modulus 
+introduces a new parallelization primitive called a ShardTensor that is designed for
```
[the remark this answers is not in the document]
coreyjadams: Done.

coreyjadams:
Target updated to 0.1.0.0-rc.  All smaller comments from @akshaysubr resolved.  Most residual bugs fixed.  

This version of the release supports ShardTensor for select models to enable domain parallelism, not yet as a general tool.

akshaysubr (approved):

pzharrington (approved):

PR #793, opened 2025-02-13
Dockerfile changes to handle onnxruntime dependency

ktangsali:
# Modulus Pull Request

## Description
Changes to Dockerfile and tests to handle onnxruntime dependency. 

Onnx is still not shipping arm wheels, hence the installation from a pre-built wheel. Additionally, the tests can now be safely changed to support any onnxruntime version > 1.19.0 as the change is upstreamed. 

## Checklist

### DIFF
--- a/test/deploy/test_onnx_fft.py
+++ b/test/deploy/test_onnx_fft.py
@@ -32,31 +32,35 @@ except ImportError:
 
 from typing import Tuple
 
+from packaging.version import Version
+
 from modulus.deploy.onnx import export_to_onnx_stream, run_onnx_inference
 
 Tensor = torch.Tensor
 logger = logging.getLogger("__name__")
 
 
-# TODO(akamenev): remove once the bug below is fixed.
-# Version "1.14.0" is the custom local build where the bug is fixed.
 def check_ort_version():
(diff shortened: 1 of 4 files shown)


test/deploy/test_onnx_fft.py
```diff
@@ -32,31 +32,35 @@
...
+
 from modulus.deploy.onnx import export_to_onnx_stream, run_onnx_inference
 
 Tensor = torch.Tensor
 logger = logging.getLogger("__name__")
 
 
-# TODO(akamenev): remove once the bug below is fixed.
-# Version "1.14.0" is the custom local build where the bug is fixed.
 def check_ort_version():
```
Alexey-Kamenev, line 43: It looks like `check_ort_version` here is the same as in the next file. Maybe extract it to utils or something to avoid having 2 copies?

Alexey-Kamenev (approved):
LGTM!

test/deploy/test_onnx_fft.py
ktangsali: Good point. Let me try to do that

ktangsali: Fixed!

PR #790, opened 2025-02-11
Cordiff usability and performance enhancements for custom dataset training

CharlelieLrt:
# Modulus Pull Request

## Description
This PR provides enhancements for usability of CorrDiff training on custom datasets. It includes:
- Documentation improvements (readme, docstrings)
- A new general purpose patching API with enhanced usability and performance
- Refactor CorrDiff samplers and losses to delegate all patching logic to the dedicated API
- Refactor CorrDiff UNets positional embeddings for performance
- Config files refactoring and simplification
- Legacy code elimination
- Support for non-square images
- Experimental support for non-square patches
- Other bugfixes

Closes #726.
Closes #447.
Closes #417.
Closes #448.
Closes #552.
Closes #542.
Closes #536.

## Note
This PR introduces breaking changes on the way corrdiff config files are handled.
For training, older version of the config files is still supported.
For generation, a new parameter `generation.patching: true/false` is required.

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/modulus/blob/main/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [x] The [CHANGELOG.md](https://github.com/NVIDIA/modulus/blob/main/CHANGELOG.md) is up to date with these changes.
- [x] An [issue](https://github.com/NVIDIA/modulus/issues) is linked to this pull request.

PR #809, opened 2025-03-14
Various Corrdiff optimizations for drastic increase of training efficiency

LostnEkko:
# Various Corrdiff optimizations for drastic increase of training efficiency

## Description

- Updated CorrDiff training code to support multiple patch iterations
to amortize regression cost and usage of `torch.compile`
- Refactored `modulus/models/diffusion/layers.py` to optimize data type casting workflow,
avoiding unnecessary casting under autocast mode
- Refactored Conv2d to enable fusion of conv2d with bias addition
- Refactored GroupNorm, UNetBlock, SongUNet, SongUNetPosEmbd to support usage of
Apex GroupNorm, fusion of activation with GroupNorm, and AMP workflow.
- Updated SongUNetPosEmbd to avoid unnecessary HtoD Memcpy of `pos_embd`
- Updated `from_checkpoint` to accommodate usage of Apex GroupNorm
- Refactored CorrDiff NVTX annotation workflow to be configurable

## Checklist


CharlelieLrt:
@simonbyrne it's currently blocked by #790 and under review, but that will be merged in the coming days.

modulus/metrics/diffusion/loss.py
```diff
@@ -651,6 +651,230 @@ def __call__(
         return loss
 
 
+class ResLoss_Opt:
```
CharlelieLrt, line 654: Would it be possible to merge it with the current `ResidualLoss`? I might be missing something, but it doesn't look like too much to add to `ResidualLoss` to unify them

modulus/models/diffusion/layers.py
```diff
@@ -143,16 +156,25 @@ def __init__(
...
+            )
+            fused_conv_bias = False
 
         super().__init__()
         self.in_channels = in_channels
         self.out_channels = out_channels
         self.up = up
         self.down = down
         self.fused_resample = fused_resample
+        self.fused_conv_bias = fused_conv_bias
```
CharlelieLrt, line 176: I am wondering if:
1. We are we loosing any useful functionality/flexibility when using `fused_conv_bias=True` if comparison to the former non-fused implementation
2. There could be a case where `fused_conv_bias=True` makes performance worse>
If the answer to these 2 questions is negative, I would advocate to always use the fused convolution and remove all code related to the non-fused case

modulus/models/diffusion/layers.py
```diff
@@ -143,16 +156,25 @@ def __init__(
         init_mode: str = "kaiming_normal",
         init_weight: float = 1.0,
         init_bias: float = 0.0,
+        fused_conv_bias: bool = False,
```
CharlelieLrt, line 159: These options are not documented in the docstring

modulus/models/diffusion/layers.py
```diff
@@ -262,28 +309,66 @@ def __init__(
         num_groups: int = 32,
         min_channels_per_group: int = 4,
         eps: float = 1e-5,
+        use_apex_gn: bool = False,
```
CharlelieLrt, line 312: Docstring update missing

modulus/models/diffusion/layers.py
```diff
@@ -262,28 +309,66 @@ def __init__(
...
+        fused_act: bool = False,
+        act: str = None,
```
CharlelieLrt, line 314: Same remarks as for the fused convolution: is there any good reason to keep the poor performance implementation?

modulus/models/diffusion/layers.py
```diff
@@ -262,28 +309,66 @@ def __init__(
...
         self.weight = torch.nn.Parameter(torch.ones(num_channels))
         self.bias = torch.nn.Parameter(torch.zeros(num_channels))
+        self.use_apex_gn = use_apex_gn
+        self.fused_act = fused_act
+        self.act = act.lower() if act else act
+        self.act_fn = None
+        self.amp_mode = amp_mode
+        if self.use_apex_gn:
+            if self.act:
+                self.gn = ApexGroupNorm(
```
CharlelieLrt, line 329: Note: generally speaking we are trying to have everything runnable on cpu. In that case could we have a fallback to the non-Apex implementation if cuda is not available?

modulus/models/diffusion/layers.py
```diff
@@ -404,6 +511,12 @@ def __init__(
         init: Dict[str, Any] = dict(),
         init_zero: Dict[str, Any] = dict(init_weight=0),
         init_attn: Any = None,
+        use_apex_gn: bool = False,
```
CharlelieLrt, line 514: Docstring to update with these aguments

modulus/models/diffusion/layers.py
```diff
@@ -404,6 +511,12 @@ def __init__(
...
+        fused_act: bool = True,
```
CharlelieLrt, line 515: Same remark as for the fused convolution: is there any good reason to keep the former, non-fused implementation?

modulus/models/diffusion/layers.py
```diff
@@ -520,20 +684,28 @@ class PositionalEmbedding(torch.nn.Module):
     """
 
     def __init__(
-        self, num_channels: int, max_positions: int = 10000, endpoint: bool = False
+        self,
+        num_channels: int,
+        max_positions: int = 10000,
+        endpoint: bool = False,
```
CharlelieLrt, line 690: To document in the docstring

modulus/models/diffusion/layers.py
```diff
@@ -557,11 +729,17 @@ class FourierEmbedding(torch.nn.Module):
         and thereby the frequency of oscillations in the embedding space. By default 16.
     """
 
-    def __init__(self, num_channels: int, scale: int = 16):
+    def __init__(self, num_channels: int, scale: int = 16, amp_mode: bool = False):
```
CharlelieLrt, line 732: To document in the docstring

modulus/models/diffusion/song_unet.py
```diff
@@ -156,6 +157,12 @@ def __init__(
         resample_filter: List[int] = [1, 1],
         checkpoint_level: int = 0,
         additive_pos_embed: bool = False,
+        use_apex_gn: bool = False,
+        fused_act: bool = True,
+        act: str = "silu",
+        fused_conv_bias: bool = True,
```
CharlelieLrt, line 163: Same comment as for the optimization options introduced in `layers.py`. IMO this model definition starts to become too complicated, with too many options. I understand that this gives very fine-grained control to the user, but if some simplification is possible then that would be great.

modulus/models/diffusion/song_unet.py
```diff
@@ -156,6 +157,12 @@ def __init__(
         resample_filter: List[int] = [1, 1],
         checkpoint_level: int = 0,
         additive_pos_embed: bool = False,
+        use_apex_gn: bool = False,
```
CharlelieLrt, line 160: To document in docstring

modulus/models/diffusion/song_unet.py
```diff
@@ -507,6 +570,13 @@ def __init__(
         gridtype: str = "sinusoidal",
         N_grid_channels: int = 4,
         checkpoint_level: int = 0,
+        additive_pos_embed: bool = False,
```
CharlelieLrt, line 573: To document in docstring + same as for `SongUnet`

modulus/models/module.py
```diff
@@ -354,14 +356,40 @@ def from_checkpoint(cls, file_name: str) -> "Module":
...
+            if model_args is not None:
+                args["__args__"].update(model_args)
+            model = cls.instantiate(args)
             # Load the model weights
             model_dict = torch.load(
                 local_path.joinpath("model.pt"), map_location=model.device
             )
-            model.load_state_dict(model_dict)
-
+            # TODO: for corrdiff model architecture specifically
```
CharlelieLrt, line 369: Two questions/comments:
1. Is this to ensure backward compatibility with previous checkpoints?
2. If this is CorrDiff-specific code we probably don't want to have it in `module.py`. Maybe in a CorrDiff "checkpoint patch" instead?
In any case, I also introduced breaking changes to the models APIs in #790 , so we will need to find a more robust mechanism to handle backward compatibility of older checkpoints

modulus/models/module.py
```diff
@@ -354,14 +356,40 @@ def from_checkpoint(cls, file_name: str) -> "Module":
...
+            if (
+                not apex_in_ckp
+                and "use_apex_gn" in args["__args__"].keys()
+                and args["__args__"]["use_apex_gn"]
+            ):
+                filtered_state_dict = {}
+                for key, value in model_dict.items():
+                    filtered_state_dict[key] = value  # Keep the original key
+                    # Duplicate weight/bias for Apex GroupNorm (without removing the original)
+                    for norm_layer in ["norm0", "norm1", "norm2", "aux_norm"]:
```
CharlelieLrt, line 379: We definitely don't want this in `module.py`

modulus/models/diffusion/song_unet.py
```diff
@@ -325,91 +367,112 @@ def __init__(
...
+                            x.shape[-1] > (self.checkpoint_threshold / 2)
+                            and "_up" in name
+                        ):
+                            x = checkpoint(block, x, emb)
+                        else:
+                            x = block(x, emb)
+            return aux
 
 
 class SongUNetPosEmbd(SongUNet):
```
CharlelieLrt, line 478: So, these optimizations are only for the `SongUNet` and the `SongUNetPosEmb`? What about the ine with lead-time embedding (`SongUNetPosLtEmb`)?
To be fair, I think the `SongUNetPosLtEmb` should have never existed because it should have inherited from `SongUNetPosEmb`, and your optimization would have been applied to both with minimal code addition to `SongUNetPosLtEmb`. But unfortunately, I think the current design does not allow that...

modulus/metrics/diffusion/loss.py
```diff
@@ -651,6 +651,230 @@ def __call__(
...
+        self.P_mean = P_mean
+        self.P_std = P_std
+        self.sigma_data = sigma_data
+        self.img_shape_x = img_shape_x
+        self.img_shape_y = img_shape_y
+        self.patch_shape_x = patch_shape_x
+        self.patch_shape_y = patch_shape_y
+        self.patch_num = patch_num
+        self.hr_mean_conditioning = hr_mean_conditioning
+        self.y_mean = None
```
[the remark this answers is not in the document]
jialusui1102: Apologies. My github account was associated with my personal emails and I wasn't notified of these during working hours. We just resolved some blockers and would like to pursue a merge soon.

y_mean here is for cacheing and reusing the regression results for following multi-iteration patching. Wondering what is our standard format for the docstrings+typing?

modulus/metrics/diffusion/loss.py
jialusui1102: It's definitely possible to merge them. I'm only cacheing the y_mean and selectively reusing them in the optimized version. We though separating them out would be clearer, but if merging them into 1 would actually be a better idea in terms of overall design, we can do that

modulus/models/diffusion/layers.py
jialusui1102: Hey,

1. In terms of `fused_conv_bias=True`, we are not loosing any functionalities. We are now integrating conv + bias as one op at pytorch level, but it still do exactly the same calculation as before.
The only place where we have to set `fused_conv_bias=False` is when we don't have `kernel` as guarded in init. Otherwise it will brings a conflict as I remembered. We can discuss further regarding this.
2. setting `fused_conv_bias=True` would not make performance worse. It would either be the same or faster. In mix-precision training with bf/fp16, if we pass bias to conv2d, pytorch will recognize these as one single op and autocast everything into bf/fp16, leading to faster calculation. In other cases performance would be the same
3.

modulus/models/diffusion/song_unet.py
jialusui1102: We can discuss further about how to potentially simplify this

modulus/models/module.py
jialusui1102: Hey, I think the previous version also loads ckp like this. Not sure whether this line is  tagged as a new edit

modulus/models/module.py
jialusui1102: This is for transfering the regression GN weights as new initialized Apex's weight. We can discuss how to make this better, maybe add as a util?

modulus/models/diffusion/song_unet.py
jialusui1102: only delete fused_conv_bias and add docstrings for the rest

modulus/models/diffusion/song_unet.py
jialusui1102: leave it for now, and submit another PR in the future with Tao to see how to combine it

modulus/metrics/diffusion/loss.py
jialusui1102: Merged with ResidualLoss

modulus/metrics/diffusion/loss.py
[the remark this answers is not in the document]
CharlelieLrt: Just one tiny detail there: is `self.y_mean` supposed to be used outside of the loss instance itself, or is it for internal use in the `__call__` method only? If for internal use only, then consider making it a private attribute with `self._y_mean`

modulus/metrics/diffusion/loss.py
CharlelieLrt: @jialusui1102 thanks for doing this refactor and merging the 2 losses!

After our discussion I thought a little bit more about this, and I realized that using the "patch-wise" gradient accumulation also requires some code outside of the loss itself (i.e. in the training loop), correct? So, automatically enabling this optimization any time the `patching` argument is provided is probably not a strategy, because some users might want to use patching without writting custom accumulation code in their training loop...

So, it's probably better to have a separate argument in the loss (something like `use_patch_grad_acc=False`), that enables the optimization. It's kind naive but we can cleanup in a later PR.

modulus/metrics/diffusion/loss.py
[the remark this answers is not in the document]
jialusui1102: We are supposed to reset loss.y_mean to None before each regression forward calls in train.py.

modulus/metrics/diffusion/loss.py
jialusui1102: Yes, in train.py we need to reset loss.y_mean before each actual regression forward (first iteration of patch accumulations). I agree that adding a separate flag would be better

modulus/models/diffusion/song_unet.py
jialusui1102: @CharlelieLrt 
Hey Charlelie, I would remove `fused_act` argument in SongUnet or UNetBlock as I just realized that for either SongUnet or UNetBlock, fused act vs non-fused act brings no difference, so it doesn't worth to create 2 paths for this choice. So I would hard-code SongUnet and UNetBlock to always fuse act with GN, but keep the `fused_act` option in case other projects have different usage.

modulus/metrics/diffusion/loss.py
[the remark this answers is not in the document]
CharlelieLrt: Got it, let's keep it as `self.y_mean` then and let's defer to future PR for something cleaner

modulus/models/diffusion/song_unet.py
CharlelieLrt: Sounds good to me

modulus/models/module.py
CharlelieLrt: @jialusui1102 
FYI I am preparing a small PR to handle backward compatibility of old checkpoints. All the conversion logic is in a `models/utils_compatibility.py`, and the `Module` class in `models/module.py` call these utility functions to perform conversions of arguments, names, etc... on old checkpoint.
For you it means you should move your transfer logic to `models/utils_compatibility.py`.

Also, could you drop a comment (or maybe even an item in the FAQ section of the `README.md`) to clarify these 2 scenarios:
1. What happens if a user loads a checkpoint with no optimization and wants to use it with the optimizations enabled. (Is it even possible?)
2. What happens if a user loads a checkpoint with optimizations and wants to use it without the optimizations (e.g. move to a machine where apex is not available)

modulus/models/diffusion/layers.py
jialusui1102: updated docstrings

modulus/models/diffusion/layers.py
jialusui1102: updated docstrings

modulus/models/diffusion/song_unet.py
jialusui1102: removed `fused_act` and `fused_conv_bias ` in init and hard-coded some values. updated docstrings

modulus/models/diffusion/song_unet.py
jialusui1102: updated docstring

modulus/models/diffusion/song_unet.py
jialusui1102: updated docstring same as SongUnet

modulus/models/diffusion/layers.py
CharlelieLrt: @jialusui1102 
Correct me if I'm wrong, but after our discussion I think we agreed that if users want to run on CPU (and/or don't have apex installed), then they should pass `use_apex_gn=False`, and everything should work... We just need to make sure this use is well documented in the docstring.

modulus/models/module.py
jialusui1102: Sure, I will add the conversion function in `models/utils_compatibility.py`

For answering the questions:
1. This is definitely possible. All user need to do is to specify: `use_apex_gn: True` in the config, and the ckp conversion function I will add in `models/utils_compatibility.py` will handle the conversion automatically
2. Thanks for bringing this up, I will add this conversion function in `models/utils_compatibility.py` as well and automatically do the conversion if the ckp loaded is  trained with apex while user set `use_apex_gn: False` in config.

I will add these 2 items in FAQ session after I'm done with writing those 2 conversion functions and make sure everything works

modulus/models/diffusion/layers.py
jialusui1102: Yes, If users want to run on CPU and apex is not supported, they need to pass use_apex_gn=False and GN will call pytorch GN in training and custom GN in inference
Added this in docstring:

> use_apex_gn : bool, optional
        A boolean flag indicating whether we want to use Apex GroupNorm for NHWC layout. 
        Need to set this as False on cpu. Defaults to False.

modulus/models/diffusion/layers.py
jialusui1102: added docstrings

modulus/models/diffusion/layers.py
jialusui1102: resolved in meetings

modulus/models/diffusion/layers.py
jialusui1102: resolved in meeting

modulus/models/diffusion/layers.py
jialusui1102: added docstring

modulus/metrics/diffusion/loss.py
jialusui1102: Hi @CharlelieLrt 
I was testing the merged Loss function and realized that in the optimized ResLoss, I have to cast some variables to channelslast whereas in the original ResLoss those variables are casted to be contiguous. If we were to combine these 2 Losses, I have to add if statements for this memory layout casting as well. Let me know how you think!

CharlelieLrt: @jialusui1102 
Sure, feel free to modify as you need! Indeed it makes more sense to have `apply(input)` return a tensor with same layout as `input`

jialusui1102: Hey @CharlelieLrt ,

I am seeing graph recompilation leading to incompetent performance after `torch.compile` due to the following reasons. Wondering do you have any insight into these?

> /usr/local/lib/python3.12/dist-packages/torch/_inductor/compile_fx.py:222: UserWarning: TensorFloat32 tensor cores for float32 matrix multiplication available but not enabled. Consider setting `torch.set_float32_matmul_precision('high')` for better performance.
  warnings.warn(
[rank0]:V0428 16:54:27.763000 544908 torch/_dynamo/guards.py:2773] [1/1] [__recompiles] Recompiling function forward in /code/modulus/examples/generative/corrdiff/physicsnemo/models/diffusion/preconditioning.py:823
[rank0]:V0428 16:54:27.763000 544908 torch/_dynamo/guards.py:2773] [1/1] [__recompiles]     triggered by the following guard failure(s):
[rank0]:V0428 16:54:27.763000 544908 torch/_dynamo/guards.py:2773] [1/1] [__recompiles]     - 1/0: L['model_kwargs']['embedding_selector'].__closure__[1].cell_contents.patch_indices[0][0] == 483
[rank0]:V0428 16:54:53.367000 544908 torch/_dynamo/guards.py:2773] [1/2] [__recompiles] Recompiling function forward in /code/modulus/examples/generative/corrdiff/physicsnemo/models/diffusion/preconditioning.py:823
[rank0]:V0428 16:54:53.367000 544908 torch/_dynamo/guards.py:2773] [1/2] [__recompiles]     triggered by the following guard failure(s):
[rank0]:V0428 16:54:53.367000 544908 torch/_dynamo/guards.py:2773] [1/2] [__recompiles]     - 1/1: Ne(L['model_kwargs']['embedding_selector'].__closure__[1].cell_contents.patch_indices[1][0], 0)  # out[B * i : B * (i + 1),] = input[  # ode/modulus/examples/generative/corrdiff/physicsnemo/utils/patching.py:312 in apply (_dynamo/utils.py:2426 in run_node)
[rank0]:V0428 16:54:53.367000 544908 torch/_dynamo/guards.py:2773] [1/2] [__recompiles]     - 1/0: L['model_kwargs']['embedding_selector'].__closure__[1].cell_contents.patch_indices[0][0] == 483
[rank0]:V0428 16:55:09.714000 544908 torch/_dynamo/guards.py:2773] [0/1] [__recompiles] Recompiling function forward in /code/modulus/examples/generative/corrdiff/physicsnemo/models/diffusion/unet.py:132
[rank0]:V0428 16:55:09.714000 544908 torch/_dynamo/guards.py:2773] [0/1] [__recompiles]     triggered by the following guard failure(s):
[rank0]:V0428 16:55:09.714000 544908 torch/_dynamo/guards.py:2773] [0/1] [__recompiles]     - 0/0: tensor 'L['self']._modules['model']._buffers['pos_embd']' dtype mismatch. expected Double, actual Float

CharlelieLrt: @jialusui1102 could you please commit your latest changes? That will help me understand which line correspond to line 312 in `patching.py`

CharlelieLrt: I think there are 2 things there, right? Some recompilations for the `EDMPrecondSuperResolution` and a recompilation for the `UNet`.

For the `EDMPrecondSuperResolution` it seems to be due to some side effects. In its forward pass, `EDMPrecondSuperResolution`:

1. Gets the positional embeddings executing the `embedding_selector` function, which is itself only a wrapper around `patching.apply` method.
2. `patching.apply` extracts patches based on an attribute `patching.patch_indices`
3. This attribute `patching.patch_indices` is dynamic and is randomly generated at every epoch.

I think dynamo is not able to handle such effects (?) A workaround might be to pass the `patch_indices` explicitly as an argument to the `embedding_selector` function.

jialusui1102: Thanks for the insight! That helps me better understand how `embedding_selector ` works! I will push my newest changes asap.
I see. So `embedding_selector ` in SongUNetPosEmbd is essentially just patching.apply. The apply function looks pretty straightforward, not sure why dynamo cannot handle this given it's working fine with the old codebase where the functionality is basically the same.

I will keep debugging in this direction, thanks so much for the info!

CharlelieLrt: Sure, let me know if you have other questions. If I compare to the old codebase I can list the following differences which might influence dynamo:

1. A couple of `if` statements in the `forward` to check the method used to extract patches.
2. The `embedding_selector` callback, which dynamically changes and depends on external attributes from the `patching` object.
3. The fact that the patch extraction is now _within_ the `forward`, whereas it was before done in the loss.

And I think that's it

jialusui1102: Would you mind elaborate more about point 2, especially how `embedding_selector` dynamically changes based on `patching`? I suspect that when it actually changes dynamically, dynamo forced a graph recompile.

CharlelieLrt: Sure. At every epoch:

1. New `patching.patch_indices` are randomly sampled ([there](https://github.com/NVIDIA/physicsnemo/blob/14129168a2151b272632317d4692d39c8202ea95/examples/generative/corrdiff/train.py#L619C45-L619C87))
2. The loss defines the `embedding_selector` wrapper around `patching.apply` ([there](https://github.com/NVIDIA/physicsnemo/blob/14129168a2151b272632317d4692d39c8202ea95/physicsnemo/metrics/diffusion/loss.py#L720))
3. `embedding_selector` is passed as a callback to `forward` ([there](https://github.com/NVIDIA/physicsnemo/blob/14129168a2151b272632317d4692d39c8202ea95/physicsnemo/metrics/diffusion/loss.py#L743))
4. The `SongUNetPosEmb` calls the `embedding_selector` to exctrat positional embeddings corresponding to the patches ([there](https://github.com/NVIDIA/physicsnemo/blob/14129168a2151b272632317d4692d39c8202ea95/physicsnemo/models/diffusion/song_unet.py#L829-L831))

CharlelieLrt: > I suspect that when it actually changes dynamically, dynamo forced a graph recompile

`embedding_selector` is definitely dynamically created at every call to the loss.
I don't know if that's the issue, though. The dynamo guard failure I am seeing seems to specifically point to the value in `patch_indices` that has changed: 

```
L['model_kwargs']['embedding_selector'].closure[1].cell_contents.patch_indices[0][0] == 483
```

CharlelieLrt: I am not sure why there a recompilation for the regression `UNet`, though.

```
tensor 'L['self']._modules['model']._buffers['pos_embd']' dtype mismatch. expected Double, actual Float
```

I assume it is because of [this type casting](https://github.com/NVIDIA/physicsnemo/blob/14129168a2151b272632317d4692d39c8202ea95/physicsnemo/models/diffusion/song_unet.py#L754-L755), but this should only occur once at the first call to `forward`. Unless the model is called repeteadly on arguments `x` with different `dtype` (but I don't think it's what we are doing).

jialusui1102: Hey @CharlelieLrt ,
Thanks for the insight!

I just check my old perf codebase without rebase, it seems like the 
`tensor 'L['self']._modules['model']._buffers['pos_embd']' dtype mismatch. expected Double, actual Float`
recompile was a old carry over which happens when we are reassigning buffer `pos_embd` as `self.pos_embd = self.pos_embd.to(x.dtype)`. In this case, dynamo would treat this as a dynamic change and force a recompile. One solution was to revert this part to the old version where we always return a casted view of the buffer at runtime, but this require casting the `pos_embd` in every forward, which doesn't make sense in terms of perf

The other recompile `L['model_kwargs']['embedding_selector'].closure[1].cell_contents.patch_indices[0][0] == 483` should be related to the refactoring. I will take a deeper look into this after I fix the first one.

Let me know if you have any suggestion regarding the first recompile issue!

jialusui1102: Hi @CharlelieLrt 

I don't think Dynamo is able to handle the logistic of `self.patch_indices`
`self.patch_indices` was precomputed and was stored as a static attribute of the class. In runtime, dynamo would treat it as invariant and force a recompile whenever it detect a change of this `self.patch_indices`
I tried to initialize this as a tensor so that it would be easier for dynamo to trace. However, this is again seen as data-dependent dynamic slicing by dynamo, and will now cause a graph break.

Wondering do you have any insights in how we should better handle this in a dynamo-friendly way?

physicsnemo/models/diffusion/layers.py
```diff
@@ -19,15 +19,23 @@
...
 """
 
+import contextlib
 from typing import Any, Dict, List
 
 import numpy as np
+import nvtx
 import torch
+import torch.cuda.amp as amp
+from apex.contrib.group_norm import GroupNorm as ApexGroupNorm
```
CharlelieLrt, line 29: @jialusui1102 Could you make the apex imports more robust with something like:

```python
is_apex_available = False
if torch.cuda.is_available():
    try:
        apex_gn = getattr(
            importlib.import_module("apex.contrib_group_norm"), "FusedAdam"
        )
        is_apex_available = True
    except ImportError:
        pass
```

Also we should raise an error when `use_apex_gn = True` and `is_apex_available` is `False`

CharlelieLrt: **EDIT:** done, I directly committed the change to your PR

loliverhennigh:
Hey @jialusui1102, @CharlelieLrt and talked about the backward compatibility issues this PR and this PR https://github.com/NVIDIA/physicsnemo/pull/790 raised. For now we can get this in but I will fix the backward compatibility stuff ASAP after. @CharlelieLrt and I discussed a solution that seems to solve all the issues. Ill need you @jialusui1102 to take a look at the PR when the time comes though to make sure this works with the corrdiff model.

CharlelieLrt (approved):
LGTM!

CharlelieLrt:
Leaving some comments about a few things that were left out and could be addressed by future PRs:

1. Optimizations are not enabled for the `SongUnetPosLetEm` architecture. Ideally we would like to merge (subclass) with the other SonUNets, such that we don't need to reimplement the optimizations.
2. Using patch-wise gradient accumulation still requires many manual changes to training loop. Would be good to automate this with something like a wrapper: `loss_fn = patch_wise_gradient_accumulation(loss_fn)`.
3. Need to restructure the `Conv2D` layers, where the case `kernel=0` is an edge case (that is actually used only once in the `SongUnet`) that complicates the implementation of these optimzations. Ideally the case `kermel=0` should be handled by a different class.

PR #810, opened 2025-03-14
Initial ReGen model release

pzharrington:
# Modulus Pull Request

## Description
Adds the ReGen model ([https://arxiv.org/abs/2406.16947](https://arxiv.org/abs/2406.16947)) training and inference code as an example, including reproducibility for paper figures.

## Checklist

## Dependencies

Additional dependencies for this example specified in `requirements.txt`


examples/weather/regen/example_inference.py
```diff
@@ -0,0 +1,209 @@
...
+)
+
+
+print("cuda available? ", torch.cuda.is_available())
+
+
+## Load Obs data
+
+
+ds_regrid = xr.open_dataset(isd_path)
```
nbren12, line 79: I thought we had discussed hardcoding a list of obs here, so that users can inference w/o the large netCDF full of NaNs.

nbren12 (approved):
LGTM. Would be nice to delete the pdfs though. There are definitely a lot of follow-up focused on usability, but  there probably isn't much point to making the example inference usable until we release a checkpoint.

examples/weather/regen/full_inference.py
```diff
@@ -0,0 +1,382 @@
...
+import netCDF4 as nc
+
+
+# Initialize distributed manager
+DistributedManager.initialize()
+dist = DistributedManager()
+device = dist.device
+
+
+class NetCDFWriter:
```
CharlelieLrt, line 59: There are already NetCDF writers in (e.g. in `modulus/corrdiff/utils/`) Would it be possible to re-use it or do you really need something case-specific here?

examples/weather/regen/obs/station_locations_on_grid.nc
CharlelieLrt, line 1: Should this binary be untracked instead?

examples/weather/regen/paper_figures/climate.py
CharlelieLrt, line 1: Are the scripts to make these figures required for the paper submission?
If not, would be good to remove them now, otherwise to be cleaned-up in a future PR...

examples/weather/regen/paper_figures/evenmore_random_val_stations.npy
CharlelieLrt, line 1: I think it would be better to remove this .npy (?)

examples/weather/regen/training/torch_utils/distributed.py
CharlelieLrt, line 1: Is the distributed.py used/necessary? I think the Modulus distributed manager does pretty much the same thing, right? Would be great to replace it, if possible

examples/weather/regen/training/torch_utils/misc.py
```diff
@@ -0,0 +1,146 @@
...
+import re
+import contextlib
+import numpy as np
+import torch
+import warnings
+import dnnlib
+
+
+try:
+    nan_to_num = torch.nan_to_num  # 1.8.0a0
```
CharlelieLrt, line 27: Why do we need to redefine the `nan_to_num` instead of directly using the one from pytorch?

examples/weather/regen/training/torch_utils/misc.py
```diff
@@ -0,0 +1,146 @@
...
+    decorator.__name__ = fn.__name__
+    return decorator
+
+
+# ----------------------------------------------------------------------------
+# Sampler for torch.utils.data.DataLoader that loops over the dataset
+# indefinitely, shuffling items as it goes.
+
+
+class InfiniteSampler(torch.utils.data.Sampler):
```
CharlelieLrt, line 70: There is already an InfiniteSampler in `modulus/utils/generative/utils.py`. Would it be possible to re-use it instead of defining a new one?

examples/weather/regen/training/torch_utils/misc.py
```diff
@@ -0,0 +1,146 @@
...
+def named_params_and_buffers(module):
+    assert isinstance(module, torch.nn.Module)
+    return list(module.named_parameters()) + list(module.named_buffers())
+
+
+# ----------------------------------------------------------------------------
+# Check DistributedDataParallel consistency across processes.
+
+
+def check_ddp_consistency(module, ignore_regex=None):
```
CharlelieLrt, line 120: Probably not necessary?

examples/weather/regen/training/utils/diffusions/networks.py
```diff
@@ -0,0 +1,766 @@
...
+        self.out_channels = out_channels
+        self.up = up
+        self.down = down
+        self.fused_resample = fused_resample
+        init_kwargs = dict(
+            mode=init_mode,
+            fan_in=in_channels * kernel * kernel,
+            fan_out=out_channels * kernel * kernel,
+        )
+        self.weight = (
```
CharlelieLrt, line 108: Change not required for now, but FYI having separate bias is very bad for perf. Future PR should move this to fused call to `nn.Conv2D`. See #809 for details

CharlelieLrt: Anyways, I guess ultimately this should re-use as much as possible layers from SongUNET used for CorrDiff

CharlelieLrt (approved):
Just left a few comments, mostly about:
1. A few binary files. I'm not sure if those are required for the paper submission but it would be better not to have them in the `main`
2. Some utilities that are already implemented in modulus, which could be re-used if possible.
3. The model that has many poor performance operations. This could re-use many of the layers of the SongUNet defined for CorrDiff.
I understand it's for paper submission, so feel free to address these later in another PR. Otherwise, LGTM.

examples/weather/regen/example_inference.py
pzharrington: I thought we were leaving that for later refactoring

pzharrington:
@CharlelieLrt thanks for the review, I agree with all your comments. As Noah mentioned most of this is in there for direct paper reproducibility/journal submission -- I have a version of the training loop etc in development that uses the physicsnemo `SongUNet` and other utils, just got stuck in the process of verifying accuracy but we can hopefully pick that up again after this is merged.

Also thanks @nbren12 for the comments -- should be addressed now.

Issue #817, opened 2025-03-19
🐛[BUG]: Type checking is not catching Entrypoint from importlib_metadata


Tristanjmeyers:
### Version

1.0.0

### On which installation method(s) does this occur?

Pip

### Describe the issue

Hi,

I'm running the example for corrdiff mini, following the instructions in `examples/generative/corrdiff/`. The first step, training the regression model, works well. However, when I tried to train the generative model, I ran into the following error:

`TypeError: 'EntryPoint' object is not callable` at 
```
physicsnemo/models/module.py", line 203, in instantiate:
return _cls(**arg_dict["__args__"]
```

This instantiate method recognises that the `cls` is in the `model_registry`. This is the because it is the following: 
```
EntryPoint(name='UNet', value='physicsnemo.models.diffusion:UNet', group='physicsnemo.models')
```

So, it calls to the method `factory` in the `model_registry` module. In this method, the `model` is checked with an `isisntance` on line 122:
```
        if model is not None:
            if isinstance(model, EntryPoint):
                model = model.load()
```
Where `EntryPoint` is imported from `importlib` on line 18 of `model_registry`:
```
from importlib.metadata import EntryPoint, entry_points
```

However, the `model` is actually of type `importlib_metadata.Entrypoint`. Hence it passes this `if` statement without a fail, and just returns the original object. Hence the `TypeError`.

This can be easily fixed, can I create a pull request for this, but basically in `model_registry` do another import:
```
from importlib_metadata import EntryPoint as EntryPoint_
```
And then add another check, and probably best to have an `else` clause here too:

```
        if model is not None:
            if isinstance(model, EntryPoint):
                print(f'prior to loading, the model is: {model}')
                model = model.load()
                print(f'after loading, the model is: {model}')
            elif isinstance(model, EntryPoint_):
                print(f'prior to loading, the model is: {model}')
                model = model.load()
                print(f'after loading, the model is: {model}')
            else raise KeyError(f'Model name {name} is not a valid entrypoint')
```

### Minimum reproducible example

```shell
Try the corrdiff example in the README.md, I'm on python 3.11.
```

### Relevant log output

```shell
[rank0]:   File "<UNAME>/bin/corrdiff/physicsnemo-corrdiff/train.py", line 242, in main
[rank0]:     regression_net = Module.from_checkpoint(regression_checkpoint_path)
[...]

PR #818, opened 2025-03-20
Bug entry point

loliverhennigh:
https://github.com/NVIDIA/physicsnemo/issues/817


physicsnemo/registry/model_registry.py
```diff
@@ -120,8 +121,10 @@ def factory(self, name: str) -> "physicsnemo.Module":
 
         model = self._model_registry.get(name)
         if model is not None:
-            if isinstance(model, EntryPoint):
+            if isinstance(model, EntryPoint) or isinstance(model, EntryPointOld):
                 model = model.load()
+            else:
+                raise ValueError(f"Model {name} is not a valid entry point.")
```
NickGeneva, line 127: TypeError

physicsnemo/registry/model_registry.py
```diff
@@ -16,6 +16,7 @@
 
 import warnings
 from importlib.metadata import EntryPoint, entry_points
+from importlib_metadata import EntryPoint as EntryPointOld
```
ktangsali, line 19: Can we avoid this import from `importlib_metadata`?  This has been deprecated in the newer versions of Python

NickGeneva, line 19: Can you add a small comment explaining what this is, why it here?
We will forget in a month.

importlib_metadata is a back port of new APIs for previous python vesions, also users dont necessarily need this installed.  If it happens to be installed, we should check it.

Or does this come with base python (I dont think so but maybe)

NickGeneva: Best guess is somehow older versions of importlib are getting being used and some how this impacts the return type on `entry_points`

NickGeneva: What about this solution?

https://github.com/sphinx-doc/sphinx/issues/10157#issuecomment-1028068202

NickGeneva (approved):

PR #831, opened 2025-03-31
fixing member reference

azrael417:
# PhysicsNeMo Pull Request

## Description
This fixes a wire error with new torch, because the wrong member variable is referenced in init process group, so the except branch is always triggered.

## Checklist

- [] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.

### DIFF
--- a/physicsnemo/distributed/manager.py
+++ b/physicsnemo/distributed/manager.py
@@ -551,6 +551,8 @@ class DistributedManager(object):
             f"cuda:{manager.local_rank}" if torch.cuda.is_available() else "cpu"
         )
 
+        print(manager._device)
+
         if manager._distributed:
             # Setup distributed process group
             try:
@@ -558,7 +560,7 @@ class DistributedManager(object):
                     backend,
                     rank=manager.rank,
                     world_size=manager.world_size,
-                    device_id=manager.device,
+                    device_id=manager._device,
                 )
             except TypeError:
                 # device_id only introduced in PyTorch 2.3


coreyjadams (commented):
Casting local_rank to int before checking it's not-`None` seems like a good idea.  What error were you encountering where it needed this?

Changing the access of rank/local_rank/world_size/device from the `property` access to private attributes  seems counterintuitive.  Can you elaborate on the reason for that change?

physicsnemo/distributed/manager.py
```diff
@@ -311,7 +311,7 @@ def initialize_env():
         rank = int(os.environ.get("RANK"))
         world_size = int(os.environ.get("WORLD_SIZE"))
         if "LOCAL_RANK" in os.environ:
-            local_rank = os.environ.get("LOCAL_RANK")
+            local_rank = int(os.environ.get("LOCAL_RANK"))
```
akshaysubr, line 314: Isn't this int conversion handled in the next `if` block? And won't this break when `LOCAL_RANK` is `None`?

physicsnemo/distributed/manager.py
```diff
@@ -548,30 +548,30 @@ def setup(
                 manager._local_rank = local_rank
 
         manager._device = torch.device(
-            f"cuda:{manager.local_rank}" if torch.cuda.is_available() else "cpu"
+            f"cuda:{manager._local_rank}" if torch.cuda.is_available() else "cpu"
```
akshaysubr, line 551: Any specific reason to switch to using private attributes here instead of the public property versions of these as was used before?

azrael417: If the external ones alias those internal ones, that is OK, but isn't it cleaner to use the internal ones for internal tasks?

physicsnemo/distributed/manager.py
azrael417: can that happen? if LOCAL_RANK is set, can it still be None: like with 

export LOCAL_RANK=

?

PR #834, opened 2025-03-31
adding t-edm

mnabian:
# PhysicsNeMo Pull Request

## Checklist

### DIFF
--- a/examples/generative/corrdiff/train.py
+++ b/examples/generative/corrdiff/train.py
@@ -547,7 +547,11 @@ def main(cfg: DictConfig) -> None:
                                         loss = loss_fn(**loss_fn_kwargs)
 
                                 loss = loss.sum() / batch_size_per_gpu
-                                loss_accum += loss / num_accumulation_rounds
+                                loss_accum += (
+                                    loss
+                                    / num_accumulation_rounds
+                                    / len(patch_nums_iter)
+                                )
                                 with nvtx.annotate(f"loss backward", color="yellow"):
                                     loss.backward()
 
@@ -684,7 +688,7 @@ def main(cfg: DictConfig) -> None:
                                     for patch_num_per_iter in patch_nums_iter:
                                         if patching is not None:
                                             patching.set_patch_num(patch_num_per_iter)
-                                            loss_fn_kwargs.update(
+                                            loss_valid_kwargs.update(
                                                 {"patching": patching}
                                             )
                                         with torch.autocast(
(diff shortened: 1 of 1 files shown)


physicsnemo/models/diffusion/preconditioning.py
```diff
@@ -1174,3 +1174,183 @@ def voriticity_residual(self, w, re=1000.0, dt=1 / 32):
...
+        self,
+        img_resolution,
+        img_channels,
+        nu=10,
+        label_dim=0,
+        use_fp16=False,
+        sigma_min=0,
+        sigma_max=float("inf"),
+        sigma_data=0.5,
+        model_type="DhariwalUNet",
```
CharlelieLrt, line 1251: Are we sure that we want `DhariwalUNet` to be the default architecture? I thought `SongUNet` would make more sense?
In any case, would it be possible to document the possible values for the `model_type` argument?
If for now we want to restrict it to `DhariwalUNet ` we should remove this option.

physicsnemo/models/diffusion/preconditioning.py
```diff
@@ -1174,3 +1174,183 @@ def voriticity_residual(self, w, re=1000.0, dt=1 / 32):
...
+        model_class = getattr(network_module, model_type)
+        self.model = model_class(
+            img_resolution=img_resolution,
+            in_channels=img_in_channels,
+            out_channels=img_out_channels,
+            label_dim=label_dim,
+            **model_kwargs,
+        )
+
+    def forward(
```
CharlelieLrt, line 1289: Note that this `foward` will not work with the current `deterministic_sampler`, as the `condition` kwarg will be ignored. See this snippet from `deterministic_sampler.py`:

```python
        if isinstance(net, EDMPrecond):
            # Conditioning info is passed as keyword arg
            denoised = net(
                x_hat / s(t_hat),
                sigma(t_hat),
                condition=x_lr,
                class_labels=class_labels,
            ).to(torch.float64)
        else:
            denoised = net(x_hat / s(t_hat), x_lr, sigma(t_hat), class_labels).to(
                torch.float64
            )
```

_Note:_ When working on other CorrDiff things I realized the `deterministic_sampler` is mostly broken. I am not even sure there is a single use case where it can work properly. I did not address these bugs because my understanding is that `deterministic_sampler` is almost not used at all.

CharlelieLrt (changes requested):
I left a few comments and questions mostly about:
1. More thorough docstrings, including tensor shapes and expected signatures of callable arguments. There are now many samplers and preconditioners, with different signatures and so on, and it starts to get really challenging to use them properly.
2. Test coverage for new features

I have two major comments:
1. This PR introduces `tEDMPrecond` and `tEDMLoss`, but never actually uses them in the `train.py`. I want to make sure that this is not an omission and that is really the purpose?
2. I understand that for generation, the `tEDMPrecond` is only expected to work in combination with the `deterministic_sampler`. However, I believe the `deterministic_sampler` suffers from multiple bugs (mostly arguments mismatches), so I doubt that the combination `tEDMPrecond` + `deterministic_sampler` is currently usable as it is.

PR #836, opened 2025-04-02
Add Stochastic flow matching for corrdiff

daviddpruitt:
# PhysicsNeMo Pull Request

## Description
This change adds stochastic flowmatching to the existing CorrDiff implementation

## Checklist

## Dependencies
No additional dependencies required


examples/generative/corrdiff_plus_plus/README.md
```diff
@@ -0,0 +1,149 @@
...
+To overcome these limitations, **CorDiff++** was introduced at NVIDIA. It relies on adaptive floe matching (AFM) that improves upon CorrDiff in generalization, calibration, and efficiency through several key innovations:
+
+- A joint encoder–generator architecture trained via flow matching  
+- A latent base distribution that reconstructs the large-scale, deterministic signal  
+- An adaptive noise scaling mechanism informed by the encoder’s RMSE, used to inject calibrated uncertainty  
+- A final flow matching step to refine latent samples and synthesize fine-scale physical details
+
+AFM outperforms previous methods across both real-world (e.g., 25 → 2 km super-resolution in Taiwan) and synthetic (Kolmogorov flow) benchmarks—especially for highly stochastic output channels.
+
+📄 For details, see the [AFM paper (arXiv:2410.19814)](https://arxiv.org/abs/2410.19814).
```
pzharrington, line 22: *SFM paper :)

examples/generative/corrdiff/train.py
pzharrington, line 1: These changes probably got put in the wrong place, right @daviddpruitt? I assume everything should go into `corrdiff_plus_plus` and the existing CorrDiff example should remain unchanged

daviddpruitt: Oops, I put them in the wrong file, fixed.

physicsnemo/metrics/diffusion/sfm_loss.py
```diff
@@ -0,0 +1,251 @@
...
+
+        rnd_uniform = torch.rand([img_clean.shape[0], 1, 1, 1], device=img_clean.device)
+
+        sampled_sigma = rnd_uniform * sigma_max
+        weight = (sampled_sigma**2 + self.sigma_data**2) / (
+            sampled_sigma * self.sigma_data
+        ) ** 2
+
+        # augment for conditional generaiton
+        x_tot = torch.cat((img_clean, img_lr), dim=1)
```
pzharrington, line 134: This concat has no real point, right? `x_tot` is immediately split out back into `x_1` and `x_low` and then never accessed again. Also, what does the note about augmenting and conditional generation mean?

physicsnemo/metrics/diffusion/sfm_loss.py
```diff
@@ -0,0 +1,251 @@
...
+            x=x_t_noisy,
+            sigma=sampled_sigma,
+            condition=x_low,
+        )
+        # time_weight = lambda t: 1/(1 - torch.clamp(t, 0.9))
+        # time_weight = lambda t: 1
+        def time_weight(t):
+            return 1
+
+        sfm_loss = weight * ((time_weight(time) * (D_x_t - x_1)) ** 2)
```
pzharrington, line 175: Let's drop the `time_weight` if it's just hard-coded to 1

physicsnemo/models/diffusion/encoders.py
pzharrington, line 1: I'm confused about this module. Can you explain the motivation for adding a separate `encoders` file and putting this in it, rather than just using the `Conv2d` in `layers.py`?

physicsnemo/models/diffusion/sfm_preconditioning.py
pzharrington, line 1: My high-level comment on this is that functionally `SFMPrecondSR` has the same behavior as `EDMPrecondSR` (aside from the extra `get_sigma_max` and `update_sigma_max` methods). At least as far as I can tell, correct me if I'm wrong.

Can you possibly refactor so that `EDMPrecondSR` can be used instead? The reason is we are currently suffering from a growing number of near-identical copies of EDM utilities (from CorrDiff variants and other projects), which is becoming hard to maintain and increasingly confusing for users/developers. It seems like here we can simply add the extra methods (which are simple) to the `EDMPrecondSR` along with some config/arg checking in `__init__` to ensure proper usage, and we avoid having an extra module with extra CI tests, etc. @CharlelieLrt for viz

physicsnemo/models/diffusion/sfm_preconditioning.py
```diff
@@ -0,0 +1,239 @@
...
+        torch.Tensor
+            The tensor representation of the provided sigma value(s).
+        """
+        return torch.as_tensor(sigma)
+
+
+class SFMPrecondEmpty(Module):
+    """
+    A preconditioner that does nothing
+
```
pzharrington, line 229: For `SFMPrecondEmpty` I suggest just adding a simpler one-liner `nn.Module` or something in the train script and avoid putting this in the core `physicsnemo` package code. Since the need for it is really example-specific and only supports the edge case of training an encoder-only model in that example (I assume for a point of comparison against the real SFM model which has encoder/denoiser trained jointly)

physicsnemo/utils/generative/sfm_samplers.py
```diff
@@ -0,0 +1,231 @@
...
+from physicsnemo.models.diffusion import SongUNetPosEmbd
+
+
+def sigma(t):
+    return t
+
+
+def sigma_inv(sigma):
+    return sigma
+
```
pzharrington, line 34: Let's remove these hard-coded identity functions for simplicity

physicsnemo/utils/generative/sfm_samplers.py
```diff
@@ -0,0 +1,231 @@
...
+
+@nvtx.annotate(message="SFM_encoder_sampler", color="red")
+def SFM_encoder_sampler(
+    networks: Dict[str, torch.nn.Module],
```
pzharrington, line 38: Similar comment as for `SFMPrecondEmpty`. I think this can either be stashed in a `utils` file local to the example and not in the `physicsnemo` core, or alternately support an option in the `SF_sampler` to allow running "encoder-only" sampling.

physicsnemo/utils/generative/sfm_samplers.py
```diff
@@ -0,0 +1,231 @@
...
+        u_t = (x_t_denoised - x_t_hat) / (torch.clamp(t_hat, min=cfg.t_min))
+
+        dt = t_hat - t_next  # needs to be reversed
+        x_t = x_t_hat + u_t * dt
+
+    return x_t
+
+
+@nvtx.annotate(message="SFM_Euler_sampler_Adaptive_Sigma", color="red")
+def SFM_Euler_sampler_Adaptive_Sigma(
```
pzharrington, line 155: It seems like this can be absorbed into the main `SFM_Euler_sampler` and behavior can be controlled with function args, no? Again, similar motivation, reducing duplication of nearly the same functionality.

examples/generative/corrdiff_plus_plus/README.md
```diff
@@ -0,0 +1,149 @@
...
+This will use the base configs specified in the `conf/config_generate.yaml` file.
+
+Next, to score the generated samples, run:
+
+```bash
+python score_samples.py path=<PATH_TO_NC_FILE> output=<OUTPUT_FILE>
+```
+
+Some legacy plotting scripts are also available in the `inference` directory.
+You can also bring your checkpoints to [earth2studio]<https://github.com/NVIDIA/earth2studio>
```
pzharrington, line 113: What is meant by legacy plotting scripts here? If they don't work with the output generated by the scoring or generation scripts, I suggest just removing them. Also, the earth2studio link here doesn't render properly, FYI.

examples/generative/corrdiff_plus_plus/README.md
```diff
@@ -0,0 +1,149 @@
...
+**CorrDiff** was proposed to address these issues but suffers from poor generalization—its regression-based residuals can overfit training data, leading to degraded performance on out-of-distribution inputs.
+
+To overcome these limitations, **CorDiff++** was introduced at NVIDIA. It relies on adaptive floe matching (AFM) that improves upon CorrDiff in generalization, calibration, and efficiency through several key innovations:
+
+- A joint encoder–generator architecture trained via flow matching  
+- A latent base distribution that reconstructs the large-scale, deterministic signal  
+- An adaptive noise scaling mechanism informed by the encoder’s RMSE, used to inject calibrated uncertainty  
+- A final flow matching step to refine latent samples and synthesize fine-scale physical details
+
+AFM outperforms previous methods across both real-world (e.g., 25 → 2 km super-resolution in Taiwan) and synthetic (Kolmogorov flow) benchmarks—especially for highly stochastic output channels.
```
pzharrington, line 20: *SFM

examples/generative/corrdiff_plus_plus/README.md
```diff
@@ -0,0 +1,149 @@
...
+```bash
+python train.py --config-name=config_training_sfm.yaml ++dataset.data_path=</path/to/dataset>/hrrr_mini_train.nc ++dataset.stats_path=</path/to/dataset>/stats.json
+```
+
+The training will require a few hours on a single A100 GPU. If training is interrupted, it will automatically continue from the latest checkpoint when restarted. Multi-GPU and multi-node training are supported and will launch automatically when the training is run in a `torchrun` or MPI environment.
+
+The results, including logs and checkpoints, are saved by default to `outputs/mini_generation/`. You can direct the checkpoints to be saved elsewhere by setting: `++training.io.checkpoint_dir=</path/to/checkpoints>`.
+
+> **_Out of memory?_** CorrDiff-Mini trains by default with a batch size of 256 (set by `training.hp.total_batch_size`). If you're using a single GPU, especially one with a smaller amout of memory, you might see out-of-memory error. If that happens, set a smaller batch size per GPU, e.g.: `++training.hp.batch_size_per_gpu=16`. CorrDiff training will then automatically use gradient accumulation to train with an effective batch size of `training.hp.total_batch_size`.
+
```
pzharrington, line 52: Similar comment for the CorrDiff-Mini model/dataset mentioned here -- is there a plan to implement/support that or is this example just intended to cover Taiwan/CWB for now?

examples/generative/corrdiff_plus_plus/helpers/sfm_utils.py
```diff
@@ -0,0 +1,76 @@
...
+from physicsnemo.models.diffusion import SongUNetPosEmbd
+from physicsnemo.models.diffusion import Conv2dSerializable
+import torch
+from omegaconf import DictConfig
+
+
+def get_encoder(cfg: DictConfig):
+    """
+    Helper that sets instantiates a
+
```
pzharrington, line 26: Docstring typo

pzharrington:
I think all my main comments are in. Aside from minor things, I think the main thing to focus on is how much the additional SFM stuff can be absorbed into existing EDM functionality in PhysicsNemo. This will greatly help our efforts going forward in improving the usability/readability/extensibility of these modules.

PR #838, opened 2025-04-04
Enable DoMINO parallelization via ShardTensor

coreyjadams:
## Description

This PR is enables domain parallelism for DoMINO.

Included in this PR are a number of other pending parallelization operations, many of which are required for DoMINO but not all.  These include:
- Sequence parallel attention via ring attention
- normalization layers (group Norm, layer norm via DTensor)
- ConvTranspose
- Upsampling via torch interpolate
- MaxPooling and AvgPooling

The halo passing (and new ring algorithm) are also reorganized.  This should help make the halo algorithm more readable and maintainable.

There is always more to do for performance, but nearly all cases of blocking operations when creating a shard tensor are removed.

To include testing of all the distributed algorithms would add several hundred distributed tests.  With the fork-every-test set up, this is just not feasible in the CI.  The tests currently exist in another repository for validating numerical correctness of external operations.  PhysicsNeMo-implemented operations (BallQuery) have tests implemented here.

(After the release of 25.03 and renaming, I had to do some manual merging on many files.  so there are changed files unrelated to this PR)

Finally, this commit adds two tutorials that enhance the user facing material for ShardTensor.

## Checklist

## Dependencies

External testing repo, to be opened or perhaps merged.

### DIFF
--- a/physicsnemo/distributed/shard_utils/ring.py
+++ b/physicsnemo/distributed/shard_utils/ring.py
@@ -107,6 +107,9 @@ def perform_ring_iteration(
         # Swap
         id_for_send, id_for_recv = id_for_recv, id_for_send
 
+    if not tensor.is_contiguous():
+        tensor = tensor.contiguous()
+
     if recv_shape is None:
         tensor_recv = torch.empty_like(tensor)
     else:
@@ -131,7 +134,7 @@ def perform_ring_iteration(
         p2p_op_list.append(
             dist.P2POp(
                 op=dist.isend,
-                tensor=tensor.contiguous(),
+                tensor=tensor,
                 peer=id_for_send,
                 group=local_group,
             )


docs/tutorials/domain_parallelism/fsdp_and_shard_tensor.rst
coreyjadams, line 1: I had to change the title here to get the rst linking to work ...

docs/tutorials/domain_parallelism_entry_point.rst
coreyjadams, line 1: This file is meant as the entry point for domain parallelism, so it doesn't clog up the side bar in the docs.

examples/cfd/external_aerodynamics/aero_graph_net/inference_analysis/ahmed_body.ipynb
coreyjadams, line 1: This branch started before the name change and while it was synced many times, I don't know why this change is in this PR still....

examples/cfd/external_aerodynamics/domino/src/train.py
coreyjadams, line 1: The main change here from the domino (now main) branch is that I consolidated the loss function into just one function call.  No logic is changed.  But this makes just one call in the sharded version too.

physicsnemo/distributed/_shard_redistribute.py
coreyjadams, line 1: Most of the changes in this file originate from a choice to force everything related to the sharded tensor chunks to be called `sharding_shapes` and *not* `sharding_sizes`.  I was mixing and matching and it wasn't good.

physicsnemo/distributed/_shard_redistribute.py
```diff
@@ -132,10 +140,16 @@ def _select_slice_from_replicate(
...
     else:
-        chunks = torch.tensor_split(local_tensor, sizes[:-1], dim=tensor_dim)
-
+        # Convert sizes to cumulative sum using basic Python
+        chunk_starts = []
+        running_sum = 0
+        for size in sizes[:-1]:
+            running_sum += size
+            chunk_starts.append(running_sum)
+        chunks = torch.tensor_split(local_tensor, chunk_starts, dim=tensor_dim)
```
coreyjadams, line 152: This aligns to chunk syntax from torch if no hint is given

physicsnemo/distributed/shard_tensor.py
coreyjadams, line 1: Major changes to ShardTensor:
- All "sharding_sizes"  become "sharding_shapes"
- Overrides are expected to go through __torch_function__ or __torch_dispatch__
- Additional logic to enable sharding gradients properly in a backwards pass, when converting to/from a local tensor.
- override the backward function to ensure no partial placements, and use differentiable `to_local` call.

physicsnemo/distributed/shard_utils/__init__.py
coreyjadams, line 1: I disabled natten here, since it requires monkey patching.  It will be shown in an example how to use it.

Everything else is enabled by default - but note that it's required to go through torch_function for these calls, so operations that don't use ShardTensor are unaffected.

Additionally, there are tests include to make sure the dispatch mechanism works properly.

physicsnemo/distributed/shard_utils/attention_patches.py
coreyjadams, line 1: Attention mechanism based on Ring Attention.  Needs to be optimized for fully async fwd/backward pass but this is numerically correct and performance at modest to large scales.

physicsnemo/distributed/shard_utils/index_ops.py
coreyjadams, line 1: This file intercepts index_select and select operations.  They aren't available upstream, so I intercept at the dispatch rather than function level.

physicsnemo/distributed/shard_utils/point_cloud_ops.py
coreyjadams, line 1: Note that the backwards pass of BallQuery is not implemented when the initial point cloud was sharded - it is not needed for DoMino, actually, so it's P1.

Will be added when I can do it, I haven't thought hard about it yet but it should distributing the output gradients based on the forward mapping to the proper point cloud shard.

There are passing tests for all other variations of sharding for ball query.

physicsnemo/models/domino/model.py
```diff
@@ -29,17 +29,34 @@
...
 
 from physicsnemo.models.layers.ball_query import BallQueryLayer
+from physicsnemo.utils.profiling import profile
 
 
 def fourier_encode(coords, num_freqs):
     """Function to caluculate fourier features"""
     # Create a range of frequencies
-    freqs = torch.exp(torch.linspace(0, math.pi, num_freqs))
+    freqs = torch.exp(torch.linspace(0, math.pi, num_freqs, device=coords.device))
```
coreyjadams, line 38: Avoid CPU device creation.

physicsnemo/models/domino/model.py
```diff
@@ -29,17 +29,34 @@
...
+    F = freqs.shape[0]
+
+    # freqs = torch.exp(torch.linspace(0, math.pi, num_freqs, device=coords.device))  # [F]
+    freqs = freqs[None, None, :, None]  # reshape to [*, F, 1] for broadcasting
+
+    coords = coords.unsqueeze(-2)  # [*, 1, D]
+    scaled = (coords * freqs).reshape(*coords.shape[:-2], D * F)  # [*, D, F]
+    features = torch.cat([torch.sin(scaled), torch.cos(scaled)], dim=-1)  # [*, D, 2F]
+
+    return features.reshape(*coords.shape[:-2], D * 2 * F)  # [*, D * 2F]
```
coreyjadams, line 59: This is a better way to do this with shard tensor, sicne it's less overhead.

physicsnemo/models/domino/model.py
```diff
@@ -1130,42 +1141,116 @@ def calculate_solution_with_neighbors(
...
-                else:
-                    volume_m_c = surface_mesh_neighbors[:, :, p - 1] + 1e-6
-                    noise = surface_mesh_centers - volume_m_c
-                    dist = torch.sqrt(
-                        noise[:, :, 0:1] ** 2.0
-                        + noise[:, :, 1:2] ** 2.0
-                        + noise[:, :, 2:3] ** 2.0
+        if (
+            self.solution_calculation_mode == "one-loop"
+            or self.solution_calculation_mode == "compare"
```
coreyjadams, line 1146: I've left a "compare" handle in the model in case anyone wants to validate these are equivalent.  But it should be removed before approval, please request that change :)

physicsnemo/models/domino/model.py
```diff
@@ -1130,42 +1141,116 @@ def calculate_solution_with_neighbors(
...
+            if self.solution_calculation_mode != "compare":
+                return output_all
 
+        if self.solution_calculation_mode == "compare":
+            print(
+                f"NEIGHBORS: 2-loop vs. 1-loop Agreement? {torch.allclose(one_loop_output_all, output_all)}"
+            )
         return output_all
 
+    # @to_local_tensors
```
coreyjadams, line 1253: This needs to be deleted.

physicsnemo/models/layers/ball_query.py
```diff
@@ -26,244 +237,91 @@ class BallQuery(torch.autograd.Function):
...
-                ctx.outputs,
-            ],
-            dim=ctx.dim,
+        # Apply the primitive.  Note the batch index is removed.
+        mapping, num_neighbors, outputs = _ball_query_forward_primative_(
+            points1[0],
+            points2[0],
+            k,
+            radius,
+            hash_grid,
```
coreyjadams, line 275: Handling the ball query fwd/bwd operations as "primitives" let's me reuse the implementation in the sharded version without also reusing the context saving.

physicsnemo/models/layers/ball_query.py
```diff
@@ -26,244 +237,91 @@ class BallQuery(torch.autograd.Function):
...
+    radius: float,
+    hash_grid: wp.HashGrid,
+):
+    """
+    Wrapper for BallQuery.apply to support a functional interface.
+    """
+    if has_torch_function((points1, points2)):
+        return handle_torch_function(
+            ball_query_layer, (points1, points2), points1, points2, k, radius, hash_grid
+        )
```
coreyjadams, line 321: This is needed to enable torch_function to work on this operation, and therefore overload in ShardTensor.

docs/test_scripts/domain_parallelism/new_layers/reshape_subtract.py
```diff
@@ -0,0 +1,41 @@
...
+dm = DistributedManager()
+
+# We'll make these 3D tensors to represent 3D points
+a = torch.randn(N1, 3, device=dm.device)
+b = torch.randn(N2, 3, device=dm.device)
+
+mesh = dm.initialize_mesh([-1,], ["domain"])
+placements = (Shard(0),)
+a_sharded = scatter_tensor(a, 0, mesh, placements)
+b_sharded = scatter_tensor(b, 0, mesh, placements)
```
peterdsharpe, line 26: In general, when it comes to tutorial scripts that are introducing new functions, I would lean towards keyword args when possible - it makes it far easier for a new user to understand what's happening. Positional arguments save time for the writer but can cost time for the reader (if it's not a function they're already well-acquainted with, like `torch.randn`).

To be specific: here, it would be useful to explain to the reader what the placements argument does, and how it is formatted (e.g., why is it a length-1 tuple, what does the `0` in `Shard(0)` do, etc.) - this is not necessarily obvious here

coreyjadams: Resolved, thanks!

physicsnemo/distributed/shard_tensor.py
```diff
@@ -305,132 +339,123 @@ def __repr__(self) -> str:
...
+        with annotate(f"__torch_dispatch___{func.__name__}"):
+            # Leverage DTensor Dispatch as much as possible, but, enable
+            # the ability to operate on this output in the future:
+            if func in cls._dispatch_registry:
+                res = cls._dispatch_registry[func](*args, **kwargs)
+                return res
+
+            # We assume that if we reach this point, the operator has not been
+            # intercepted by a wrapper or in the registry.  So the DTensor
+            # default behavior is likely to be correct.
```
peterdsharpe, line 403: Would this be worth a `warnings.warn()`? How often do you expect to have ShardTensors use functions that are not in the registry, and where they cannot fall back on the DTensor implementation and still sort-of-work? As they say: being loudly wrong is ok, being silently wrong is not. This particular case is a judgment call that you're in a better position to make than I am, but just food for thought

physicsnemo/distributed/shard_utils/attention_patches.py
```diff
@@ -0,0 +1,695 @@
...
+    RingPassingConfig,
+    perform_ring_iteration,
+)
+
+aten = torch.ops.aten
+
+
+def add_log_sumexp(
+    log_a: Optional[torch.Tensor], log_b: Optional[torch.Tensor]
+) -> torch.Tensor:
```
peterdsharpe, line 44: Notably, this function currently returns None if both log_a and log_b are None, which the type hints indicate is permissible. I think we should error if both are None, rather than allow a later error when a variable expecting a tensor becomes None.

physicsnemo/models/layers/ball_query.py
```diff
@@ -26,244 +237,91 @@ class BallQuery(torch.autograd.Function):
...
-                ctx.points2,
-                ctx.mapping,
-                ctx.num_neighbors,
-            ],
-            outputs=[
-                ctx.outputs,
-            ],
-            dim=ctx.dim,
+        # Apply the primitive.  Note the batch index is removed.
+        mapping, num_neighbors, outputs = _ball_query_forward_primative_(
```
peterdsharpe: Super nit-pick, sorry
```suggestion
        mapping, num_neighbors, outputs = _ball_query_forward_primitive_(
```

peterdsharpe (approved):
Excellent work - some really clever stuff in here. Overall looks good. Going to mark it as "Approve" just so that we don't bog down the overall process of getting to 25.06 code freeze, but before merging, please take a look through the comments and address those that you feel are worthwhile. Nice job!

examples/cfd/external_aerodynamics/domino/README.md
```diff
@@ -64,6 +64,50 @@ To train and test the DoMINO model on AWS dataset, follow these steps:
...
+`domain_size: 1` is not advised since that is the standard training regime,
+but with extra overhead.  `shard_grid` and `shard_points` will enable domain
+parallelism over the latent space and input/output points, respectively.
+
+Please see `src/train_sharded.py` for more details regarding the changes
+from the standard training script required for domain parallel DoMINO training.
+
+As one last note regarding domain-parallel training: in the phase of the DoMINO
+where the output solutions are calculated, the model can used two different
+techniques (numerically identical) to calculate the output.  Due to the
```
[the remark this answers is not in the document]
coreyjadams: There are not really algorithmic differences at a math level, it's all in the implementation.  Imagine

```
a = torch.arange(5,20)
b = torch.zeros(5)
for i in range(5):
  b += a[i].sum(1)
```

versus
```
a = torch.arange(5,20)
b = torch.sum(a, 1).sum()
```

(neither of these are clever examples, sorry)

In the first case, you have a loop with a GPU kernel each time.  But it's low memory overhead, you never allocate anything more than needed.  In the second case, you have no loop but you allocate an intermediate tensor - more memory.  In DoMINO, this is a high level difference.  I did this because the overhead on the loop is a little higher with shard tensor - though I'll reevaluate that it's as big an impact as I recall, I later found CPU copy that blocks execution.

FYI, the domino test can be adapted to do both and validate the results.  I'll add that.

physicsnemo/models/domino/model.py
coreyjadams: A little, for single device.  There's already sync points in building the hashmap for ball query, so it's not a crazy hit here.

In domain parallel mode it is brutal for performance!

docs/test_scripts/domain_parallelism/sharded_conv_example.py
```diff
@@ -0,0 +1,119 @@
...
+
+########################################
+# Single GPU - forward + loss + backward
+########################################
+single_gpu_output = conv(original_tensor)
+
+# This isn't really a loss, just a pretend one that's scalar!
+single_gpu_output.mean().backward()
+# Copy the gradients produced here - so we don't overwrite them later.
+original_tensor_grad = original_tensor.grad.data.clone()
```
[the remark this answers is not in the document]
coreyjadams: I decided to let it compute on all devices - but lower in that script when I compare, I only check on rank 0 and have a short comment about it.

physicsnemo/models/layers/ball_query.py
coreyjadams: Yes, this one was pointed out to me to fix elsewhere too  :)

docs/test_scripts/domain_parallelism/new_layers/vector_dot_sharded.py
```diff
@@ -0,0 +1,109 @@
...
+    # If you wanted to write a generic sharding handler for this type of operation, 
+    # you could do:
+    # local_dot_product = func(local_x, local_y)
+    # But it's over kill here...
+    
+    # SUM_Reduce the local result across all ranks:
+    dist.all_reduce(local_dot_product, op=dist.ReduceOp.SUM, group=mesh.get_group())
+
+    # We do want to return the result as a ShardTensor, for consistency.
+    # We can easily create one on the same mesh as a "Replicated" tensor:
```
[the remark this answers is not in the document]
coreyjadams: Comment about placements added.

physicsnemo/distributed/custom_ops/_reductions.py
```diff
@@ -14,7 +14,18 @@
...
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    TypeVar,
+    Union,
+)
```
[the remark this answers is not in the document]
coreyjadams: Thanks for sharing this!

physicsnemo/distributed/custom_ops/_reductions.py
```diff
@@ -27,102 +38,543 @@
...
+    "avg": mean_wrapper,
+}
+
+
+def register_reduction_functions() -> None:
+    """
+    Register reduction functions with the ShardTensor class.
+
+    This function gatekeeps the registration of these ops
+    to ensure they don't fire unless wanted.
```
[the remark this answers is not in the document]
coreyjadams: The previous implementation was using more invasive monkey-patching instead of dispatch.  So in that case, yes, we did not want them registered by default.

Now, it's generally safe though and you're right, this function can be removed.

physicsnemo/distributed/shard_utils/halo.py
```diff
@@ -14,179 +14,447 @@
...
+    # gets discarded on the edge ranks.  But, it would have to wait
+    # for the other ranks to make this selection anyways.
 
     # Select halo regions to exchange
-    left_indices = torch.arange(0, halo_t).to(local_tensor.device)
-    max_index = local_tensor.shape[tensor_dim]
+    left_indices = torch.arange(0, halo_config.halo_size, device=local_tensor.device)
+    max_index = local_tensor.shape[halo_config.tensor_dim]
     right_indices = max_index - 1 - left_indices
     right_indices = torch.flip(right_indices, (0,))
```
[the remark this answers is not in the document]
coreyjadams: Nice catch, I'll update.

physicsnemo/models/domino/model.py
```diff
@@ -442,12 +449,16 @@ def __init__(self, input_features: int, model_parameters=None):
         self.fc1 = nn.Linear(input_features_calculated, base_layer)
         self.fc2 = nn.Linear(base_layer, int(base_layer))
         self.fc3 = nn.Linear(int(base_layer), int(base_layer))
-        self.bn1 = nn.BatchNorm1d(base_layer)
-        self.bn2 = nn.BatchNorm1d(int(base_layer))
-        self.bn3 = nn.BatchNorm1d(int(base_layer))
+        # self.bn1 = nn.BatchNorm1d(base_layer)
+        # self.bn2 = nn.BatchNorm1d(int(base_layer))
+        # self.bn3 = nn.BatchNorm1d(int(base_layer))
```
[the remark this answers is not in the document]
coreyjadams: Cut!

physicsnemo/distributed/shard_utils/pooling_patches.py
```diff
@@ -0,0 +1,397 @@
...
+    # Handle return_indices case
+    return_indices = pool_kwargs.get("return_indices", False)
+    if return_indices:
+        local_pooled_output, indices = local_pooled_output
+
+    # Everything below here is computing output meta data
+
+    # Reject cases where stride != kernel_size
+    if pool_kwargs.get("stride") != pool_kwargs.get("kernel_size"):
+        raise MissingShardPatch("Stride must equal kernel_size for pooling operations")
```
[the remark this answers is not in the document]
coreyjadams: You're right, this is my imposed limitation based on the implementation.  I'll update the error message thanks.

physicsnemo/distributed/shard_utils/unpooling_patches.py
```diff
@@ -0,0 +1,311 @@
...
+    *args: Any,
+    **kwargs: Any,
+) -> Tuple[Union[torch.Tensor, ShardTensor], Dict[str, Any]]:
+    """Repackages interpolation arguments into standard format.
+
+    Args:
+        input: Input tensor to interpolate
+        size: Output spatial size
+        scale_factor: Multiplier for spatial size
+        mode: Algorithm used for upsampling
```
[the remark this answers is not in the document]
coreyjadams: Instead of copying, I've linked to upstream docs.

physicsnemo/distributed/shard_utils/unpooling_patches.py
```diff
@@ -0,0 +1,311 @@
...
+    unhalo_padding,
+)
+
+
+def repackage_interpolate_args(
+    input: Union[torch.Tensor, ShardTensor],
+    size: Optional[Union[int, Tuple[int, ...]]] = None,
+    scale_factor: Optional[Union[float, Tuple[float, ...]]] = None,
+    mode: str = "nearest",
+    align_corners: Optional[bool] = None,
```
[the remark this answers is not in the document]
coreyjadams: Same - linking to upstream docs.

physicsnemo/distributed/shard_tensor.py
coreyjadams: In this case, we definitely don't want to emit a warning.  In fact, I rely on DTensor for so many of these operations that to put a warning here would just be ignored, since it would warn with just about every single operation.

As an example: `a + b` when both are shard tensors -> fall back to DTensor.
Nearly every single-tensor function or operation (`2*a`, `torch.nn.functional.relu(a)`, etc) -> fall back to DTensor.

The reason I have overloads, in general, is that some assumptions in DTensor break things (chunk sharding sizes are mandatory) while other behavior is incorrect for domain parallelism (like in the reductions, making sure grads go the right way and the mean is weighted properly).

tl;dr falling back to DTensor is a key design feature, actually :)

physicsnemo/models/domino/model.py
```diff
@@ -563,10 +574,10 @@ def __init__(
...
         self.fc4 = nn.Linear(int(base_layer), int(base_layer))
         self.fc5 = nn.Linear(int(base_layer), self.output_features)
-        self.bn1 = nn.BatchNorm1d(base_layer)
-        self.bn2 = nn.BatchNorm1d(int(base_layer))
-        self.bn3 = nn.BatchNorm1d(int(base_layer))
-        self.bn4 = nn.BatchNorm1d(int(base_layer))
+        # self.bn1 = nn.BatchNorm1d(base_layer)
+        # self.bn2 = nn.BatchNorm1d(int(base_layer))
+        # self.bn3 = nn.BatchNorm1d(int(base_layer))
+        # self.bn4 = nn.BatchNorm1d(int(base_layer))
```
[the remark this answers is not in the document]
coreyjadams: Cut

examples/cfd/external_aerodynamics/domino/src/train_sharded.py
```diff
@@ -0,0 +1,595 @@
...
+        vol_factors = np.load(vol_save_path)
+    else:
+        vol_factors = None
+
+    if os.path.exists(surf_save_path):
+        surf_factors = np.load(surf_save_path)
+    else:
+        surf_factors = None
+
+    train_dataset = create_domino_dataset(
```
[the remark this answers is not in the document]
coreyjadams: Done.

physicsnemo/distributed/shard_utils/pooling_patches.py
```diff
@@ -0,0 +1,397 @@
...
+from physicsnemo.distributed import ShardTensor
+from physicsnemo.distributed.shard_utils.patch_core import (
+    MissingShardPatch,
+    UndeterminedShardingError,
+)
+
+aten = torch.ops.aten
+
+
+def compute_output_shape(input_shape, pool_kwargs):
```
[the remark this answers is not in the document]
coreyjadams: Yes, but deferring to next release.

physicsnemo/distributed/shard_utils/unpooling_patches.py
```diff
@@ -0,0 +1,311 @@
...
+def compute_halo_configs_from_interpolate_args(
+    input: ShardTensor,
+    interp_kwargs: Dict[str, Any],
+) -> List[HaloConfig]:
+    """Compute halo configurations for a sharded tensor based on interpolation arguments.
+
+    Args:
+        input: The sharded tensor that will be used in interpolation
+        interp_kwargs: Dictionary of interpolation arguments including mode, size,
+                      scale_factor, etc.
```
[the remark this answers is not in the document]
coreyjadams: Yes, but deferring.

physicsnemo/distributed/shard_utils/unpooling_patches.py
```diff
@@ -0,0 +1,311 @@
...
+    input_shape: Tuple[int, ...],
+    placements: Sequence[Any],
+    interp_kwargs: Dict[str, Any],
+) -> Dict[int, Tuple[int, int]]:
+    """Compute the necessary halo sizes for different interpolation modes.
+
+    Args:
+        input_shape: Shape of the input tensor
+        placements: Placements from the ShardTensor spec
+        interp_kwargs: Keyword arguments for the interpolation operation
```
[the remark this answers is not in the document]
coreyjadams: Yes but deferring.

physicsnemo/distributed/shard_utils/attention_patches.py
```diff
@@ -0,0 +1,695 @@
...
+    Returns:
+        torch.Tensor: Result of log(exp(log_a) + exp(log_b)) computed in a numerically stable way
+
+    Think of this function as taking two values, A and B,
+    passed in via log form: log(A) and log(B).  This function
+    will return log(A+B) in a numerically stable way.
+
+    """
+    if log_a is None or log_b is None:
+        return log_a if log_a is not None else log_b
```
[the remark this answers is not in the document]
coreyjadams: Yes but deferring.

physicsnemo/distributed/shard_utils/attention_patches.py
coreyjadams: Yes, but deferring.

physicsnemo/distributed/shard_utils/conv_patches.py
```diff
@@ -411,59 +501,39 @@ def backward(
...
-        Convolution result as either torch.Tensor or ShardTensor
+        The result of the convolution operation
 
     Raises:
-        UndeterminedShardingError: If input tensor types are invalid
+        UndeterminedShardingError: If input, weight, or bias have invalid types
     """
-    input, weight, bias, conv_kwargs = repackage_conv_args(*args, **kwargs)
+
+    if "transpose" in func.__name__:
```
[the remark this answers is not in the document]
coreyjadams: Noted.  I think it's safer than first appearances suggest - func here is from pytorch `torch.nn.functional.conv[_transpose]Nd` where N  in 1, 2, 3.  those are the only ones wrapped.

I will leave it, for now, but welcome better suggestions in the future?

physicsnemo/distributed/shard_utils/attention_patches.py
```diff
@@ -0,0 +1,695 @@
...
+
+    diff = torch.abs(log_a - log_b)
+    return torch.max(log_a, log_b) + torch.log(torch.exp(-diff) + 1.0)
```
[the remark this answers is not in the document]
coreyjadams: Yes but deferring.

physicsnemo/distributed/shard_tensor.py
```diff
@@ -305,132 +339,123 @@ def __repr__(self) -> str:
...
-        return dispatch_res
+            return dispatch_res
 
     @staticmethod
     def from_local(
         local_tensor: torch.Tensor,
         device_mesh: Optional[DeviceMesh] = None,
         placements: Optional[Sequence[Placement]] = None,
-        infer_shape: Optional[bool] = True,
+        sharding_shapes: Union[str, Dict[int, List[Tuple[int, ...]]]] = "infer",
```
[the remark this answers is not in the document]
coreyjadams: Hmmm the docstring was elsewhere in another function using this.  But that one was out of date!

I've updated and duplicated so it's everywhere.  But for reference:
```
            sharding_shapes: Controls how shard tensor spec is generated:
                - "chunk": Use torch.chunk shapes to infer shapes from global shape (no communication)
                - "infer": Use collective communication to infer shapes from mesh neighbors.
                - Manual dict mapping mesh dim to list of shard shapes: Use provided shapes.  Must pass on each rank!
```

PR #839, opened 2025-04-07
DoMINO Model Refactor

peterdsharpe:
# PhysicsNeMo Pull Request

Refactor of DoMINO for dead code removal, performance tweaks, documentation/type hints, etc. Also fixes an important typo in the convolutional network setup of GeoProcessor, which used `conv2` twice in the downsampling and never used `conv3`.

Not to be merged until (at minimum) Rishi + Corey can review.

## Checklist

### DIFF
--- a/physicsnemo/models/domino/model.py
+++ b/physicsnemo/models/domino/model.py
@@ -47,20 +47,39 @@ def scale_sdf(sdf: torch.Tensor) -> torch.Tensor:
 
 
 class BQWarp(nn.Module):
-    """Warp based ball-query layer"""
+    """
+    Warp-based ball-query layer for finding neighboring points within a specified radius.
+    
+    This layer uses an accelerated ball query implementation to efficiently find points 
+    within a specified radius of query points.
+    """
 
     def __init__(
         self,
         input_features: int,
-        grid_resolution: Sequence[int, int, int] = [256, 96, 64],
+        grid_resolution: Sequence[int] = [256, 96, 64],
         radius: float = 0.25,
         neighbors_in_radius: int = 10,
     ):
+        """
+        Initialize the BQWarp layer.
+        
+        Args:
+            input_features: Number of feature dimensions for input points
+            grid_resolution: Resolution of the grid in each dimension [nx, ny, nz]
+            radius: Radius for ball query operation
+            neighbors_in_radius: Maximum number of neighbors to return within radius
+        """
         super().__init__()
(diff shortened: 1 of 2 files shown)


physicsnemo/models/domino/model.py
```diff
@@ -310,7 +290,8 @@ def forward(self, x, p_grid, sdf):
         # Binary sdf
         binary_sdf = binarize_sdf(sdf)
         # Gradients of SDF
-        sdf_x, sdf_y, sdf_z = calculate_gradient(sdf)
+        sdf_x, sdf_y, sdf_z = torch.gradient(sdf, dim=[2, 3, 4])
```
peterdsharpe, line 293: This replaces calculate_gradient() function for the sdf gradients. Note that this is not exactly numerically identical, as:

 a) the old method did not divide the central difference by 2 to account for the doubled dx on a central difference, while this does.

 b) the old method replaced the ends with zero gradients, while the new method replaces them with one-sided first-order finite differences.

In general, this is probably better - but checkpoints trained on the old model architecture will produce incorrect answers if inferenced with the new architecture. (This shouldn't be an issue for us, assuming we're pinning PhysicsNeMo versions in actual deployments, but we can't guarantee that all end-users are doing the same. This may be a motivation to save a PhysicsNeMo version ID in checkpoint files, but that's a separate discussion.)

physicsnemo/models/domino/model.py
```diff
@@ -375,7 +437,16 @@ def __init__(self, input_features, model_parameters=None):
 
         self.activation = F.relu
 
-    def forward(self, x, padded_value=-10):
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
```
peterdsharpe, line 440: Note: removed `padded_value=-10` here, which seems like dead code. Worth double-checking.

physicsnemo/models/domino/model.py
```diff
@@ -199,49 +237,66 @@ def __init__(self, input_filters, model_parameters):
...
-        else:
-            x = self.activation((self.conv2(x)))
+            x = self.conv_bn2(x)
+        x = self.activation(x)
         x = self.max_pool(x)
 
         x2 = x
+        x = self.conv3(x)
         if self.batch_norm:
-            x = self.activation(self.conv_bn3(self.conv2(x)))
```
peterdsharpe, line 219: Note: this fixes an important bug: in the 3rd level of the downsampling, conv2 was accidentally used twice, and conv3 was never used (in the batch_norm branch).

PR #840, opened 2025-04-07
DoMINO Model Refactor

peterdsharpe:
# PhysicsNeMo Pull Request

Refactor of DoMINO for dead code removal, performance tweaks, documentation/type hints, etc. Also fixes an important typo in the convolutional network setup of GeoProcessor, which used `conv2` twice in the downsampling and never used `conv3`.

Not to be merged until (at minimum) Rishi + Corey can review.

## Checklist


physicsnemo/models/domino/model.py
```diff
@@ -305,12 +393,13 @@ def forward(self, x, p_grid, sdf):
...
 
-        # Scaled sdf to emphasis on surface
+        # Scaled sdf to emphasize near surface
         scaled_sdf = scale_sdf(sdf)
         # Binary sdf
-        binary_sdf = binarize_sdf(sdf)
+        binary_sdf = torch.where(sdf >= 0, 0.0, 1.0)
         # Gradients of SDF
-        sdf_x, sdf_y, sdf_z = calculate_gradient(sdf)
+        sdf_x, sdf_y, sdf_z = torch.gradient(sdf, dim=[2, 3, 4])
```
peterdsharpe, line 401: This replaces calculate_gradient() function for the sdf gradients. Note that this is not exactly numerically identical, as:

a) the old method did not divide the central difference by 2 to account for the doubled dx on a central difference, while this does.

b) the old method replaced the ends with zero gradients, while the new method replaces them with one-sided first-order finite differences.

In general, this is probably better - but checkpoints trained on the old model architecture will produce incorrect answers if inferenced with the new architecture. (This shouldn't be an issue for us, assuming we're pinning PhysicsNeMo versions in deployments, but we can't guarantee that all end-users are doing this correctly. This may be a motivation to save a PhysicsNeMo version ID in checkpoint files, but that's a separate discussion.)

physicsnemo/models/domino/model.py
```diff
@@ -349,7 +438,16 @@ def __init__(self, input_features, model_parameters=None):
 
         self.activation = F.relu
 
-    def forward(self, x, padded_value=-10):
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
```
peterdsharpe, line 441: Note: removed padded_value=-10 here, which seems like dead code. Worth double-checking.

physicsnemo/models/domino/model.py
```diff
@@ -199,49 +237,66 @@ def __init__(self, input_filters, model_parameters):
...
-        else:
-            x = self.activation((self.conv2(x)))
+            x = self.conv_bn2(x)
+        x = self.activation(x)
         x = self.max_pool(x)
 
         x2 = x
+        x = self.conv3(x)
         if self.batch_norm:
-            x = self.activation(self.conv_bn3(self.conv2(x)))
```
peterdsharpe, line 219: Note: this fixes an important functionality bug: in the 3rd level of the downsampling, conv2 was accidentally used twice, and conv3 was never used (in the batch_norm branch).

physicsnemo/models/domino/model.py
```diff
@@ -21,76 +21,87 @@
...
-        vec.append(torch.cos(nx / 10000 ** (2 * (k) / d)))
-    return vec
-
-
-def scale_sdf(sdf):
-    """Function to scale SDF"""
-    return sdf / (0.4 + abs(sdf))
-
-
-def calculate_gradient(sdf):
```
peterdsharpe, line 50: Replaced below with `torch.gradient()`

peterdsharpe:
Main changes to be reviewed are called out in comments above. Thanks!

physicsnemo/models/domino/model.py
```diff
@@ -55,55 +51,80 @@ def calculate_pos_encoding(nx, d=8):
...
+    within a specified radius of query points.
+    """
 
     def __init__(
         self,
-        input_features,
-        grid_resolution=[256, 96, 64],
-        radius=0.25,
-        neighbors_in_radius=10,
+        input_features: int,
```
coreyjadams, line 81: Is input_features even used?

physicsnemo/models/domino/model.py
```diff
@@ -55,55 +51,80 @@ def calculate_pos_encoding(nx, d=8):
...
+        """
         batch_size = x.shape[0]
-        nx, ny, nz = (
-            self.grid_resolution[0],
-            self.grid_resolution[1],
-            self.grid_resolution[2],
-        )
+        nx, ny, nz = self.grid_resolution
 
         p_grid = torch.reshape(p_grid, (batch_size, nx * ny * nz, 3))
```
coreyjadams, line 129: Do we need to supply nx/ny/nz (and, grid_resolution in the constructor)?  This should work on arbitrary point clouds, the only requirements should be that the shape is [batch_size, (anything), 3] and the forward function could quietly reshape appropriately.

physicsnemo/models/domino/model.py
```diff
@@ -146,7 +184,12 @@ def __init__(self, input_features, model_parameters, grid_resolution=[256, 96, 6
 
         self.activation = F.relu
 
-    def forward(self, x, radius=0.025, neighbors_in_radius=10):
+    def forward(
+        self, x: torch.Tensor, radius: float = 0.025, neighbors_in_radius: int = 10
+    ) -> torch.Tensor:
+        """
```
coreyjadams, line 190: I don't fully understand this layer yet.  The fully-connecteds make sense, but the expand and reshape I don't totally follow.  Can we add a description of what is happening?

Also, could remove `nx, ny, nz = (self.grid_resolution[0] ...` and just do
```
x = torch.resshape(x, (batch_size, x.shape[-1], *self.grid_resolution)
```

physicsnemo/models/domino/model.py
```diff
@@ -211,49 +261,66 @@ def __init__(self, input_filters, model_parameters):
         self.activation = F.relu
         self.batch_norm = False
```
coreyjadams, line 262: batch norm is set to false here, and is not configurable at the user level.  Further, batch_norm doesn't make sense in this model since the batch_size is always 1.  

If no normalization will be used, we should remove calls in the forward bass and not construct the layers, it's just extra parameters.  If normalization will be used, LayerNorm or GroupNorm would make more sense in this scenario.

examples/cfd/external_aerodynamics/domino/src/test.py
```diff
@@ -472,9 +476,7 @@ def main(cfg: DictConfig):
             surface_coordinates = np.array(mesh.cell_centers().points, dtype=np.float32)
 
             interp_func = KDTree(surface_coordinates)
```
coreyjadams, line 478: I haven't looked at this code.  But we might replace this with a cuml call to find the neighbors on the GPU.

coreyjadams (approved):
Thank you for going through the model like this @peterdsharpe !  I left some comments but by and large consider them "optional", I think this is ready to merge.

physicsnemo/models/domino/model.py
peterdsharpe: Fixed in b572a2c279210ec3e25e1af902cee9384108773c

physicsnemo/models/domino/model.py
peterdsharpe: Yup, I buy this argument - good call. Removing.

peterdsharpe: Fixed in 46a09ddbb43fce1a4db1044b1f428dd97a3d8099

PR #863, opened 2025-04-30
A few fixes for the domino pipeline.

coreyjadams:
- initialize the distributed manager, if it isn't already.
- For partial datasets (surface only, volumen only) don't move "None" objects to cupy.
- When sampling/shuffling, if the number of points is too high then don't error.  Instead, shuffle and rely on padding.

# PhysicsNeMo Pull Request

## Checklist

### DIFF
--- a/physicsnemo/utils/domino/utils.py
+++ b/physicsnemo/utils/domino/utils.py
@@ -314,8 +314,6 @@ def pad_inp(arr: ArrayType, npoin: int, pad_value: float = 0.0) -> ArrayType:
     return arr_padded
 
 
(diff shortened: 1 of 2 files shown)


physicsnemo/utils/domino/utils.py
```diff
@@ -314,15 +314,16 @@ def pad_inp(arr: ArrayType, npoin: int, pad_value: float = 0.0) -> ArrayType:
     return arr_padded
 
 
-# CJA NOTE: the `sample_array` function is the same optimization I put into shuffle_array.
-# Keeping the cupy compatible version (mine)
 @profile
```
peterdsharpe, line 317: Is keeping the `@profile` decorator intended?

PR #866, opened 2025-05-02
Making the unit tests' nfs-data-dir configurable

abokov-nv:
# PhysicsNeMo Pull Request

## Description
Some of the unit tests require test data, currently it's hardcoded to `/data/nfs/modulus-data` and defined in a few places independently. `@nfsdata_or_fail` decorator is used to check if the expected path is present on the system and marks test as skipped if it needs data, but it is not there.
The problem with that approach is that you need root permissions to be able to create this `/data` folder on the system.
We should be more flexible and allow user to override that default path if needed.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -214,14 +214,16 @@ The pipeline has following stages:
     test, you will have to review your changes and fix the issues.
     To run pytest locally you can simply run `pytest` inside the `test` folder.
 
-    While writing these tests, we encourage you to make use of the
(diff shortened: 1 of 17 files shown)


ktangsali (approved):
Looks good! Great to have the ability to configure the dataset location. 

Should we also update the Makefile `get-data` command to make it more configurable? 

Also, we should update the Contribution Guide: https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md#ci

PR #871, opened 2025-05-05
Fea backward compatability

loliverhennigh:
# PhysicsNeMo Pull Request

## Checklist


physicsnemo/models/diffusion/unet.py
```diff
@@ -84,25 +84,68 @@ class UNet(Module):  # TODO a lot of redundancy, need to clean up
...
+        self.sigma_max = sigma_params.get("sigma_max", float("inf"))
+        self.sigma_data = sigma_params.get("sigma_data", 0.5)
+
         # for compatibility with older versions that took only 1 dimension
         if isinstance(img_resolution, int):
             self.img_shape_x = self.img_shape_y = img_resolution
         else:
-            self.img_shape_y = img_resolution[0]
-            self.img_shape_x = img_resolution[1]
+            self.img_shape_x = img_resolution[0]
```
CharlelieLrt, line 147: `img_resolution` is given as `[height, width]`, so this should be:

```python
self.img_shape_y = img_resolution[0]
self.img_shape_x = img_resolution[1]
```

loliverhennigh: Sorry wait didn't push the latest code for this...

physicsnemo/models/module.py
```diff
@@ -374,7 +414,7 @@ def from_checkpoint(
             # Open the tar file and extract its contents to the temporary directory
```
CharlelieLrt, line 414: In `from_checkpoint`, are you sure that you want to keep the argument `model_args`?

```python
    def from_checkpoint(
        cls, file_name: str, model_args: Optional[Dict] = None
    ) -> "Module":
```

AFAIK this was only implemented as a temporary workaround for the CorrDiff checkpoints

physicsnemo/models/diffusion/preconditioning.py
```diff
@@ -941,6 +942,186 @@ def amp_mode(self, value: bool):
...
+        img_resolution,
+        img_channels,
+        img_in_channels,
+        img_out_channels,
+        use_fp16=False,
+        sigma_min=0.0,
+        sigma_max=float("inf"),
+        sigma_data=0.5,
+        model_type="SongUNetPosEmbd",
+        scale_cond_input=True,
```
CharlelieLrt, line 1022: To avoid code duplication with the new `EDMPrecondSuperResolution`, I would actually subclass it and only override its `__init__` with the legacy arguments. Something like a legacy interface:

```python
class EDMPrecondSR(EDMPrecondSuperResolution):
    def __init__(self, ..., img_channels, ..., sigma_max, ...)
        super().__init__(self, ....)  # Args with popped legacy args img_channels, sigma_min, ...
    # Nothing else to do because everything is the same as in the new EDMPrecondSuperResolution
````
 
This has the significant advantage that we can enable perf optimization (e.g. `amp_mode`), even on the legacy checkpoints.

File: physicsnemo/models/module.py [hunk 4]
```diff
@@ -386,11 +426,35 @@ def from_checkpoint(
...
             # Merge model_args (adding new keys and updating existing ones)
             if model_args is not None:
                 args["__args__"].update(model_args)
 
-            model = cls.instantiate(args)
+            # Instantiate the model
+            model = Module.instantiate(args)
 
             # Load the model weights
             model_dict = torch.load(
```
CharlelieLrt, line 460: Just want to make sure that you are aware of this that is called in the `from_checkpoint`:

```python
model_dict = convert_ckp_apex(ckp_args, model_args, model_dict)
```

This is very specific to the SongUnet models (and was also a temporary workaround). AFAIK we would like to keep the core `module.py` as neat as possible, so maybe remove this?

That means we need to figure out a mechanism to still perform this conversion. IMO the conversion to apex should be handled in the specific classes that use it, so maybe a conversion `classmethod` there?

Also I'm not sure whether we want to keep `convert_ckp_apex` in `from physicsnemo.models.util_compatibility`. There is probably a better place for that...

CharlelieLrt (changes requested):
Overall looks like a neat handling of backward compatibility. My concerns are mostly about these two things:

1. A bunch of stuff was pushed recently as temporary fixes for CorrDiff checkpoints. Those kinda _polluted_ the `module.py`, so I want to make sure we leave it as neat as it used to be.
2. It's good to still be able to load CorrDiff checkpoints (which I think is the case with this PR), but we also want to be able to enable recent perf optimizations on these old checkpoints. I think the`UNet` in this PR supports this, but not the legacy class `EDMPrecondSR`.

CharlelieLrt (approved):
> 1. A bunch of stuff was pushed recently as temporary fixes for CorrDiff checkpoints. Those kinda polluted the module.py, so I want to make sure we leave it as neat as it used to be.

After discussion, we agreed that this is not critical and the cleanup of `module.py` (e.g. remove the apex conversion from `from_checkpoint`) can happen in a later PR.

Backward compatibility for both `EDMPrecondSR` and `UNet` now looks good to me, and should work with CorrDiff checkpoints.

PR #880, opened 2025-05-08
Stormcast customization conditions

jleinonen:
# PhysicsNeMo Pull Request

## Description

This PR adds support for configurable model conditions to the StormCast recipe.

**Main changes**:
1. Add configuration parameters `model.regression_conditions` and `model.diffusion_conditions`. These are lists consisting of `'state'`, `'background'`, `'invariant'` and/or `'regression'` (`'regression'` can only be specified in `model.diffusion_conditions`). The defaults are set to correspond to the StormCast paper.
2. Eliminate redundant parameters and configuration options that are now handled by the new two options: `previous_step_conditioning`, `use_regression_net`.
3. Streamline code to utilize the new parameters, eliminating redundancies. Particularly `utils.nn. build_network_condition_and_target` and `utils.nn.regression_model_forward`, as well as the calculation of model input dimension in `utils.trainer.py`, have been refactored.

**Features enabled by the changes**:
1. Training background-only or state-only models for benchmarking and debugging purposes, or to obtain a CorrDiff-like model to initialize a StormCast model state.
2. Other changes to model inputs, such as passing the background directly to the diffusion model.
3. Training pure diffusion (i.e. no regression step) models (if `'regression'` is not in `model.diffusion_conditions`).

## Checklist

## Dependencies

No new dependencies are introduced.

### DIFF
--- a/examples/generative/stormcast/config/inference/stormcast.yaml
+++ b/examples/generative/stormcast/config/inference/stormcast.yaml
@@ -15,7 +15,6 @@
 # limitations under the License.
 
 # General inference config items
-outdir: 'rundir' # Root path under which to save inference outputs
 experiment_name: 'stormcast-inference' # Name for the inference experiment being run
 run_id: 0 # Unique identifier for the inference run
 rundir: ./${inference.outdir}/${inference.experiment_name}/${inference.run_id} # Path where experiement outputs will be saved
--- a/examples/generative/stormcast/config/inference/stormcast.yaml
+++ b/examples/generative/stormcast/config/inference/stormcast.yaml
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 # General inference config items
+outdir: 'rundir' # Root path under which to save inference outputs
 experiment_name: 'stormcast-inference' # Name for the inference experiment being run
 run_id: 0 # Unique identifier for the inference run
 rundir: ./${inference.outdir}/${inference.experiment_name}/${inference.run_id} # Path where experiement outputs will be saved
--- a/examples/generative/stormcast/utils/nn.py
+++ b/examples/generative/stormcast/utils/nn.py
@@ -65,27 +65,41 @@ def get_preconditioned_architecture(
 
 
 def build_network_condition_and_target(
-    background, state, invariant_tensor, regression_net=None, train_regression_unet=True
+    background,
+    state,
+    invariant_tensor,
+    regression_net=None,
+    condition_list=("state", "background"),
+    regression_condition_list=("state", "background"),
 ):
-    assert not (train_regression_unet and (regression_net is not None))
+    assert not (("regression" in condition_list) and (regression_net is None))
     target = state[1]
-    if regression_net is not None:
-        # Inference regression model
-        with torch.no_grad():
-            reg_out = regression_model_forward(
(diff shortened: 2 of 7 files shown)


examples/generative/stormcast/config/inference/stormcast.yaml
```diff
@@ -15,7 +15,6 @@
 # limitations under the License.
 
 # General inference config items
-outdir: 'rundir' # Root path under which to save inference outputs
```
pzharrington, line 18: Was removing this intentional?

examples/generative/stormcast/utils/nn.py
```diff
@@ -65,27 +65,41 @@ def get_preconditioned_architecture(
...
 
 def build_network_condition_and_target(
-    background, state, invariant_tensor, regression_net=None, train_regression_unet=True
+    background,
+    state,
+    invariant_tensor,
+    regression_net=None,
+    condition_list=("state", "background"),
+    regression_condition_list=("state", "background"),
 ):
```
pzharrington, line 74: I think we should add a full docstring describing the behavior and inputs/outputs for this method since it is pretty important.

pzharrington (changes requested):
Overall the changes look nice and should improve usability/flexibility. I think the only needed additions are to flesh out the documentation a bit more:

- Let's add a complete docstring for `build_network_condition_and_target` since it is core to the training/inference functionality
- Document the new options and general setup/behavior in the README. Currently there are instructions for the old method that should be changed, and the descriptions of the model configs should also describe the/define the different components in the input condition list in this new approach

examples/generative/stormcast/config/inference/stormcast.yaml
jleinonen: This wasn't used anymore in the code (at least since the last PR, I'm not sure if it was used even before). All outputs are saved in `rundir` which I think is sufficient.

jleinonen: Ah - I see, this is still needed in the config file even though it's not referred to in the code. I'll put it back in the next commit.

jleinonen: Restored in https://github.com/NVIDIA/physicsnemo/pull/880/commits/4059d421f3028eebd628245e52f827e6d40aba9e.

examples/generative/stormcast/utils/nn.py
jleinonen: Improved docstring and type hints added in https://github.com/NVIDIA/physicsnemo/pull/880/commits/0dd9c7e56868cbd11689edf5072b59c01e1107aa.

jleinonen:
> Overall the changes look nice and should improve usability/flexibility. I think the only needed additions are to flesh out the documentation a bit more:
> 
> * Let's add a complete docstring for `build_network_condition_and_target` since it is core to the training/inference functionality
> * Document the new options and general setup/behavior in the README. Currently there are instructions for the old method that should be changed, and the descriptions of the model configs should also describe the/define the different components in the input condition list in this new approach

I have now added a more complete docstring and described the new options in the README. I also added in the README more information about adding custom datasets and improved the docstring of the `StormCastDataset` class.

pzharrington (approved):
Nice work! Good to go

PR #901, opened 2025-05-20
Merge SongUNetPosLtEmb with SongUNetPosEmb and add support for batch>1

### LINKED ISSUE #900: 🐛[BUG]: lt_aware_ce_regression training currently not working
### Version

main

Pip

### Describe the issue

train.py fails in line 265 because UNet got an unexpected keyword argument 'use_apex_gn'.
This is caused by the missing arguments 'use_apex_gn', 'act', 'profile_mode', 'amp_mode' of SongUNetPosLtEmbd in physicsnemo.models.diffusion.song_unet

Fixing this leads to another error in line 1126 in song_unet.py due to a dimension mismatch.  
RuntimeError: shape '[4, 160, 160]' is invalid for input of size 409600

```shell

```

### Relevant log output

```shell

```

### Environment details

```shell

```

jialusui1102:
# PhysicsNeMo Pull Request

## Description

Merged SongUNetPosLtEmb with SongUNetPosEmb and added support for batch>1 for lead-time models
Updated positional_embedding_indexing for faster bw
No graph recompile/graph break when running model with torch.compile

Closes #900.

## Checklist


physicsnemo/models/diffusion/song_unet.py
```diff
@@ -829,441 +947,8 @@ def positional_embedding_selector(
...
-            grid.requires_grad = False
-        elif self.gridtype == "test" and self.N_grid_channels == 2:
-            idx_x = torch.arange(self.img_shape_y)
-            idx_y = torch.arange(self.img_shape_x)
-            mesh_x, mesh_y = torch.meshgrid(idx_x, idx_y)
-            grid = torch.stack((mesh_x, mesh_y), dim=0)  # (2, img_shape_y, img_shape_x)
-        else:
-            raise ValueError("Gridtype not supported.")
-        return grid
-
```
CharlelieLrt, line 893: I might be missing something there, but why has this method totally disappeared?

CharlelieLrt (changes requested):
My main comment is about the item we discussed offline, which is to preserve an interface to the original `SongUNetPosLtEmb` class. Ideally:
1. Its `__init__` should have the same arguments as the original class (we can handle new arguments and/or renaming).
2. Its forward should have the same signature as before

physicsnemo/models/diffusion/song_unet.py
jialusui1102: It's still in the codebase. Not sure why the git history indicated a deletion

examples/generative/corrdiff/conf/base/model/lt_aware_ce_regression.yaml
```diff
@@ -31,7 +31,7 @@ model_args:
   # Number of channels for lead-time embeddings
   lead_time_steps: 9
   # Number of lead-time steps
-  model_type: "SongUNetPosLtEmbd"
```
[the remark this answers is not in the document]
jialusui1102: added SongUNetPosLtEmbd class and updated the config

examples/generative/corrdiff/train.py
```diff
@@ -256,6 +257,12 @@ def main(cfg: DictConfig) -> None:
         }
```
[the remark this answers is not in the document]
jialusui1102: I can moved the model arg update for `use_apex_gn` and `profile_mode` into the preceding if statement.

For `amp_mode `, can we assume that users always train the same series of ckp with the same precision?

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -1062,278 +1210,18 @@ def __init__(
...
-        grid = torch.nn.Parameter(
-            torch.randn(
-                self.lead_time_steps,
-                self.lead_time_channels,
-                self.img_shape_y,
-                self.img_shape_x,
-            )
-        )  # (lead_time_steps, lead_time_channels, img_shape_y, img_shape_x)
-        return grid
+    # Nothing else is re-implemented, because everything is already in the parent SongUNetPosEmb
```
CharlelieLrt, line 1227: Ok now I think the `SongUNetPosLtEmb` looks in good shape. Just a detail: to be more explicit you could redefine the forward with something similar to this:

```python
def forward(self, x, y, z):
    super().forward(x=x, y=y, z=z)
```

Calling the `super().forward` makes the code clearer and more maintainable.

examples/generative/corrdiff/train.py
```diff
@@ -245,16 +246,24 @@ def main(cfg: DictConfig) -> None:
         use_torch_compile = cfg.training.perf.torch_compile
     if hasattr(cfg.training.perf, "use_apex_gn"):
         use_apex_gn = cfg.training.perf.use_apex_gn
+        model_args.update(
+            {
+                "use_apex_gn": use_apex_gn,
+            }
+        )
```
CharlelieLrt, line 253: Did you mean `model_args["use_apex_gn"] = use_apex_gn` ?

jialusui1102: Yes, updated accordingly

physicsnemo/models/diffusion/song_unet.py
jialusui1102: Added forward()

CharlelieLrt (approved):
LGTM!

PR #902, opened 2025-05-21
Fixes for test collection and guards for missing PhysicsNeMo-Sym

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Addresses two bugs:

- Previously, running `pytest .` from root caused an error in test collection due to path issues in `test_runner.py`. Fixed this, and refactored to use the more-modern `pathlib` builtin library instead of `os.path`. (In general, going forward, [we should typically prefer `pathlib` over `os.path`](https://medium.com/@barila/choose-your-path-pythons-pathlib-vs-os-path-4de0b1e752dd))
- There was (and still is) a bug that [one of the docs-folder test scripts in the PhysicsNeMo repo depends on PhysicsNeMo-Sym](https://github.com/NVIDIA/physicsnemo/blob/main/docs/test_scripts/test_custom_model_demo_1.py). This fails on a vanilla PhysicsNeMo install, as Sym is not included. My current understanding is that the PhysicsNeMo -> PhysicsNeMo Sym dependency was intended to be only one-way: sym depends on core, but core does not depend on sym.

This provides a fix for the first bug, and at least a more graceful failure for the second bug.

## Checklist


docs/test_scripts/test_custom_model_demo_1.py
```diff
@@ -70,11 +70,14 @@ class MdlsUNetMetaData(ModelMetaData):
 # [physicsnemo model]
 
 # [physicsnemo sym model]
-
```
ktangsali, line 73: When does this failure occur? The code snippet here is rendered in the docs, and the try-except will not look good. For running pytest in the root directory, `make pytest` should be preferred. I can see the rationale of supporting `pytest .` too, but unforutnately this does not work in the present setting. Maybe something like below is better to modify the default behavior of pytest ([reference](https://docs.pytest.org/en/stable/reference/customize.html#pyproject-toml)):

```
[tool.pytest.ini_options]
testpaths = ["test"]
```

or perhaps a `.pytest.ini` file

PR #913, opened 2025-05-28
Update lead time embedding and probabilistic variables positioning in SongUNetPosEmb

tge25:
# PhysicsNeMo Pull Request

## Description
Add lead time embedding support for `positional_embedding_selector`
Enable arbitrary positioning of probabilistic variables
## Checklist

### DIFF
--- a/physicsnemo/models/diffusion/song_unet.py
+++ b/physicsnemo/models/diffusion/song_unet.py
@@ -728,7 +728,7 @@ class SongUNetPosEmbd(SongUNet):
                 # Select positional embeddings with a selector function
                 if embedding_selector is not None:
                     selected_pos_embd = self.positional_embedding_selector(
-                        x, embedding_selector
+                        x, embedding_selector, lead_time_label=lead_time_label
                     )
                 # Select positional embeddings using global indices (selects all
                 # embeddings if global_index is None)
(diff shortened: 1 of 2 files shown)


physicsnemo/models/diffusion/song_unet.py
```diff
@@ -735,30 +743,18 @@ def forward(
             if self.lead_time_mode:
                 # if training mode, let crossEntropyLoss do softmax. The model outputs logits.
                 # if eval mode, the model outputs probability
-                all_channels = list(range(out.shape[1]))  # [0, 1, 2, ..., 10]
-                scalar_channels = [
-                    item for item in all_channels if item not in self.prob_channels
-                ]
+                out_final = out.clone()
```
CharlelieLrt, line 746: Is there a way to refactor this to avoid the `clone` and directly work on the `out` tensor?

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -899,6 +895,7 @@ def positional_embedding_selector(
         self,
         x: torch.Tensor,
         embedding_selector: Callable[[torch.Tensor], torch.Tensor],
+        lead_time_label: int = None,
```
CharlelieLrt, line 898: Please add this argument to the docstring with concise description of how it should be used. Also, does it have any influence on the shape of the output tensor? Specify the returned tensor shape if it has any influence. Would be perfect if you could also add something in the "Example" section of the docstring.

CharlelieLrt (changes requested):
Overall looks good to me, only 2 minor question/comment related to documentation improvement and the usage of `clone`.

physicsnemo/models/diffusion/song_unet.py
tge25: The clone operation has been removed—thanks! The concern was that modifying a tensor in-place could interfere with the backward pass. I’ve tested the training script with the in-place edit, and everything appears to work correctly

physicsnemo/models/diffusion/song_unet.py
tge25: docstring has been updated

PR #954, opened 2025-06-04
restructured attention into separate class and fix errors in reshaping qkv and added inference optimization and fixes

jialusui1102:
# PhysicsNeMo Pull Request

## Description
This PR:
- Restructures the attention layer used in `models/diffusion` into a separate class for code reuse. It also fix errors in reshaping qkv
- Adds multiple inference optimizations for CorrDiff and other diffusion models (model compilation, compilation of patching operations in samplers, fp16 inference,  etc...)
- Refactors the automatic mixed precision (AMP) API in layers and models defined in `physicsnemo/models/diffusion`, for improved usability.

⚠️ **BC-breaking changes:**
This PR refactors the use of AMP with layers and models defined in `physicsnemo/models/diffusion/`. This also includes any model based on the `SongUNet` family of architectures. After this PR is merged, _it will be required_ to explicitly set `model.amp_mode = True` in order to use the model in a `torch.autocast` clause.

## Checklist


CharlelieLrt:
@jialusui1102 could you merge/rebase with `main` please?

jialusui1102:
@CharlelieLrt Merged with main

examples/weather/corrdiff/generate.py
```diff
@@ -131,14 +130,31 @@ def main(cfg: DictConfig) -> None:
...
         raise ValueError(f"Invalid inference mode {cfg.generation.inference_mode}")
 
+    use_apex_gn = False
+    profile_mode = False
+
+    if hasattr(cfg.generation.perf, "use_apex_gn"):
+        use_apex_gn = cfg.generation.perf.use_apex_gn
+
+    if hasattr(cfg.generation.perf, "profile_mode"):
+        profile_mode = cfg.generation.perf.profile_mode
```
CharlelieLrt: Could you make this more compact using `getattr` instead? Also the variables `use_apex_gn ` and `profile_mode ` are only used to store in the `model_args` dict, so let's avoid creating variables for this and let's directly put them into the dict

CHANGELOG.md
```diff
@@ -59,6 +59,7 @@ arbitrary positioning of probabilistic variables
 - Enable lead time aware regression without CE loss
 - Bumped minimum PyTorch version from 2.0.0 to 2.4.0, to minimize
   support surface for `physicsnemo.distributed` functionality.
+- Moved attention logic out from UNetBlock into a separate class for clarity and modularity
```
CharlelieLrt: Could you detail that it is for diffusion model layers, and that in impacts the `SongUnet` diffusion models? Otherwise it's not clear for which model this is applied (I think we have multiple UNets implementations, in addition of the ones used for diffusion)

examples/weather/corrdiff/generate.py
```diff
@@ -131,14 +130,31 @@ def main(cfg: DictConfig) -> None:
...
+
+    model_args = {
+        "use_apex_gn": use_apex_gn,
+        "profile_mode": profile_mode,
+    }
+
```
CharlelieLrt: We use the same options for both the regression network and the diffusion network? Are we sure that there won't be a case where the regression network supports these options but the diffusion model does not?

examples/weather/corrdiff/generate.py
```diff
@@ -149,10 +165,14 @@ def main(cfg: DictConfig) -> None:
...
         reg_ckpt_filename = cfg.generation.io.reg_ckpt_filename
         logger0.info(f'Loading network from "{reg_ckpt_filename}"...')
-        net_reg = Module.from_checkpoint(to_absolute_path(reg_ckpt_filename))
+        net_reg = Module.from_checkpoint(
+            to_absolute_path(reg_ckpt_filename), model_args=model_args
+        )
         net_reg = net_reg.eval().to(device).to(memory_format=torch.channels_last)
         if cfg.generation.perf.force_fp16:
             net_reg.use_fp16 = True
+            net_reg.to(torch.float16)
```
CharlelieLrt: Same comment as before: I just want to make sure there is no case where users might attempt to use the regression with `force_fp16` while the diffusion does not use `force_fp16`?

In addition could you rename the `force_fp16` into `use_fp16` _everywhere_ (including the configs) to match the model attribute name?

examples/weather/corrdiff/generate.py
```diff
@@ -131,14 +130,31 @@ def main(cfg: DictConfig) -> None:
...
         res_ckpt_filename = cfg.generation.io.res_ckpt_filename
         logger0.info(f'Loading residual network from "{res_ckpt_filename}"...')
-        net_res = Module.from_checkpoint(to_absolute_path(res_ckpt_filename))
+        net_res = Module.from_checkpoint(
+            to_absolute_path(res_ckpt_filename), model_args=model_args
+        )
         net_res = net_res.eval().to(device).to(memory_format=torch.channels_last)
         if cfg.generation.perf.force_fp16:
             net_res.use_fp16 = True
+            net_res.to(torch.float16)
```
CharlelieLrt: A few comments here:

1. I am very confused with the fact that we need both `net_res.use_fp16 = True` and `net_res.to(torch.float16)`? What happens if I do `net_res.use_fp16 = False` and `net_res.to(torch.float16)`? Or vice-versa what happens when `net_res.use_fp16 = True` and `net_res.to(torch.float32)`? Are those supposed to raise errors or is it supposed to run? If it runs, what's the expected behavior?

2. AFAIK we don't have any CI tests for the `UNet` and `EDMPrecondSuperResolution` in `fp16` precision. It would be good to duplicate and adapt [this test for the UNet](https://github.com/NVIDIA/physicsnemo/blob/main/test/models/diffusion/test_unet_wrappers.py#L34-L56) and  [this test for the EDMPrecondSuperResolution](https://github.com/NVIDIA/physicsnemo/blob/main/test/models/diffusion/test_preconditioning.py#L31-L56). To test everything end-to-end, we should create the models with `use_fp16=False`, then do `net.use_fp16 = True`, and finally run the forward. I guess we should also check that the results given in `fp16` and `fp32` are not too far from each other.
**Note**: we might even need a _non-regression_ test in `fp16`, otherwise this feature will most likely break at some point.

physicsnemo/utils/generative/stochastic_sampler.py
```diff
@@ -23,6 +23,16 @@
...
+@torch.compile()
+def apply_wrapper(patching, input, additional_input=None):
+    return patching.apply(input=input, additional_input=additional_input)
+
+
+@torch.compile()
+def fuse_wrapper(patching, input, batch_size):
+    return patching.fuse(input=input, batch_size=batch_size)
+
+
```
CharlelieLrt: So, this will always be compiled, right? Are we certain that this won't cause any problem (e.g. perf degradation)? I am think of cases where we might not have many patches, or when the patches are small?

physicsnemo/models/diffusion/layers.py
```diff
@@ -476,6 +477,74 @@ def backward(ctx, dw):
         return dq, dk
 
 
+class Attention(torch.nn.Module):
+    """
+    Unified U-Net block with optional up/downsampling and self-attention.
+    Represents the union of all features employed by the DDPM++, NCSN++, and
+    ADM architectures.
+    """
```
CharlelieLrt: This docstring is not accurate. Also we are trying to improve documentation , so docstrings should preferably be in the format:

```python
class Attention(torch.nn.Module):
    r"""
    A brief description of the class.
    
    Parameters
    ------------
    param1 : type of param 1, indicate if optional, if optional write default value
        Description of param1, including its expected shape if it is a tensor.
    param2 : etc...

    Forward
    --------
    argument1 : type of argument1, indicate if optional, if optional write default value
        Description of argument1, including expected shape if it is a tensor.
    
    Outputs
    --------
    torch.Tensor (or other type if the forward doesn't return a tensor)
        Description of output, including shape if it is a tensor.
    """
```

See [this example](https://github.com/NVIDIA/physicsnemo/blob/b93d124ad5c305c14fcf2b9f6c34dea70b83e90d/physicsnemo/models/diffusion/song_unet.py#L62-L248) for docstring formatting (doesn't need to be as long of course)

physicsnemo/models/diffusion/layers.py
```diff
@@ -476,6 +477,74 @@ def backward(ctx, dw):
...
+        )
+        self.proj = Conv2d(
+            in_channels=out_channels,
+            out_channels=out_channels,
+            kernel=1,
+            fused_conv_bias=fused_conv_bias,
+            amp_mode=amp_mode,
+            **init_zero,
+        )
+        self.num_heads = num_heads
```
CharlelieLrt, line 604: I think we should add parameter validation for `num_heads` (ie. raise an error if `num_heads` is not an integer > 0)

CharlelieLrt (changes requested):
@jialusui1102 those look like some very nice additions. I left some minor comments/clarifications, and a few major ones:

1. The use of `net.use_fp16 = True` seems inconsistent to me, because it seems that to make it work properly the user still need to manually do `net.to(torch.float16)`. What would happen if the user doesn't do it is not 100% clear to me. IMO, a good design would be to transform `use_fp16` into a `@property` of the U-Net wrappers, that automatically does the type conversion `self.to(torch.float16)` when enabled.

2. A few CI tests missing for the `use_fp16` feature (actually non-regression would be even better).

3. Some documentation missing.

examples/weather/corrdiff/generate.py
jialusui1102: Sure. Updated in code

examples/weather/corrdiff/generate.py
jialusui1102: Yes, the regression and diffusion net are essentially of the same SongUnetPosEmbd architecture, and both support such options (same as training)

examples/weather/corrdiff/generate.py
jialusui1102: I couldn't think of a case where the user want to use different dtype for inferencing regression and diffusion model. The inference data type should be consistent for both.
Sure I will rename `force_fp16 ` into `use_fp16 ` for clearer usage

examples/weather/corrdiff/generate.py
jialusui1102: I think we wouldn't really expect users to manually change this section of code, especially about the model dtype casting. According to the current logistic, if and only if user set `use_fp16`=True in config will we explicitly cast the model into `float16`. Setting `net_reg.use_fp16 = True` will only cast the inputs to float16 in model forward, not the model itself. If we do not explicitly converting the model (regression net AND diffusion net) to float16, we would be sure to encounter a type mismatch error in model forward.

physicsnemo/utils/generative/stochastic_sampler.py
jialusui1102: I don't think we would have any perf degradation by compiling these function. Without torch.compile, the `patching.apply` would call `torch.nn.functional.fold` and `torch.nn.functional.unfold`, leading to long `img2col` kernels with barely no DRAM activity and tensorcore activity. After torch.compile, we are able to reduce runtime for such `img2col` kernels from 420ms to ~2ms for patch shape (448,448). For bigger patches, I would expect the speedup effect to be even bigger, for smaller patches, it will be either of the same speed or much faster.

physicsnemo/models/diffusion/layers.py
jialusui1102: Sure. This is some old comments carried over from the cBottle codebase, I will update the docstring according to this standard

physicsnemo/models/diffusion/layers.py
jialusui1102: that makes sense. added a check in Attention init

CHANGELOG.md
jialusui1102: updated CHANGELOG for more detailed explanation

jialusui1102:
Made a few changes based on the feedback collected last week:
1. added attention ckp conversion function
2. updated the use_fp16 logistics to make all conversions inside model level (including `init` and `model.use_fp16`)
3. added unit tests for use_fp16 (all passed locally)

physicsnemo/utils/patching.py
```diff
@@ -463,9 +465,78 @@ def fuse(self, input: Tensor, batch_size: int) -> Tensor:
...
+        Returns
+        -------
+        None
+
+        Notes
+        -----
+        - The input tensor is a dummy tensor of ones with shape [1, 1, H, W],
+          where H and W are the padded image dimensions.
+        - The output is a tensor of the same shape indicating how many times
+          each pixel is visited by patches.
```
CharlelieLrt: This part of the docstring seems wrong? Why is it mentioning a "a dummy tensor of ones" as input? Also the `Returns` section should be a tensor with shape (H, W), or something similar, correct?

physicsnemo/utils/patching.py
```diff
@@ -615,6 +684,7 @@ def image_fuse(
     batch_size: int,
     overlap_pix: int,
     boundary_pix: int,
+    overlap_count: Tensor,
```
CharlelieLrt: I expect the existing tests for `image_fuse` to fail, because it seems they were not updated with this additional argument. One easy way to make them pass could be to make `overlap_count` an optional argument. If it is not provided (`None`), then it could be recomputed in `image_fuse`, as it was the case before. This way we keep the original functionality + the optional performance gain by providing a pre-computed `overlap_count`. What do you think?

CharlelieLrt (changes requested):
LGTM, just a small detail about a docstring and some tests that might not pass (but let's see if they do first)

physicsnemo/utils/patching.py
jialusui1102: That makes sense! I will adjust the code and update the PR

physicsnemo/utils/patching.py
jialusui1102: Input tensor was intended to refer to `input_ones`, and the returned `overlap_count` was of the same shape as `input_ones` (just double checked the shape)
I would revise the docstring for better clarity

CharlelieLrt (approved):
LGTM

PR #956, opened 2025-06-05
Add tutorial on physics addition

ktangsali:
# PhysicsNeMo Pull Request

## Description
Adds tutorial on adding physics to the model training. 

Also, adding some text to the landing page (index.rst) based on the technical writer feedback

## Checklist


docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,670 @@
+Adding Physics-based constraints using PhysicsNeMo
```
peterdsharpe: This is perhaps more of a big-picture discussion, but: would it maybe be clearer to communicate this as "Physics-based loss terms" or "Physics-Guided training" rather than "Physics-based constraints"? To me, I worry that:

* The less-technical audience will see "constraints" here and assume that this means that a model will *exactly* obey physical laws, when that's not really true - we're adding a term that encourages the model to obey the governing equations locally, but those physical laws are only approximately enforced, rather than exactly constrained. In a way, it's more like "guiding" rather than "constraining". Maybe "Physics-guided training" is another way we could phrase this?

* I worry that the technical audience (especially those from more of an applied math background than a ML background) will assume we're doing something like a Lagrange-multiplier based primal-dual constraint method, when really we're just adding exterior penalty functions. This is a bit less rigorous, in that it requires a heuristic weighting parameter and leads to poorly-scaled optimization formulations (i.e., training) if this is not tuned well by the user. "Constraint" implies a projection of the admissible function space onto the manifold of functions that respect the governing equations - a) this approach only achieves this in the limit of infinite `physics_weight`, and b) this is only possible if the physics loss is driven to exactly zero, which is unlikely to happen for practical problems.

(Edit: tagging @mnabian for his input here, and @ram-cherukuri per our discussion earlier today)

docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,670 @@
...
+Adding this, enforces an equilibrium condition on the model's prediction and is a
+demonstration of how physics knowledge can be incorporated in your training workflow. 
+A full example using this loss can be found in the `Molecular Dynamics Example <../examples/molecular_dynamics/lennard_jones/README.rst>`_
+
+Adding PDE constraints
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Sometimes, the governing equations for the problem can include Partial Differential Equations.
+This is especially true for problems in physics and other scientific domains. Computing these
+PDEs on model predictions involves computing spatial and temporal gradients and the methods
```
peterdsharpe: Maybe instead of: "Computing these PDEs on model predictions" it would be more precise to say "Computing local residuals of these PDEs on a model prediction"

ram-cherukuri (approved):
Looks comprehensive and detailed @ktangsali. I love it. One comment - given the multiple options, do we have guidance on when to use a particular method vs other.

ktangsali:
> Looks comprehensive and detailed @ktangsali. I love it. One comment - given the multiple options, do we have guidance on when to use a particular method vs other.

Thanks @ram-cherukuri . So I have tried providing some guidance in the notes section of individual methods. For example:
```
The code below shows an example of using the ``autodiff`` method to compute the residuals.
A few things to note when using the ``autodiff`` method:

- Ensure the model is differentiable, e.g. a model that uses ReLU activation
  function will have it's second derivatives zero. 
- Set ``.requires_grad_()`` to ``True``. 
- Coordinates is a (N, 1) shaped tensor for 1D, (N, 2) for 2D and (N, 3) for 3D.
- This method is accurate but more computationally expensive compared to some 
  other numerical methods due to automatic differentiation. 
```

However, I can also add a tabulated representation to simply it further. Let me add that.

ktangsali:
> Looks comprehensive and detailed @ktangsali. I love it. One comment - given the multiple options, do we have guidance on when to use a particular method vs other.

Added

docs/api/physicsnemo.models.rst
```diff
@@ -313,7 +313,7 @@ model from the ``.mdlus`` file.
 
 
 .. note::
-   In order to make use of this functionality, the model must have json serializable
+   In order to make use of this functionality, the model must have .json serializable
```
peterdsharpe:
```suggestion
   In order to make use of this functionality, the model must have `.json` serializable
```

docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,693 @@
+Adding Physics-based constraints using PhysicsNeMo
+===================================================
+
+In this tutorial, we'll learn how to add physics-based constraints to your model training
+using PhysicsNeMo. `PhysicsNeMo-Sym <https://github.com/NVIDIA/physicsnemo-sym>`_
+is a submodule of PhysicsNeMo that provides algorithms and utilities to physics-inform
+the training of AI models. In this tutorial, we will explore the different utilities from
+PhysicsNeMo-Sym, followed by sample end-to-end training workflows.
+
+Adding physics-based constraints
```
peterdsharpe:
```suggestion
Adding Physics-Based Constraints
```

docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,693 @@
...
+    loss = loss_data + physics_weight * loss_physics
+    loss.backward()
+    optimizer.step()
+    ...
+
+Adding this, enforces an equilibrium condition on the model's prediction and is a
+demonstration of how physics knowledge can be incorporated in your training workflow. 
+A full example using this loss can be found in the `Molecular Dynamics Example <../examples/molecular_dynamics/lennard_jones/README.rst>`_
+
+Adding PDE constraints
```
peterdsharpe:
```suggestion
Adding PDE Constraints
```

docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,693 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Sometimes, the governing equations for the problem can include Partial Differential Equations.
+This is especially true for problems in physics and other scientific domains. Computing these
+PDEs on model predictions involves computing spatial and temporal gradients and the methods
+used can vary based on the model architecture. 
+
+PhysicsNeMo-Sym aims to simplify this process. 
+The `PhysicsInformer <https://docs.nvidia.com/deeplearning/physicsnemo/physicsnemo-sym/api/physicsnemo.sym.eq.html#module-physicsnemo.sym.eq.phy_informer>`_
+utility can compute the PDE losses on point clouds, grids or graphs/meshes using
```
peterdsharpe:
```suggestion
utility can compute the PDE losses on point clouds, grids, or graphs/meshes using
```

docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,693 @@
...
+techniques such as Automatic Differentiation, Finite Difference, Meshless Finite Difference,
+Least Squares, Spectral Differentiation, etc. 
+
+The ``PhysicsInformer`` class requires equations to compute as a parameter. The equation
+must be in the form of PhysicsNeMo-Sym's PDE. Custom PDEs are also supported. 
+
+Some general comments for the ``PhysicsInformer``:
+
+
+- Using PhysicsInformer, you can physics-inform almost any model architecture
```
peterdsharpe:
```suggestion
- Using `PhysicsInformer`, you can add physics loss terms to almost any model architecture
```

docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,693 @@
...
+  (point-cloud based, grid based, graph based, etc.), e.g. MLPs, DeepONets, FNOs, CNNs, Diffusion Models,
+  Graph networks, etc.
+- The utility supports gradients via Automatic Differentiation, Spectral,
+  Finite Difference, Meshless Finite Difference and Least Squares methods.
```
peterdsharpe: Can we add a link here to more detail on these? Some of these are relatively straightforward (AD, finite-diff), but others are less conventional (e.g., meshless finite-diff), or users may have questions about which cases this can be applied to (e.g., spectral derivatives).

docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,693 @@
...
+- Given the equations and required_outputs, this utility constructs the computational graph,
```
peterdsharpe:
```suggestion
- Given the PDE governing equations and `required_outputs`, this utility constructs the computational graph,
```

docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,693 @@
...
+- This utility simplifies the spatial derivative computation. 
+  If using transient PDEs, the temporal derivative will have to be computed separately
+  and passed as an input to the ``forward()`` call.
+  Refer `API Doc <https://docs.nvidia.com/deeplearning/physicsnemo/physicsnemo-sym/api/physicsnemo.sym.eq.html#module-physicsnemo.sym.eq.phy_informer>`_ for more details.
+- Different spatial gradient computing methods require different inputs to the ``forward()`` call. 
+  To identify the inputs that need to be passed to the ``forward()`` call, you 
+  can access the value of ``.required_inputs`` property.
+
+
+The various methods and their applicability based on the model output type can be summarized below
```
peterdsharpe:
```suggestion
The various spatial derivative methods and their applicability based on the model output type can be summarized below:
```

docs/tutorials/physics_addition.rst
peterdsharpe: Nevermind, seen below.
```suggestion
  Finite Difference, Meshless Finite Difference and Least Squares methods. (See the following section for more info.)
```

docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,693 @@
...
+
+- **Automatic Differentiation**: Suitable for outputs from models which are differentiable w.r.t model inputs.
+  Model inputs must include coordinates as inputs. Ideal for MLP type of architectures, and can operate on point clouds,
+  structured grids or unstructured meshes as long as the differentiability and input constraints are satisfied.
```
peterdsharpe:
```suggestion
  structured grids, or unstructured meshes as long as the differentiability and input constraints are satisfied.
```

docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,693 @@
...
+  Computationally expensive but more accurate than other numerical gradient methods.
+
+- **Meshless Finite Difference**: Numerical gradient method suitable for models that can
+  predict field values on stencil points in addition to the original points. The points can come from a point cloud, 
+  grid (structured or unstructured), but most suitable for point clouds.
```
peterdsharpe: "The points can come from a point cloud, grid (structured or unstructured), but most suitable for point clouds."

Would be clearer as:

"The points can come from a either point clouds or grids (structured or unstructured); point clouds are the most suitable applications."

docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,693 @@
...
+
+
+Computing PDE losses using Automatic Differentiation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The code below shows an example of using the ``autodiff`` method to compute the residuals.
+A few things to note when using the ``autodiff`` method:
+
+- Ensure the model is differentiable, e.g. a model that uses ReLU activation
+  function will have it's second derivatives zero.
```
peterdsharpe: "- Ensure the model is differentiable, e.g. a model that uses ReLU activation function will have it's second derivatives zero. "

This is unclear - after all, a ReLU function is infinitely differentiable. I think what we mean here is not that the function should be differentiable, but rather that it should be at least $C^1$ continuous. So, maybe we write:

"Ensure the model is a $C^1$-continuous mapping. In practice, this often means replacing any ReLU activation functions with higher-continuity variants, such as Swish, Softplus, or GeLU."

docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,693 @@
...
+Computing PDE losses using Automatic Differentiation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The code below shows an example of using the ``autodiff`` method to compute the residuals.
+A few things to note when using the ``autodiff`` method:
+
+- Ensure the model is differentiable, e.g. a model that uses ReLU activation
+  function will have it's second derivatives zero. 
+- Set ``.requires_grad_()`` to ``True``. 
+- Coordinates is a (N, 1) shaped tensor for 1D, (N, 2) for 2D and (N, 3) for 3D.
```
peterdsharpe:
```suggestion
- Coordinates is a tensor of shape `(N, D)`, where `D` is the number of spatial dimensions.

File: docs/tutorials/physics_addition.rst [hunk 15]
```diff
@@ -0,0 +1,693 @@
...
+    phy_informer = PhysicsInformer(
+        required_outputs=["continuity", "momentum_x"],
+        equations=ns,
+        grad_method="autodiff",
+        device=coords.device,
+    )
+
+    # model forward pass
+    # this needs to be fully differentiable for auto-diff gradients to work
+    # if the model is not fully differentiable, follow along this tutorial to
```
peterdsharpe: "fully differentiable" should be replaced by "$C^1$-continuous" here.

docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,693 @@
...
+- In Addition to the outputs at the original data points, outputs are needed on the 
+  stencil points. The stencil points can be computed using 
+  ``physicsnemo.sym.eq.spatial_grads.spatial_grads.compute_stencil3d`` function
+  from PhysicsNeMo Sym. Stencil points are defined using the following convention
+  "u>>x::1": u(i+1, j) "u>>x::-1": u(i-1, j) "u>>x::1&&y::1": u(i+1, j+1) "u>>x::-1&&y::-1": u(i-1, j-1) etc.
+  To identify the inputs that need to be passed to the ``forward()`` call, you 
+  can access the value of ``.required_inputs`` property.
+- ``fd_dx`` is a hyperparameter. Smaller value typically yields more accurate
+  gradients, but can lead to numerical instability. A value of 0.001 is a good
+  value to start.
```
peterdsharpe:
```suggestion
  value to start, assuming the variation of spatial coordinates in the problem is $\mathcal{O}(1)$.
```

docs/tutorials/physics_addition.rst
peterdsharpe: Actually, on second thought:
* First-order PDEs need $C^1$-continuity
* Second-order PDEs (e.g., Navier-Stokes) need $C^2$-continuity.
* And so on...

The example activation functions we've given here are $C^\infty$-continuous, so this is fine, but this may be something to highlight to the user here.

docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,693 @@
...
+- Apply boundary conditions
+- ...
+
+Let's review some of these below.
+
+Computing Signed Distance Fields
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Mathematically, signed distance field or signed distance function (SDF) is defined as the orthogonal distance
+of a given point to the boundary / surface of a geometric shape. It is widely used to describe the geometry
```
peterdsharpe:
```suggestion
of a given point to the nearest boundary / surface of a geometric shape. It is widely used to describe the geometry
```

docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,693 @@
...
+                "z": np.array([[0]]),
+            },
+        params={}
+    )["sdf"]
+
+A few examples using SDF during training / inference can be found in the 
+`External Aerodynamics using DoMINO Example <../examples/cfd/external_aerodynamics/domino/README.rst>`_, 
+`Datacenter CFD example <../examples/cfd/datacenter/README.rst>`_.
+
+Sampling point clouds
```
peterdsharpe:
```suggestion
Sampling Point Clouds
```

docs/tutorials/physics_addition.rst
peterdsharpe: Related: we can recommend that users who want to use ReLU functions use numerical derivatives instead - but in general, this will produce poor results for second-order PDEs if you try to use physics loss terms, even if one uses numerical derivatives. Fundamentally, the function space needs to admit curvature in order for a solution of a second-order PDE to be valid (in the non-trivial case), so I worry we're steering users in the wrong direction here.

docs/tutorials/physics_addition.rst
```diff
@@ -0,0 +1,693 @@
...
+
+Computing PDE losses using Automatic Differentiation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The code below shows an example of using the ``autodiff`` method to compute the residuals.
+A few things to note when using the ``autodiff`` method:
+
+- Ensure the model is differentiable, e.g. a model that uses ReLU activation
+  function will have it's second derivatives zero. 
+- Set ``.requires_grad_()`` to ``True``.
```
peterdsharpe:
```suggestion
- For all spatial coordinate tensors (e.g., `x`, `y`, and `z`), call the method ``x.requires_grad_(True)`` to enable gradient tracking. 
```

docs/tutorials/physics_addition.rst
ktangsali: Great points. Added some opening text ad modified where applicable. Please have a look.

docs/tutorials/physics_addition.rst
ktangsali: Great point. I also provided links to API docs for individual sections, just in case.

docs/tutorials/physics_addition.rst
ktangsali: Great point. Added.

PR #969, opened 2025-06-11
Resuming for `InfiniteSampler`

juliusberner:
# PhysicsNeMo Pull Request

## Description
When resuming before iterating through the whole dataset, the `InfiniteSampler` would restart and a certain part of the dataset would be ignored during training. We also adapt the seed based on `cur_nimg` to avoid the same randomness in, e.g., dropout, after resuming.

## Checklist


examples/weather/corrdiff/train.py
```diff
@@ -153,8 +153,17 @@ def main(cfg: DictConfig) -> None:
             cfg.training.hp.total_batch_size // dist.world_size
         )
 
+    # Load the current number of images for resuming
+    try:
+        cur_nimg = load_checkpoint(
+            path=checkpoint_dir,
+            device=dist.device,
```
CharlelieLrt: We probably don't need the `device` argument since we are only reading `cur_nimg`

examples/weather/corrdiff/datasets/dataset.py
```diff
@@ -110,14 +111,19 @@ def init_train_valid_datasets_from_config(
...
     """
 
     config = copy.deepcopy(dataset_cfg)
     (dataset, dataset_iter) = init_dataset_from_config(
-        config, dataloader_cfg, batch_size=batch_size, seed=seed
+        config,
+        dataloader_cfg,
+        batch_size=batch_size,
+        seed=seed,
+        sampler_start_idx=sampler_start_idx,
```
CharlelieLrt, line 126: Why is this not applied to the `validation_dataset` just below?

CharlelieLrt (changes requested):
Just two minor questions/comments, otherwise everything looks good

examples/weather/corrdiff/datasets/dataset.py
juliusberner: I could also include resuming for the validation sampler. However,
1) we currently do not count the nimg for the validation set 
2) we typically validate over the full validation dataset or do not care so much if we resume from the previous index. 

What do you think?

examples/weather/corrdiff/train.py
juliusberner: I changed it, thanks!

examples/weather/corrdiff/datasets/dataset.py
CharlelieLrt: Agree with 2. Thanks for the clarifications!

CharlelieLrt (approved):
LGTM, thanks for the fixes!

PR #971, opened 2025-06-12
Add patch-based support for deterministic sampler

### LINKED ISSUE #345: 🚀[FEA]: patch-based corrdiff generation for deterministic sampler

New Feature

Low (would be nice)

_No response_

_No response_

tge25:
# Patch-base support for deterministic sampler

## Description
Enable patch-base inference with deterministic sampler.
Stochastic sampler cleanup.

Closes #345 .

## Checklist


physicsnemo/utils/diffusion/deterministic_sampler.py
```diff
@@ -81,6 +85,21 @@ def deterministic_sampler(
         during the stochastic sampling. Must have the same signature as
         torch.randn_like and return torch.Tensor. Defaults to
         torch.randn_like.
+    patching : Optional[GridPatching2D], optional
+        A patching utility for patch-based diffusion. Implements methods to
+        extract patches from an image and batch the patches along `dim=0`.
+        Should also implement a `fuse` method to reconstruct the original image
+        from a batch of patches. See :class:`physicsnemo.utils.patching.GridPatching2D`
+        for details. By default None, in which case non-patched diffusion is used.
```
CharlelieLrt: A few details that might not render properly in the docs:
- use double `` of single ` for inline variable or code
- use ~ in :class:&#96;~physicsnemo.....&#96;

physicsnemo/utils/diffusion/deterministic_sampler.py
```diff
@@ -81,6 +85,21 @@ def deterministic_sampler(
...
+    mean_hr : Optional[Tensor], optional
+        Optional tensor containing mean high-resolution images for
+        conditioning. Must have same height and width as `img_lr`, with shape
+        (B_hr, C_hr, img_lr_shape_y, img_lr_shape_x)  where the batch dimension
```
CharlelieLrt: Formatting missing for all mathematical variable names. For example: :math:&#96;(B_{hr}, C_{hr}, H, W)&#96;

physicsnemo/utils/diffusion/deterministic_sampler.py
```diff
@@ -213,6 +260,22 @@ def deterministic_sampler(
     sigma_min = max(sigma_min, net.sigma_min)
     sigma_max = min(sigma_max, net.sigma_max)
 
+    batch_size = img_lr.shape[0]
+    # input and position padding + patching
+    if patching:
+        # Patched conditioning [x_lr, mean_hr]
+        # (batch_size * patch_num, C_in + C_out, patch_shape_y, patch_shape_x)
+        x_lr = patching.apply(input=x_lr, additional_input=img_lr)
```
CharlelieLrt: After rebasing, you will see that the `stochastic_sampler` does not directly use `patching.apply` and `patching.fuse`, but some wrappers around those. Those are used for compilation purposes, so it would be good to follow the same workflow in the `deterministic_sampler`

physicsnemo/utils/diffusion/deterministic_sampler.py
```diff
@@ -213,6 +260,22 @@ def deterministic_sampler(
...
+
+        # Function to select the correct positional embedding for each patch
+        def patch_embedding_selector(emb):
+            # emb: (N_pe, image_shape_y, image_shape_x)
+            # return: (batch_size * patch_num, N_pe, patch_shape_y, patch_shape_x)
+            return patching.apply(emb[None].expand(batch_size, -1, -1, -1))
```
CharlelieLrt, line 331: I know that you adapted the code from the `stochastic_sampler`, but could you double-check that this `patch_embedding_selector` works for lead-time aware models like the `SongUNetPosLtEmdb`? I think there is an additional argument to the embedding selector in the lead-time aware case. If it _does not_ work, a fix in `stochastic_sampler` is also going to be required.

I also strongly suggest adding a CI test for `deterministic_sampler` used with `SongUNetPosLtEmdb`, if there is not one already...

physicsnemo/utils/diffusion/deterministic_sampler.py
```diff
@@ -295,20 +365,39 @@ def deterministic_sampler(
...
-            denoised = net(x_hat / s(t_hat), x_lr, sigma(t_hat), class_labels).to(
-                torch.float64
-            )
+            denoised = net(
+                x_hat_batch / s(t_hat),
+                x_lr,
+                sigma(t_hat),
+                class_labels,
+                **optional_args,
+            ).to(torch.float64)
```
CharlelieLrt: If the `.to(torch.float64)` is not absolutely necessary, please remove it (for this line but also others)

physicsnemo/utils/diffusion/deterministic_sampler.py
```diff
@@ -319,18 +408,34 @@ def deterministic_sampler(
...
             if isinstance(net, EDMPrecond):
                 # Conditioning info is passed as keyword arg
                 denoised = net(
-                    x_prime / s(t_prime),
+                    x_prime_batch / s(t_prime),
                     sigma(t_prime),
                     condition=x_lr,
                     class_labels=class_labels,
+                    **optional_args,
                 ).to(torch.float64)
```
CharlelieLrt: Same comment, remove the `.to(torch.float64)` if not absolutely necessary

test/utils/generative/test_deterministic_sampler.py
```diff
@@ -224,3 +224,146 @@ def round_sigma(self, sigma):
...
+        self,
+        x,
+        x_lr,
+        t,
+        class_labels,
+        global_index=None,
+        embedding_selector=None,
+    ) -> torch.Tensor:
+        # Mock behavior: return input tensor for testing purposes
+        return x * 0.9
```
CharlelieLrt: IMO these tests with some mock networks are not good enough to really test things. The minimum we should have in terms of test:
- In `test_generation_steps.py`: add cases with the deterministic sampler + patching to `test_diffusion_step_rectangle` (should also cover `mean_hr` conditioning). Note that these tests use an actual network instead of a mock network.
- In `test_deterministic_sampler.py`: duplicate and adapt tests from `test_stochastic_sampler.py` (these two: `test_stochastic_sampler_patching_differentiable` and `test_stochastic_sampler_rectangle_patching`).

In addition of those, I would strongly suggest to add a test for `deterministic_sampler` + patching + an actual `SongUNetPosLtEmbd` (see [these helper functions](https://github.com/NVIDIA/physicsnemo/blob/632f16e8c1f2029440274716d33924dd8f48b9bc/test/models/diffusion/test_song_unet_pos_lt_embd.py#L31-L80) to setup a small model and test data)... because I suspect a possible bug in the current implementation of the deterministic sampler with lead-time aware models (to double-check).

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Added new `vortex_shedding_mgn` example that uses PyTorch Geometric instead.
 - HEALPixLayer can now use earth2grid HEALPix padding ops, if desired
 - Migrated Vortex Shedding Reduced Mesh example to PyTorch Geometric.
+- Patch-based inference support with deterministic sampler
```
CharlelieLrt: Need more specific context, eg: `- Diffusion utils: patch-based inference support with deterministic sampler.`

Also, your PR is enabling lead-time aware models with the `deterministic_sampler`, correct? If so, please add it here.

CharlelieLrt (changes requested):
Overall looks good to me. A few things to improve:
- Better tests that use actual networks rather than mock networks. Also need to test patching + lead-time + deterministic sampler.
- A possible bug in the embedding selector defined in the deterministic (and the stochastic) samplers --> to double-check
- A few details in the docs that will not render properly and that need to be fixed.

examples/weather/corrdiff/generate.py
```diff
@@ -169,15 +169,12 @@ def main(cfg: DictConfig) -> None:
 
     # Partially instantiate the sampler based on the configs
     if cfg.sampler.type == "deterministic":
-        if cfg.generation.hr_mean_conditioning:
-            raise NotImplementedError(
-                "High-res mean conditioning is not yet implemented for the deterministic sampler"
-            )
```
CharlelieLrt, line 190: Could you also double-check all the parameters in the config files related to the deterministic sampler and make sure that they are up-to-date with recommended values. If they're not, please update them.

physicsnemo/utils/diffusion/deterministic_sampler.py
tge25: lead time embedding is concatenated to the regular positional embedding before `patch_embedding_selector` is applied, and this is handled in `positional_embedding_selector` in network, so it seems `patch_embedding_selector` does not need additional argument

CHANGELOG.md
tge25: edited

test/utils/generative/test_deterministic_sampler.py
tge25: unit tests with the real preconditioner and model have been added to test_deterministic_sampler

physicsnemo/utils/diffusion/deterministic_sampler.py
tge25: I just re-did copying and pasting from the stochastic sampler. Hopefully it solves the formating issue

physicsnemo/utils/diffusion/deterministic_sampler.py
tge25: I just re-did copying and pasting from the stochastic sampler. Hopefully it solves the formating issue

examples/weather/corrdiff/generate.py
tge25: Confirm that new patch-based deterministic sampler works with patched.yaml

physicsnemo/utils/diffusion/deterministic_sampler.py
CharlelieLrt: Right, thanks for the explanation!

physicsnemo/utils/diffusion/deterministic_sampler.py
```diff
@@ -319,18 +418,34 @@ def deterministic_sampler(
         if solver == "euler" or i == num_steps - 1:
             x_next = x_hat + h * d_cur
         else:
+            # Patched input
+            # (batch_size * patch_num, C_out, patch_shape_y, patch_shape_x)
+            x_prime_batch = (patching.apply(input=x_prime) if patching else x_prime).to(
```
CharlelieLrt: I think we need to use the compiled wrapper instead of `patching.apply` here (the one for inputs with Cout channels I believe)

physicsnemo/utils/diffusion/deterministic_sampler.py
```diff
@@ -319,18 +418,34 @@ def deterministic_sampler(
...
+                    x_lr,
+                    sigma(t_prime),
+                    class_labels,
+                    **optional_args,
+                )
+
+            if patching:
+                # Un-patch the denoised image
+                # (batch_size, C_out, img_shape_y, img_shape_x)
+                denoised = patching.fuse(input=denoised, batch_size=batch_size)
```
CharlelieLrt: I think we need to use the fuse wrapper instead of `patching.fuse`

physicsnemo/utils/diffusion/deterministic_sampler.py
```diff
@@ -295,20 +372,42 @@ def deterministic_sampler(
...
+                x_lr,
+                sigma(t_hat),
+                class_labels,
+                **optional_args,
             )
+
+        if patching:
+            # Un-patch the denoised image
+            # (batch_size, C_out, img_shape_y, img_shape_x)
+            denoised = patching.fuse(input=denoised, batch_size=batch_size)
```
CharlelieLrt: I think we need to use the fuse wrapper instead of `patching.fuse`

physicsnemo/utils/diffusion/deterministic_sampler.py
CharlelieLrt: @tge25 I unresolved this comment, as I found a few more lines where the compiled wrappers could be used. I marked these lines in some other comments below

physicsnemo/utils/diffusion/deterministic_sampler.py
CharlelieLrt: @tge25 I unresolved this comment as I found two other lines where `float64` is still used:
- `step_indices = torch.arange(num_steps, dtype=torch.float64, device=latents.device)`
- `u = torch.zeros(M + 1, dtype=torch.float64, device=latents.device)`
I know that these lines are not directly related to your code changes, but I just want to make sure that removing `float64` in some places and keeping it in these lines will not cause any issue. Ideally, and if lower precision is not a concern, we would like to be consistent and have everything in `float32`. I'll leave it up to you to decide what's best to do.

test/utils/generative/test_deterministic_sampler.py
```diff
@@ -224,3 +226,182 @@ def round_sigma(self, sigma):
...
+        img_lr=img_lr,
+        patching=None,
+        mean_hr=mean_hr,
+        num_steps=2,
+        lead_time_label=torch.tensor([1]),
+    )
+
+    assert (
+        result_mean_hr.shape == latents.shape
+    ), "Mean HR conditioned output shape does not match expected shape"
```
CharlelieLrt: I'm not sure what is this test for? Unless I am missing something, it seems everything it does is already covered by the following test?
Also size 448 is too large (we try to keep the test runtime as low as possible)

test/utils/generative/test_deterministic_sampler.py
```diff
@@ -224,3 +226,182 @@ def round_sigma(self, sigma):
...
+        channel_mult=[1, 2, 2],
+        channel_mult_emb=2,
+        num_blocks=2,
+        attn_resolutions=[attn_res],
+        gridtype="learnable",
+        lead_time_mode=True,
+        N_grid_channels=N_pos,
+        lead_time_steps=lt_steps,
+        lead_time_channels=lt_channels,
+    )
```
CharlelieLrt: This is creating a preconditioner that wraps a `SongUNetPosEmbd`, _**and not**_ a `SongUNetPosLtEmbd`. The right way to create a lead-time aware model is to use a `SongUNetPosLtEmbd` ( in the preconditioner wrapper, this can be done by passing `model_type=SongUNetPosLtEmbd` and removing the `lead_time_mode` argument).

I know that these APIs are equivalent under their current implementation, but if there is any AAPI change to the `SongUNetPosLtEmbd`, then those will not propagate to this test, which would cause some issues.

test/utils/generative/test_deterministic_sampler.py
```diff
@@ -224,3 +226,182 @@ def round_sigma(self, sigma):
...
+        result_mean_hr.shape == latents.shape
+    ), "Mean HR conditioned output shape does not match expected shape"
+
+
+# The test function for edm_sampler with rectangular domain and patching
+@import_or_fail("cftime")
+def test_deterministic_sampler_rectangle_patching(pytestconfig):
+    from physicsnemo.utils.generative import deterministic_sampler
+    from physicsnemo.utils.patching import GridPatching2D
+
```
CharlelieLrt, line 294: This test needs a `torch._dynamo.reset()` here, because the `deterministic_sampler` is using `torch.compile` and it might fail because of graph breaks in the CI pipeline

test/utils/generative/test_deterministic_sampler.py
```diff
@@ -224,3 +226,182 @@ def round_sigma(self, sigma):
...
+    patching = GridPatching2D(
+        img_shape=(img_shape_y, img_shape_x),
+        patch_shape=(patch_shape_y, patch_shape_x),
+        overlap_pix=4,
+        boundary_pix=2,
+    )
+    net.to(device)
+    # Test with mean_hr conditioning
+    mean_hr = torch.randn(2, 3, img_shape_y, img_shape_x, device=device)
+    result_mean_hr = stochastic_sampler(
```
CharlelieLrt: This is `stochastic_sampler`, not `deterministic_sampler`

test/utils/generative/test_deterministic_sampler.py
```diff
@@ -224,3 +226,182 @@ def round_sigma(self, sigma):
...
+    )
+
+    assert (
+        result_mean_hr.shape == latents.shape
+    ), "Mean HR conditioned output shape does not match expected shape"
+
+
+# The test function for edm_sampler with rectangular domain and patching
+@import_or_fail("cftime")
+def test_deterministic_sampler_rectangle_patching(pytestconfig):
```
CharlelieLrt: This test needs a be parametrized with a `device` parameter and it needs to create tensors and models on the appropriate `device`

test/utils/generative/test_deterministic_sampler.py
```diff
@@ -224,3 +226,182 @@ def round_sigma(self, sigma):
...
+    assert (
+        result_mean_hr.shape == latents.shape
+    ), "Mean HR conditioned output shape does not match expected shape"
+
+
+# Test that the stochastic sampler is differentiable with rectangular patching
+# (tests differentiation through the patching and fusing)
+@import_or_fail("cftime")
+@pytest.mark.parametrize("device", ["cuda:0", "cpu"])
+def test_stochastic_sampler_patching_differentiable(device, pytestconfig):
```
CharlelieLrt: Not "stochastic", but "deterministic"

test/utils/generative/test_stochastic_sampler.py
```diff
@@ -257,3 +257,49 @@ def __call__(
...
+    result = stochastic_sampler(
+        net=net,
+        latents=latents,
+        img_lr=img_lr,
+        patching=None,
+        mean_hr=torch.ones_like(img_lr),
+        lead_time_label=[0],
+    )
+
+    assert result.shape == latents.shape, "Output shape does not match expected shape"
```
CharlelieLrt: Same question as before: I am not sure I understand what is this test covering that isn't covered in other tests?

test/utils/generative/test_deterministic_sampler.py
CharlelieLrt: @tge25  thanks for adding these tests! I added a few more questions/comments about the tests that you added.

CharlelieLrt (changes requested):
Thanks for addressing my previous comments!
I spotted a few more details in `deterministic_sampler` (missing calls to compiled wrappers, ...) as well as a couple of issues in the new tests that you added.

test/utils/generative/test_stochastic_sampler.py
tge25: That was used to test lead time embedding args in the stochastic sampler. This is to ensure that lead time label can be passed to the network without patching operator

test/utils/generative/test_deterministic_sampler.py
tge25: This test is for testing full-domain determinsitic sampler with lead time embedding arg to ensure that the full-domain sampler still works. I have changed the test function name to better reflect that. The size is changed to (64,64).

test/utils/generative/test_deterministic_sampler.py
tge25: Thanks. I recalled there was a discussion about whether we should maintain multiple variants of SongUNet or consolidate into a single version that handles all use cases. I used SongUNetPosEmbd here as part of the effort to unify the network, with the goal of potentially deprecating SongUNetPosLtEmbd in the future.
It sounds like we'd still like to keep SongUNetPosEmbd. In that case I have changed it to SongUNetPosEmbd.

physicsnemo/utils/diffusion/deterministic_sampler.py
tge25: Thanks. I think I've resolved all of them

physicsnemo/utils/diffusion/deterministic_sampler.py
tge25: torch.float64 has been changed to torch.float32 for consistancy

physicsnemo/utils/diffusion/deterministic_sampler.py
tge25: Changed. Thanks for pointing it out

test/utils/generative/test_deterministic_sampler.py
tge25: device arg has been add to test functions

test/utils/generative/test_deterministic_sampler.py
tge25: torch._dynamo.reset() has been added

test/utils/generative/test_deterministic_sampler.py
CharlelieLrt: > I recalled there was a discussion about whether we should maintain multiple variants of SongUNet or consolidate into a single version

That's correct, there is ongoing work to consolidate multiple variants into a single class, but _not all_ of them.
To give you more context:
- positional embeddings is general purpose in the sense that it can be beneficial for many applications and domains, so it makes sense to merge the SongUNetPosEmb with the base `SongUNet`.
- lead-time embeddings is something useful for weather applications, but not in other domains, so there is no plan to merge `SongUNetPosLtEmbd` with the base `SongUNet`. So, in short, we plan to keep a separate `SongUNetPosLtEmbd` class, that inherits most of its logic from a base `SongUNet` and only adds the lead-time embeddings on top of it.

Thank you for addressing my comment nontheless.

test/utils/generative/test_deterministic_sampler.py
```diff
@@ -224,3 +226,180 @@ def round_sigma(self, sigma):
...
+    )
+    return model
+
+
+# The test function for patch-based deterministic_sampler
+@import_or_fail("cftime")
+@pytest.mark.parametrize("device", ["cuda:0", "cpu"])
+def test_deterministic_sampler_full_domain_lead_time(device, pytestconfig):
+
+    from physicsnemo.utils.generative import deterministic_sampler
```
CharlelieLrt: @tge25 `physicsnemo.utils.generative` is deprecated, is has been renamed into `physicsnemo.utils.generative`. There are a couple of other places where you'll need to change it

tge25: Thanks! They have been changed

PR #973, opened 2025-06-13
Changes to support DoMINO Design Sensitivities work + DoMINO model code fixes

peterdsharpe:
# PhysicsNeMo Pull Request

## Checklist

### DIFF
--- a/examples/cfd/external_aerodynamics/domino_sensitivity/__init__.py
+++ b/examples/cfd/external_aerodynamics/domino_sensitivity/__init__.py
@@ -1,0 +1,15 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2024 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.


physicsnemo/utils/domino/utils.py
```diff
@@ -376,15 +374,6 @@ def combine_dict(old_dict, new_dict):
...
 
-def merge(*lists):
-    """Function to merge lists"""
-    newlist = lists[:]
-    for x in lists:
-        if x not in newlist:
-            newlist.extend(x)
-    return newlist
-
-
```
peterdsharpe: Review note: this PR deletes `merge()`; this function is (a) not used anywhere, (b) can and should be replaced simply by the built-in `sum(lists, start=[])`, and (c) as-written will always create an error, since `newlist` is a tuple and hence does not have a `.extend()` method.

physicsnemo/utils/domino/utils.py
```diff
@@ -18,7 +18,7 @@
 Important utilities for data processing and training, testing DoMINO.
 """
 
-import os
```
peterdsharpe: Note for future work: in general, `pathlib` is preferred over `os.path`, since as far back as the Python 3.4 days:
- https://medium.com/@barila/choose-your-path-pythons-pathlib-vs-os-path-4de0b1e752dd
- https://www.reddit.com/r/learnpython/comments/i29h6j/when_or_why_should_i_use_pathlibpath_over_ospath/

peterdsharpe:
Review note: this introduces changes to the activation functions of the DoMINO model, and hence, existing DoMINO checkpoints will need to be retrained.

Last remaining to-do to pass CI: update the cached DoMINO checkpoint at `./test/models/data/domino_output.pth` so that it can be loaded and validated at test-time. Going to hold off on this until review comments come back, in case other changes need to be made before generating this.

examples/cfd/external_aerodynamics/domino/src/conf/config.yaml
```diff
@@ -38,9 +38,9 @@ resume_dir: ${output}/models
 # └───────────────────────────────────────────┘  
 data_processor: # Data processor configurable parameters
   kind: drivaer_aws # must be either drivesim or drivaer_aws
-  output_dir: /lustre/rranade/aws_data_all/
-  input_dir: /lustre/datasets/drivaer_aws/drivaer_data_full/
-  cached_dir: /lustre/cached/drivaer_aws/drivaer_data_full/
+  output_dir: /user/aws_data_all/
+  input_dir: /data/drivaer_aws/drivaer_data_full/
+  cached_dir: /user/cached/drivaer_aws/drivaer_data_full/
```
coreyjadams, line 43: We already declare both `project_dir` and `output` in this config.  Is it possible to reuse those with `${output}`?  I don't want to leave a config that requires many manual changes, it'd be nice to format all these paths as sensible relative paths.

examples/cfd/external_aerodynamics/domino_sensitivity/conf/config.yaml
coreyjadams, line 1: So much of this is copy/paste from the original config.  If it's possible to avoid, can we extend that config rather than duplicate it?  It would make maintenance simpler, and users who train a  model with custom hyperparameters might find a single config surface easier to manage.

examples/cfd/external_aerodynamics/domino_sensitivity/design_datapipe.py
coreyjadams, line 1: We did a decent amount of work in the last release to accelerate the data pipeline for DoMINO.  I see a number of pieces here using CPU-only methods.  If we have time to do it before merge, we should isolate the pieces from the training datapipe to reuse them here on CPU/GPU agnostic codes.

Really, I'm staring at the KDTree.  That's a CPU op but `cuml` has a faster implementation on GPU.

examples/cfd/external_aerodynamics/domino_sensitivity/gradient_checking_results/drag_gradients_raw.txt
coreyjadams, line 1: I checked your math, might have a decimal error on line 62 column 17?

Just kidding.  But - are these meant to be for unit testing and math checks?   Might be better placed in the test area if so?

examples/cfd/external_aerodynamics/domino/src/conf/config.yaml
peterdsharpe: This is a good consideration! (Actually, I hadn't realized that you could use `${}`-variables in YAML, so I learned something new today!)

Let me investigate what I can do here and get back to you on this one.

examples/cfd/external_aerodynamics/domino_sensitivity/design_datapipe.py
coreyjadams, line 1: Also - the way it's implemented here should be pretty easy to do with cupy, too.  Worth doing?

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
```diff
@@ -0,0 +1,491 @@
...
+                param.requires_grad = False
+
+            self.model = torch.compile(self.model, disable=True)  # TODO review
+
+            if self.model_checkpoint_path is not None:
+                with open(self.model_checkpoint_path, "rb") as f:
+                    self.model.load_state_dict(torch.load(f, map_location=self.device))
+                print("Model loaded with checkpoint...")
+            else:
+                print("Model loaded without checkpoint...")
```
coreyjadams: Should there perhaps be something more aggressive here?  Running inference without checkpoints doesn't make any sense unless you're just unit testing.  This could raise an exception, unless some unit testing flag / env variable is set?

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
```diff
@@ -0,0 +1,491 @@
...
+                        0.00381244,
+                    ],
+                    [
+                        -1.9028450e00,
+                        -1.0032533e00,
+                        -1.0505041e00,
+                        -1.4412953e00,
+                        1.5563720e-18,
+                        -2.7427445e-20,
+                    ],
```
coreyjadams, line 184: You say in the comment that this is meant to be saved at training time and later used to unnormalize predictions.  Should this instead be loading from a file?  Hardcoding the numbers seems like it will lead to numerical errors as soon as someone uses a different dataset.

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
```diff
@@ -0,0 +1,491 @@
...
+        These are saved at training time based on statistics of the training data, and re-used at each inference call.
+        """
+        return torch.from_numpy(
+            np.array(
+                [
+                    [0.98881036, 0.00550783, 0.00854675, 0.00452144],
+                    [-2.4203062, -0.00740275, -0.00848471, -0.00448634],
+                ],
+                dtype=np.float32,
+            )
```
coreyjadams, line 204: Same comment as for the volume factors.

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
```diff
@@ -0,0 +1,491 @@
...
+            ...     mesh=mesh,
+            ...     stream_velocity=30.0,
+            ...     stencil_size=7,
+            ...     air_density=1.205
+            ... )
+            >>>
+            >>> # Access results
+            >>> forces = results['aerodynamic_force']
+            >>> print(f"Drag force: {forces[0]:.2f} N")
+        """
```
coreyjadams: First, I really like this interface: pass in the mesh, and this takes care of loading the model, running inference, and unpacking the results back to sensible numbers.

The docstring is also good.  Minor nit: as presented, it isn't clear if can pass a second mesh.  Can you?  Or is that not the intended workflow?

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
```diff
@@ -0,0 +1,491 @@
...
+            bounding_box_surface=self.bounding_box_surface_min_max,
+            grid_resolution=self.cfg.model.interp_res,
+            stencil_size=stencil_size,
+        )
+        dataloader = torch.utils.data.DataLoader(
+            datapipe, batch_size=2**13, shuffle=False
+        )
+
+        input_dict = {
+            k: torch.from_numpy(np.expand_dims(np.float32(v), axis=0)).to(self.device)
```
coreyjadams: Nit: Consider doing the conversion to torch + transfer to the device in the datapipe.  Would enable numpy/cupy agnostic loading more easily.

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
```diff
@@ -0,0 +1,491 @@
...
+        geometry_coordinates = (
+            input_dict["geometry_coordinates"].detach().cpu().numpy()[0]
+        )
+        geometry_sensitivity: np.ndarray = np.zeros_like(geometry_coordinates)
+
+        def _print_memory_usage(label: str) -> None:
+            return
+            print(
+                f"VRAM usage after {label:<20}: {(torch.cuda.memory_allocated()/(1024**3)):.2f} GB"
+            )
```
coreyjadams: Dead code?

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
```diff
@@ -0,0 +1,491 @@
...
+            input_dict_batch = {
+                **input_dict,
+                **{
+                    k: torch.unsqueeze(sample_batched[k], dim=0).to(self.device)
+                    for k in surface_keys
+                },
+            }
+            input_dict_batch["geometry_coordinates"].requires_grad_(True)
+
+            _print_memory_usage("data loading")
```
coreyjadams: Should be trimmed?

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
```diff
@@ -0,0 +1,491 @@
...
+
+            prediction_vol_batch, prediction_surf_batch = self.model(input_dict_batch)
+
+            _print_memory_usage("model forward")
```
coreyjadams: And here

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
```diff
@@ -0,0 +1,491 @@
...
+
+            # Required to free memory
+            del prediction_vol_batch
```
coreyjadams: It's unusual to have to manually call `del`.  I know the reason this is here, but it might help to expand the comment for future maintenance?

examples/cfd/external_aerodynamics/domino_sensitivity/main_gradient_checking.py
coreyjadams, line 1: Should this be a unit test?

examples/cfd/external_aerodynamics/domino_sensitivity/requirements.txt
```diff
@@ -0,0 +1,4 @@
+git+https://github.com/NVIDIA/apex.git
+hydra-core
+cupy-cuda12x
+numba
```
coreyjadams, line 4: nit: what brings the numba requirement?

examples/cfd/external_aerodynamics/domino_sensitivity/conf/cached.yaml
coreyjadams, line 1: Is this different from the main domino cached config?  If not, can we reuse the original source instead of duplicating?

examples/cfd/external_aerodynamics/domino_sensitivity/utilities/mesh_postprocessing.py
```diff
@@ -0,0 +1,410 @@
...
+# limitations under the License.
+
+import pyvista as pv
+import numpy as np
+from tqdm import tqdm
+import numba
+from typing import Literal
+
+
+@numba.njit
```
coreyjadams, line 24: OK well this answers my numba question!

examples/cfd/external_aerodynamics/domino_sensitivity/utilities/mesh_postprocessing.py
coreyjadams, line 1: This file appears to have some good stuff.  I don't want it to get buried and reimplemented later, and anyways it looks like it could use some tests implemented too.

Could this get reorganized into, say, `physicsnemo.utils.mesh` or similar?

physicsnemo/models/domino/model.py
coreyjadams, line 1: I think wholesale switching to `gelu` makes sense for your design sensitivities work but it's going to be a bumpy shift for everyone who has a pretrained or fine tuned model.  Can I propose, instead, the activation is configurable at the model level?  Default to relu (for backward compatibility) and set it to gelu in your configs.  We can target a pivot from relu -> gelu after the next trained model is released, presumably with `gelu`?

coreyjadams (changes requested):
I think this is a great example use of DoMINO and a really cool extension to it's utility.  As written, I think it's great, but it will be easier to maintain and extend with a few changes.  I know you have another PR open on the utils cleanup, so I won't focus there too much. My main thoughts are:
- The mesh utility functions should live somewhere they can get unit tested and used more broadly.
- The config files have a lot of repetition (from the original configs) that could be cleaned up and consolidated
- `gelu` -> `relu` is a good idea but we should do it more cautiously so as not to break the model for existing users.
- The second half of the README is mostly just a list of bullets.  Can we elaborate on those where they are useful?  And a few are repetitive, perhaps consolidate?
- The datapipe uses scipy KDTree which is good, but we can also use cuml and do it on the GPUs.

examples/cfd/external_aerodynamics/domino_sensitivity/requirements.txt
peterdsharpe: Answered: https://github.com/NVIDIA/physicsnemo/pull/973/files/fa462297600b63d1ed04c209ba0d95f1a35df475#diff-36d58f93d8b670465ba77a5c26a1d7faa09e155c637dc0ea9ef8e1af0b501cb6

examples/cfd/external_aerodynamics/domino_sensitivity/conf/cached.yaml
peterdsharpe: In this case the cached config is actually not used, so I can go ahead and delete this - good call!

peterdsharpe: Finished: https://github.com/NVIDIA/physicsnemo/pull/973/commits/900427bb74467b35568a0a534d91f92c55e09d44

examples/cfd/external_aerodynamics/domino_sensitivity/conf/config.yaml
peterdsharpe: Good call, fixed (now imports from DoMINO config): https://github.com/NVIDIA/physicsnemo/pull/973/commits/900427bb74467b35568a0a534d91f92c55e09d44

physicsnemo/models/domino/model.py
peterdsharpe: This is a valid concern!

I agree that there may be growing pains here, and I actually think it's symptomatic of a much larger problem that goes beyond this instance. Namely, the fundamental issue is that saved checkpoints currently do not encode much info about the model architecture, so we're loading fixed checkpoints (via saved files) into a moving-target architecture (via the repository).
- On the GeLU/ReLU issue: I agree that making activations configurable for DoMINO is a workable fix for this instance of the issue. I'm open to doing this, though I think it's a bit of a band-aid fix to the larger true issue.
- For the larger issue, it's an open question whether/how we tackle it:
   - One possible fix would be if, when the user saves a model checkpoint, a hash of the PhysicsNeMo repo is also saved (most convenient could be the latest Git commit hash). This has some implementation issues (what if the user isn't using Git?), but something conceptually similar would be nice. This way, when the checkpoint is loaded, we can throw a warning if the architecture is different than when it was saved. (PyTorch will already throw this for some model changes, but not for all changes.)
   - I think something like ONNX is not a great option here, since some of our models end up using a lot of bespoke code that doesn't fit nicely into ONNX primitives (e.g., anything with Warp)
   - We could just leave this as-is and expect users to pin commit hashes / versions for serious deployments. In an ideal world this should be happening anyway, but I'm not sure how reliable this solution is "in the wild".

More long-term: when it comes to new model architectures, I would recommend defaulting to $C^\infty$-continuous modeling where possible - at least for problems where the underlying PDE is expected to be smooth. This shows up most notably in activation functions, but also in more subtle places like KNN kernels (depending on how the results are then handled). Reworking how we handle KNN-based layers to be higher-order continuous is a challenge, but for the low-hanging fruit (like activations), I think encouraging continuity would be a good default policy.
- This is because many applications benefit from this continuity property: for example, @ktangsali 's current work on adding PDE-residual-based physics loss to DoMINO also fundamentally requires meaningful spatial gradients, which any architecture that passes spatial information through ReLU cannot provide via AD (this is explicitly true for derivatives of order 2+, but even first-derivative accuracy will suffer via loss of smoothness).

peterdsharpe: My personal preference would be as follows:
* Make GeLU the default for now, since higher-order continuity is required for multiple downstream applications and we expect the true solution of the RANS PDE to exhibit at least $C^2$-continuity
* Not add activation as an exposed configuration option, trusting that users should be pinning their PhysicsNeMo repo versions when using a fixed saved checkpoint.
* Choose whether/how to address the fundamental issue of separation-of-checkpoints-and-architectures later. 

However, I'm definitely open to being convinced - can other reviewers weigh in here? @ktangsali @RishikeshRanade

examples/cfd/external_aerodynamics/domino_sensitivity/utilities/mesh_postprocessing.py
peterdsharpe: Sure! It's actually based on some utilities that @ktangsali and I were working on in PhysicsNeMo-CFD, which I had included here since I needed our neighbor-finding algorithm to implement the Laplacian smoothing kernel, and I couldn't depend on end-users of this specific example having PhysicsNeMo-CFD installed.

Perhaps one alternative strategy (rather than including these utilities here) would be to have this example list PN-CFD in its `requirements.txt` - but I'd really like to keep the dependency direction one-way: PhysicsNeMo-CFD depends on PhysicsNeMo, but not vice versa. It gets super messy when you have a bidirectional dependency and need to enforce mutual version pins. So, I don't love that alternative.

I do think this would be a good thing to upstream into something like `physicsnemo.utils.mesh`, since it's just generally useful and not CFD-specific - but I'd like to run it by @ktangsali first.

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
peterdsharpe: I see what you mean, though I'd rather raise a warning than an exception! 

There are some (though few) valid reasons to run inference without a checkpoint - in addition to unit testing, the end user might be monkey-patching weights/biases in from some other source. I also think it's good policy to allow dataclasses to be instantiated with as little validation as possible, since they're mutable - it's hard to anticipate all the ways an end-user might hack at it.

I'll implement a warning!

peterdsharpe: Done: https://github.com/NVIDIA/physicsnemo/pull/973/commits/085077fb30d4ae70343c7023cf92fa1b9ec9516a

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
peterdsharpe: Thank you! Yeah, I think callable dataclasses are criminally underutilized for many situations that involve "a function with some data attached" - thank [Equinox](https://github.com/patrick-kidger/equinox/tree/main) for showing the full power of that interface!

Re:

>it isn't clear if can pass a second mesh. Can you? Or is that not the intended workflow?

I'm not sure I fully understand the question, but if the question is whether you can call the function multiple times, then the answer is yes! Something like:

```python
domino = DoMINOInference(...)
result_a = domino(mesh=mesh_a, ...)
result_b = domino(mesh=mesh_b, ...)
```

will work fine, all without re-loading the checkpoint. Actually, `main_gradient_checking.py` gives an example of this.

I'm not sure if this needs to be explicitly stated in the `__call__` docstring, but if you think it does, we definitely can do so!

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
peterdsharpe: Good catch, sorry! Yeah, I was using this for debugging. I think I'll refactor the dataclass to accept a `verbose` boolean flag to guard this (and other prints) behind.

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
peterdsharpe: Agreed, super unusual. I can make this comment clearer!

physicsnemo/models/domino/model.py
```diff
@@ -187,7 +187,7 @@ def __init__(
 
         self.grid_resolution = grid_resolution
 
-        self.activation = F.relu
+        self.activation = F.gelu
```
RishikeshRanade: Is it possible to provide this as an option in config?

examples/cfd/external_aerodynamics/domino/src/conf/config.yaml
RishikeshRanade: Sometimes users save data and project files on different mounts and having this flexibility helps. The project dir corresponds to the path where all the meta data generated from training and testing is saved but the data might not necessarily reside here.

examples/cfd/external_aerodynamics/domino_sensitivity/design_datapipe.py
```diff
@@ -0,0 +1,236 @@
...
+            use_sign_winding_number=True,
+        )
+        surf_sdf_grid = np.array(surf_sdf_grid).reshape(nx, ny, nz)
+
+        # Sample surface_vertices
+        grid = normalize(grid, v_max, v_min)
+        s_grid = normalize(s_grid, s_max, s_min)
+
+        surface_coordinates = stl_centers
+        interp_func = KDTree(surface_coordinates)
```
RishikeshRanade: Should we use cuml here (or provide an option) similar to training datapipe?

examples/cfd/external_aerodynamics/domino_sensitivity/design_datapipe.py
```diff
@@ -0,0 +1,236 @@
...
+            ValueError: If bounding_box or bounding_box_surface are not 2x3 arrays
+        """
+        if len(grid_resolution) != 3:
+            raise ValueError("grid_resolution must contain exactly 3 values")
+
+        self.mesh = mesh
+        rng = np.random.RandomState(seed)
+
+        # Compute basic mesh properties
+        length_scale = np.amax(self.mesh.points, 0) - np.amin(self.mesh.points, 0)
```
RishikeshRanade: Can we wrap this preprocessing code (lines 89 to 191) as a function call?

examples/cfd/external_aerodynamics/domino_sensitivity/design_datapipe.py
```diff
@@ -0,0 +1,236 @@
...
+        surface_neighbors_normals = surface_neighbors_normals[:, 1:]
+        surface_neighbors_area = surface_areas[ii]
+        surface_neighbors_area = surface_neighbors_area[:, 1:]
+
+        pos_normals_com_surface = surface_coordinates - center_of_mass
+
+        surface_coordinates = normalize(surface_coordinates, s_max, s_min)
+        surface_neighbors = normalize(surface_neighbors, s_max, s_min)
+
+        # Volume processing
```
RishikeshRanade, line 154: In the scenario that the user has trained a surface only model the volume processing is not required. Can we use the model.model_type flag from config.yaml to omit the volume calculation in this scenario?

RishikeshRanade:
This looks great, Peter. Just a general comment: As this is an inference workflow specific to cfd at the moment, we should have this example in physicsnemo-cfd instead of physicsnemo. What do you think?

physicsnemo/models/domino/model.py
peterdsharpe: A good suggestion! Linking to this open comment thread so we can resolve it there:

https://github.com/NVIDIA/physicsnemo/pull/973#discussion_r2167530404

tl;dr: my preference would be to default to GeLU (or anything else $C^\infty$-continuous) for theory reasons, but I'm open to making this configurable if the consensus is that this would be better.

examples/cfd/external_aerodynamics/domino_sensitivity/design_datapipe.py
peterdsharpe: Good call - yeah, that should work and is a free speedup we should take. I'll implement it!

examples/cfd/external_aerodynamics/domino/src/conf/config.yaml
peterdsharpe: ^ Fair take as well. I think I'll just leave it as-is -- after all, this path structure is just an example for end-users, so it's expected that they'll need to change it anyway. While we could simplify it with `${}` expressions, the current bare path might be more readable and expressive for end users.

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
peterdsharpe: Fixed: 7d0eecdd8fcc98125a73db0d31da8a05468ad7cd

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
peterdsharpe: Fixed: 7d0eecdd8fcc98125a73db0d31da8a05468ad7cd

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
peterdsharpe: Fixed: 7d0eecdd8fcc98125a73db0d31da8a05468ad7cd

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
peterdsharpe: Fixed: 7d0eecdd8fcc98125a73db0d31da8a05468ad7cd

examples/cfd/external_aerodynamics/domino_sensitivity/design_datapipe.py
peterdsharpe: Is there an instance where we would not want to perform the data preprocessing at the time of DesignDatapipe initialization?

examples/cfd/external_aerodynamics/domino_sensitivity/design_datapipe.py
peterdsharpe: Fixed: fc1b48b9dc9f181e39df41f68aad0f48be11daae

peterdsharpe: Fixed: fc1b48b9dc9f181e39df41f68aad0f48be11daae

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
peterdsharpe: Good call, thanks! Fixed: f79d2698c9fe4ac8fd9123b3aac76422d7519f6a

examples/cfd/external_aerodynamics/domino_sensitivity/main_gradient_checking.py
peterdsharpe: I don't think so, especially considering that for DoMINO, the finite-diff gradient doesn't actually converge to the AD-derived one due to model nonsmoothness issues. This is really meant to be exposed to the end-user as an example diagnostic tool, not a unit test.

examples/cfd/external_aerodynamics/domino_sensitivity/design_datapipe.py
peterdsharpe: Done

examples/cfd/external_aerodynamics/domino_sensitivity/gradient_checking_results/drag_gradients_raw.txt
peterdsharpe: I don't think so, especially considering that for DoMINO, the finite-diff gradient doesn't actually converge to the AD-derived one due to model nonsmoothness issues. This is really meant to be exposed to the end-user as an example diagnostic tool, not a unit test.

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
peterdsharpe: This is copied from analogous code in the main DoMINO inference example:

https://github.com/NVIDIA/physicsnemo/blob/9cdfe789c5bd594f2a22cee735c41911167fe2c9/examples/cfd/external_aerodynamics/domino/src/inference_on_stl.py#L751

Given that this example is really a subset of that main example, I'll keep the logic as-is here until we can fix it there first in a later PR.

examples/cfd/external_aerodynamics/domino_sensitivity/main.py
peterdsharpe: This is copied from analogous code in the main DoMINO inference example:

https://github.com/NVIDIA/physicsnemo/blob/9cdfe789c5bd594f2a22cee735c41911167fe2c9/examples/cfd/external_aerodynamics/domino/src/inference_on_stl.py#L751

Given that this example is really a subset of that main example, I'll keep the logic as-is here until we can fix it there first in a later PR.

peterdsharpe:
Thanks @coreyjadams @RishikeshRanade for the reviews!

#### Re: Rishi's suggestion of including this in PhysicsNeMo-CFD rather than here:

This is a good suggestion, and I see the logic here. However, I think that I'd prefer to keep it targeted to this repo, since the underlying main motivation for presenting this example is not actually CFD-specific. More precisely, the goal isn't only to show geometry-level drag sensitivities for CFD problems, it's instead to show how you could compute geometry sensitivities for *any PDE* modeled with DoMINO (and to provide a template if users want to try this w/ other models). 

In addition, the close coupling between the DoMINO model architecture changes (e.g., GeLU) and the feasibility of this sensitivity study would make maintenance of this example difficult across two repos. For now, while architectures are still being heavily modified, examples that involve more intricate work might benefit from inclusion in this repo, even if they are inference-focused. @ktangsali thoughts?

physicsnemo/models/domino/model.py
peterdsharpe: Pinging @ktangsali @RishikeshRanade for their take on the GeLU/ReLU issue here

examples/cfd/external_aerodynamics/domino_sensitivity/design_datapipe.py
peterdsharpe: Actually, now profiling this, the volume pre-processing is around 100x-1000x faster than the actual DoMINO model inference, so I think it's safe to leave this as-is without introducing additional complexity.

physicsnemo/models/domino/model.py
peterdsharpe: Update: GeLU/ReLU now configurable. For now, the default is ReLU for back-compatibility; in the future, let's default to smooth activations.

peterdsharpe:
That could work too! I chose to implement it this way because:

- (Main reason) The individual constructors currently only have visibility into their respective sub-section of `model_parameters`, so implementing the configurable activations like this required fewer overall changes to `model.py`.
- This allows a bit more flexibility for switching activations in different subsections of the model. (Granted, I can't imagine a theory-backed reason why one would want to do this, but who knows. As we were discussing, if it were up to me, the activations would all be hard-coded to GeLU or similar)

Issue #898, opened 2025-05-19
🐛[BUG]: `Mesh_Reduced` model cannot be run because of missing `torch_cluster` dependency.


peterdsharpe:
### Version

Latest Github `main` branch (hash: 609e03e )

### On which installation method(s) does this occur?

Source

### Describe the issue

I'm getting test errors in part because the [model at `physicsnemo/models/mesh_reduced/mesh_reduced.py`](https://github.com/NVIDIA/physicsnemo/blob/main/physicsnemo/models/mesh_reduced/mesh_reduced.py) imports `torch_cluster`, which is not listed in `pyproject.toml` (under any group). Normally I'd fix this by adding torch_cluster to the TOML - but it looks like this model hasn't been modified in over a year, so I'm wondering if it's dead code and should be removed. Does anyone know if this model is used anywhere / have insight into the best path forward here?

Note that, while `torch_cluster` is currently installed within the pre-packaged container, it is not listed as a dependency in `pyproject.toml`, which means that source installations will not pass tests.

### Minimum reproducible example

```shell
git clone <physicsnemo>
cd physicsnemo
pip install -e .[all]

python ./physicsnemo/models/mesh_reduced/mesh_reduced.py

OR

pytest ./examples/cfd/vortex_shedding_mesh_reduced/test_sequence.py
```

### Relevant log output

```shell
ImportError while importing test module '/home/psharpe/GitHub/physicsnemo/examples/cfd/vortex_shedding_mesh_reduced/test_sequence.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../.local/share/uv/python/cpython-3.10.16-linux-x86_64-gnu/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
examples/cfd/vortex_shedding_mesh_reduced/test_sequence.py:33: in <module>
    from physicsnemo.models.mesh_reduced.mesh_reduced import Mesh_Reduced
physicsnemo/models/mesh_reduced/mesh_reduced.py:18: in <module>
    import torch_cluster
E   ModuleNotFoundError: No module named 'torch_cluster'
```

### Environment details

```shell
Python 3.10, Linux, installed from latest source with all dependencies via `pip install -e .[all]`
```

satologicbot:
[...]

PR #977, opened 2025-06-14
fix(deps): add torch-cluster and torch-scatter to pyproject.toml

sohil-sisodiya:
# Fix: Add `torch-cluster` and `torch-scatter` dependencies for `mesh_reduced` model

The `mesh_reduced.py` model depends on `torch-cluster` and `torch-scatter` libraries, but these dependencies were not listed in `pyproject.toml`. This caused source-based installations and tests to fail since those packages were missing.

## Description
- Added `torch-cluster` and `torch-scatter` to `pyproject.toml` under dependencies.
- Verified the `vortex_shedding_mesh_reduced` example works end-to-end.
- Kept the `mesh_reduced.py` model and associated example intact as they are still used.

### Impact
This change resolves installation and test failures related to missing `torch-cluster` and `torch-scatter` dependencies and improves user experience for source installs.

## Checklist

- [x] An [issue](https://github.com/NVIDIA/physicsnemo/issues/898) is linked to this pull request.

## Dependencies
- `torch-cluster`
- `torch-scatter`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.1.0] - 2025-06-05
 
+### Dependencies
+- Added `torch-cluster` and `torch-scatter` to `pyproject.toml` under dependencies.
+
+[1.0.2] - 2025-06-14
+
 ### Added
 
 - Added ReGen score-based data assimilation example
@@ -390,3 +395,6 @@ weather models
 ### Added
 
 - Initial public release.
+
+
+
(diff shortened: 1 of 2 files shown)


coreyjadams:
Hi @sohil-sisodiya - thanks for the PR.  I do not think we should add `torch-cluster` and `torch-scatter` to the global list of dependencies.  They sometimes require careful builds for all of the cuda libraries to work together, and since it's not required for most of `physicsnemo` it will be unnecessary burden on many models.

Can you clarify which model this is for?  It might make more sense to add it to the requriements.txt for that example.

coreyjadams:
Thanks for the info!  If you want to make a PR for that example to add a requirements.txt file, we can take a look.

Let's close this one to make it a clean PR for `requirements.txt`, if you choose to do that one.

PR #980, opened 2025-06-17
Improve lead time support for diffusion models

jleinonen:
# PhysicsNeMo Pull Request

## Description

Adds/fixes support for lead-time labels in various places where it was missing or not working:

- `SongUNetPosEmbd` now works properly using either, both or neither of positional embedding and lead-time embedding. In the previous version some pieces of code could try to access properties of these even when set to `None`.
- `deterministic_sampler` now accepts lead-time labels and passes them through to the model, if given.
- `EDMLoss` also now supports lead-time labels.
- Added tests for the above features. 

## Checklist

## Dependencies

No new dependencies needed.


physicsnemo/metrics/diffusion/loss.py
```diff
@@ -220,7 +220,15 @@ def __init__(
...
 
-    def __call__(self, net, images, condition=None, labels=None, augment_pipe=None):
+    def __call__(
+        self,
+        net,
+        images,
+        condition=None,
+        labels=None,
+        augment_pipe=None,
+        lead_time_label=None,
```
CharlelieLrt, line 230: `lead_time_label` needs to be added in the "Parameters" section of the docstring. Also, for tensor parameters, ideally we would like the expected shape to be detailed in the docstring. Also, all parameters here are missing type hints. It would be good to fix that.

physicsnemo/utils/diffusion/deterministic_sampler.py
```diff
@@ -49,6 +49,7 @@ def deterministic_sampler(
     S_min: float = 0.0,
     S_max: float = float("inf"),
     S_noise: float = 1.0,
+    lead_time_label: torch.Tensor | None = None,
```
CharlelieLrt: Same comment as for the `EDMLoss`: new parameter `lead_time_label ` needs to be documented in the docstring.

physicsnemo/utils/diffusion/deterministic_sampler.py
```diff
@@ -167,6 +168,11 @@ def deterministic_sampler(
     # conditioning
     x_lr = img_lr
 
+    # do not pass lead time labels to nets that may not support them
+    lead_time_label = (
```
CharlelieLrt: This assignment is causing a type checking warning, I suggest adding a type hint `lead_time_label : Dict[str, torch.Tensor] = ...`
IN terms of naming convention it might also be clearer to call this `additional_labels` to be more consistent with the `EDMLoss`

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -826,7 +826,7 @@ def __init__(
 
         self.gridtype = gridtype
         self.N_grid_channels = N_grid_channels
-        if self.gridtype == "learnable":
+        if (self.gridtype == "learnable") or (self.N_grid_channels == 0):
```
CharlelieLrt, line 870: So, if I understand correctly, the goal is to still be able to disable positional embeddings by passing `N_grid_channels = 0` to the `SongUNetPosEmbd`, correct? Could you please detail that in the section "Parameters" of the docstring and explain that in this case, the model behaves similarly to the base class `SongUNet`?

I also think the class docstrings as well as the docstrings for the embeddings selection methods (`positional_embedding_indexing` and `positional_embedding_selector`) should be edited to document and explain the added flexibility when combining/using separately positional embeddings and lead-time embeddings.
**Note:** some of the documentation might be better placed in the `SongUNetPosLtEmbd`'s docstrings rather than `SongUNetPosEmdb`, since this class is the expected entry-point for lead-time aware model.

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -840,6 +840,8 @@ def __init__(
                 self.scalar = torch.nn.Parameter(
                     torch.ones((1, len(self.prob_channels), 1, 1))
                 )
+        else:
+            self.lt_embd = None
```
CharlelieLrt: I know this is not totally in the scope of the PR, but I would prefer if we get rid of the `lead_time_mode` argument (and attribute) altogether. Because, right now we can have `lead_time_mode = True` and `lead_time_channels = 0`, and conversely `lead_time_mode = False` and `lead_time_channels > 0` which are both incompatible.
In addition, since it is now possible to disable positional embeddings by setting `N_grid_channels = 0`, users should expect to be able to do the same for lead-time embeddings by setting `lead_time_channels = 0`.

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -989,23 +991,26 @@ def positional_embedding_indexing(
...
+                    (1, 0, 2, 3),
+                )  # (P, C_pe, X, Y)
+
+                selected_pos_embd = selected_pos_embd.repeat(
+                    B, 1, 1, 1
+                )  # (B*P, C_pe, X, Y)
+
+                embeds = [selected_pos_embd]
+            else:
+                embeds = []
```
CharlelieLrt: Not totally in the scope of this PR, but can we slightly refactor this if statement as well as [the previous one](https://github.com/NVIDIA/physicsnemo/blob/21b93afaef2cadfbb868b3859a927b14c54f37ab/physicsnemo/models/diffusion/song_unet.py#L957-L981) (`if global_index is None`) to make the logic more similar in both? For example:
1. Rename variables in the previous part of the statement into `embeds` instead of `selected_pos_embd` (since it contains also lead-time embeddings, it doesn't really make sense to call it `selected_pos_embd`)
2. Use the same append to list, followed by `torch.cat` in both statements

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -1026,8 +1031,8 @@ def positional_embedding_indexing(
                     )  # (B*P, C_pe, X, Y)
                     embeds.append(selected_lt_pos_embd)
 
-                if len(embeds) > 0:
-                    selected_pos_embd = torch.cat(embeds, dim=1)
+            if len(embeds) > 0:
+                selected_pos_embd = torch.cat(embeds, dim=1)
```
CharlelieLrt: So, what is the method returning if both `self.pos_embd` and `self.lt_emdb` are `None`? If I didn't miss anything it's going to fail with an `UnboundLocalError`...
I think we need better parameters validation here. This could be done in the `__init__`, such that an exception is raised if _both_ positional embeddings and lead-time embeddings are disabled (it doesn't really make sense to use this class instead of the base class `SongUNet` if not using any of these embeddings)

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -1090,8 +1104,9 @@ def positional_embedding_selector(
         ...     return patching.apply(emb[None].expand(batch_size, -1, -1, -1))
         >>>
         """
-        if x.dtype != self.pos_embd.dtype:
+        if (self.pos_embd is not None) and (x.dtype != self.pos_embd.dtype):
```
CharlelieLrt: This change is not sufficient for `positional_embedding_selector`. This method should have equivalent functionality as `positional_embedding_indexing`, but in the present implementation it cannot handle correctly cases where positional embeddings are disabled.

physicsnemo/metrics/diffusion/loss.py
pzharrington: Similarly docstring description should be edited to note the different behavior with respect to the basic EDM loss -- with `lead_time_label` it now supports different behavior

physicsnemo/metrics/diffusion/loss.py
```diff
@@ -258,16 +266,24 @@ def __call__(self, net, images, condition=None, labels=None, augment_pipe=None):
...
+            k: v for (k, v) in additional_labels.items() if v is not None
+        }
         if condition is not None:
             D_yn = net(
                 y + n,
                 sigma,
                 condition=condition,
                 class_labels=labels,
-                augment_labels=augment_labels,
+                **additional_labels,
```
pzharrington: `class_labels` is an item that was basically copied over from EDM but never really put to use in the Earth-2 applications, as far as I'm aware. Can we add it to `additional_labels` as well here without breaking existing use-cases? Would be nice for `EDMLoss` to support networks that do not have a `class_labels` arg. Thoughts @CharlelieLrt ?

physicsnemo/utils/diffusion/deterministic_sampler.py
pzharrington: +1 to consistent naming

jleinonen:
Hi @CharlelieLrt,

> 1. As far as I understand, this PR decouples lead-time embeddings from positional embeddings, in order to allow more flexibility in using them independently from each other. This new functionality does not seem to be used in any of the training recipes/examples. It could be useful to detail in the PR description the broader context (e.g. which applications is it going to be applied to? Will there be a follow-up PR? etc...)

I would say lead-time embeddings were already decoupled from positional embeddings before the PR. This PR just includes some fixes to make sure that they can be enabled when positional embeddings are disabled, or vice versa.

> 2. The new flexibility to independently use lead-time and positional embeddings should be clearly explained in the docstrings.

As they were already implemented independently, I don't think it's a new flexibility, but I can improve the docstrings in that regard.

> 3. IMO the current implementation of the lead-time embeddings has too many failure modes to be safely exposed to broader applications. For example, in `positional_embedding_indexing`:
> 
> * `lead_time_label` can be done while `self.lt_embd` is not `None`, which leads to an error
> * Conversely, `lead_time_label` could be a user-provided tensor, while `self.lt_embd` is `None`, which leads to `lead_time_label` being silently ignored.
> 
> I strongly support better parameters validation to eliminate this failure modes, either in the `forward` method or the `__Init__` when possible.

I'll add some checks to make sure the inputs conform with the model configuration (but note that as far as I understand, these failure modes already existed before the PR).

physicsnemo/metrics/diffusion/loss.py
jleinonen: @pzharrington Not sure I understand this comment. The behavior of `EDMLoss` is not changing (unless I messed up something?). It's just passing through another argument to the model.

pzharrington: Oops, yeah you're right. Doesn't need an update if it's just arg passthrough

physicsnemo/models/diffusion/song_unet.py
jleinonen: If I'm not mistaken, if both `self.pos_embd` and `self.lt_embd` are `None`, we still set `embeds = []` here: https://github.com/NVIDIA/physicsnemo/blob/49295968b3209160e33bd5738f5380a114cdea91/physicsnemo/models/diffusion/song_unet.py#L1010 This case should also be covered by the test added to `test_song_unet_pos_embd.py`.

We could forbid disabling both, but given that it's easy to support the "neither" case, I'd vote to keep it, for instance for a situation where you want to benchmark a model with and without positional embedding while making sure that the implementation is otherwise identical.

physicsnemo/metrics/diffusion/loss.py
jleinonen: Let me check. I recall that when I was making the PR I tried something like that (not 100% sure it was here though) and it caused some of the tests to fail, so I had to revert the changes.

physicsnemo/models/diffusion/song_unet.py
CharlelieLrt: Yes, it sets `embeds = []`, but then it never goes into [this if statement](https://github.com/NVIDIA/physicsnemo/blob/49295968b3209160e33bd5738f5380a114cdea91/physicsnemo/models/diffusion/song_unet.py#L1034), right? So, `selected_pos_embd` is never declared before it is returned.

> but given that it's easy to support the "neither" case, I'd vote to keep it

Ok, let's keep it. But let's make sure we have good parameters validation in the `forward` and other methods to avoid these cases:
- `lead_time_label` is not `None` while lead-time embeddings are disabled.
- `lead_time_label` is `None` while lead-time embeddings are enabled.
(And same thing for the positional embeddings)

physicsnemo/metrics/diffusion/loss.py
CharlelieLrt: I agree with @pzharrington that `class_labels` would be better in `additional_labels`. But the problem is that it is a required (not optional) argument for the `SongUNet` (well, actually it can be set to `None`, in which case it is ignored, but it must be explicitly provided to the `forward`).

**Note:** I have a complete refactors of all diffusion losses in the works. This will allow much more flexibility in terms of conditioning for the `EDMLoss` (it will support `class_labels`, or any other type of conditioning). So, it might not be worth spending too much time on this refactor in this PR.

physicsnemo/models/diffusion/song_unet.py
jleinonen: Ah, yes you're right. But in practice this actually never happens because the call to the function is inside this if block: https://github.com/NVIDIA/physicsnemo/blob/49295968b3209160e33bd5738f5380a114cdea91/physicsnemo/models/diffusion/song_unet.py#L870

Anyway, I agree it's better to fix it in the function itself too.

physicsnemo/metrics/diffusion/loss.py
jleinonen: Resolving this conversation since the conclusion seems to be "no changes for now".

physicsnemo/utils/diffusion/deterministic_sampler.py
jleinonen: Changed to `additional_labels` in https://github.com/NVIDIA/physicsnemo/pull/980/commits/522440b625e709425fa1f0083f5e43370e45c97d.

physicsnemo/utils/diffusion/deterministic_sampler.py
jleinonen: Parameter added to docstring in https://github.com/NVIDIA/physicsnemo/pull/980/commits/522440b625e709425fa1f0083f5e43370e45c97d.

physicsnemo/metrics/diffusion/loss.py
jleinonen: Parameter added to docstring in https://github.com/NVIDIA/physicsnemo/pull/980/commits/522440b625e709425fa1f0083f5e43370e45c97d.

physicsnemo/models/diffusion/song_unet.py
jleinonen: I agree we could get rid of `lead_time_mode` but wasn't sure if this would cause some backward compatibility issues. But I added checks for this in the constructor and corresponding tests in https://github.com/NVIDIA/physicsnemo/pull/980/commits/522440b625e709425fa1f0083f5e43370e45c97d.

physicsnemo/models/diffusion/song_unet.py
jleinonen: I added an exception in https://github.com/NVIDIA/physicsnemo/pull/980/commits/522440b625e709425fa1f0083f5e43370e45c97d that is triggered if the function is called with both `self.pos_embd` and `self.lt_embd` set to `None`. Although this should not be actually called it may help guard some future changes.

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -826,12 +826,16 @@ def __init__(
...
+        if (self.gridtype == "learnable") or (self.N_grid_channels == 0):
             self.pos_embd = self._get_positional_embedding()
         else:
             self.register_buffer("pos_embd", self._get_positional_embedding().float())
         self.lead_time_mode = lead_time_mode
         if self.lead_time_mode:
+            if (lead_time_channels is None) or (lead_time_channels <= 0):
+                raise ValueError(
+                    "`lead_time_channels` must be >= 1 if `lead_time_mode` is enabled."
+                )
```
CharlelieLrt, line 879: Agree with this validation. However, there is clearly redundancy between `lead_time_mode` and `lead_time_channels`. In addition, for positional embeddings, we can disable them by setting `N_grid_channels = 0` (there is no boolean parameter `positional_embedding_mode`), so we expect the same mechanism to work for lead-time embeddings (i.e. disable by setting `lead_time_channels = 0`).

If your concern is about backward compatibility if removing `lead_time_mode`:
1. This parameters was introduced only very recently (<2 months), so it could be okay to remove it
2. If still a concern, can we keep `lead_time_mode` but add a deprecation warning in this if statement:
```python
if self.lead_time_mode:
    warnings.warn(
        f"The parameter `lead_time_mode` will be deprecated in a future version. "
        f"The recommended way to enable (disable) lead-time embeddings is to set `lead_time_channels > 0` (`lead_time_channels = 0`)")
    [...]
```

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -1090,15 +1121,24 @@ def positional_embedding_selector(
...
+            # both positional and lead-time embedding
             # all patches share same lead_time_label
             embeddings = torch.cat(
                 [self.pos_embd, self.lt_embd[lead_time_label[0].int()]]
             )
-        else:
+        elif self.pos_embd is None:  # positional embedding only
             embeddings = self.pos_embd
+        elif lead_time_label is None:  # lead time embedding only
+            embeddings = self.lt_embd[lead_time_label[0].int()]
```
CharlelieLrt: Why using only `lead_time_label[0]` here? I know it follows the original implementation, but isn't it an error?

pzharrington: The usage of `[0]` does seem like an error to me...wouldn't that be either selecting only the first lead time in the batch or lead_time dimension (whichever is first in that tensor)? I see it was introduced in #913, maybe @tge25 can chime in?

pzharrington: Unrelated, I think there is a second bug here. The `elif` clauses should read `elif self.pos_embd is not None` and `elif self.lead_time_label is not None`, right?

physicsnemo/metrics/diffusion/loss.py
```diff
@@ -245,6 +253,10 @@ def __call__(self, net, images, condition=None, labels=None, augment_pipe=None):
             An optional data augmentation function that takes images as input and
             returns augmented images. If not provided, no data augmentation is applied.
 
+        lead_time_label: torch.Tensor, optional
+            Lead-time labels to pass to the model, shape (batch_size, 1).
```
CharlelieLrt: Are you sure about the shape `(batch_size, 1)` of `lead_time_label`? I've seen other places (not in your PR), where it's `(batch_size,)`, some other places where it's `(batch_size, 1, 1, 1)`

physicsnemo/models/diffusion/song_unet.py
CharlelieLrt: I confirmed with @tge25 that the `[0]` is some prototyping code, so we should get rid of it.

This `elif self.lead_time_label is not None` is not possible, because `lead_time_label` is not an attribute. But I agree that the `elif` clause should be different. I think it should be `elif self.lt_embd is not None`, right?

CharlelieLrt (approved):
LGTM

PR #982, opened 2025-06-17
Fix casting in SongUNetPosEmbd and shape in CorrDiff generation

juliusberner:
# PhysicsNeMo Pull Request

## Description
1. Fix regression output shape in CorrDiff 
2. Only use act if `fused_act` is `True` in `ApexGroupNorm`
3. Avoid dtype change of attributes (since `self.pos_embd` can be buffer *or* parameter) and fix dtype of softmax output (which is `fp32`) for `SongUNetPosEmbd`
4. Avoid changing the dtype of the `.data` attribute of `self.scalar` to enable `.compile` for `SongUNetPosEmbd`.

## Checklist


physicsnemo/models/diffusion/song_unet.py
```diff
@@ -877,18 +874,19 @@ def forward(
 
             out = super().forward(x, noise_labels, class_labels, augment_labels)
 
-            if self.lead_time_mode:
+            if self.lead_time_mode and self.prob_channels:
```
juliusberner, line 913: @CharlelieLrt Do you know why we need `if self.lead_time_mode` here?

CharlelieLrt: We don't need it _per-se_, but the self.lead_time_mode indicates whether we are dealing with a `SongUNetPosEmdb`, or with a lead-time aware version of it (ie. a `SongUNetPosLtEmdb`). For historical reasons, the `self.prob_channels` is only used with the latter, because lead-time aware model and probability channels have been implemented at the same time, in the same class. Since probability channels should **only** be enabled in lead-time aware model, if `self.lead_time_mode` is `False`, then there is no `prob_channels` and we don;t enter the `if` statement.

In reality `prob_channels` and lead-time have nothing to do with each other and their implementation should be totally decoupled. We have some WiP that is refactoring those into a cleaner more modular implementation.

juliusberner: Makes sense! So we can either remove `if self.lead_time_mode` here or in the future refactoring.

CharlelieLrt: There will be no `self.lead_time_mode` at all in the future implementation. There will only be:
1. One unified `SongUNet` class that optionally supports position embeddings and probability channels. IMO these two features are sufficiently general purpose to be lifted to the base parent class.
2. A specialized class `SongINetPosLtEmdb` that subclasses the base class and only add the lead-time embeddings capability to it.

juliusberner: Sounds good!

examples/weather/corrdiff/generate.py
```diff
@@ -196,7 +196,7 @@ def generate_fn():
                         net=net_reg,
                         img_lr=img_lr,
                         latents_shape=(
-                            cfg.generation.seed_batch_size,
+                            sum(map(len, rank_batches)),
```
CharlelieLrt, line 211: @juliusberner could you explain the reason for this change? AFAIK the batch dimension of `latents_shape` is never really used, right?

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -877,18 +874,19 @@ def forward(
...
+                if out.dtype != scalar.dtype:
+                    scalar = scalar.to(out.dtype)
+                if self.training:
+                    out[:, self.prob_channels] = out[:, self.prob_channels] * scalar
+                else:
                     out[:, self.prob_channels] = (
-                        out[:, self.prob_channels] * self.scalar
+                        (out[:, self.prob_channels] * scalar)
+                        .softmax(dim=1)
+                        .to(out.dtype)
```
CharlelieLrt, line 925: LGTM, but could you just post below an MRE of the bug you encountered with the former casting logic?

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -947,15 +945,16 @@ def positional_embedding_indexing(
         """
         # If no global indices are provided, select all embeddings and expand
         # to match the batch size of the input
-        if x.dtype != self.pos_embd.dtype:
-            self.pos_embd = self.pos_embd.to(x.dtype)
+        pos_embd = self.pos_embd
+        if x.dtype != pos_embd.dtype:
+            pos_embd = pos_embd.to(x.dtype)
```
CharlelieLrt, line 986: Two remarks:
1. Same as above: could you post below a MRE of the bug that you would get with the former casting logic (the MRE could be grouped with the one above)
2. Is there some logic problem here? We are accessing `pos_embd.dtype` and right below we are checking `if pos_embd is not None`? I think if `self.pos_embd` is `None` then we do `return None` right away?

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -1086,15 +1085,12 @@ def positional_embedding_selector(
         ...     return patching.apply(emb[None].expand(batch_size, -1, -1, -1))
         >>>
         """
-        if x.dtype != self.pos_embd.dtype:
-            self.pos_embd = self.pos_embd.to(x.dtype)
+        embeddings = self.pos_embd
+        if x.dtype != embeddings.dtype:
+            embeddings = embeddings.to(x.dtype)
```
CharlelieLrt: Two remarks:
1. Same as above, it would be great if you could post an MRE below (can be grouped with other RMEs for these castings bugs).
1. Is there a specific reason to call it `embeddings` here, whereas it was called `pos_embd` in the `positional_embedding_indexing` method? If not, let's try to remain consistent in the names

CharlelieLrt (changes requested):
Overall looks good! Just need a few clarifications and a few MREs for the bugs that this PR fixes.

examples/weather/corrdiff/generate.py
juliusberner: This is the batch-size that the output of `regression_step` is expanded to. Since we compute `image_out = image_reg + image_res` later, this needs to match the batch-size of the output of `diffusion_step`.

physicsnemo/models/diffusion/song_unet.py
juliusberner: In `amp-bf16` training, the output of `softmax` is `float32`, while `out.dtype = bfloat16`, which gives a `RuntimeError: Index put requires the source and destination dtypes match, got BFloat16 for the destination and Float for the source.`.

physicsnemo/models/diffusion/song_unet.py
juliusberner: 1. The assignment `self.pos_embd = self.pos_embd.to(dtype)` only works if `pos_embd` is a buffer but not if it is a parameter (which is the case if `self.gridtype == "learnable"`). Thus, we define a new local variable which works in both cases.
2. `positional_embedding_indexing` is only called in the `forward` if `self.pos_embd is not None`. If it is called from outside, it would return an empty list `[]`. How should we handle it?

physicsnemo/models/diffusion/song_unet.py
juliusberner: 1. Copying from above: The assignment `self.pos_embd = self.pos_embd.to(dtype)` only works if `pos_embd` is a buffer but not if it is a parameter (which is the case if `self.gridtype == "learnable"`). Thus, we define a new local variable which works in both cases.
2. I took it from the existing code, but it makes sense to rename it to `pos_embd`.

CharlelieLrt (approved):
LGTM

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -913,18 +910,19 @@ def forward(
```
juliusberner, line 910: @CharlelieLrt I think we also need to change the line above to `x = torch.cat((x, selected_pos_embd.to(x.dtype)), dim=1)` for the `use_fp16` option to work.

CharlelieLrt: Yes!

PR #984, opened 2025-06-19
Patch for deprecation of generative utils

CharlelieLrt:
# PhysicsNeMo Pull Request

## Description
In a previous PR `physicsnemo.utils.generative` has been renamed into `physicsnemo.utils.diffusion`. This PR adds a patch that:
- Defines aliases, such that all objects that used to exist in `physicsnemo.utils.generative` are still accessible and point to their counterparts in `physicsnemo.utils.diffusion`
- Prints a warning message when importing from `physicsnemo.utils.generative`, to inform the user about this deprecation process.

## Checklist

### DIFF
--- a/physicsnemo/utils/generative/__init__.py
+++ b/physicsnemo/utils/generative/__init__.py
@@ -1,0 +1,58 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2024 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ruff: noqa
+
+import warnings
+
+warnings.warn(
+    "physicsnemo.utils.generative is deprecated and will be removed in a future version. "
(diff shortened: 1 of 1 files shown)


physicsnemo/utils/generative/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+# ruff: noqa
+
+import warnings
+
+warnings.warn(
+    "physicsnemo.utils.generative is deprecated and will be removed in a future version. "
+    "Please use physicsnemo.utils.diffusion instead."
+)
+
+from physicsnemo.utils.diffusion.deterministic_sampler import deterministic_sampler
```
jleinonen, line 25: It's a bit confusing that the module name and function name for the sampler is the same. I may be wrong but it seems to me that if an external code is using
```
from physicsnemo.utils.generative import deterministic_sampler
```
it probably works with this patch but if it was using
```
from physicsnemo.utils.generative.deterministic_sampler import deterministic_sampler
```
it probably used to work with the old version but now won't? (And same for the stochastic sampler).

All external imports I'm aware of use the first form though.

jleinonen:
Let's discuss the import semantics briefly, otherwise looks good to me!

physicsnemo/utils/generative/__init__.py
CharlelieLrt: IMO any external code that is doing the second option:

```python
from physicsnemo.utils.generative.deterministic_sampler import deterministic_sampler
```

should stop doing that because that's not the right way to do it. Lifting the APIs to the module-level (i.e. in the `__init__.py`) is not only to have shorter imports, but also to distinguish public APIs exposed to the user vs semi-private APIs. In summary:
1. Anything in `physicsnemo/utils/generative/__init__.py` is public (or exposed to the user)
2. Anything in `physicsnemo/utils/generative/some_module.py` is semi-private. Users can access it, but at their own risk.

To make this distinction more explicit we could rename the `some_module.py` into `_some_module.py` (this would also avoid the overloaded name `deterministic_sampler`).

I also think the two samplers could be re-organized into a single module `samplers.py` instead of their own modules, but this would be for another PR.

PR #985, opened 2025-06-20
Refactor on DoMINO `utils.py` to improve code quality

peterdsharpe:
# PhysicsNeMo Pull Request

Major changes:
- Adds complete function docstrings with argument definitions, return type definitions, notes on what exceptions can raise, type hints, and example usages.
- Refactors to use `pathlib.Path` over `os.path` for file path management (preferred since Python 3.4) [[1](https://www.reddit.com/r/learnpython/comments/i29h6j/when_or_why_should_i_use_pathlibpath_over_ospath/), [2](https://medium.com/@barila/choose-your-path-pythons-pathlib-vs-os-path-4de0b1e752dd)]
- Fixes functions that have mutable default values (e.g., defaulting an argument to a list, rather than a tuple) - [this should essentially-never be used](https://www.geeksforgeeks.org/python/use-mutable-default-value-as-an-argument-in-python/), and leads to very-difficult-to-trace bugs.
- Removes a usage of `from {package} import *`, which is not necessary here and [should essentially-never be used](https://www.geeksforgeeks.org/python/why-import-star-in-python-is-a-bad-idea/).
- Fixes missing requirements from DoMINO `requirements.txt`, to facilitate reproducible training
- Removes logic throughout that handled cases where PyVista/VTK/CuPy is not installed; this is unnecessary as the top level package dependencies already assert that these should be present (in "all"), and the pre-packaged container already contains these.
- Updates type hinting style to use latest Black/Ruff style recommendations for modern Python (e.g., `list` instead of `from typing import List`; use `|` rather than `Union`; prefer `Type | None` to `Optional`)
- Applies `black` formatting to `utils.py`, to bring it into compliance with repository-wide [pre-commit hooks](.pre-commit-config.yaml) requirements. (Note: all future PRs should comply with this per CI, and we should not allow PRs to merge that bypass this check.)

Changes not yet implemented in this PR; these should be implemented in a future PR:

- DoMINO `utils.py` itself has no unit test coverage (and is only indirectly tested by the DoMINO forward pass test).
- Randomness in data processing and forward passes of model architectures should always be seeded for reproducibility; currently, it is not.
- Many VTK utilities could be refactored onto PyVista via `pv.wrap()`, which would reduce verbosity and help long-term maintainability.
- In DoMINO `model.py`, fixing instances that do `torch.cat()` on each iteration of Python-level `for` loops to incrementally append to tensors (we should instead build up a `list[torch.Tensor]` to `cat` all-at-once, or vectorize)

## Checklist

### DIFF
--- a/physicsnemo/utils/domino/utils.py
+++ b/physicsnemo/utils/domino/utils.py
@@ -79,10 +79,12 @@ def calculate_center_of_mass(centers: ArrayType, sizes: ArrayType) -> ArrayType:
         ValueError: If centers and sizes have incompatible shapes.
 
     Examples:
-        >>> centers = np.array([[0, 0, 0], [1, 1, 1], [2, 2, 2]])
+        >>> import numpy as np
+        >>> centers = np.array([[0.0, 0.0, 0.0], [1.0, 1.0, 1.0], [2.0, 2.0, 2.0]])
         >>> sizes = np.array([1.0, 2.0, 3.0])
         >>> com = calculate_center_of_mass(centers, sizes)
-        >>> print(com)  # [[1.5, 1.5, 1.5]]
+        >>> np.allclose(com, [[1.5, 1.5, 1.5]])
+        True
     """
     xp = array_type(centers)
 
@@ -115,12 +117,15 @@ def normalize(
         ZeroDivisionError: If max_val equals min_val (zero range).
 
     Examples:
+        >>> import numpy as np
         >>> field = np.array([1.0, 2.0, 3.0, 4.0, 5.0])
         >>> normalized = normalize(field, 5.0, 1.0)
-        >>> print(normalized)  # [-1, -0.5, 0, 0.5, 1]
-
+        >>> np.allclose(normalized, [-1.0, -0.5, 0.0, 0.5, 1.0])
+        True
         >>> # Auto-compute min/max
-        >>> normalized = normalize(field)
+        >>> normalized_auto = normalize(field)
+        >>> np.allclose(normalized_auto, [-1.0, -0.5, 0.0, 0.5, 1.0])
+        True
     """
     xp = array_type(field)
 
@@ -150,9 +155,11 @@ def unnormalize(
         Field values restored to their original physical range.
 
     Examples:
+        >>> import numpy as np
         >>> normalized = np.array([-1.0, -0.5, 0.0, 0.5, 1.0])
         >>> original = unnormalize(normalized, 5.0, 1.0)
-        >>> print(original)  # [1, 2, 3, 4, 5]
+        >>> np.allclose(original, [1.0, 2.0, 3.0, 4.0, 5.0])
+        True
     """
     field_range = max_val - min_val
     return (normalized_field + 1.0) * field_range * 0.5 + min_val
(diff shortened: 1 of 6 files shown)


peterdsharpe:
Note: this PR should ideally be merged after https://github.com/NVIDIA/physicsnemo/pull/973; however, the changes are largely orthogonal between these two, so there should be minimal merge issues.

physicsnemo/utils/domino/utils.py
```diff
@@ -15,219 +15,544 @@
...
+mesh manipulation, field normalization, and geometric computations. It supports both
+CPU (NumPy) and GPU (CuPy) operations with automatic fallbacks.
 """
 
-import os
-from typing import Any, List, Optional, Sequence, Tuple, Union
+from pathlib import Path
+from typing import Any, Sequence
 
+import cupy as cp
```
coreyjadams: Do we really want a bare cupy import?  Its not in the requirements, and while it's a dep of cuml which is, we might not actually want to enforce it for CPU execution?

physicsnemo/utils/domino/utils.py
```diff
@@ -15,219 +15,544 @@
...
 
 from physicsnemo.utils.profiling import profile
 
-try:
-    import cupy as cp
+# Type alias for arrays that can be either NumPy or CuPy
+ArrayType = np.ndarray | cp.ndarray
+
+
+def array_type(array: ArrayType) -> type[np] | type[cp]:
```
coreyjadams: I originally made this function because I didn't force a cupy dependency.  If we *are* forcing a cupy dependency ... we should trim this and use `cupy.get_array_module` everywhere.  This wrapper doesn't serve any purpose if we know we have cupy, I think,

physicsnemo/utils/domino/utils.py
```diff
@@ -15,219 +15,544 @@
...
+    Examples:
+        >>> import numpy as np
+        >>> arr = np.array([1, 2, 3])
+        >>> xp = array_type(arr)
+        >>> result = xp.sum(arr)  # Uses numpy.sum
+    """
+    return cp.get_array_module(array)
 
 
-try:
```
coreyjadams, line 37: This is only used in one function, I think.  Should we force the `pv` dependency for everything else just for this?

peterdsharpe:
Tests added: daac91a0a48ddb68d1b240466008c479f456a693

peterdsharpe:
CuML hard-dependency removed: d2b11c7583a45e0b161891e3c55718b5f89022f4

physicsnemo/utils/domino/utils.py
peterdsharpe: CuML hard-dependency removed: d2b11c7583a45e0b161891e3c55718b5f89022f4

physicsnemo/utils/domino/utils.py
peterdsharpe: CuML hard-dependency removed: d2b11c7583a45e0b161891e3c55718b5f89022f4

physicsnemo/utils/domino/utils.py
peterdsharpe: Fixed: 02849e0462a551e4bc637b257146c147ca3da9cb

PR #989, opened 2025-06-24
Refactor GroupNorm and log unmatched state_dict keys

### LINKED ISSUE #1001: 🐛[BUG]: GroupNorm creates unused parameters when `use_apex_gn=True`
### Version

1.1.0

Pip

### Describe the issue

The GroupNorm class in physicsnemo defines the affine parameters here: https://github.com/pytorch/pytorch/blob/134179474539648ba7dee1317959529fbd0e7f89/torch/nn/modules/normalization.py#L291-L292

But Apex defines those internally as well: https://github.com/NVIDIA/apex/blob/d8200a1bec8d29fe3f44d96b0a9e7b7fba850c77/apex/contrib/group_norm/group_norm.py#L321-L324

Gradients are only computed for the Apex defined weights and the physicsnemo defined weights are unused, so they are indeed unused parameters that shouldn’t be there in the first place. This causes the physicsnemo defined weights to have `None` gradients and as a result, DDP hangs waiting for an allreduce for those unused weight gradients. A workaround might is to use `find_unused_parameters=True` with DDP, but that is very unsatisfactory. There is another issue that a model checkpoint that was trained used `use_apex_gn=False` will produce wrong results if inferenced using `use_apex_gn=True`.

```shell
In [1]: from physicsnemo.models.diffusion.layers import GroupNorm

In [2]: import torch

In [3]: device = torch.device("cuda")

In [4]: m = GroupNorm(num_channels=128, eps=1e-6, use_apex_gn=True).to(device)

In [5]: inp = torch.ones(1, 128, 64, 64, device=device)

In [6]: optimizer = torch.optim.SGD(m.parameters(), lr=1e-4, momentum=0.9)

In [7]: out = m(inp)

In [9]: loss = out.mean()

In [10]: loss.backward()

In [11]: optimizer.step()

In [12]: for name, param in m.named_parameters():
    ...:     if param.grad is None:
    ...:         print(f"{name} has `None` gradients")
    ...:     else:
    ...:         print(f"{name} has valid gradients")
    ...: 
weight has `None` gradients
bias has `None` gradients
gn.weight has valid gradients
gn.bias has valid gradients
```

### Relevant log output

```shell

```

### Environment details

```shell

```

juliusberner:
# PhysicsNeMo Pull Request

## Description
- Diffusion models: consolidate `ApexGroupNorm` and `GroupNorm` in `models/diffusion/layers.py` with a factory `get_group_norm` that can be used to instantiate either one of them. this ensures that the state_dict remains consistent with previous versions.
- Log missing and unexpected keys when loading checkpoints.
- In `SongUNet`-based diffusion models, add `persistent=False` for deterministic, non-learnable positional embeddings.

Closes #1001 .

## Checklist


physicsnemo/models/util_compatibility.py
CharlelieLrt, line 1: ~~Keep this file (empty with the license is fine), as other PRs are adding back sone content there.~~

physicsnemo/models/module.py
```diff
@@ -30,11 +29,36 @@
...
+        logging.warning(
+            f"Missing keys when loading {module.__class__.__name__}: {missing_keys}"
+        )
+    if unexpected_keys:
+        logging.warning(
+            f"Unexpected keys when loading {module.__class__.__name__}: {unexpected_keys}"
+        )
+    return missing_keys, unexpected_keys
+
+
```
CharlelieLrt: @juliusberner just to confirm: this logging is only as a precaution for future usage, right? Given your changes with the `get_group_norm`, we should not have any missing or unexpected keys ate any moment, correct?

juliusberner: Yes, exactly!

physicsnemo/models/module.py
```diff
@@ -461,8 +483,7 @@ def from_checkpoint(
                 local_path.joinpath("model.pt"), map_location=model.device
             )
 
-            model_dict = convert_ckp_apex(ckp_args, model_args, model_dict)
-            model.load_state_dict(model_dict, strict=False)
+            load_state_dict_with_logging(model, model_dict, strict=False)
```
pzharrington: I'm generally uncomfortable with `strict=False` here. I realize it was injected in #809 rather than this PR, but since this is a widely used function affecting all trained models in physicsnemo, can we revert it to `strict=True` here? Seems like backwards-compat handling should be taken care of by the time this line is run. Thoughts @CharlelieLrt ?

CharlelieLrt: @pzharrington good point! I was actually wondering why we were setting `strict=False` here, but I thought it was there since the beginning and I've never realized it was introduced by #809 . We should definitely revert it back to `True`. AFAIK it's only useful when fine-tuning some parts of the model, or other things like this...

@juliusberner does your application requires `strict=False` there, or would it be okay to revert to `True`?

pzharrington: Probably also worth checking with @LostnEkko -- was it introduced in #809 as part of the apex checkpoint handling, and would you be able to test your use-case off of this PR to see if `strict=True` works?

juliusberner: I think the best option would be to expose `strict` (similar as in the `load` method above). We can then default it to `True` if we want --- if the user sets it to `False`, we would still log unexpected/missing keys with this PR (preventing silent errors which are happening right now since the checkpoint conversion has a bug in https://github.com/NVIDIA/physicsnemo/blob/d1c9391f0f594f7279c8990bff70b8227a6d1f93/physicsnemo/models/util_compatibility.py#L92C17-L92C50)

CharlelieLrt: > Probably also worth checking with @LostnEkko

@jialusui1102 for viz

CharlelieLrt: @juliusberner sounds good to me! Feel free to update your PR accordingly

physicsnemo/models/diffusion/layers.py
```diff
@@ -290,6 +290,64 @@ def forward(self, x):
...
+        The activation function to use when fusing activation with GroupNorm. Defaults to None.
+    amp_mode : bool, optional
+        A boolean flag indicating whether mixed-precision (AMP) training is enabled. Defaults to False.
+    Notes
+    -----
+    If `num_channels` is not divisible by `num_groups`, the actual number of groups
+    might be adjusted to satisfy the `min_channels_per_group` condition.
+    """
+
+    num_groups = min(num_groups, num_channels // min_channels_per_group)
```
akshaysubr: Should include the fix from https://github.com/NVIDIA/physicsnemo/pull/996 here as well

physicsnemo/models/module.py
juliusberner: @CharlelieLrt I adapted it

physicsnemo/models/diffusion/layers.py
juliusberner: Thanks, I rebased and included it now!

physicsnemo/models/diffusion/layers.py
```diff
@@ -322,6 +322,72 @@ def forward(self, x):
...
+            act=act,
+        )
+    else:
+        return GroupNorm(
+            num_groups=num_groups,
+            num_channels=num_channels,
+            eps=eps,
+            act=act,
+            amp_mode=amp_mode,
+        )
```
CharlelieLrt: @juliusberner if I am not mistaken, the `ApexGroupNorm` and `GroupNorm` do not have the same functionality, because of the `amp_mode` argument.

## Problem

For `GroupNorm`:
1. when `amp_mode=False`, the parameters of the `GroupNorm` layer are casted to the dtype of the input tensor during the forward pass.
2. when `amp_mode=True`, no casting is manually performed, as `torch.autocast` is supposed to be enabled

However, since `ApexGroupNorm` does not have an `amp_mode` argument, it will not be able to do casting as in case 1 above (case 2 is not a concern). From the user perspective that means that the code below:
```python
x_bf16  # bf16 input
model_fp32 = MyModel(..., amp_mode=False)  # fp32 model
y = model_fp32(x_bf16)
```
will work if `model_fp32` uses `GroupNorm` but will fail if it uses `ApexGroupNorm`.
IMO, this is unexpected behavior and should be avoided.

## Solution

An easy solution could be to subclass `ApexGroupNorm` with something like:

```python
[...]
_ApexGroupNorm = getattr(apex_gn_module, "GroupNorm")

class ApexGroupNorm(_ApexGroupNorm):
    def __init__(..., amp_mode=False):
        super().__init__(...)
        self.amp_mode = amp_mode
    
    def forward(self, x):

        # Do the casting the same way as in `GroupNorm`
        weight, bias = self.weight, self.bias
        _validate_amp(self.amp_mode)
        if not self.amp_mode:
            if weight.dtype != x.dtype:
                weight = self.weight.to(x.dtype)
            if bias.dtype != x.dtype:
                bias = self.bias.to(x.dtype)

       # Call forward from parent class
       super().forward(x)        
```

Maybe @jialusui1102 can chime in if this ApexGN subclass is a viable solution or not.

juliusberner: Good catch! I think this check would have also failed before this MR. One question is whether we want to support the example that you provided? Typically, one requires the user to call `model_fp32.to(dtype)` before feeding inputs with a different `dtype`.

CharlelieLrt: >  think this check would have also failed before this MR

Yes, very possible. But let's try to fix it here.

> Typically, one requires the user to call model_fp32.to(dtype) before feeding inputs with a different dtype

I thought so as well, but then what is the purpose of the `amp_mode` argument? The only thing it does are these manual casting... @jialusui1102 I think you introduced the `amp_mode` argument, any thought on this?

CharlelieLrt:
**Note:** need to add tests (see [this thread](https://github.com/NVIDIA/physicsnemo/pull/1038#discussion_r2246023578))

physicsnemo/models/diffusion/layers.py
CharlelieLrt: After discussion, we will accept this for now, since the `amp_mode` will be removed in a future PR anyways.

physicsnemo/models/diffusion/__init__.py
```diff
@@ -20,6 +20,7 @@
     Conv2d,
     FourierEmbedding,
     GroupNorm,
+    get_group_norm,
```
CharlelieLrt, line 23: @juliusberner do we still want to expose `GroupNorm` or replace it by `get_group_norm` as the recommended API for instantiating GN layers?

physicsnemo/models/diffusion/layers.py
```diff
@@ -361,73 +418,33 @@ class GroupNorm(torch.nn.Module):
 
     def __init__(
         self,
+        num_groups: int,
```
CharlelieLrt: @juliusberner in `GroupNorm` the parameter `num_groups` went from an optiomal keyword argument to first required positional argument. This change is BC-breaking and would be better to avoid. There are 2 options:
1. Ideally, we would like to keep the same API for `GroupNorm` and have the `get_group_norm` factory handle the difference of APIs between `GroupNorm` and `ApexGroupNorm`
2. If this BC-breaking change is really needed, that's okay, but in that case we should not expose `GroupNorm` in the `physicsnemo.models.diffusion` module (see [comment above](https://github.com/NVIDIA/physicsnemo/pull/989/files#r2249051321)). Instead we should have a clear message indicating that `GroupNorm` is not the recommended API for GN, because subject to API changes, and that users should instead use `get_group_norm`, which is the recommended API.

## ⚠️  Notes: 
- Actually I think changing the API of `GroupNorm` will not pass CI tests, because it breaks the TopoDiff example (in `examples/generative/topodiff/topodiff.py`) that uses directly the `GroupNorm` class.
- This also applies to the `min_channels_per_group` parameter that has been removed.

physicsnemo/models/diffusion/__init__.py
juliusberner: I left it for backwards compatibility but I am fine with removing it.

physicsnemo/models/diffusion/layers.py
juliusberner: The change is not needed and could be reverted. I tried to simplify the API for `GroupNorm` (and stay close to the PyTorch API https://docs.pytorch.org/docs/stable/generated/torch.nn.GroupNorm.html), since `get_group_norm` can take care of the necessary conversions and checks. I think we have two options:
1. Use the factory `get_group_norm` also in TopoDiff.
2. Revert the API of `GroupNorm` and add the conversions and checks (which will produce duplicate code in `GroupNorm` and `get_group_norm`).

CharlelieLrt: I would prefer 2 (there are other external users of `GroupNorm`, so we don't want to break their code without warning). No problem for the code duplicate, as long as it's consistent between `get_group_norm` and `GroupNorm`.

juliusberner: What functionality of `GroupNorm` are they using, since even option 2. is BC-breaking (as it does not include the apex option anymore)?

CharlelieLrt: Yeah good point... So, I think it's fine to have BC-breaking changes for the `use_apex_gn`, because it was added relatively recently. But even in this case we still need a deprecation message. So, I propose to keep all parameters and functionalities of `GroupNorm` the same, except for `use_apex_gn`, for which we do something like:

```python
if use_apex_gn:
    warnings.warn(
        "'use_apex_gn' is deprecated. Please use 'get_group_norm' to enable Apex-based group norm."
    )
```

physicsnemo/models/diffusion/__init__.py
CharlelieLrt: Ok. If we can ensure BC-compatibility, we can keep the `GroupNorm` here

physicsnemo/models/diffusion/layers.py
juliusberner: Sounds good, I have committed a fix.

CharlelieLrt (approved):

prem-krishnan:
@CharlelieLrt Can we associate this task with an epic?

PR #992, opened 2025-06-30
[Minor] Profiling Cleanup

coreyjadams:
Reduce profiling output by not printing profiles for functions that don't get called.  

# PhysicsNeMo Pull Request

## Checklist

### DIFF
--- a/physicsnemo/utils/profiling/line_profile.py
+++ b/physicsnemo/utils/profiling/line_profile.py
@@ -89,7 +89,7 @@ class LineProfileWrapper(PhysicsNeMoProfilerWrapper, metaclass=_Profiler_Singlet
         # Get the output directory:
         out_top = self.output_dir(output_top)
         with open(out_top / Path("profiler_stats.txt"), "w") as stats:
-            self._profiler.print_stats(stream=stats)
+            self._profiler.print_stats(stream=stats, stripzeros=True)
 
         # Make this profiler completed:
         self.finalized = True


coreyjadams:
FYI - `stripzeros` is a parameter that cleans up the output of `line_profiler`.  For example, if you have the following instrumented code:

```python

from physicsnemo.utils.profiling import profile

@profile
def foo(*args, **kwargs):
    # Do something here ...

@profile
def bar(*args, **kwargs):
    # Do something different here ...

if __name__ == "__main__":
    for i in range(50):
        bar(i)

```

Then, if you run this with the profiling utilities and `stripzeros=False`, you will get the line-by-line profile of both functions,  Of course, `foo` will have no information for any lines (it was never called).  With `stripzeros=True`, `foo` will be excluded from the output file since it was never used.

The upshot is this makes it substantially more "leave-in" friendly: users will not see the profiles of functions they don't call at some point in their application.  It makes the output files less cluttered.

physicsnemo/utils/profiling/line_profile.py
```diff
@@ -89,7 +89,7 @@ def finalize(self, output_top: Path) -> None:
         # Get the output directory:
         out_top = self.output_dir(output_top)
         with open(out_top / Path("profiler_stats.txt"), "w") as stats:
-            self._profiler.print_stats(stream=stats)
+            self._profiler.print_stats(stream=stats, stripzeros=True)
```
laserkelvin, line 92: Would it be desirable to have `stripzeros` as an argument to `finalize`?

Alternatively, unpacking `kwargs` into `print_stats` to access other functionality?

coreyjadams: No - if it's going to be configurable, it should go into a config_override for this profiling module.  IMO we should set it to `True` by default.  The only information gained with using it as False is to see if your function that you instrumented was not called.  But if it's not called, that information is already clear since the function will not show up in the outputs.

laserkelvin (approved):
LGTM

Issue #972, opened 2025-06-13
🐛[BUG]: GroupNorm can set num_groups to 0 in an edge case


akshaysubr:
### Version

1.1.0

### On which installation method(s) does this occur?

Pip

### Describe the issue

In cases where the channel dimension is <4, num_groups can be set to 0 which then leads to this Apex error:
```
self = <[AttributeError("'GroupNorm' object has no attribute 'act'") raised in repr()] GroupNorm object at 0x7add64483c50>, num_groups = 0, num_channels = 2, eps = 1e-06, affine = True, device = None, dtype = None, act = 'silu'

    def __init__(self,
                 num_groups: int,
                 num_channels: int,
                 eps: float = 1e-5,
                 affine: bool = True,
                 device=None,
                 dtype=None,
                 act=None) -> None:
        factory_kwargs = {'device': device, 'dtype': dtype}
        super().__init__()
>       if num_channels % num_groups != 0:
E       ZeroDivisionError: integer modulo by zero

/usr/local/lib/python3.12/dist-packages/apex/contrib/group_norm/group_norm.py:313: ZeroDivisionError
```
Looks like that issue stems from this line:
https://github.com/NVIDIA/physicsnemo/blob/b0f58ed59937a8cf737c652c1ccb4dab23fb4028/physicsnemo/models/diffusion/layers.py#L345

Maybe it should instead be something like this instead?
```python
self.num_groups = min(num_groups, (num_channels + min_channels_per_group - 1) // min_channels_per_group) 
```

### Minimum reproducible example

```shell
from physicsnemo.models.diffusion.layers import GroupNorm
norm = GroupNorm(num_channels=3, use_apex_gn=True,)
```

### Relevant log output

```shell
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
Cell In[2], line 1
----> 1 norm = GroupNorm(num_channels=3, use_apex_gn=True,)

File /usr/local/lib/python3.12/dist-packages/physicsnemo/models/diffusion/layers.py:367, in GroupNorm.__init__(self, num_channels, num_groups, min_channels_per_group, eps, use_apex_gn, fused_act, act, amp_mode)
    358         self.gn = ApexGroupNorm(
    359             num_groups=self.num_groups,
    360             num_channels=num_channels,
   (...)    363             act=self.act,
    364         )
    366     else:
--> 367         self.gn = ApexGroupNorm(
    368             num_groups=self.num_groups,
[...]

PR #996, opened 2025-07-02
Fixing GroupNorm edge case with num_channels < min_channels_per_group or not perfectly divisible

akshaysubr:
# PhysicsNeMo Pull Request

## Description

Closes https://github.com/NVIDIA/physicsnemo/issues/972

## Checklist

## Dependencies

None

### DIFF
--- a/physicsnemo/models/diffusion/layers.py
+++ b/physicsnemo/models/diffusion/layers.py
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 """
-Model architecture layers used in the paper "Elucidating the Design Space of 
+Model architecture layers used in the paper "Elucidating the Design Space of
 Diffusion-Based Generative Models".
 """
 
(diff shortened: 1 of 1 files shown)


physicsnemo/models/diffusion/layers.py
```diff
@@ -342,7 +342,10 @@ def __init__(
             raise ValueError("'act' must be specified when 'fused_act' is set to True.")
 
         super().__init__()
-        self.num_groups = min(num_groups, num_channels // min_channels_per_group)
+        self.num_groups = min(
+            num_groups,
+            (num_channels + min_channels_per_group - 1) // min_channels_per_group,
+        )
```
CharlelieLrt, line 348: So let's say I do:

```python
gn  = GroupNorm(num_channels=4, min_channels_per_group=8)
```
This is actually going to use 1 group with only 4 channels inside. Is there a reason not to raise an error, like [what pytorch does for GroupNorm](https://github.com/pytorch/pytorch/blob/134179474539648ba7dee1317959529fbd0e7f89/torch/nn/modules/normalization.py#L291-L292)?

akshaysubr: Good point, that might be an easier fix

physicsnemo/models/diffusion/layers.py
```diff
@@ -342,7 +342,14 @@ def __init__(
...
+        if num_channels % self.num_groups != 0:
+            raise ValueError(
+                "num_channels must be divisible by num_groups or min_channels_per_group"
+            )
```
CharlelieLrt, line 352: @akshaysubr is the if statement wrong? It doesn't seem to match the error message

akshaysubr: No, I think it is correct. The if will be triggered either if the original `num_channels` and `num_groups` aren't compatible or if the second arg of the `min` is chosen, then if `min_channels_per_group` isn't a factor of `num_channels`.

CharlelieLrt: Oh yeah right

PR #998, opened 2025-07-02
Consolidate functionality to overrides arguments in checkpoint loading

CharlelieLrt:
# PhysicsNeMo Pull Request

## Description
🔴 PR #809 introduced a new mechanism to override `__init__`'s arguments saved in a checkpoint file when loading the checkpoint. This was achieved with `Module.from_checkpoint("chkpt.mdlus", models_args)`, where `model_args` is a dictionary of arguments that overrides those saved in the checkpoint *before the model instantiation*. However this mechanism can lead to many incompatibilities when overriding arguments that should not. It should therefore be used with extreme caution, and only in last resort.

🔵 This PR implements guardrails such that the override mechanism can only be applied to specific arguments specified in a `_overridable_args` class attribute. By default, all `__init__` arguments are protected and cannot be overridden by `Module.from_checkpoint`. In addition this, in this PR:
- New tests for `from_checkpoint` with and without overridable arguments are added.
- These changes are propagated to the CorrDiff model architectures (`SongUnet`) and CorrDiff model wrappers (`EDMPercondSuperResolution` and `Unet`). For those, the only case where overriding is necessary is the `use_apex_gn` argument, which is needed for proper instantiation of the model's layers.
- Other `__init__` argument like `profile_mode` do not require this mechanism and are therefore handled with a `@property`
- CorrDiff training recipe is updated accordingly
- Tests for the `profile_mode` property are added

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Improved documentation for diffusion models and diffusion utils.
+- Safe API to override `__init__`'s arguments saved in checkpoint file with
+  `Module.from_checkpoint("chkpt.mdlus", models_args)`.
 
 ### Changed
 
 - physicsnemo.utils.generative renamed into physicsnemo.utils.diffusion
+- In CorrDiff model wrappers (`EDMPercondSuperResolution` and `UNet`), the
+  arguments `profile_mode` and `amp_mode` cannot be overriden by
+  `from_checkpoint`. They are now properties that can be dynamically changed
+  *after* the model instantiation.
 
 ### Deprecated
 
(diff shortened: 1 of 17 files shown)


CHANGELOG.md
```diff
@@ -11,10 +11,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Improved documentation for diffusion models and diffusion utils.
+- Safe API to override `__init__`'s arguments saved in checkpoint file with
+  `Module.from_checkpoint("chkpt.mdlus", models_args)`.
 
 ### Changed
 
 - physicsnemo.utils.generative renamed into physicsnemo.utils.diffusion
+- In CorrDiff model wrappers (`EDMPercondSuperResolution` and `UNet`), the
```
pzharrington: *EDMPrecondSuperResolution

pzharrington (approved):

PR #1000, opened 2025-07-03
Solve checkpoint and validation bugs in training corrdiff unet

### LINKED ISSUE #994: 🐛[BUG]: CorrDiff training with SongUnet checkpoint support
### Version

Main

Pip

### Describe the issue

In songunet the checkpoint_threshold is computed using the number of rows, however in the forward pass the number of columns is used to activate the checkpoint. When the input image is rectangular and n_cols > n_rows checkpointing is activated always.

In the init:
`self.img_resolution = img_resolution
  if isinstance(img_resolution, int):
      self.img_shape_y = self.img_shape_x = img_resolution
  else:
      self.img_shape_y = img_resolution[0]
      self.img_shape_x = img_resolution[1]

  # set the threshold for checkpointing based on image resolution
  self.checkpoint_threshold = (self.img_shape_y >> checkpoint_level) + 1`

Then in the forward function:
`if isinstance(block, UNetBlock):
  if x.shape[-1] > self.checkpoint_threshold:
      # self.checkpoint = checkpoint?
      # else: self.checkpoint  = lambda(block,x,emb:block(x,emb))
      x = checkpoint(block, x, emb)
  else:
      # AssertionError: Only support NHWC layout.
      x = block(x, emb)`

This causes the checkpoint to be always activated in case n_cols > n_rows. Checkpoint and DDP results in a "RuntimeError: Expected to mark a variable ready only once" if the graph is not static. Using pytorch==2.6.0.

```shell

```

### Relevant log output

```shell
RuntimeError: Expected to mark a variable ready only once. This error is caused by one of the following reasons: 1) Use of a module parameter outside the `forward` function. Please make sure model parameters are not shared across multiple concurrent forward-backward passes. or try to use _set_static_graph() as a workaround if this module graph does not change during training loop.2) Reused parameters in multiple reentrant backward passes. For example, if you use multiple `checkpoint` functions to wrap the same part of your model, it would result in the same set of parameters been used by different reentrant backward passes multiple times, and hence marking a variable ready multiple times. DDP does not support such use cases in default. You can try to use _set_static_graph() as a workaround if your module graph does not change over iterations.
Parameter at index 624 with name model.dec.512x512_block4.skip.weight has been marked as ready twice. This means that multiple autograd engine  hooks have fired for this particular parameter during this iteration.
```

### Environment details

```shell

```

albertocarpentieri:
- solved bug in training to enable validation
- add use_reentrant=False to enable checkpointing and compiling in song unet
- make dimensions consistent in song unet for checkpoint_threshold computation

Closes #994 .


albertocarpentieri:
@CharlelieLrt Can you please review it?

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -336,7 +336,7 @@ def __init__(
             self.img_shape_x = img_resolution[1]
 
         # set the threshold for checkpointing based on image resolution
-        self.checkpoint_threshold = (self.img_shape_y >> checkpoint_level) + 1
+        self.checkpoint_threshold = (self.img_shape_x >> checkpoint_level) + 1
```
CharlelieLrt: I understand this modification is sufficient to make it work for your application, but there might be issues in some other cases, for example when the image has very high/low aspect ratio. IMO a more robust solution would be to use the _size of the feature map_, defined as the _geometric mean_ of `img_shape_x` and `img_shape_y`. This would require the following modifications:
1. This line should be changed into: 
```python
self.checkpoint_threshold = (math.sqrt(self.img_shape_y * img_shape_x) >> checkpoint_level) + 1
```
2. In the `forward` method, all statements `if x.shape[-1] > self.checkpoint_threshold` should be changed into:
```python
if math.sqrt(x.shape[-2] * x.shape[-1]) > self.checkpoint_threshold
```

Let me know what you think of this.

CharlelieLrt (changes requested):
LGTM, just left a suggestion to make the checkpointing more robust for high/low aspect ratio images.

physicsnemo/models/diffusion/song_unet.py
albertocarpentieri: Yes I agree, however it could make it more difficult to understand when / where the checkpoint is active, no?

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -1153,10 +1148,10 @@ def _get_positional_embedding(self):
...
             grid.requires_grad = False
         elif self.gridtype == "test" and self.N_grid_channels == 2:
-            idx_x = torch.arange(self.img_shape_y)
-            idx_y = torch.arange(self.img_shape_x)
-            mesh_x, mesh_y = torch.meshgrid(idx_x, idx_y)
-            grid = torch.stack((mesh_x, mesh_y), dim=0)  # (2, img_shape_y, img_shape_x)
+            idx_x = torch.arange(self.img_shape_x)
+            idx_y = torch.arange(self.img_shape_y)
+            mesh_y, mesh_x = torch.meshgrid(idx_y, idx_x)
+            grid = torch.stack((mesh_y, mesh_x), dim=0)  # (2, img_shape_y, img_shape_x)
```
CharlelieLrt, line 1165: Your changes are consistent with the original implementation, but I think the original implementation is wrong.
If you look at the shape of grid, I think it is transposed in comparison to the case `self.gridtype == "linear"`.

The case `self.gridtype == "test"` is only used in CI tests, but it might be better to fix the implementation + fix any test that is failing (probably not many because we have only very few tests for non-square images)

albertocarpentieri: It is not transposed, torch.meshgrid and numpy meshgrid have different default grid layouts

CharlelieLrt: Oh right, didn't notice this one used meshgrid from torch

CharlelieLrt (approved):
LGTM

PR #1006, opened 2025-07-08
Add conditional loss for precip training and option to disable skip connection in Symmetric ConvNeXt block

raulantmor:
# Precipitation model PR

This pull request incorporates two features necessary for training a precipitation diagnostic model with the physicsnemo framework.
1. addition of a conditional loss function in healpix_loss.py that weights cases of zero precipitation less than cases of non-zero precipitation
2. functionality to remove the skip connection in the Symmetric ConvNeXt block via config argument, which is necessary for training the diagnostic model

@pzharrington - Please review when you have a moment. Thanks!


pzharrington:
Hi @raulantmor -- can you submit this PR against `dev` branch of AtmosSci-DLESM/modulus-uw rather than NVIDIA/physicsnemo? It will be easier to review/approve there, and we can upstream it to physicsnemo main later on. Thanks for putting it together!

PR #1008, opened 2025-07-10
Cleanup on SDF functions for maintainability

peterdsharpe:
Small PR that cleans up the SDF function for maintainability.

# PhysicsNeMo Pull Request

## Checklist

### DIFF
--- a/physicsnemo/utils/sdf.py
+++ b/physicsnemo/utils/sdf.py
@@ -152,7 +152,9 @@ def signed_distance_field(
         indices=wp.array(mesh_indices, dtype=wp.int32, device=device),
     )
 
-    N = len(input_points)
+    warp_input_points = wp.array(input_points, dtype=wp.vec3f, device=device)
+
+    N = len(warp_input_points)
 
     sdf = wp.empty(shape=(N,), dtype=wp.float32, device=device)
     sdf_hit_point = wp.empty(shape=(N,), dtype=wp.vec3f, device=device)
@@ -163,7 +165,7 @@ def signed_distance_field(
         dim=N,
         inputs=[
             mesh.id,
-            wp.array(input_points, dtype=wp.vec3f, device=device),
+            warp_input_points,
             max_dist,
             sdf,
             sdf_hit_point,


physicsnemo/utils/sdf.py
```diff
@@ -112,41 +135,32 @@ def signed_distance_field(
...
-    # Get the current warp device string (e.g., "cuda:2")
-    try:
-        device = str(wp.get_device())
-    except Exception:
-        device = "cuda"
+    if return_cupy is None:
+        return_cupy = any(
+            isinstance(arr, cp.ndarray)
+            for arr in (mesh_vertices, mesh_indices, input_points)
+        )
```
coreyjadams, line 145: So much better haha

physicsnemo/utils/sdf.py
```diff
@@ -112,41 +135,32 @@ def signed_distance_field(
...
-        return_cupy = True
-    if isinstance(mesh_indices, cp.ndarray):
-        return_cupy = True
-    if isinstance(input_points, cp.ndarray):
-        return_cupy = True
-
-    # Convert numpy to warp arrays:
-    mesh_vertices = wp.array(mesh_vertices, dtype=wp.vec3, device=device)
-    mesh_indices = wp.array(mesh_indices, dtype=wp.int32, device=device)
+    wp.init()
```
coreyjadams, line 147: I know it was in the function before;  that said, I wonder if this call has perf impacts and should it be outside the function?  I would assume it's ok.

No action needed just pointing this out.

(Honestly I think `import warp as wp` should do the init ...)

physicsnemo/utils/sdf.py
```diff
@@ -112,41 +135,32 @@ def signed_distance_field(
...
 
-    mesh = wp.Mesh(mesh_vertices, mesh_indices)
+    N = len(input_points)
 
-    sdf = wp.zeros(shape=sdf_points.shape, dtype=wp.float32, device=device)
-    sdf_hit_point = wp.zeros(shape=sdf_points.shape, dtype=wp.vec3f, device=device)
-    sdf_hit_point_id = wp.zeros(shape=sdf_points.shape, dtype=wp.int32, device=device)
+    sdf = wp.empty(shape=(N,), dtype=wp.float32, device=device)
+    sdf_hit_point = wp.empty(shape=(N,), dtype=wp.vec3f, device=device)
+    sdf_hit_point_id = wp.empty(shape=(N,), dtype=wp.int32, device=device)
```
coreyjadams, line 161: Does the sdf kernel below hit every point here?  or is it possible some of these points won't be visited by warp?

If its possible that not every single output location will be assigned in the kernel, we should init to the default values and not empty tensors, which is probably undefined in values.

coreyjadams, line 161: Another comment on these tensors too.  They are the output data, which has to be converted from warp.  On the CPU, I don't care, but on the GPU it makes more sense to initialize them in the host framework.  For example:
- cupy inputs -> cupy outputs, init these in cupy and cast to warp via dlpack memory sharing.
- torch inputs -> torch outputs (yeah, we should support this too probably), init in torch and transfer to wp with memory sharing again.

It's not totally an idle point.  Warp and Cupy and Torch all use different memory management and none of them are cudaMalloc directly.  They're all caching memory managers.  So using the "native" memory management makes more sense.

On cpu it doesn't matter.

physicsnemo/utils/sdf.py
```diff
@@ -156,25 +170,20 @@ def signed_distance_field(
...
-            return (sdf.numpy(), sdf_hit_point.numpy())
-        elif include_hit_points_id:
-            return (sdf.numpy(), sdf_hit_point_id.numpy())
+    def convert(array: wp.array) -> np.ndarray | cp.ndarray:
+        """Converts a Warp array to CuPy/NumPy based on the `return_cupy` flag."""
+        if return_cupy:
+            return cp.asarray(array)
         else:
-            return sdf.numpy()
+            return array.numpy()
```
coreyjadams, line 183: Good use case for lambdas?

```
convert = lambda arr : cp.asarray(arr) if return_cupy else arr.numpy()
```

coreyjadams (approved):
I think this looks good.  I think we should consider a pytorch interface to this function in the near future too (and I can explain why, I promise :) ).

No reason to hold this up though.

physicsnemo/utils/sdf.py
peterdsharpe: PEP8 recommends "Always use a def statement instead of an assignment statement that binds a lambda expression directly to an identifier"

https://docs.astral.sh/ruff/rules/lambda-assignment/

physicsnemo/utils/sdf.py
peterdsharpe: I'm not familiar enough with the Warp internals to say for sure, but one would hope that repeated `init` calls should be skipped after the first one.

> (Honestly I think import warp as wp should do the init ...)

I agree, yeah. I suppose maybe the goal of this design pattern is to be able to defer initialization, if GPU resources change between import and first Warp call, but it's tough to imagine a practical case where that would happen (while still living in the same Python environment)...

physicsnemo/utils/sdf.py
peterdsharpe: Yes, the SDF kernel will overwrite every element of this array.

PR #1012, opened 2025-07-11
CorrDiff bugfix: Removed unnecessary check in args overriding

CharlelieLrt:
# PhysicsNeMo Pull Request

## Description
 PR #998 introduced a bug in CorrDiff `train.py` due to an unnecessary check on the overridable arguments of the `UNet` wrapper. This PR removes this check.

## Checklist


physicsnemo/models/module.py
```diff
@@ -177,10 +177,6 @@ def _override_args(
                 raise ValueError(
                     f"Argument '{key}' cannot be overridden for " f"{cls.__name__}."
                 )
-            if key not in args:
```
Alexey-Kamenev, line 180: Just curious, would it make sense to replace exception with logging a warning rather than removing this check altogether?

CharlelieLrt: @Alexey-Kamenev very good point! I think a warning totally makes sense, I'll update the PR

PR #1020, opened 2025-07-18
Fea transient mesh datapipe

loliverhennigh:
# PhysicsNeMo Pull Request

## Description
Transient mesh datapipe that support. Also support for vtm files.


physicsnemo/datapipes/cae/mesh_datapipe.py
```diff
@@ -70,8 +76,8 @@ class MeshDatapipe(Datapipe):
         If provided, the statistics are used to normalize the attributes
     batch_size : int, optional
         Batch size, by default 1
-    num_steps : int, optional
```
loliverhennigh, line 73: Is this ever needed? This seems like copy pasta from ERA5 or something?

PR #1022, opened 2025-07-18
Update Transolver + Add Transolver external_aero example

coreyjadams:
# PhysicsNeMo Pull Request
This PR updates the transolver model and introduces an external aero example that uses it.  Summary of changes to the model code:
- Transolver is extended to irregular meshes and 3D data, previously this wasn't actually usable.
- Transolver's interface is improved for readability.
- The PhysicsAttention layer has been consolidated to improve readability and reduce code duplication between 2D, 3D, and irregular inputs.
- The PhysicsAttention layer has been modified in several cases:
  - Using einops rearrange to manipulate data shapes
  - Using matmuls directly instead of einsum
  - Reordering the weight normalization and projection; this improves numerical stability for lower precisions.
- The Transolver_block implementation can now use transformer_engine for most layers.

Additionally, the darcy_transolver example was updated to accomodate these changes, and still converges like the paper result.

A summary of the external aero example:
- The model can be trained on the physicsnemo curator outputs from domino; if you have a dataset that works for domino in zarr format, it can work for transolver.
- The model uses an irregular mesh, and only surface OR volume data can be used at a given time.
- The surface data uses the mesh centers + normals, the volume data uses just mesh centers.  SDF could be added but hasn't been.
- The training script supports fp32, bf16, fp16, and fp8 "in principle".  In reality, only fp32 and bf16 are known to be stable for the surface data.  Volumetric experiments are ongoing.
- The dataloading is ready to use domain parallelism, but the training script has not enabled it yet.  The model should be uniquely scalable due to the physics-attention state projections.
- the training normalizes the targets to mean 0, std 1, but un-normalizes the targets and predictions for relative L2 calculations.

There is not yet a good interface for inference with a trained model.  For the surface data, considering the simplicity of the preprocessing and reuse of domino inputs, it should be straightforward.

## Checklist


physicsnemo/models/transolver/transolver.py
```diff
@@ -56,7 +72,9 @@
 
 
 class MLP(nn.Module):
-    def __init__(self, n_input, n_hidden, n_output, n_layers=1, act="gelu", res=True):
+    def __init__(
+        self, n_input, n_hidden, n_output, n_layers=1, act="gelu", res=True, use_te=True
```
mnabian, line 77: `n_layers` is not used here, right? 
I'm curious, why did you change this MLP to have a single hidden layer?

physicsnemo/models/transolver/transolver.py
```diff
@@ -102,92 +124,284 @@ def __init__(
...
+    mechanism replacing the standard attention mechanism.
+
+    For more architecture details, see: https://arxiv.org/pdf/2402.02366 and https://arxiv.org/pdf/2502.02414
+
+    Transolver can work on structured or unstructured data points as a model construction choice:
+    - unstructured data (like a mesh) should provide some sort of positional encoding to accompany inputs
+    - structured data (2D and 3D grids) can provide positional encodings optionally
+
+    When constructing Transolver, you can choose to use "unified position" or not.  If you select "unified
+    position" (`unified_pos=True`), then
```
mnabian, line 248: Incomplete sentence

physicsnemo/models/transolver/transolver.py
```diff
@@ -102,92 +124,284 @@ def __init__(
...
+    Length 2 tuples are assumed to be image-like, length 3 tuples are assumed to be 3D voxel like.
+    Other structured shape sizes are not supported.  Passing a structured_shape of None assumes irregular data.
+
+    Output shape will have the same spatial shape as the input shape, with potentially more features
+
+    Also can support Transolver++ implementation.  When using the distributed algorithm
+    of Transolver++, use PhysicsNeMo's ShardTensor implementation to support automatic
+    domain parallelism and 2D parallelization (data parallel + domain parallel, for example).
+
+    Note
```
mnabian, line 260: Remove or complete this

physicsnemo/models/transolver/transolver.py
```diff
@@ -102,92 +124,284 @@ def __init__(
...
+
+        # This MLP is the initial projection onto the hidden space
+        self.preprocess = MLP(
+            mlp_input_dimension,
+            n_hidden * 2,
+            n_hidden,
+            n_layers=0,
+            res=False,
+            act=act,
+            use_te=False,
```
mnabian: Why is this hardcoded to False?

physicsnemo/models/transolver/transolver.py
```diff
@@ -34,14 +34,30 @@
 import numpy as np
 import torch
 import torch.nn as nn
-from timm.layers import trunc_normal_
+
+try:
+    import transformer_engine.pytorch as te
```
mnabian, line 40: I think we should extend the transolver test and include TE: https://github.com/NVIDIA/physicsnemo/blob/main/test/models/test_transolver.py

physicsnemo/models/transolver/transolver.py
```diff
@@ -102,92 +124,284 @@ def __init__(
...
+        from the input data + embeddings in the early preprocessing, before the
+        PhysicsAttention layers.
+    dropout : float
+        The dropout rate, applied across the PhysicsAttention Layers.  Default is 0.0
+    n_head : int
+        The number of attention heads in each PhysicsAttention Layer.  Default is 8.  Note
+        that the number of heads must evenly divide the `n_hidden` parameter to yield an
+        integer head dimension.
+    act : str
+        The activation function, default is Gelu.
```
mnabian: Gelu -> gelu

physicsnemo/models/transolver/transolver.py
```diff
@@ -249,153 +483,73 @@ def get_grid(self, batchsize=1):
...
-    # Physics informed
-    var_dim: int = 1
-    func_torch: bool = False
-    auto_grad: bool = False
-
+    def forward(
+        self,
+        fx: torch.Tensor,
+        embedding: torch.Tensor | None = None,
+        T: torch.Tensor = None,
```
mnabian: `T: torch.Tensor | None = None,`

physicsnemo/models/transolver/transolver.py
```diff
@@ -249,153 +483,73 @@ def get_grid(self, batchsize=1):
...
+                    )
+
+            if self.unified_pos:
+                # Extend the embedding to the batch size:
+                embedding = (
+                    self.embedding.repeat(embedding.shape[0], 1, 1)
+                    # .reshape(x.shape[0], -1, self.embedding_dim)
+                )
+            # Combine the embedding and functional input:
+            fx = torch.cat((embedding, fx), -1)
```
mnabian: I think this won't work if `embedding` is `None`?

physicsnemo/models/transolver/transolver.py
```diff
@@ -249,153 +483,73 @@ def get_grid(self, batchsize=1):
...
+            fx = self.preprocess(fx)
 
-    def forward(
-        self, x: torch.Tensor, fx: torch.Tensor = None, T: torch.Tensor = None
-    ) -> torch.Tensor:
-        """Forward pass.
+        if T is not None:
+            with torch.autograd.profiler.record_function("Time_Embedding"):
+                Time_emb = timestep_embedding(T, self.n_hidden).repeat(
+                    1, embedding.shape[1], 1
```
mnabian, line 534: I think this won't work if `embedding` is `None`

physicsnemo/models/transolver/transolver.py
coreyjadams: I will undo this change, but, the reasoning since you asked :)

- In the Transolver code, this is never actually set > 0.  So all of the model code was already using n_layers=0.
- I was debugging TE errors, trying to remove unneeded linear layers.

physicsnemo/models/transolver/transolver.py
coreyjadams: From debugging fp8 issues - it crashes.  I will let it follow the overall pattern.

physicsnemo/models/transolver/transolver.py
coreyjadams: I've reworked the readme nearly entirely.

physicsnemo/models/transolver/transolver.py
coreyjadams: Done!

coreyjadams:
Some updates on this PR:
- fp8 training appears to be functional and stable.  It does not, however, produce a significant speedup yet.
- No training curves or results are included in the example README, since they are still being validated.
- Some volumetric functions are present but in general volumetric training is WIP.  Its not supported yet, and the README makes that clear.
- The same can be said for domain parallelism: WIP, not supported yet.

The goal here was to have a transformer-engine leveraging model, so focus was on training stability, model compatibility, and inference.

examples/cfd/external_aerodynamics/transolver/README.md
```diff
@@ -0,0 +1,210 @@
+# Transolver for External Aerodynamics on Irregular Meshes
```
ram-cherukuri, line 1: Looks good and comprehensive overall. I am not calling out minor knit picks. We should have a section on customization - if a user needs to adapt this to a new problem, say CFD data from internal flow problem, what is the guidance in terms of where the user needs to look to customize - assume its VTK data, so can reuse the datapipe but what other elements would need modification.

examples/cfd/external_aerodynamics/transolver/README.md
```diff
@@ -0,0 +1,210 @@
...
+states back to each input point.
+
+By stacking multiple PhysicsAttention layers, the Transolver model learns to map from
+the functional input space to the output space with high fidelity. The PhysicsNeMo
+implementation closely follows the original Transolver architecture
+([https://github.com/thuml/Transolver](https://github.com/thuml/Transolver)), but
+introduces modifications for improved numerical stability and compatibility with NVIDIA
+TransformerEngine.
+
+The training workflow for Transolver leverages the same input datasets as DoMINO. For
```
ram-cherukuri, line 20: same input datasets as other models perhaps. We are specifically mentioning things in the context of DoMINO - can be avoided.

coreyjadams:
This has been divided into two PRS and merged separately.  Closing.

PR #1023, opened 2025-07-21
Refactoring and updates to Domino

RishikeshRanade:
# PhysicsNeMo Pull Request

## Description
- Added several user configs to have more control on the ball query operation
- Fixed bugs with training and inference on stencil_size = 1
- Added Unet + Attention Module in Geometry Processor
- Modified ball query projection to be distance weighted and optimized for memory

## Checklist


physicsnemo/datapipes/cae/domino_datapipe.py
```diff
@@ -166,7 +166,7 @@ class DoMINODataConfig:
     # Surface-specific variables:
     surface_variables: Optional[Sequence] = ("pMean", "wallShearStress")
     surface_points_sample: int = 1024
-    num_surface_neighbors: int = 11
+    num_surface_neighbors_surface: int = 11
```
peterdsharpe: Two notes here:

- Is this renaming redundant, given there is no `num_surface_neighbors_volume` (or similar)?
- If we proceed with the renaming, the docstring in `DoMINODataConfig` needs to be updated too (line 110 of this file) - both with the new name, but also ideally with some longer explanation for end-users about what this controls

physicsnemo/datapipes/cae/domino_datapipe.py
```diff
@@ -1448,7 +1456,7 @@ def create_domino_dataset(
             model_type=cfg.model.model_type,
             bounding_box_dims=cfg.data.bounding_box,
             bounding_box_dims_surf=cfg.data.bounding_box_surface,
-            num_surface_neighbors=cfg.model.num_surface_neighbors,
+            num_surface_neighbors_surface=cfg.model.num_neighbors_surface,
```
peterdsharpe: What's the motivation for having separate names here? Having the same name across the config file and Python variables makes debugging way easier. Unless there's a compelling reason not to have identical names in this constructor mapping, I'd strongly recommend that we do so. 

(Same goes for `bounding_box_dims` vs. `cfg.data.bounding_box` above, `grid_resolution` vs. `cfg.model.interp_res`, and several others, but it's okay if we don't want to add those to the scope of this PR)

physicsnemo/models/domino/model.py
```diff
@@ -192,17 +193,35 @@ def __init__(
         if grid_resolution is None:
             grid_resolution = [256, 96, 64]
         base_neurons = model_parameters.base_neurons
+        self.fourier_features = model_parameters.fourier_features
+        self.num_modes = model_parameters.num_modes
+
+        if self.fourier_features:
+            input_features_calculated = (
+                input_features + input_features * self.num_modes * 2
```
peterdsharpe:
```suggestion
                input_features * (1 + 2 * self.num_modes)
```
We also might consider renaming `input_features` -> `n_input_features` (or similar) in this constructor for clarity, but this is a discretionary change as the `int` type signature hints at this already

physicsnemo/models/domino/model.py
```diff
@@ -192,17 +193,35 @@ def __init__(
...
+            )
+        else:
+            input_features_calculated = input_features
 
-        self.fc1 = nn.Linear(input_features, base_neurons)
+        self.fc1 = nn.Linear(input_features_calculated, base_neurons)
         self.fc2 = nn.Linear(base_neurons, int(base_neurons / 2))
```
peterdsharpe:
```suggestion
        self.fc2 = nn.Linear(base_neurons, base_neurons // 2)
```

physicsnemo/models/domino/model.py
```diff
@@ -192,17 +193,35 @@ def __init__(
...
-        self.fc3 = nn.Linear(int(base_neurons / 2), model_parameters.base_neurons_out)
+        self.fc3 = nn.Linear(int(base_neurons / 2), model_parameters.base_neurons_in)
```
peterdsharpe:
```suggestion
        self.fc3 = nn.Linear(base_neurons // 2, model_parameters.base_neurons_in)
```

physicsnemo/models/domino/model.py
```diff
@@ -192,17 +193,35 @@ def __init__(
...
     def forward(
-        self, x: torch.Tensor, radius: float = 0.025, neighbors_in_radius: int = 10
+        self,
+        x: torch.Tensor,
+        grid: torch.Tensor,
+        radius: float = 0.025,
+        neighbors_in_radius: int = 10,
     ) -> torch.Tensor:
         """
         Process and project geometric features onto a 3D grid.
```
peterdsharpe, line 225: Can we expand this docstring with details about expected array shapes for `x`, `grid`, and the return array?

physicsnemo/models/domino/model.py
```diff
@@ -360,14 +395,57 @@ def __init__(self, input_features: int, radii, model_parameters=None):
...
+                        pool_size=2,
+                        normalization="layernorm",
+                        use_attn_gate=self.self_attention,
+                        attn_decoder_feature_maps=[8 * h, 4 * h, 2 * h],
+                        attn_feature_map_channels=[4 * h, 2 * h, h],
+                        attn_intermediate_channels=4 * h,
+                        gradient_checkpointing=True,
+                    )
+                )
+            else:
```
peterdsharpe: Rather than:

```python
if geometry_rep.geo_processor.processor_type == "unet":
   <unet code>
else:
   <conv code>
```

let's guard it:
```python
if geometry_rep.geo_processor.processor_type == "unet":
   <unet code>
elif geometry_rep.geo_processor.processor_type == "conv":
   <conv code>
else:
   raise ValueError("add message here")
```

In general, we want to fail loudly when working with user-given strings from configs. If the user sets their config to "UNet", we want to error - not silently use the conv strategy.

This comment also applies below circa lines 460-547 in this file

physicsnemo/models/domino/model.py
```diff
@@ -409,11 +569,12 @@ def forward(
             x_encoding = []
             for j, p in enumerate(self.radii):
                 mapping, k_short = self.bq_warp[j](x, p_grid)
-                x_encoding_inter = self.geo_conv_out[j](k_short)
+                x_encoding_inter = self.geo_conv_out[j](k_short, p_grid)
                 # Propagate information in the geometry enclosed BBox
-                for _ in range(self.hops):
+                for i in range(self.hops):
```
peterdsharpe:
```suggestion
                for _ in range(self.hops):
```

(if unused, which appears to be the case)

physicsnemo/models/domino/model.py
```diff
@@ -646,6 +810,57 @@ def forward(self, x):
         return out
 
 
+class PositionEncoder(nn.Module):
+    """Positional encoding of point clouds"""
+
+    def __init__(self, input_features: int, model_parameters=None):
+        super(PositionEncoder, self).__init__()
```
peterdsharpe:
```suggestion
        super().__init__()
```
In Python 3, `super()` implicitly does this. (In addition to readability, this notation is slightly preferred in some edge cases where we further extend `PositionEncoder` but don't overwrite the child's `__init__`)

physicsnemo/models/domino/model.py
```diff
@@ -646,6 +810,57 @@ def forward(self, x):
         return out
 
 
+class PositionEncoder(nn.Module):
```
peterdsharpe, line 805: Given that this just looks like a vanilla MLP with optional concatenation of fourier features on the input, does it need its own `nn.Module` or could we just use a standard `MLP` module?

This is a discretionary change, but one of those things that might cut down on duplicate code.

physicsnemo/models/domino/model.py
```diff
@@ -1240,7 +1537,7 @@ def calculate_solution_with_neighbors(
                         noise = surface_mesh_centers - volume_m_c
                         dist = torch.norm(noise, dim=-1, keepdim=True)
 
-                    basis_f = nn_basis[f](volume_m_c)
+                    basis_f = nn_basis[0](volume_m_c)
```
peterdsharpe: In general, this section of the model has many repeated variations of this similar code pattern:

```python
for i in range(N):
    if i == 0:
        x = process(output[i])
    else:
        x = torch.cat((x, process(output[i])))
```

These repeated `cat` operations are costly. In general, we should pre-allocate:

```python
x = torch.empty([len(N), *output.shape])
for i in range(N):
    x[i, ...] = process(output[i])
```

or, when the final size is not known (not applicable here), we should track the pieces and `cat` once:

```python
x_pieces: list[torch.tensor] = []
for i in range(N):
    x_pieces.append(process(output[i]))
x = torch.cat(x_pieces)
```

It's very rare that we should be doing conditionals with `if i == 0` inside a for-loop (usually, this indicates we should be pre-allocating something before the loop)

examples/cfd/external_aerodynamics/domino/src/conf/config.yaml
```diff
@@ -145,16 +154,21 @@ model:
...
+    fourier_features: true
+    num_modes: 5
   geometry_local: # Hyperparameters for local geometry extraction
-    volume_neighbors_in_radius: [64, 128, 256] # [64, 128]
-    surface_neighbors_in_radius: [64, 128, 256] # [64]
-    volume_radii: [0.05, 0.25, 1.0] # [0.05. 0.1]
-    surface_radii: [0.05, 0.25, 1.0] # [0.05]
+    volume_neighbors_in_radius: [32, 64, 128] # [64, 128]
+    surface_neighbors_in_radius: [16, 32, 64] # [64]
+    # volume_radii: [0.1, 0.5, 2.0] # [0.05. 0.1]
```
peterdsharpe:
```suggestion
```

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -103,10 +103,12 @@ def loss_fn(
     num = torch.sum(mask * (output - target) ** 2.0, dims)
     if loss_type == "rmse":
         denom = torch.sum(mask * target**2.0, dims)
+        loss = torch.mean(torch.sqrt(num) / torch.sqrt(denom))
```
peterdsharpe:
```suggestion
        loss = torch.mean(torch.sqrt(num / denom))
```

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -103,10 +103,12 @@ def loss_fn(
...
     else:
```
peterdsharpe:
```suggestion
    elif loss_type == "mse":
```

And add `else: raise ValueError("...")`

physicsnemo/datapipes/cae/domino_datapipe.py
```diff
@@ -744,7 +751,8 @@ def preprocess_surface(self, data_dict, core_dict, center_of_mass, s_min, s_max)
             if self.config.normalize_coordinates:
                 core_dict["surf_grid"] = normalize(core_dict["surf_grid"], s_max, s_min)
                 surface_coordinates = normalize(surface_coordinates, s_max, s_min)
-                surface_neighbors = normalize(surface_neighbors, s_max, s_min)
+                if self.config.num_surface_neighbors_surface > 1:
+                    surface_neighbors = normalize(surface_neighbors, s_max, s_min)
```
peterdsharpe: Why only normalize in the case of multiple surface neighbors? Given the flag name `normalize_coordinates`, this seems like it could lead to unexpected behavior for end users

physicsnemo/models/domino/model.py
```diff
@@ -333,15 +358,21 @@ class GeometryRep(nn.Module):
...
 
     The combined encoding enables the model to reason about both local and global
     geometric properties.
     """
 
-    def __init__(self, input_features: int, radii, model_parameters=None):
+    def __init__(
+        self,
+        input_features: int,
+        radii,
```
peterdsharpe:
```suggestion
        radii: Sequence[float],
```
and `from typing import Sequence`

physicsnemo/models/domino/model.py
```diff
@@ -360,14 +395,57 @@ def __init__(self, input_features: int, radii, model_parameters=None):
                 BQWarp(
```
peterdsharpe, line 403: Above, on the line `for j, p in enumerate(radii):`

It looks like `p` isn't used, so why use the `enumerate`?

Preferred:
```
for i in range(len(radii)):
```

Or:
```
for radius, n_neighbors in zip(radii, neighbors_in_radius):
```

physicsnemo/models/domino/model.py
```diff
@@ -833,26 +1129,31 @@ def __init__(
...
+            neighbors_in_radius=model_parameters.geometry_rep.geo_conv.volume_neighbors_in_radius,
             model_parameters=model_parameters,
         )
 
         # Basis functions for surface and volume
         base_layer_nn = model_parameters.nn_basis_functions.base_layer
         if self.output_features_surf is not None:
             self.nn_basis_surf = nn.ModuleList()
-            for _ in range(self.num_variables_surf):
+            for _ in range(1):  # Have the same basis function for each variable
```
peterdsharpe: Why keep the for loop here?

physicsnemo/models/domino/model.py
```diff
@@ -862,7 +1163,7 @@ def __init__(
 
         if self.output_features_vol is not None:
             self.nn_basis_vol = nn.ModuleList()
-            for _ in range(self.num_variables_vol):
+            for _ in range(1):  # Have the same basis function for each variable
```
peterdsharpe: For loop should be eliminated in this case

physicsnemo/models/domino/model.py
```diff
@@ -880,25 +1181,20 @@ def __init__(
             else:
                 inp_pos_vol = 7 if model_parameters.use_sdf_in_basis_func else 3
 
-            self.fc_p_vol = nn.Linear(inp_pos_vol, position_encoder_base_neurons)
+            # self.fc_p_vol = nn.Linear(inp_pos_vol, position_encoder_base_neurons)
```
peterdsharpe:
```suggestion
```

physicsnemo/datapipes/cae/domino_datapipe.py
RishikeshRanade: My bad. Thanks for catching this. This change is not required. Not sure what I did to add that extra _surface tag but I will remove it.

physicsnemo/datapipes/cae/domino_datapipe.py
RishikeshRanade: Thanks, that makes sense. I will address this this part of another PR.

physicsnemo/models/domino/model.py
RishikeshRanade: We are working on this as part of another PR for all the sub modules.

physicsnemo/models/domino/model.py
RishikeshRanade: Thanks, I have added this.

physicsnemo/models/domino/model.py
RishikeshRanade: There is a plan to expand this. It was an MLP before, I packaged it as a class in expectation of that change.

physicsnemo/models/domino/model.py
RishikeshRanade: This should not be range(1). I have fixed it in my commit.

physicsnemo/models/domino/model.py
RishikeshRanade: This is a bug. This should not be range(1). I have fixed it in my commit.

physicsnemo/models/domino/model.py
RishikeshRanade: This is fixed everywhere now.

examples/cfd/external_aerodynamics/domino/src/train.py
RishikeshRanade: Done

physicsnemo/datapipes/cae/domino_datapipe.py
RishikeshRanade: That's right. Fixed it.

physicsnemo/models/domino/model.py
RishikeshRanade: Thanks. This make sense. I will add this as part of a different PR.

examples/cfd/external_aerodynamics/domino/src/test.py
```diff
@@ -397,13 +397,16 @@ def main(cfg: DictConfig):
 
     dirnames = get_filenames(input_path)
     dev_id = torch.cuda.current_device()
-    num_files = int(len(dirnames) / dist.world_size)
+    num_files = int(len(dirnames) / dist.world_size) + 1
```
peterdsharpe, line 400: Where does this +1 come from? Fine if this is needed; just ensuring this is intentional.

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -302,7 +306,7 @@ def compute_loss_dict(
         # Needs to be taken from the dataset
         stream_velocity = batch_inputs["global_params_values"][:, 0, :]
 
-        loss_surf = loss_fn_surface(
+        loss_surf = loss_fn(
```
peterdsharpe: This is a pretty big change, is this intentional?

If so, I believe there are no longer any remaining uses of the `loss_fn_surface()` function, so we should:

1. Delete `loss_fn_surface()`
2. Remove the reference to `loss_fn_surface()` in `examples/cfd/external_aerodynamics/domino/README.md`, line 356

If not, we should revert this.

physicsnemo/models/domino/model.py
peterdsharpe: Sounds good, thanks!

physicsnemo/models/domino/model.py
```diff
@@ -352,22 +393,69 @@ def __init__(self, input_features: int, radii, model_parameters=None):
...
+                            output_filters=geometry_rep.geo_conv.base_neurons_out,
+                            model_parameters=geometry_rep.geo_processor,
+                        ),
+                    )
+                )
+            else:
+                raise ValueError("Invalid prompt. Specify unet or conv ...")
 
         self.geo_conv_out = nn.ModuleList()
         for j, p in enumerate(radii):
```
peterdsharpe: Neither `j` nor `p` appear used in the loop body. This should be:

```python
self.geo_conv_out.extend(GeoConvOut(...) for _ in range(len(radii))
```

physicsnemo/models/domino/model.py
peterdsharpe: Got it, sounds good!

peterdsharpe (changes requested):
Looking great! All prior issues look good to me. Only one comment on new changes, regarding the change from `loss_surface_fn` to `loss_fn` in `train.py`:
- If this is intended, we can remove `loss_surface_fn`, as no uses remain
- If this is not intended, we should revert

Other than that looks good to go!

examples/cfd/external_aerodynamics/domino/src/test.py
RishikeshRanade: Yes this is a simple fix for when the total number of test files is not divisible by 8 or less than 8.

examples/cfd/external_aerodynamics/domino/src/train.py
RishikeshRanade: This was an experiment and forgot to revert it. Good catch.

physicsnemo/models/domino/model.py
RishikeshRanade: Done

peterdsharpe (approved):
LGTM!

Issue #890, opened 2025-05-13
🐛[BUG]: DeprecationWarning: Python 3.14 will, by default, filter extracted tar archives and reject files or modify their metadata. Use the filter argument to control this behavior.


simonbyrne:
### Version

25.03 container

### On which installation method(s) does this occur?

Docker

### Describe the issue

When loading a trained model using `Module.from_checkpoint`, I get the following warning:

```
DeprecationWarning: Python 3.14 will, by default, filter extracted tar archives and reject files or modify their metadata. Use the filter argument to control this behavior.
```
See https://peps.python.org/pep-0706/ for details.

### Minimum reproducible example

```shell

```

### Relevant log output

```shell

```

### Environment details

```shell

```

loliverhennigh:
@simonbyrne, thanks for the report. I did not know this about 3.14. Taking a look now...

loliverhennigh:
This should fix it.

PR #1024, opened 2025-07-21
Fea tar filter deprecation fix

loliverhennigh:
# PhysicsNeMo Pull Request
Fixes deprecation warning for extract all tar.

## Description
https://github.com/NVIDIA/physicsnemo/issues/890


physicsnemo/models/module.py
```diff
@@ -404,9 +404,18 @@ def load(
 
             # Open the tar file and extract its contents to the temporary directory
             with tarfile.open(cached_file_name, "r") as tar:
-                tar.extractall(
-                    path=local_path, members=list(Module._safe_members(tar, local_path))
+                # Safely extract while supporting Python versions < 3.12 that lack the
+                # ``filter`` keyword.  Starting with 3.12, ``filter="data"`` is the
+                # recommended way to avoid unsafe members; in older versions we fall
+                # back to the explicit ``_safe_members`` logic.
```
NickGeneva: one question, "we fall back to the explicit ``_safe_members`` logic"

what does this mean here?

NickGeneva (approved):
Makes sense to me, data filter is more secure than None too for python 3.12 - 3.13

The default will then be "data" in python 3.14
https://docs.python.org/3/library/tarfile.html#tarfile-extraction-filter

PR #1034, opened 2025-07-28
Add capability to introduce PDE losses to DoMINO

ktangsali:
# PhysicsNeMo Pull Request

## Description
Adds ability to introduce Physics losses to DoMINO training using PhysicsNeMo-Sym

Corresponding updates to Sym: https://github.com/NVIDIA/physicsnemo-sym/pull/254

## Checklist


examples/cfd/external_aerodynamics/domino/src/conf/config.yaml
```diff
@@ -168,7 +170,9 @@ train: # Training configurable parameters
   sampler:
     shuffle: true
     drop_last: false
-  checkpoint_dir: /user/models/ # Use only for retraining
+  checkpoint_dir: /lustre/rranade/models/ # Use only for retraining
```
peterdsharpe:
```suggestion
  checkpoint_dir: /user/models/ # Use only for retraining
```
We should never be hardcoding paths that are user-specific.

Instead, we should use generic paths, and then when we kick off training runs on EOS, we should *ourselves* map paths to generic path names inside the container.

examples/cfd/external_aerodynamics/domino/README.md
```diff
@@ -191,6 +191,154 @@ launch overhead at the cost of more memory use.  For non-sharded
...
+
+The computation of Physics residuals is supported using the PhysicsNeMo-Sym
+library. Install it using
+
+```bash
+pip install "Cython"
+pip install "nvidia-physicsnemo.sym>2.1.0" --no-build-isolation
+```
+
+To execute the training using physics losses, run the `train.py` with below
```
peterdsharpe: change to "with the configuration below"

examples/cfd/external_aerodynamics/domino/src/conf/config.yaml
```diff
@@ -79,7 +79,7 @@ data: # Input directory for training and validation data
   bounding_box_surface: # Bounding box dimensions for car surface
     min: [-1.1, -1.2 , -0.32]
     max: [4.5 , 1.2  , 1.2]
-  gpu_preprocessing: true
+  gpu_preprocessing: false
```
peterdsharpe: Do we want to default this to false? Remember that almost-all end-users will use this config as a template, so we always want to leave it with best-practices settings as the default. For changes needed for one-off experiments, it's fine to have local uncommitted changes!

examples/cfd/external_aerodynamics/domino/src/conf/config.yaml
peterdsharpe: Or if you want to make changes to paths, that's fine - but these should not be committed unless these are intended to be globally-applicable to all end-users.

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -33,7 +33,7 @@
 import torch
 import torchinfo
 
-from typing import Literal
+from typing import Literal, Any, Optional, Tuple
```
peterdsharpe: `Optional` should not be imported, as any optional arguments should be `Type | None` rather than `Optional[Type]`.

`Tuple` should not be imported, as we should be using the builtin `tuple` for type hinting instead.

(Both suggestions are consistent with Ruff formatting guidance)

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -72,6 +72,19 @@
 
 from physicsnemo.utils.profiling import profile, Profiler
 
+
+def get_physics_imports(add_physics_loss):
```
peterdsharpe: Two changes:

1. Is this function necessary? Can we instead just have this logic inline where imports are needed? (Remember that Python's builtin `importlib` caches, so don't be shy about putting in import statements at point-of-use - it's nearly-free from a performance perspective.) It's a quite odd function type signature to have a boolean-dependent return that's either [Module, Type, Type] or [None, None, None]; I think it'd be clearer to just do this inline.
2. If we do keep this function, the `add_physics_loss` should be type-hinted.

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -81,6 +94,11 @@ def loss_fn(
     target: torch.Tensor,
     loss_type: Literal["mse", "rmse"],
     padded_value: float = -10,
+    first_deriv: Optional[torch.nn.Module] = None,
```
peterdsharpe:
```suggestion
    first_deriv: torch.nn.Module | None = None,
```
Same change to be made on other optional args

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -89,9 +107,14 @@ def loss_fn(
...
         padded_value: Value used for padding in the tensor
+        first_deriv: First derivative calculator (only used if add_physics_loss=True)
+        ns_eqn: Navier-Stokes equations (only used if add_physics_loss=True)
+        bounding_box: Bounding box for normalization (only used if add_physics_loss=True)
+        vol_factors: Volume factors for normalization (only used if add_physics_loss=True)
+        add_physics_loss: Whether to add physics-based loss terms
 
     Returns:
-        Calculated loss as a scalar tensor
+        Calculated loss as a scalar tensor or tuple of losses if physics loss is enabled
```
peterdsharpe:
```suggestion
        Calculated loss as a scalar tensor or tuple of losses if `add_physics_loss` is True
```
Minor change; a little more explicit for readability

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -100,13 +123,161 @@ def loss_fn(
...
         dims = None
 
-    num = torch.sum(mask * (output - target) ** 2.0, dims)
-    if loss_type == "rmse":
-        denom = torch.sum(mask * target**2.0, dims)
+    if not add_physics_loss:
+        num = torch.sum(mask * (output - target) ** 2.0, dims)
+        if loss_type == "rmse":
+            denom = torch.sum(mask * target**2.0, dims)
+        else:
```
peterdsharpe:
```suggestion
        elif loss_type == "mse":
```
And later below, add a quick `else: raise ValueError(f"Bad value of {loss_type=}")`

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -100,13 +123,161 @@ def loss_fn(
...
+        du = (
+            grad_neighbors_center[:, 0:1]
+            - grad_neighbors_center[:, neighbor_ids_tensor]
+        )
+        dv = (
+            coords_total_unnormalized[:, 0:1]
+            - coords_total_unnormalized[:, neighbor_ids_tensor]
+        )
+
+        ggrads_center = first_deriv.forward(
```
peterdsharpe: Can you add a comment about what `ggrads` is? Why the double-g naming convention? (Fine to keep this, but let's explain it)

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -100,13 +123,161 @@ def loss_fn(
...
+
+        # Get the outputs on the original nodes
+        fields_center_unnormalized = output_total_unnormalized[:, 0, :]
+        grad_center = grad_neighbors_center[:, 0, :, :]
+        grad_grad_uvw_center = ggrad_center[:, :, :9]
+
+        nu = 1.507 * 1e-5
+
+        dict_mapping = {
+            "u": fields_center_unnormalized[:, 0:1],
```
peterdsharpe:
```suggestion
            "u": fields_center_unnormalized[:, [0]],
```
If you want to index an array and keep the dimension, it's more readable to index it like this.

(Comment also applies to subsequent entries of `dict_mapping`.)

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -100,13 +123,161 @@ def loss_fn(
...
+
+        weighted_continuity = weight_continuity * torch.abs(continuity)
+        weighted_momentum_x = weight_momentum_x * torch.abs(momentum_x)
+        weighted_momentum_y = weight_momentum_y * torch.abs(momentum_y)
+        weighted_momentum_z = weight_momentum_z * torch.abs(momentum_z)
+
+        num = torch.sum(mask * (output - target) ** 2.0, dims)
+        if loss_type == "rmse":
+            denom = torch.sum(mask * target**2.0, dims)
+        else:
```
peterdsharpe:
```suggestion
        elif loss_type == "mse":
```
And add the `else: raise ValueError(...)`

examples/cfd/external_aerodynamics/domino/src/train.py
peterdsharpe: The point is not that we expect other values (as you've done a good job type hinting it as `Literal[...]`) - it's for readers to quickly see which block corresponds to which argument, without having to go back and check the Literal.

Also, if end users accidentally give `loss_type="RMSE"`, the as-written code will silently use "mse", which is very undesirable - with the new code, this results in an early error, which is what we want.

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -356,15 +559,25 @@ def validation_step(
     loss_fn_type=None,
     vol_loss_scaling=None,
     surf_loss_scaling=None,
+    first_deriv: Optional[torch.nn.Module] = None,
```
peterdsharpe:
```suggestion
    first_deriv: torch.nn.Module | None = None,
```
And same for below

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -265,7 +436,12 @@ def compute_loss_dict(
     integral_scaling_factor: float,
     surf_loss_scaling: float,
     vol_loss_scaling: float,
-) -> tuple[torch.Tensor, dict]:
+    first_deriv: Optional[torch.nn.Module] = None,
```
peterdsharpe:
```suggestion
    first_deriv: torch.nn.Module | None = None,
```
And same for below

examples/cfd/external_aerodynamics/domino/README.md
```diff
@@ -191,6 +191,154 @@ launch overhead at the cost of more memory use.  For non-sharded
 training, the `two-loop` setting is more optimal. The difference in `one-loop`
 or `two-loop` is purely computational, not algorithmic.
 
+### Training with Physics Losses
```
RishikeshRanade, line 192: Need a section on the steps a user needs to follow to add other equations and physical constraints.

examples/cfd/external_aerodynamics/domino/README.md
```diff
@@ -191,6 +191,154 @@ launch overhead at the cost of more memory use.  For non-sharded
...
+equations. Hence, even if `num_volume_neighbors` is set to 8, for the fields,
+it will sample `num_volume_neighbors (num_volume_neighbors // 2 + 1)` (in this
+case 40) total points.
+
+The results of physics addition can be found below (using the DrivAerML
+dataset). The results are computed on the design ID 419 and 439 from the
+validation set and averaged.
+
+We observe that, addition of physics losses improves the model
+predictions' ability to respect the governing laws better.
```
RishikeshRanade, line 233: Can we also add CD plots that you showed?

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -100,13 +123,161 @@ def loss_fn(
...
+            output_total, vol_factors[0], vol_factors[1]
+        )
+        coords_total_unnormalized = unnormalize(
+            coords_total, bounding_box[0], bounding_box[1]
+        )
 
-    return torch.mean(num / denom)
+        # compute first order gradients on all the nodes from the neighbors_list
+        grad_list = {}
+        for parent_id, neighbor_ids in neighbors_list.items():
```
RishikeshRanade: Can we wrap this code as a function to calculate Navier-Stokes PDE residuals? So that users can write similar functions for other PDEs instead of modifying this loss function here.

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -356,15 +559,25 @@ def validation_step(
...
+    vol_factors: Optional[torch.Tensor] = None,
+    add_physics_loss=False,
 ):
     running_vloss = 0.0
     with torch.no_grad():
         for i_batch, sample_batched in enumerate(dataloader):
             sampled_batched = dict_to_device(sample_batched, device)
 
-            with autocast(enabled=True):
+            with autocast(enabled=False):
```
RishikeshRanade, line 628: We should add this to config.yaml and provide guidance to set to False for Physics based training

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -397,6 +615,11 @@ def train_epoch(
     loss_fn_type,
     vol_loss_scaling=None,
     surf_loss_scaling=None,
+    first_deriv: Optional[torch.nn.Module] = None,
+    ns_eqn: Any = None,
```
RishikeshRanade: Lets name this argument pde instead of ns_eqn

examples/cfd/external_aerodynamics/domino/src/train.py
```diff
@@ -397,6 +615,11 @@ def train_epoch(
...
+    bounding_box: Optional[torch.Tensor] = None,
+    vol_factors: Optional[torch.Tensor] = None,
+    add_physics_loss=False,
```
RishikeshRanade, line 677: This seems redundant. Can we check if ns_eqn (or pde) is None or not?

physicsnemo/models/domino/model.py
```diff
@@ -1264,6 +1282,41 @@ def calculate_solution_with_neighbors(
 
             return output_all
 
+    def sample_sphere(self, center, r, num_points):
```
RishikeshRanade, line 1583: Are these utilities specific to domino? If yes, then can we move them to domino/utils and if not then to general utils?

physicsnemo/models/domino/model.py
```diff
@@ -1391,18 +1445,62 @@ def calculate_solution(
             return one_loop_output_all
 
         if self.solution_calculation_mode == "two-loop":
```
RishikeshRanade, line 1772: Can we specify in config.yaml that physics loss is not supported with one-loop?

examples/cfd/external_aerodynamics/domino/src/conf/config.yaml
```diff
@@ -79,7 +79,7 @@ data: # Input directory for training and validation data
   bounding_box_surface: # Bounding box dimensions for car surface
     min: [-1.1, -1.2 , -0.32]
     max: [4.5 , 1.2  , 1.2]
-  gpu_preprocessing: true
```
RishikeshRanade: Rebase with config.yaml in main. We want gpu processing to be true.

examples/cfd/external_aerodynamics/domino/src/conf/config.yaml
```diff
@@ -107,7 +107,9 @@ model:
   surface_sampling_algorithm: area_weighted # random or area_weighted
   geom_points_sample: 300_000 # Number of points to sample on STL per epoch
   surface_neighbors: true # Pre-compute surface neighborhood from input data
-  num_surface_neighbors: 7 # How many neighbors?
+  num_surface_neighbors: 7 # How many surface neighbors?
+  num_volume_neighbors: 20 # How many volume neighbors?
+  return_volume_neighbors: True # Whether to return volume neighbors or not
```
RishikeshRanade, line 112: Are return_volume_neighbors and add_physics_loss correlated in a way? Separate configs can confuse users.

examples/cfd/external_aerodynamics/domino/src/test.py
```diff
@@ -401,11 +401,12 @@ def main(cfg: DictConfig):
     dirnames_per_gpu = dirnames[int(num_files * dev_id) : int(num_files * (dev_id + 1))]
 
     pred_save_path = cfg.eval.save_path
+
     if dist.rank == 0:
         create_directory(pred_save_path)
 
     for count, dirname in enumerate(dirnames_per_gpu):
-        # print(f"Processing file {dirname}")
+        print(f"Processing file {dirname}")
```
RishikeshRanade: Let's remove this line

physicsnemo/models/domino/model.py
```diff
@@ -1264,6 +1282,41 @@ def calculate_solution_with_neighbors(
...
+        directions = directions / torch.norm(directions, dim=-1, keepdim=True)
+
+        radii = r * torch.pow(
+            torch.rand(size=(num_points, 1), device=center.device), 1 / 3
+        )
+        return center.unsqueeze(2) + directions * radii[None, None, :, :].expand(
+            -1, center.shape[1], num_points, -1
+        )
+
+    def sample_shpere_shell(self, center, r_inner, r_outer, num_points):
```
RishikeshRanade, line 1612: Information of inputs/outputs to this function will help.

RishikeshRanade (changes requested):

examples/cfd/external_aerodynamics/domino/src/conf/config.yaml
ktangsali: Rebased. This was due to a bug that got fixed in this PR: https://github.com/NVIDIA/physicsnemo/pull/1023

examples/cfd/external_aerodynamics/domino/src/conf/config.yaml
ktangsali: This was added due to a bug. The bug got fixed here: https://github.com/NVIDIA/physicsnemo/pull/1023 and hence I am reverting it back.

examples/cfd/external_aerodynamics/domino/README.md
ktangsali: Added a note. Please have a look

examples/cfd/external_aerodynamics/domino/README.md
ktangsali: Let's add them after we re-run the experiments with the latest changes? What do you think?

examples/cfd/external_aerodynamics/domino/src/train.py
ktangsali: Done!

examples/cfd/external_aerodynamics/domino/src/train.py
ktangsali: Added a note in the logs.

examples/cfd/external_aerodynamics/domino/src/train.py
ktangsali: Done

physicsnemo/models/domino/model.py
ktangsali: Done

examples/cfd/external_aerodynamics/domino/src/conf/config.yaml
ktangsali: Not really. Physics Losses requires volume_neighbors, but that is not the only scenario that they might be required.

examples/cfd/external_aerodynamics/domino/src/test.py
ktangsali: Done

examples/cfd/external_aerodynamics/domino/src/conf/config.yaml
ktangsali: Done!

examples/cfd/external_aerodynamics/domino/src/train.py
ktangsali: Thanks! Done

examples/cfd/external_aerodynamics/domino/src/train.py
ktangsali: Done!

examples/cfd/external_aerodynamics/domino/src/train.py
ktangsali: Done!

examples/cfd/external_aerodynamics/domino/src/train.py
ktangsali: Thanks! Done.

examples/cfd/external_aerodynamics/domino/src/train.py
ktangsali: Done

examples/cfd/external_aerodynamics/domino/src/train.py
ktangsali: Done

examples/cfd/external_aerodynamics/domino/src/train.py
ktangsali: Done!

examples/cfd/external_aerodynamics/domino/src/train.py
ktangsali: Done

examples/cfd/external_aerodynamics/domino/src/train.py
ktangsali: Done!

examples/cfd/external_aerodynamics/domino/src/train.py
ktangsali: Done!

examples/cfd/external_aerodynamics/domino/README.md
ktangsali: Done

examples/cfd/external_aerodynamics/domino/src/train.py
ktangsali: Umm, I don't think that is a good idea. Being explicit is better than silently adding the PDEs. This will allow control via flag too.

physicsnemo/models/domino/model.py
ktangsali: I think these are very specific to the model and not to the domino training recipe. I recommend keeping them here.

physicsnemo/models/domino/model.py
ktangsali: Added.

physicsnemo/models/domino/model.py
```diff
@@ -1107,6 +1108,23 @@ def __init__(
         self.encode_parameters = model_parameters.encode_parameters
         self.geo_encoding_type = model_parameters.geometry_encoding_type
 
+        if hasattr(model_parameters, "num_volume_neighbors"):
```
peterdsharpe, line 1111: Don't worry about fixing it here, but as an FYI for the future: `getattr` supports inline defaults for this:

```python
self.num_volume_neighbors = getattr(model_parameters, "num_volume_neighbors", 50)
```

ktangsali: Sounds good. Since not of the highest priority, let's get addressed in a separate PR. I believe we should take a look at all other instances (if any) in the code base for such occurrences.

examples/cfd/external_aerodynamics/domino/README.md
RishikeshRanade: Yes, we can modify this later.

RishikeshRanade (approved):
LGTM

peterdsharpe (approved):
Also LGTM!

PR #1036, opened 2025-07-29
MeshGraphNet Performance: Automaticaly Use transformer engine for LayerNorm.

coreyjadams:
# PhysicsNeMo Pull Request

## Description

This PR provides performance enhancements for mesh graph net on GPUs by automatically using transformer engine for layer norm.  Here's what happens:
- The MeshGraphMLP modules used to have a parameter to select which norm backend to use - the default was pytorch.
- Switching between the two backends (after a save and restore, for example), previously, caused issues because transformer engine adds an extra component of the state dict for fp8 normalizations.
- Now, physicsnemo has a dynamic layer norm layer - use TE if available, and fall back to torch.  This also handles the state dict differences to ensure the results are consistent after reload.  Since MGN does not use fp8, it doesn't affect anything.

A note about test changes: transformer engine is not supported on CPU.  So the tests are modified to let physicsnemo select the layernorm backend optimally for performance, except if the test is a CPU test: then it is forced to torch instead of TE.  MGN tests therefore use both torch and TE backends during testing, and both show agreement after restoring from file.

On synthetic data, the performance improvement is pretty good on large graphs.  I measured with PyG and DGL, up to 200k nodes and up to 1M+ edges (more goes out of memory).  At small graphs, the performance is similar.  For large graphs, transformer engine is better - especially during training.

Additionally, PyG is faster than DGL in all measurements.

Including performance measurements here in the PR for posterity.

## Float 32

Training
image

Inference
image

## Float16

Training
image

Inference
image

## BFloat16

Training
image

Inference
image

## Checklist


physicsnemo/models/gnn_layers/mesh_graph_mlp.py
```diff
@@ -147,21 +142,14 @@ def __init__(
...
-                    )
-                else:
-                    norm_layer = getattr(nn, norm_type)
+                warnings.warn(
+                    "The MeshGraphNet 'norm_type' argument is deprecated and will be removed in a future release."
+                    "In the future, transformer engine will be used automatically for layer norm, if it is installed."
+                    "Override this behavior by setting the PHYSICSNEMO_FORCE_TE environment variable to 'False'.",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
```
coreyjadams: Reviewers: please comment on if you want to include this deprecation warning or not?

test/distributed/distributed_utils_for_testing.py
coreyjadams, line 1: This contained tools to modify test env variables, and was moved to the physics nemo test utilities.

test/models/common/optimization.py
coreyjadams, line 1: This is actually unrelated to layer norm, but fixes an api deprecation warning from torch.

test/models/conftest.py
coreyjadams, line 1: This adds a global fixture in the models test folder.  After every test, the layer norm module is deleted - this helps make sure the tests that are validating it works on CPU/GPU or torch/TE actually use the right version.

coreyjadams:
FYI - this PR looks bigger than it is.  Many tests are updated but most are not _really_ changed: I have some env variables to force the torch version of layernorm on tests that explicitly use CPU.  This prevents transformer engine from being used on CPU models.

There are new additional tests for layer norm, however, that should get reviewed before merge.

physicsnemo/models/gnn_layers/mesh_graph_mlp.py
Alexey-Kamenev: I would remove it. I don't think anyone would want to explicitly change the LN implementation (well, maybe if TE has a bug?). Fewer knobs is better :)

physicsnemo/models/layers/layer_norm.py
```diff
@@ -0,0 +1,154 @@
...
+        TE_AVAILABLE  # make a local copy to avoid changing the global variable
+    )
+
+    # Can't use transformer engine without cuda:
+    if not torch.cuda.is_available():
+        te_available = False
+
+    # Let the users force the setting no matter what:
+    if force_te_setting is not None:
+        if force_te_setting.lower() == "true" or force_te_setting.lower() == "1":
```
Alexey-Kamenev, line 111: Nit: consider adding `else` clause. What happens if `force_te_setting` is not in {"true", "1", "false", "0"}? Or it's not possible?

physicsnemo/models/meshgraphnet/meshgraphkan.py
```diff
@@ -124,6 +124,13 @@ class MeshGraphKAN(Module):
 
     Example
     -------
+    >>> # `norm_layer` is in MGM Layers is deprecated,
```
Alexey-Kamenev: `MGM` → `MGN`?

test/models/meshgraphnet/test_meshgraphnet.py
```diff
@@ -36,42 +36,46 @@
 def test_meshgraphnet_forward(device, pytestconfig):
     """Test mehsgraphnet forward pass"""
 
-    from physicsnemo.models.meshgraphnet import MeshGraphNet
+    if device == "cpu":
```
Alexey-Kamenev: Question: it seems we have this check and subsequent `with modify_environment` in many tests, is there a way to reduce duplication by representing it as a fixture? Like `pytest` `monkeypatch` fixture?

physicsnemo/models/layers/layer_norm.py
coreyjadams: Good idea - if the env variable is set, but doesn't match the expected settings, that should be a warning at the minimum.

physicsnemo/models/meshgraphnet/meshgraphkan.py
coreyjadams: In this case, it'd be "MGK" though in this case it's referring to MeshGraphMLP ... I'll switch to writing it out fully :)

test/models/meshgraphnet/test_meshgraphnet.py
coreyjadams: Yes.  There was.  Thanks for pushing for this.

PR #1038, opened 2025-07-31
Fix load_state_dict in diffusion UNetBlocks for loading checkpoints

CharlelieLrt:
# PhysicsNeMo Pull Request

## Description
This PR fixes a bug in the `UNetBlock` class defined in `models/diffusion/layers.py`. It should allow the loading of older checkpoints where the attention parameters were defined at the root of the `UNetBlock` instead of in a in a dedicated `attn` submodule. It does so by registering a pre-load_state_dict hook that maps old keys to new keys in the state_dict. It also adds a non-regression test to cover this behavior and prevent future failures.

## Checklist


CHANGELOG.md
```diff
@@ -63,7 +63,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   warnings are raised when attempting to use Apex group norm on CPU.
 - Diffusion utils: systematic compilation of patching operations in `stochastic_sampler`
   for improved performance.
-- Diffusion utils: patch-based inference and lead time support with deterministic sampler
```
pzharrington, line 67: Did you intend to remove this?

test/models/diffusion/test_layers_unet_block.py
```diff
@@ -0,0 +1,133 @@
...
+    assert not torch.allclose(out_0, out_2, atol=1e-4)
+
+    # after loading the state_dict of model_0, model_2 should be the same as model_0
+    model_2.load_state_dict(model_0.state_dict())
+    out_2 = model_2(x, emb)
+    assert torch.allclose(out_0, out_2, atol=1e-4)
+
+    # TODO: add identical test with use_apex_gn=True once it's fixed
+    # TODO: add test with fused_conv_bias=True once it's fixed (nor now not
+    # working, expected behavior or not?)
```
pzharrington: Not sure if it will help you remember -- I tend to forget things like this -- should we capture these TODOs in an issue or do you think you'll remember to add these?

CHANGELOG.md
CharlelieLrt: Yes, @Alexey-Kamenev noticed that these lines were duplicated in the CHANGELOG

test/models/diffusion/test_layers_unet_block.py
CharlelieLrt: Right. I linked the first TODO to the relevant PR that is already open. For the second TODO, I discovered some strange behavior when trying to write tests with `fused_conv_bias=True`. Not sure if I am doing something wrong or if there is some other bug somewhere else. I am discussion this with @jialusui1102 and I'll update the test as soon as we figure out what's going on.

CharlelieLrt:
Merged into #989 to have a single PR, as both depend on each other.

PR #1039, opened 2025-08-01
Update header_check.py

### LINKED ISSUE #1009: 🐛[BUG]: License pre-commit breaks when deleting files
### Version

1.2.0

Source

### Describe the issue

If files are removed via git - when other files are added in the same commit - it breaks the pre-commit workflow.  The removed files fail the license check, though of course they shouldn't need a license if they are deleted.

One workaround is to add files in one commit, and remove files in a separate commit.  But if we have a chance to fix this we should.

```shell

```

### Relevant log output

```shell

```

### Environment details

```shell

```

coreyjadams:
Fix license header check: when files are deleted while other files are modified, it fails.  This should make sure that the license check only runs on files not marked `D` for deleted - those get filtered out of the committed files list now.

Fixes #1009 

# PhysicsNeMo Pull Request

## Checklist

### DIFF
--- a/test/ci_tests/header_check.py
+++ b/test/ci_tests/header_check.py
@@ -17,11 +17,11 @@
 
 """A script to check that copyright headers exists"""
 
-import os
 import argparse
 import fnmatch
 import itertools
 import json
+import os
 import re
 import shutil
 import subprocess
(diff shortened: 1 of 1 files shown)


ktangsali (approved):
LGTM! Thanks!

PR #1061, opened 2025-08-13
Added backend context manager to select SDPA implementation

CharlelieLrt:
# PhysicsNeMo Pull Request

### Context

PR #954 changed the computation of the `Attention` forward defined in `physicsnemo/models/diffusion/layers.py` such that it is now based on `torch.nn.scaled_dot_product_attention` (instead of the former custom python implementation). This Pytorch API offers improved performance, but it is still in beta and is known to be hardware dependent and sensitive to numerical errors.

### Changes

The present PR re-introduces the former attention computation, based on a custom python implementation, as an option. In comparison to `torch.nn.scaled_dot_product_attention`, this implementation offers worse performance, but better stability and sensitivity to numerical errors. The default forward pass of the `Attention` class remains based on `torch.nn.scaled_dot_product_attention`, but the custom python implementation can now be set with a context manager:

```python
from physicsnemo.models.diffusion.layers import Attention
from torch.nn.attention import SDPBackend

# Default: use torch.nn.functional.scaled_dot_product_attention
# without specific backend
y = model(x)

# Use custom python implementation of attention
with Attention.SDPA_backend("python"):
    y = model(x)
    
# Use specific pytorch backend
# for torch.nn.functional.scaled_dot_product_attention
with Attention.SDPA_backend(SDPBackend.FLASH_ATTENTION):
    y = model(x)
```

## Checklist

### DIFF
--- a/physicsnemo/models/diffusion/layers.py
+++ b/physicsnemo/models/diffusion/layers.py
@@ -22,12 +22,14 @@ Diffusion-Based Generative Models".
 import contextlib
 import importlib
 import math
-from typing import Any, Dict, List, Set
+from contextvars import ContextVar
+from typing import Any, Dict, Generator, List, Literal, Set
 
 import numpy as np
 import nvtx
 import torch
 from einops import rearrange
+from torch.nn.attention import SDPBackend, sdpa_kernel
 from torch.nn.functional import elu, gelu, leaky_relu, relu, sigmoid, silu, tanh
 
 from physicsnemo.models.diffusion import weight_init
@@ -656,6 +658,14 @@ class Attention(torch.nn.Module):
     -------
     torch.Tensor
         Output tensor of the same shape as input: :math:`(B, C, H, W)`.
+
+    .. important::
+
+        This implementation uses by default the implementation of scaled dot product attention (SDPA) from
+        `torch.nn.functional.scaled_dot_product_attention`<https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html>_.
+        This operator is optimized for performance, but is still in beta and its results might be affected by numerical errors.
+        To use a pure python implementation, set the SDPA backend to "python" using the
+        :meth:`SDPA_backend` context manager.
     """
 
     def __init__(
@@ -682,6 +692,7 @@ class Attention(torch.nn.Module):
                 f"`out_channels` must be divisible by `num_heads`, but got {out_channels} and {num_heads}"
             )
         self.num_heads = num_heads
+        self._sdpa_backend = None
         self.norm = get_group_norm(
             num_channels=out_channels,
             eps=eps,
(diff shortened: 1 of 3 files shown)


physicsnemo/models/diffusion/layers.py
```diff
@@ -656,6 +658,14 @@ class Attention(torch.nn.Module):
...
     torch.Tensor
         Output tensor of the same shape as input: :math:`(B, C, H, W)`.
+
+    .. important::
+
+        This implementation uses by default the implementation of scaled dot product attention (SDPA) from
+        `torch.nn.functional.scaled_dot_product_attention`<https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html>_.
+        This operator is optimized for performance, but is still in beta and its results might be affected by numerical errors.
+        To use a pure python implementation, set the SDPA backend to "python" using the
+        :meth:`SDPA_backend` context manager.
```
coreyjadams, line 668: A nit, but only because I've been working on docs so much lately: I think rst indentation expects 3, not 4, spaces in blocks like this.

physicsnemo/models/diffusion/layers.py
```diff
@@ -705,29 +716,94 @@ def __init__(
...
+        # Implementation based on torch.nn.functional.scaled_dot_product_attention
+        else:
+            q, k, v = (
+                (
+                    x1.reshape(
+                        x.shape[0], self.num_heads, x.shape[1] // self.num_heads, 3, -1
+                    )
                 )
+                .permute(0, 1, 4, 3, 2)
+                .unbind(-2)
```
coreyjadams, line 801: It might be more easy to maintain and follow the reshaping and permuting if we used einops here?

physicsnemo/models/diffusion/layers.py
```diff
@@ -705,29 +716,94 @@ def __init__(
...
-                    x.shape[0], self.num_heads, x.shape[1] // self.num_heads, 3, -1
+        # Get SDPA backend
+        _sdpa_backend = self.__class__._get_sdpa_backend()
+
+        # Custom python implementation (V1.0.1 implementation)
+        if _sdpa_backend == "python":
+            q, k, v = x1.reshape(
+                x.shape[0] * self.num_heads, x.shape[1] // self.num_heads, 3, -1
+            ).unbind(2)
+            w = AttentionOp.apply(q, k)
```
coreyjadams, line 790: I don't know if this matters in your numerical stability issues. But the pytorch sdpa appears to use a scale factor:

```
scale_factor = 1 / math.sqrt(query.size(1)) if scale is None else scale
```

On the other hand, AttentionOp uses:

```
(k / torch.sqrt(torch.tensor(k.shape[1]))).to(torch.float32)
```

Is the sequence length of q / k in these models always equal?

coreyjadams (commented):
I don't want to hold up your review @CharlelieLrt .  But a question - pytorch *already* supports a context manager to change the behavior of attention layers.  In particular, they offer a pytorch c++ implementation that claims to provide the "classic" style implementation, but in a C++ one-layer op.

[torch.backends.cuda.enable_math_sdp()](https://docs.pytorch.org/docs/stable/backends.html#torch.backends.cuda.enable_math_sdp)

Have you tried the sdpa with that backend to see if you get better numerical agreement?

Additionally, regardless of the results, could you add some details on the level of instabilities you're seeing?  I think it will help us in the future to have it saved somewhere saying "these layers diverged by X amount on CPU vs GPU0 vs GPU1", etc.

PR #1067, opened 2025-08-14
Add the inference code for x-meshgraphnet for external aerodynamics

mnabian:
# PhysicsNeMo Pull Request

## Checklist

### DIFF
--- a/examples/cfd/external_aerodynamics/xaeronet/surface_dgl/inference.py
+++ b/examples/cfd/external_aerodynamics/xaeronet/surface_dgl/inference.py
@@ -1,0 +1,480 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2024 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import os
+import sys
+import json
+import dgl
+import pyvista as pv
+import torch
+import hydra
+import numpy as np
+from sklearn.neighbors import NearestNeighbors
+from scipy.interpolate import Rbf, griddata
+from hydra.utils import to_absolute_path
+from torch.cuda.amp import GradScaler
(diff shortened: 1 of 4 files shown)


examples/cfd/external_aerodynamics/xaeronet/surface_dgl/inference.py
```diff
@@ -0,0 +1,480 @@
...
+        coordinates_denorm_np = coordinates_denorm.cpu().numpy()
+        surface_vertices_np = surface_vertices[i]
+
+        # Fit the kNN model
+        nbrs_surface = NearestNeighbors(n_neighbors=k, algorithm="ball_tree").fit(
+            coordinates_denorm_np
+        )
+
+        # Find the k nearest neighbors and their distances
+        distances, indices = nbrs_surface.kneighbors(surface_vertices_np)
```
coreyjadams, line 287: We really need to enable a pytorch-native kNN wrapper around cuml to accelerate these calls.  This shows up in DoMINO too.

Not asking a change here, just pointing it out since I've already got it on my to-do list for the next release, and we can come back and speed this up.

examples/cfd/external_aerodynamics/xaeronet/surface_dgl/inference.py
```diff
@@ -0,0 +1,480 @@
...
+        pressure_l2_error_list = []
+        shear_stress_x_l2_error_list = []
+        shear_stress_y_l2_error_list = []
+        shear_stress_z_l2_error_list = []
+        pressure_l1_error_list = []
+        shear_stress_x_l1_error_list = []
+        shear_stress_y_l1_error_list = []
+        shear_stress_z_l1_error_list = []
+        for j in range(cfg.num_partitions):
+            part = test_graphs[i][j].to(device)
```
coreyjadams, line 206: Alexey noticed a memory leak that was, I think, coming from a PyG / DGL difference in how the data is preloaded (line 130 here) and then how long it remains on the device after it's usage completes.

I doubt it will impact inference practically, since the memory needs are much lower.  But worth checking that we aren't introducing a leak here - if someone tries this with very very large graphs they might OoM after a few iterations.

Alexey-Kamenev: Corey is right - while working on DGL to PyG migration, I've noticed that pre-loading graphs often causes OOM during training.
It seems PyG/PyTorch might be doing some sort of caching which for some reason causes OOM.
Just a small refactoring, like I've done in [PyG version](https://github.com/NVIDIA/physicsnemo/blob/main/examples/cfd/external_aerodynamics/xaeronet/surface/train.py#L174), that avoids pre-loading, fixes this issue.
Since this is DGL version (and it's inference, not training), it may not have this problem. But I think fixing it is a good idea anyway: pre-loading will not scale well if the number of graphs increases.

mnabian: Let me merge this PR to unblock Sheel, since it is already approved and ci is passed. I will open a separate PR right away to fix this graph preloading.

PR #1077, opened 2025-08-18
Adding DiT 

Dibyajyoti-Chakraborty:
# PhysicsNeMo Pull Request

## Description

Added DiT to experimental  - https://github.com/facebookresearch/DiT/blob/main/models.py 
Changes from the original DiT implementation
1. Added a constructor flag to use MultiheadAttention from transformer_engine
2. Made the conditioning optional based on a flag in the constructor
3. Added capability to handle non-square images
4. Added a flag to use learnable positional embedding

Updated Mlp code in physicsnemo.layers.mlp_layers with option to use Conv2d, bias, and also a norm layer in the middle.
## Checklist


physicsnemo/experimental/models/dit/dit.py
```diff
@@ -0,0 +1,453 @@
...
+            embedding = torch.cat([embedding, torch.zeros_like(embedding[:, :1])], dim=-1)
+        return embedding
+
+    def forward(self, t):
+        t_freq = self.timestep_embedding(t, self.frequency_embedding_size)
+        t_emb = self.mlp(t_freq)
+        return t_emb
+
+
+class LabelEmbedder(nn.Module):
```
pzharrington: Maybe rename to `ConditionEmbedder` since we are intending this to be usable with more conditioning info. other than just data "labels"?

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -0,0 +1,453 @@
...
+#################################################################################
+
+class DiTBlock(nn.Module):
+    """
+    A DiT block with adaptive layer norm zero (adaLN-Zero) conditioning.
+    """
+    def __init__(self, hidden_size, num_heads,use_transformer_engine_attn, mlp_ratio=4.0, **block_kwargs):
+        super().__init__()
+        self.norm1 = nn.LayerNorm(hidden_size, elementwise_affine=False, eps=1e-6)
+        if use_transformer_engine_attn:
```
pzharrington: Looking ahead to more attention variants, e.g. natten, we may want to control this with a string rather than a boolean to support more than just torch fsdp or TE attention. Thoughts @coreyjadams ?

Dibyajyoti-Chakraborty: Agreed. It is better to change it to a 'str'. What keywords to put ?  I'm thinking like : 
attention_backbone (str): If 'transformer_engine', uses MultiheadAttention from transformer_engine. Defaults to 'timm' that uses timm Atenntion block

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -0,0 +1,453 @@
...
+        use_transformer_engine_attn: bool = False,
+        class_dropout_prob=0.1,
+        condition_dim=None,
+        learn_sigma=True,
+        embedding_type='sin-cos',
+    ):
+        """
+        Initializes the DiT model.
+
+        Args:
```
pzharrington: In general the docstrings should be formatted like other PNM docstrings, same spacing and indentation layout

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -0,0 +1,453 @@
...
+            input_size (Tuple[int, int]): Height and width of the input images. Default is (32, 32).
+            patch_size (int): The size of each image patch. Defaults to 2.
+            in_channels (int): The number of input channels. Defaults to 4.
+            out_channels (None, int): The number of output channels. Defaults to None. If None, it is either in_channels or 2*in_channels based on learn_sigma
+            hidden_size (int): The dimensionality of the transformer embeddings. Defaults to 1152.
+            depth (int): The number of transformer blocks. Defaults to 28.
+            num_heads (int): The number of attention heads. Defaults to 16.
+            mlp_ratio (float): The ratio of the MLP hidden dimension to the embedding dimension. Defaults to 4.0.
+            use_transformer_engine_attn (bool): If True, use MultiheadAttention from transformer_engine. Defaults to False.
+            class_dropout_prob (float): The dropout probability for classifier-free guidance. Defaults to 0.1.
```
pzharrington: Related to renaming `LabelEmbedding`, this should be called something like `condition_embedding_dropout`

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -0,0 +1,453 @@
...
+            condition_dim (int, optional): The number of classes for conditioning. If None, the model is unconditional. Defaults to None.
```
pzharrington: Similar comment, not necessarily a "number of classes"

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -0,0 +1,453 @@
...
+            learn_sigma (bool): If True, the model learns the variance of the diffusion process. Defaults to True.
+            embedding_type (str): The type of positional embedding to use. 'sin-cos' for fixed sinusoidal embeddings,
+                                  'positional' for learnable positional embeddings. Defaults to 'sin-cos'.
```
pzharrington: I think the learnable pos embed should be called "learnable" rather than "positional"

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -0,0 +1,453 @@
...
+            in_channels (int): The number of input channels. Defaults to 4.
+            out_channels (None, int): The number of output channels. Defaults to None. If None, it is either in_channels or 2*in_channels based on learn_sigma
+            hidden_size (int): The dimensionality of the transformer embeddings. Defaults to 1152.
+            depth (int): The number of transformer blocks. Defaults to 28.
+            num_heads (int): The number of attention heads. Defaults to 16.
+            mlp_ratio (float): The ratio of the MLP hidden dimension to the embedding dimension. Defaults to 4.0.
+            use_transformer_engine_attn (bool): If True, use MultiheadAttention from transformer_engine. Defaults to False.
+            class_dropout_prob (float): The dropout probability for classifier-free guidance. Defaults to 0.1.
+            condition_dim (int, optional): The number of classes for conditioning. If None, the model is unconditional. Defaults to None.
+            learn_sigma (bool): If True, the model learns the variance of the diffusion process. Defaults to True.
```
pzharrington: I don't see a good reason for us to support this option -- all our internal works are using an EDM or related diffusion training framework, so predicting the the variance is not needed.

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -0,0 +1,453 @@
...
+            c = t                                                         # (N, D)
+
+        for block in self.blocks:
+            x = block(x, c)                      # (N, T, D)
+        x = self.final_layer(x, c)                # (N, T, patch_size ** 2 * out_channels)
+        x = self.unpatchify(x)                   # (N, out_channels, H, W)
+        return x
+    
+if __name__ == "__main__":
+    device = "cuda" if torch.cuda.is_available() else "cpu"
```
pzharrington: This should eventually be removed, the unit tests will go under the `test` folder in physicsnemo

physicsnemo/experimental/models/dit/dit.py
Dibyajyoti-Chakraborty: Updated

physicsnemo/experimental/models/dit/dit.py
Dibyajyoti-Chakraborty: Updated

physicsnemo/experimental/models/dit/dit.py
Dibyajyoti-Chakraborty: Updated

physicsnemo/experimental/models/dit/dit.py
Dibyajyoti-Chakraborty: Updated

physicsnemo/experimental/models/dit/dit.py
Dibyajyoti-Chakraborty: Updated

physicsnemo/experimental/models/dit/dit.py
Dibyajyoti-Chakraborty: Updated

physicsnemo/experimental/models/dit/dit.py
Dibyajyoti-Chakraborty: Updated

coreyjadams:
I will take a close look tomorrow morning.  First, I love the idea of bringing new models where the API may change under an "experimental" heading. Great call.

We should chase down a couple things:
- All functions should have docstrings and type hints.  In fact, I'm surprised precommit is letting you commit this.
- We need unit tests on the functions you're committing (as much as possible, within reason of course) and some end to end tests on the model.
- The CHANGELOG should be updated.

Thanks for throwing this together so quickly!

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -0,0 +1,739 @@
...
+        Parameters
+        -----------
+        hidden_size (int): The dimensionality of the input and output.
+        num_heads (int): The number of attention heads.
+        attention_backbone (str): The attention implementation to use ('transformer_engine' or 'timm').
+        mlp_ratio (float, optional): The ratio for the MLP's hidden dimension. Defaults to 4.0.
+        **block_kwargs (Any): Additional keyword arguments for the attention layer.
+        """
+        super().__init__()
+        self.norm1 = nn.LayerNorm(hidden_size, elementwise_affine=False, eps=1e-6)
```
coreyjadams, line 341: These should default to transformer_engine LayerNorm on GPU for better performance.  You can use `physicsnemo.models.layers.layer_norm` for this, which will enable automatic selection.  Check out MeshGraphMLP for an example:

https://github.com/NVIDIA/physicsnemo/blob/main/physicsnemo/models/gnn_layers/mesh_graph_mlp.py#L147-L149

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -0,0 +1,739 @@
...
+        if attention_backbone == "transformer_engine":
+            self.attn = MultiheadAttention(
+                hidden_size=hidden_size, num_attention_heads=num_heads, **block_kwargs
+            )
+        else:
+            self.attn = Attention(
+                dim=hidden_size, num_heads=num_heads, qkv_bias=True, **block_kwargs
+            )
+
+        self.norm2 = nn.LayerNorm(hidden_size, elementwise_affine=False, eps=1e-6)
```
coreyjadams, line 351: Here too, with layer norm.

physicsnemo/models/layers/mlp_layers.py
```diff
@@ -14,31 +14,38 @@
...
     def __init__(
         self,
         in_features,
         hidden_features=None,
         out_features=None,
         act_layer=nn.GELU,
+        norm_layer=None,
+        bias=True,
         drop=0.0,
+        use_conv=False,
```
coreyjadams: What's the motivation to enable kernel_size==1 convs as the linear backend?  It's not used directly in your model, that I can find.  If it's more performant, we should look at it, but I am not sure if it is.

physicsnemo/models/layers/mlp_layers.py
```diff
@@ -14,31 +14,38 @@
...
         super().__init__()
         out_features = out_features or in_features
         hidden_features = hidden_features or in_features
-        self.fc1 = nn.Linear(in_features, hidden_features)
+        linear_layer = partial(nn.Conv2d, kernel_size=1) if use_conv else nn.Linear
+
+        self.fc1 = linear_layer(in_features, hidden_features, bias=bias)
         self.act = act_layer()
-        self.fc2 = nn.Linear(hidden_features, out_features)
+        self.norm = norm_layer(hidden_features) if norm_layer is not None else nn.Identity()
```
coreyjadams: Same question for the norm layer here: it's off by default, and not enabled in your code.  Is it needed?

physicsnemo/models/layers/mlp_layers.py
Dibyajyoti-Chakraborty: This is now the default MLP block in timm. 
(https://github.com/huggingface/pytorch-image-models/blob/main/timm/layers/mlp.py#L27)
So I decided to update it (with backwards compatibility). Future architectures can use it.

physicsnemo/models/layers/mlp_layers.py
Dibyajyoti-Chakraborty: For the same reason, this is the MLP block in timm.

physicsnemo/models/layers/mlp_layers.py
coreyjadams: Ah, the reason is to apply it to image-like models without reshaping data:

```python
    # NOTE: When use_conv=True, expects 2D NCHW tensors, otherwise N*C expected.
```

I'm not sure we should follow this convention.  We have ConvMlp and ConvMlp3d in `dpot` too, now, which are doing similar things.  Do you think this is necessary to enable the DiT architecture?

physicsnemo/experimental/models/dit/dit.py
pzharrington: @coreyjadams any idea how the performance of TE matches up against the apex fusedlayernorm? The `SongUNet` in physicsnemo uses the apex groupnorm for better perf over vanilla PyTorch. Might be nice to support all three options in `physicsnemo.models.layers.layer_norm`?

physicsnemo/experimental/models/dit/dit.py
pzharrington: Nit: still a bit of formatting to do for the docstrings -- the description should be in the following line, and indented. See e.g.: https://github.com/NVIDIA/physicsnemo/blob/77b33c5028bde9e10f4e8ef4a9fdbad9becce0fc/physicsnemo/models/meshgraphnet/meshgraphnet.py#L83

physicsnemo/experimental/models/dit/dit.py
Dibyajyoti-Chakraborty: Can't be use due to `elementwise_affine=False`

physicsnemo/experimental/models/dit/dit.py
Dibyajyoti-Chakraborty: same, can't be use due to `elementwise_affine=False`

physicsnemo/models/layers/mlp_layers.py
Dibyajyoti-Chakraborty: Not in DiT. The usual Mlp shall be fine here.

physicsnemo/models/layers/mlp_layers.py
Dibyajyoti-Chakraborty: Reverted to original

physicsnemo/models/layers/mlp_layers.py
Dibyajyoti-Chakraborty: Reverted to orginal

physicsnemo/experimental/models/dit/dit.py
Dibyajyoti-Chakraborty: Removed

test/models/test_dit.py
```diff
@@ -0,0 +1,174 @@
...
+
+import common
+import onnxruntime
+import pytest
+import torch
+
+from physicsnemo.experimental.models.dit import DiT
+
+script_path = os.path.abspath(__file__)
+sys.path.append(os.path.join(os.path.dirname(script_path), ".."))
```
pzharrington, line 29: What's going on with these?

Dibyajyoti-Chakraborty: Just to import modules from the root directory as if they were in the same directory.
 Took it from here.
https://github.com/NVIDIA/physicsnemo/blob/main/test/models/diffusion/test_dhariwal_unet.py#L23

PR #1079, opened 2025-08-19
Physicsnemo  kNN Utility

coreyjadams:
# PhysicsNeMo Pull Request

Add knn utilities with optimized backend selection.  Primarily targeting data pipes and preprocessing, this kNN is not differentiable on the distances returned.  However, you could use the indexes returned and select on the `points` tensor, and *that* is differentiable.

The knn op has several backends:
  - torch (brute force)
  - scipy.spatial.KDTree (cpu only)
  - cuml.Neighbors (GPU only)

The default backend is "auto" which will dispatch correctly.

## Checklist


physicsnemo/utils/neighbors/knn/_cuml_impl.py
```diff
@@ -0,0 +1,85 @@
...
+        indices = torch.from_dlpack(indices)
+
+        # Return torch objects.
+        return indices, distance
+
+    @knn_impl.register_fake
+    def _(
+        points: torch.Tensor, queries: torch.Tensor, k: int = 3
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        assert points.device == queries.device  # noqa S101
```
laserkelvin: Instead of the `noqa`, would you rather:

```python
if points.device != queries.device:
    raise RuntimeError("Points and queries must be on the same device")
```

or coerce (promote) one to the other?

physicsnemo/utils/neighbors/knn/_cuml_impl.py
```diff
@@ -0,0 +1,85 @@
...
+    @torch.library.custom_op("physicsnemo::knn_cuml", mutates_args=())
+    def knn_impl(
+        points: torch.Tensor, queries: torch.Tensor, k: int = 3
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        # Use dlpack to move the data without copying between pytorch and cuml:
+        points = cp.from_dlpack(points)
+        queries = cp.from_dlpack(queries)
+
+        # Construct the knn:
+        knn = cuml.neighbors.NearestNeighbors(n_neighbors=k)
```
laserkelvin, line 36: Is it okay to not be able to parametrize this at all? e.g. different `p`, distance measurements

physicsnemo/utils/neighbors/knn/_scipy_impl.py
```diff
@@ -0,0 +1,83 @@
...
+        # This reshape is to prevent scipy from eating the second dimension whten k ==1
+        return indices.reshape(queries.shape[0], k), distance.reshape(
+            queries.shape[0], k
+        )
+
+    @knn_impl.register_fake
+    def _(
+        points: torch.Tensor, queries: torch.Tensor, k: int = 3
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        assert points.device == queries.device  # noqa S101
```
laserkelvin: Same comment about `noqa` as the `cuml` module

physicsnemo/utils/neighbors/knn/_scipy_impl.py
```diff
@@ -0,0 +1,83 @@
...
+
+if SCIPY_AVAILABLE:
+    from scipy.spatial import KDTree
+
+    @torch.library.custom_op("physicsnemo::knn_scipy", mutates_args=())
+    def knn_impl(
+        points: torch.Tensor, queries: torch.Tensor, k: int = 3
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        # Use dlpack to move the data without copying between pytorch and cuml:
+        points = points.numpy()
```
laserkelvin: Should/do you need to call `detach` first, just in case?

physicsnemo/utils/neighbors/knn/_scipy_impl.py
```diff
@@ -0,0 +1,83 @@
...
+        queries = queries.numpy()
+
+        interp_func = KDTree(points)
+        distance, indices = interp_func.query(queries, k=k)
+
+        # Ensure dtype compatibility: cast distances to the dtype of queries:
+        distance = distance.astype(queries.dtype)
+
+        indices = torch.from_numpy(indices)
```
laserkelvin, line 40: Move `indices` and `distance` back to the same device as `points` or `queries`?

physicsnemo/utils/neighbors/knn/_torch_impl.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    Perform kNN search with torch.
+
+    Args:
+        points (torch.Tensor): Query points, shape ( M, D)
+        queries (torch.Tensor): Reference points, shape (N, D)
+        k (int): Number of neighbors
+
+    Returns:
+        torch.Tensor: Aggregated features at p1, shape (B, M, D_feat)
```
laserkelvin: This doesn't sound right?

physicsnemo/utils/neighbors/knn/knn.py
```diff
@@ -0,0 +1,89 @@
...
+    # auto will dispatch to scipy if points.device==cpu and cuml if points.device==cuda
+    # the brute-force torch backend will never be reached unless requested explicitly.
+
+    if backend == "auto":
+        if points.is_cuda:
+            backend = "cuml"
+        else:
+            backend = "scipy"
+
+    if backend == "scipy":
```
laserkelvin: This is more of a personal style, but the implementation might look cleaner if you were to use `match`/`case` instead; the minimum required version of `physicsnemo` is 3.10 so this should be supported.

```python
match backend:
    case "scipy":
        if points.device.type != "cpu":
            raise ValueError(f"`knn` scipy backend only supports CPU, got {points.device=}")
        method = knn_scipy
    case "cuml":
        if not points.is_cuda:  # just for consistency
            raise ValueError(f"`knn` cuml backend requires CUDA, got {points.device=}"
        method = knn_cuml
    case "torch":
        method = "knn_torch"
    case _:  # this will unlikely ever resolve but is good to catch
         raise NotImplementedError
     indices, distances = method(points, queries, k)
```

test/utils/neighbors/test_knn.py
```diff
@@ -0,0 +1,155 @@
...
+import torch
+
+from physicsnemo.utils.neighbors import knn
+from physicsnemo.utils.neighbors.knn._cuml_impl import knn_impl as knn_cuml
+from physicsnemo.utils.neighbors.knn._scipy_impl import knn_impl as knn_scipy
+
+
+@pytest.mark.parametrize("device", ["cpu", "cuda"])
+@pytest.mark.parametrize("k", [1, 5])
+@pytest.mark.parametrize("backend", ["cuml", "torch", "scipy", "auto"])
```
laserkelvin, line 27: Parametrize over data types, too?

test/utils/neighbors/test_knn.py
```diff
@@ -0,0 +1,155 @@
...
+    else:
+        op = knn_scipy
+
+    torch.library.opcheck(op, args=(points, queries, k))
+
+
+@pytest.mark.parametrize("device", ["cuda", "cpu"])  # cuml only works on CUDA
+def test_knn_comparison(device):
+    torch.manual_seed(42)
+    if device == "cuda":
```
laserkelvin: How come you only seed this run, and not the others?

physicsnemo/utils/neighbors/knn/_cuml_impl.py
coreyjadams: Fixed.

physicsnemo/utils/neighbors/knn/_cuml_impl.py
coreyjadams: I think it's OK; the defaults (euclidean distance, auto alg selection) are correct and any parameters changed might be unexpected from the high level function.

The one to consider is the cuml handle, which controls the stream.  It might be worth it to ensure this runs on the current pytorch stream.  But I think that is a problem for later.

physicsnemo/utils/neighbors/knn/_scipy_impl.py
coreyjadams: Fixed.

physicsnemo/utils/neighbors/knn/_scipy_impl.py
coreyjadams: Well, it won't hurt. Sure.

physicsnemo/utils/neighbors/knn/_scipy_impl.py
coreyjadams: They are all on CPU by design here, and calling `from_numpy` will stay on CPU.  I don't think it's necessary.

physicsnemo/utils/neighbors/knn/_torch_impl.py
coreyjadams: Good catch, thanks!

physicsnemo/utils/neighbors/knn/knn.py
coreyjadams: Done.

test/utils/neighbors/test_knn.py
coreyjadams: Arrrg cursor did this to me.  There should not be a need for a seed at all.  Fixed.

test/utils/neighbors/test_knn.py
coreyjadams: This was a good call.  I caught a bug that would have surfaced quickly: bf16 is not supported in cuml or scipy.  Scipy errors, cuml just ... core dumps.  Love it.

I now quietly promote to fp32 since it has the same dynamic range as bf16, and pass that to the worker functions.  Distances is returned in the original dtype.  Indices is int64 anyways, no change there.

physicsnemo/utils/neighbors/knn/_torch_impl.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    Perform kNN search with torch.
+
+    Args:
+        points (torch.Tensor): Query points, shape ( M, D)
+        queries (torch.Tensor): Reference points, shape (N, D)
+        k (int): Number of neighbors
+
+    Returns:
+        torch.Tensor: Aggregated features at p1, shape (N, k)
```
laserkelvin: I thought I pointed this out in the last round of reviews, but apparently not?

This docstring return signature doesn't seem correct

physicsnemo/utils/neighbors/knn/_torch_impl.py
```diff
@@ -0,0 +1,45 @@
...
+# limitations under the License.
+
+import torch
+
+
+def knn_impl(
+    points: torch.Tensor,
+    queries: torch.Tensor,
+    k: int = 3,
+) -> torch.Tensor:
```
laserkelvin: This return type annotation isn't correct

physicsnemo/utils/neighbors/knn/knn.py
```diff
@@ -0,0 +1,110 @@
...
+from ._scipy_impl import knn_impl as knn_scipy
+from ._torch_impl import knn_impl as knn_torch
+
+
+def knn(
+    points: torch.Tensor,
+    queries: torch.Tensor,
+    k: int,
+    backend: Literal["cuml", "torch", "scipy", "auto"] = "auto",
+) -> tuple[torch.Tensor]:
```
laserkelvin: Type annotation should be two `torch.Tensor`

laserkelvin (changes requested):
Sorry to have to go another round, but these are actually minor!

laserkelvin (approved):
:shipit:

PR #1087, opened 2025-08-21
Migrate Stokes MGN example to PyTorch Geometric

Alexey-Kamenev:
# PhysicsNeMo Pull Request

## Key Changes

1. Added PyG-based Stokes MGN example.
2. Renamed existing DGL-based example to `stokes_mgn_dgl`
3. Added DGL/PyG equivalency test.
4. Fixed issue with PI script.
5. Fixed issue with LR scheduler.

### Inference

image

## Checklist


examples/cfd/stokes_mgn/train.py
```diff
@@ -123,9 +131,21 @@ def __init__(self, cfg: DictConfig, dist, rank_zero_logger):
             rank_zero_logger.info("Using FusedAdam optimizer")
         except:
             self.optimizer = torch.optim.Adam(self.model.parameters(), lr=cfg.lr)
-        self.scheduler = torch.optim.lr_scheduler.LambdaLR(
-            self.optimizer, lr_lambda=lambda epoch: cfg.lr_decay_rate**epoch
+        # If lr_decay_rate is not set, calculate it based on the number of epochs
```
ktangsali, line 134: Thanks! This is much cleaner way to handle LR!

examples/cfd/stokes_mgn/conf/config.yaml
```diff
@@ -41,7 +41,8 @@ num_validation_samples: 10
 num_test_samples: 10
 
 lr: 1e-4
-lr_decay_rate: 0.99985
+# If not set, will be calculated based on the number of epochs.
+lr_decay_rate:
```
coreyjadams: Should we set this to `null` for explicit-ness?

examples/cfd/stokes_mgn/pi_fine_tuning_gnn.py
```diff
@@ -196,14 +196,11 @@ def parabolic_inflow(self, y, U_max=0.3):
...
-        out = self.model(
-            self.dgl_graph.ndata["x"], self.dgl_graph.edata["x"], self.dgl_graph
-        )
+        out = self.model(self.pyg_graph.x, self.pyg_graph.edge_attr, self.pyg_graph)
 
         # inflow points
         mask_inflow = (
-            self.dgl_graph.ndata["marker"]
-            == torch.tensor([0, 1, 0, 0, 0]).to(self.device)
+            self.pyg_graph.marker == torch.tensor([0, 1, 0, 0, 0]).to(self.device)
```
coreyjadams: maybe use `torch.tensor([0,1,0,0,0], device=self.device)` so this initializes right on the gpu and is non-blocking?

examples/cfd/stokes_mgn/pi_fine_tuning_gnn.py
```diff
@@ -212,12 +209,10 @@ def loss(self):
 
         # no-slip points
         mask_1 = (
-            self.dgl_graph.ndata["marker"]
-            == torch.tensor([0, 0, 0, 1, 0]).to(self.device)
+            self.pyg_graph.marker == torch.tensor([0, 0, 0, 1, 0]).to(self.device)
```
coreyjadams: Here too?

examples/cfd/stokes_mgn/pi_fine_tuning_gnn.py
```diff
@@ -212,12 +209,10 @@ def loss(self):
...
         ).all(dim=1)
         mask_2 = (
-            self.dgl_graph.ndata["marker"]
-            == torch.tensor([0, 0, 0, 0, 1]).to(self.device)
+            self.pyg_graph.marker == torch.tensor([0, 0, 0, 0, 1]).to(self.device)
```
coreyjadams: Here too?

examples/cfd/stokes_mgn/pi_fine_tuning_gnn.py
```diff
@@ -227,15 +222,14 @@ def loss(self):
 
         # interior points
         mask_int = (
-            self.dgl_graph.ndata["marker"]
-            == torch.tensor([1, 0, 0, 0, 0]).to(self.device)
+            self.pyg_graph.marker == torch.tensor([1, 0, 0, 0, 0]).to(self.device)
```
coreyjadams: Here too?

examples/cfd/stokes_mgn/train.py
```diff
@@ -222,10 +242,11 @@ def main(cfg: DictConfig) -> None:
     rank_zero_logger.info("Training started...")
 
     for epoch in range(trainer.epoch_init, cfg.epochs):
+        trainer.dataloader.sampler.set_epoch(epoch)
         loss_agg = 0
         for graph in trainer.dataloader:
             loss = trainer.train(graph)
-            loss_agg += loss.detach().cpu().numpy()
+            loss_agg += loss.item()
```
coreyjadams: This emits a warning without the `detach()`.  Might be work adding back in?

```
/user_data/workdir/physicsnemo-pyg-pr-1087/examples/cfd/stokes_mgn/train.py:249: UserWarning: Converting a tensor with requires_grad=True to a scalar may lead to unexpected behavior.
```

examples/cfd/stokes_mgn/train.py
Alexey-Kamenev: Thank you!

examples/cfd/stokes_mgn/conf/config.yaml
Alexey-Kamenev: Done.

examples/cfd/stokes_mgn/pi_fine_tuning_gnn.py
Alexey-Kamenev: Done. There are other places where such optimization can be done too:
```py
self.ref_u = torch.tensor(ref_u).float().to(self.device)
```
but I'll address it in a separate PR.

examples/cfd/stokes_mgn/pi_fine_tuning_gnn.py
Alexey-Kamenev: Done.

examples/cfd/stokes_mgn/pi_fine_tuning_gnn.py
Alexey-Kamenev: Done.

examples/cfd/stokes_mgn/pi_fine_tuning_gnn.py
Alexey-Kamenev: Done.

examples/cfd/stokes_mgn/train.py
Alexey-Kamenev: That's a good one! As I understand, the purpose of this warning is to make sure the result of conversion is not used accidentally somewhere down the line which expects a valid computation graph.
In this case it's a false positive since loss_agg is supposed to be just a scalar which aggregates the loss (e.g. it's not used in backprop).
However, I agree that the warning is a bit annoying, so I'm updating the change to use `detach`.

Alexey-Kamenev:
> I have one significant change to ask. The README does not appear to be updated from the DGL readme, and gives confusing instructions for installation of DGL, etc. I think we need to resolve that before merge.

Good catch - I've updated README accordingly, also fixed typos and other minor things.

coreyjadams (approved):
Looks good!

Alexey-Kamenev:
> Looks great! Thanks! Quick question - did the Physics informed inference work for the new PyTorch Geometric code for both MLP and MGN too?

Yes! Verified that both scripts are working as expected.

PR #1095, opened 2025-08-26
Speeding up distributed tests

coreyjadams:
# PhysicsNeMo Pull Request

his PR is meant to enable a fast-path for distributed tests, for checks on numerical accuracy and functionality of functions such as `scatter_v` (differentiable scatter collective), as well as `ShardTensor` layers.  PhysicsNeMo tests now come in 3 categories:

- `pytest.mark.multigpu_static` represent tests that can use a static, init-once-and-reuse test environment.
- `pytest.mark.multigpu_dynamic` represent tests that need a fresh distributed environment at run time.
- SingleGPU are everything that *isn't* marked as one of the multi-gpu types.

## Motivation

The distributed test path currently runs all distributed tests with process spawning, which takes several seconds to spawn and coordinate torch distributed.  One single test is not a big deal; thousands of tests are problematic for testing performance, making unit testing of distributed tools impractical.  Because it's slow and impractical, many distributed functionality items have tests that aren't regularly run, and this has led to functionality regressions.

Some tests *have* to have dynamic start up, to accommodate testing 2D parallelism as well as testing the distributed manager, etc.  Other tests, such as validating numerical accuracy of sharded layers, can and should use static parallelism.  With just one process launch, these tests run very quickly.

For static-sized distributed tests, such as layer checks, we can use just one `torchrun` (or similar) command to spawn all groups.

## Tooling

Several items are needed to enable this functionality:
- New markers for static and dynamic multigpu tests, and migrating most `multigpu` marks to `multigpu_dynamic` marks.
- A distributed printing plugin, to ensure we only print out once per rank in multigpu.
- Several updates to conftest.py, and a distributed conftest.py to enable a new fixture
	- When using `--multigpu-static` on the command line, the `pytest_configure` function will initialize via DistributedManager.
	- The collection modification will now sort tests into groups based on markings:
		- distributed tests will only run under their marks.
		- single gpus tests are everything else, and won't run in distributed runs.
	- The distributed test folder has `distributed_mesh` and `distributed_mesh_2d` fixtures to use in static tests.

## Changes

- multigpu-static tests no longer need `modify_environment`, since that's only need for starting up the distributed manager.
- They also no longer need to use torch.multiprocessing, and the tests can run directly without worker functions.
- 

## Pain points

- The DistributedManager needs to be initialized at the start of the `multigpu_static` job.  If it breaks, that whole pipeline breaks.
	- So ... the `multigpu-dynamic` pipeline should really be first and a prereq in CI.
- It's probably advisable to separate the tests by file (not mix and match dynamic / static).  It would probably work, but might cause confusion when trying to debug.

## Wishlist

- The single-gpu tests should probably get accelerated with pytest-xdist if we can get reliable access to multiple GPUs.
- The coverage calculation does not include distributed tests.  It'd be nice to include it.
- The ops supported by ShardTensor should all get their tests ported into a testing tool here, in a folder under distributed.
	- They currently live elsewhere and need to get ported.

## Performance

In testing, I saw tests running double-digit speedups faster (minutes to seconds) due to removing the launch and init overhead.

## How to run it

Previously, tests were run like this:
```
py.test [--multigpu] test/
```

Now, it would be a two-step process:

```
py.test --multigpu-dynamic test/ #run the dynamic tests
torchrun --nproc-per-node 2 -m pytest --multigpu-static test/ # run the fast path tests
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added lead-time aware training support to the StormCast example.
 - Add a device aware kNN method to physicsnemo.utils.neighbors. Works with CPU or GPU
   by dispatching to the proper optimized library, and torch.compile compatible.
+- Added additional testing of the DoMINO datapipe.
 
 ### Changed
 
--- a/physicsnemo/datapipes/cae/domino_datapipe.py
+++ b/physicsnemo/datapipes/cae/domino_datapipe.py
@@ -522,8 +522,6 @@ class DoMINODataPipe(Dataset):
         if mesh_indices_flattened.dtype != xp.int32:
             mesh_indices_flattened = mesh_indices_flattened.astype(xp.int32)
 
-        length_scale = xp.amax(xp.amax(stl_vertices, 0) - xp.amin(stl_vertices, 0))
-
         center_of_mass = calculate_center_of_mass(stl_centers, stl_sizes)
 
         if self.config.bounding_box_dims_surf is None:
@@ -570,7 +568,6 @@ class DoMINODataPipe(Dataset):
         surf_grid_max_min = xp.stack([s_min, s_max])
 
         return_dict = {
-            "length_scale": length_scale,
             "surf_grid": surf_grid,
             "sdf_surf_grid": sdf_surf_grid,
             "surface_min_max": surf_grid_max_min,
@@ -651,7 +648,7 @@ class DoMINODataPipe(Dataset):
                     (s_max[2] - s_min[2]) / nz,
                 )
                 pos_normals_com_surface = calculate_normal_positional_encoding(
-                    surface_coordinates, center_of_mass, cell_length=[dx, dy, dz]
+                    surface_coordinates, center_of_mass, cell_dimensions=[dx, dy, dz]
                 )
             else:
                 pos_normals_com_surface = surface_coordinates - xp.asarray(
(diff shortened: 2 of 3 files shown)


coreyjadams:
It's actually a little tricky to say *exactly* how much this speeds tests up, because the previous multigpu setup runs many tests that are not actually tagged multigpu.  So it's taking quite awhile, 20 minutes maybe?

The updated tests:

multigpu-dynamic
```
================ 2 failed, 32 passed, 1523 skipped, 17 warnings in 481.07s (0:08:01) ================
```

(note by the way some of these are failing on 8 GPUs!).  2/3 of that 481s time is spent in just two files:

```
test/models/graphcast/test_graphcast_snmg.py: 246.52 seconds
test/models/meshgraphnet/test_meshgraphnet_snmg.py: 73.93 seconds
test/models/meshgraphnet/test_meshgraphnet_snmg.py: 73.93 seconds
test/distributed/test_mesh.py: 44.13 seconds
test/distributed/test_manager.py: 31.36 seconds
test/models/test_distributed_graph.py: 24.11 seconds
test/distributed/test_utils.py: 19.51 seconds
test/distributed/test_config.py: 12.06 seconds
test/distributed/test_distributed_fft.py: 11.92 seconds
test/test_multi_gpu_sample.py: 0.97 seconds
```
Which we could probably improve in the future.

As for the static tests, total test time on 8 GPUs is 17 seconds:

multigpu-static
```
========================== 67 passed, 1490 skipped, 17 warnings in 16.70s ===========================
```

To be fair, though, I plan to use some of this regained performance to add comprehensive testing on Domain Parallelism.

PR #1096, opened 2025-08-27
fix bug in positional embedding

jialusui1102:
# PhysicsNeMo Pull Request

Fix bug in `SongUNetPosEmb` with sinusoidal positional embedding introduced by #1000 .

This PR fixes the bug but @jialusui1102 also uncovered gaps in the non-regression CI tests that let this bug through when #1000 was merged.

## :warning: Summary and remaining TODOs

1. The implementation introduced by #1000 for sinusoidal embeddings has it stack x and y dim in the wrong order, leading to incorrect model output. The current PR fixed the bug while preserving the consistent naming convention introduced by the previous PR.

2. The fixed implementation is able to reproduce the same metrics as CorrDiff v4 paper.

3. :warning: More robust unit tests are needed to ensure the correctness of each new implementation since both the correct and incorrect implementation of sinusoidal embeddings can pass the current tests (e.g. in `test/models/diffusion/test_song_unet_pos_embd.py`, see [there](https://github.com/NVIDIA/physicsnemo/blob/59f7af001c43c91bd9f571957901c526ff61e83b/test/models/diffusion/test_song_unet_pos_lt_embd.py#L104-L133)), where the max and mean difference is both below the tolerance. However, the cosine similarity of both version comparing to the stored output are all low (~0.3-0.37).

4. The input to a lot of these accuracy tests are tensor of ones, which are highly error tolerant. Updated tests with random inputs could be useful in identifying potential bugs in new PRs.

## Checklist


physicsnemo/models/diffusion/song_unet.py
```diff
@@ -1224,24 +1224,28 @@ def _get_positional_embedding(self):
         elif self.gridtype == "linear":
             if self.N_grid_channels != 2:
                 raise ValueError("N_grid_channels must be set to 2 for gridtype linear")
-            y = np.meshgrid(np.linspace(-1, 1, self.img_shape_y))
-            x = np.meshgrid(np.linspace(-1, 1, self.img_shape_x))
-            grid_y, grid_x = np.meshgrid(x, y)
+            x = np.meshgrid(np.linspace(-1, 1, self.img_shape_y))
+            y = np.meshgrid(np.linspace(-1, 1, self.img_shape_x))
+            grid_x, grid_y = np.meshgrid(y, x)
```
CharlelieLrt: I don't think these lines should be reversed

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -1224,24 +1224,28 @@ def _get_positional_embedding(self):
...
-            y2 = np.meshgrid(np.cos(np.linspace(0, 2 * np.pi, self.img_shape_y)))
-            grid_y1, grid_x1 = np.meshgrid(x1, y1)
-            grid_y2, grid_x2 = np.meshgrid(x2, y2)
-            grid = torch.from_numpy(
-                np.stack((grid_x1, grid_y1, grid_x2, grid_y2), axis=0)
-            )
+            x1 = np.meshgrid(np.sin(np.linspace(0, 2 * np.pi, self.img_shape_y)))
+            x2 = np.meshgrid(np.cos(np.linspace(0, 2 * np.pi, self.img_shape_y)))
+            y1 = np.meshgrid(np.sin(np.linspace(0, 2 * np.pi, self.img_shape_x)))
+            y2 = np.meshgrid(np.cos(np.linspace(0, 2 * np.pi, self.img_shape_x)))
```
CharlelieLrt: Same comment: these lines should not be reversed

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -1224,24 +1224,28 @@ def _get_positional_embedding(self):
...
+            grid_x1, grid_y1 = np.meshgrid(y1, x1)
+            grid_x2, grid_y2 = np.meshgrid(y2, x2)
```
CharlelieLrt: Only these two lines should changed into: `grid_y1, grid_x1 = np.meshgrid(y1, x1)`. Please check that it works like this

physicsnemo/models/diffusion/song_unet.py
```diff
@@ -1262,10 +1266,10 @@ def _get_positional_embedding(self):
...
             grid.requires_grad = False
         elif self.gridtype == "test" and self.N_grid_channels == 2:
-            idx_x = torch.arange(self.img_shape_x)
-            idx_y = torch.arange(self.img_shape_y)
-            mesh_y, mesh_x = torch.meshgrid(idx_y, idx_x)
-            grid = torch.stack((mesh_y, mesh_x), dim=0)  # (2, img_shape_y, img_shape_x)
+            idx_x = torch.arange(self.img_shape_y)
+            idx_y = torch.arange(self.img_shape_x)
+            mesh_x, mesh_y = torch.meshgrid(idx_x, idx_y)
+            grid = torch.stack((mesh_x, mesh_y), dim=0)  # (2, img_shape_y, img_shape_x)
```
CharlelieLrt: No need to reverse those (please check that this is correct)

CharlelieLrt (changes requested):
IMO the previous PR should not be totally reversed because it had the merit of using clearer naminhg conventions that improved code maintainability. I think the only thing to change is the order of the tensors retuned by `np.meshgrid`

jialusui1102:
@CharlelieLrt updated the PR based on comments.
unit test pending

jialusui1102:
Current findings:
1. the previous implementation for sinusoidal embeddings as it stack x and y dim in the wrong order, leading to incorrect model output. The current PR fixed the bug while preserving the consistent naming convention
2. The current implementation is able to generate the same output as the version in Julius's branch and reproduce the same metrics as CorrDiff v4 paper.
3. More robust unit test is needed to ensure the correctness of each new implementation since and both the correct and incorrect inplementation of sinusoidal embeddings can pass the current test, where the max and mean difference is both below the tolerance. However, the cosine similarity of both version comparing to the stored output are all low (~0.3-0.37).
4. The input to a lot of the accuracy tests are tensor of ones, which are highly error tolerant. Updated tests with random inputs  could be useful in identifying potential bugs in new PRs.

PR #1102, opened 2025-09-03
Add significantly more coverage of the domino datapipe to catch more errors

coreyjadams:
This PR is mostly adding further test coverage of Domino's datapipe, to enhance out ability to catch errors.

Only single-gpu tests here.

# PhysicsNeMo Pull Request

## Checklist

### DIFF
--- a/physicsnemo/datapipes/cae/domino_datapipe.py
+++ b/physicsnemo/datapipes/cae/domino_datapipe.py
@@ -570,7 +570,7 @@ class DoMINODataPipe(Dataset):
         surf_grid_max_min = xp.stack([s_min, s_max])
 
         return_dict = {
-            "length_scale": length_scale,
+            "length_scale": xp.array(length_scale, dtype=xp.float32),
             "surf_grid": surf_grid,
             "sdf_surf_grid": sdf_surf_grid,
             "surface_min_max": surf_grid_max_min,
@@ -651,7 +651,7 @@ class DoMINODataPipe(Dataset):
                     (s_max[2] - s_min[2]) / nz,
                 )
                 pos_normals_com_surface = calculate_normal_positional_encoding(
-                    surface_coordinates, center_of_mass, cell_length=[dx, dy, dz]
+                    surface_coordinates, center_of_mass, cell_dimensions=[dx, dy, dz]
                 )
             else:
                 pos_normals_com_surface = surface_coordinates - xp.asarray(
@@ -744,7 +744,13 @@ class DoMINODataPipe(Dataset):
 
             else:
                 # We are *not* sampling, kNN on ALL points:
-                ii = knn.kneighbors(surface_coordinates, return_distance=False)
+                if self.array_provider == cp:
+                    ii = knn.kneighbors(surface_coordinates, return_distance=False)
+                else:
+                    _, ii = interp_func.query(
+                        surface_coordinates,
+                        k=self.config.num_surface_neighbors,
+                    )
 
                 # Construct the neighbors arrays:
                 surface_neighbors = surface_coordinates[ii][:, 1:]
@@ -892,10 +898,10 @@ class DoMINODataPipe(Dataset):
                 pos_normals_closest_vol = calculate_normal_positional_encoding(
                     volume_coordinates,
                     sdf_node_closest_point,
-                    cell_length=[dx, dy, dz],
+                    cell_dimensions=[dx, dy, dz],
                 )
                 pos_normals_com_vol = calculate_normal_positional_encoding(
-                    volume_coordinates, center_of_mass, cell_length=[dx, dy, dz]
+                    volume_coordinates, center_of_mass, cell_dimensions=[dx, dy, dz]
                 )
             else:
                 pos_normals_closest_vol = volume_coordinates - sdf_node_closest_point
--- a/physicsnemo/datapipes/cae/domino_datapipe.py
+++ b/physicsnemo/datapipes/cae/domino_datapipe.py
@@ -522,8 +522,6 @@ class DoMINODataPipe(Dataset):
         if mesh_indices_flattened.dtype != xp.int32:
             mesh_indices_flattened = mesh_indices_flattened.astype(xp.int32)
 
-        length_scale = xp.amax(xp.amax(stl_vertices, 0) - xp.amin(stl_vertices, 0))
-
         center_of_mass = calculate_center_of_mass(stl_centers, stl_sizes)
 
         if self.config.bounding_box_dims_surf is None:
@@ -570,7 +568,6 @@ class DoMINODataPipe(Dataset):
         surf_grid_max_min = xp.stack([s_min, s_max])
 
         return_dict = {
-            "length_scale": xp.array(length_scale, dtype=xp.float32),
             "surf_grid": surf_grid,
             "sdf_surf_grid": sdf_surf_grid,
             "surface_min_max": surf_grid_max_min,
--- a/test/datapipes/test_domino_datapipe.py
+++ b/test/datapipes/test_domino_datapipe.py
@@ -14,9 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import tempfile
 from dataclasses import dataclass
 from typing import List
 
+import numpy as np
 import pytest
 import torch
 from pytest_utils import import_or_fail
@@ -27,7 +29,7 @@ Tensor = torch.Tensor
 @dataclass
 class ConcreteBoundingBox:
     """
-    Really simple bounding box to mimic a structured config;  Don't use elsewhere.
+    Really simple bounding box to mimic a structured config; Don't use elsewhere.
     """
 
     min: List[float]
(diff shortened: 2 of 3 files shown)


physicsnemo/datapipes/cae/domino_datapipe.py
```diff
@@ -744,7 +744,13 @@ def preprocess_surface(self, data_dict, core_dict, center_of_mass, s_min, s_max)
 
             else:
                 # We are *not* sampling, kNN on ALL points:
-                ii = knn.kneighbors(surface_coordinates, return_distance=False)
+                if self.array_provider == cp:
+                    ii = knn.kneighbors(surface_coordinates, return_distance=False)
+                else:
```
Alexey-Kamenev, line 746: Maybe use `elif` just in case? And assert/throw in `else`?

test/datapipes/test_domino_datapipe.py
```diff
@@ -39,92 +41,418 @@ def data_dir(nfs_data_dir):
...
+def create_basic_dataset(data_dir, model_type, **kwargs):
+    """Helper function to create a basic DoMINODataPipe with default settings."""
     from physicsnemo.datapipes.cae.domino_datapipe import DoMINODataPipe
 
-    if model_type == "surface":
-        input_path = data_dir / "surface/"
-    elif model_type == "volume":
+    assert model_type in ["volume", "surface", "combined"]
+
+    if model_type == "volume":
```
Alexey-Kamenev: Since you already have `assert` above, you can replace this `if` block with this:
```py
input_path = data_dir / model_type
```
?

test/datapipes/test_domino_datapipe.py
```diff
@@ -39,92 +41,418 @@ def data_dir(nfs_data_dir):
...
+# Core test - smaller matrix focusing on essential device/model combinations
+@import_or_fail(["warp", "cupy", "cuml"])
+@pytest.mark.parametrize("gpu_preprocessing", [True, False])
+@pytest.mark.parametrize("gpu_output", [True, False])
+@pytest.mark.parametrize("model_type", ["surface", "volume", "combined"])
+def test_domino_datapipe_core(
+    data_dir, gpu_preprocessing, gpu_output, model_type, pytestconfig
+):
+    """Core test for basic functionality with different device and model configurations."""
+    # if gpu_preprocessing and model_type in ["surface", "combined"]:
```
Alexey-Kamenev: Should this commented out code be there?

test/datapipes/test_domino_datapipe.py
```diff
@@ -39,92 +41,418 @@ def data_dir(nfs_data_dir):
...
+@pytest.mark.parametrize("model_type", ["combined"])
+@pytest.mark.parametrize("normalize_coordinates", [True, False])
+def test_domino_datapipe_coordinate_normalization(
+    data_dir, model_type, normalize_coordinates, pytestconfig
+):
+    """Test coordinate normalization functionality."""
+    dataset = create_basic_dataset(
+        data_dir,
+        model_type,
+        gpu_preprocessing=True,
```
Alexey-Kamenev, line 352: Does it make sense to make `gpu_preprocessing` a test parameter as well?

test/datapipes/test_domino_datapipe.py
```diff
@@ -39,92 +41,418 @@ def data_dir(nfs_data_dir):
...
+    sample = dataset[0]
+    validate_sample_structure(sample, model_type, gpu_output=True)
+
+    v_coords = sample["volume_mesh_centers"]
+    s_coords = sample["surface_mesh_centers"]
+    # If normalization is enabled, coordinates should be in [-2, 2] range
+    if normalize_coordinates:
+        assert all(torch.min(v_coords, dim=0).values >= -2.0) and all(
+            torch.max(v_coords, dim=0).values <= 2.0
+        ), "Normalized coordinates should be in [-1,1]"
```
Alexey-Kamenev: This one says `[-1,1]` but the code and other comment - `[-2,2]`.

test/datapipes/test_domino_datapipe.py
```diff
@@ -39,92 +41,418 @@ def data_dir(nfs_data_dir):
...
+):
+    """Test bounding box sampling functionality."""
+    dataset = create_basic_dataset(
+        data_dir, model_type, gpu_preprocessing=False, sample_in_bbox=sample_in_bbox
+    )
+
+    sample = dataset[0]
+    validate_sample_structure(sample, model_type, gpu_output=True)
+
+    # if sample_in_bbox:
```
Alexey-Kamenev: Remove/uncomment?

test/datapipes/test_domino_datapipe.py
```diff
@@ -39,92 +41,418 @@ def data_dir(nfs_data_dir):
...
-        scaling_type=None,
-        bounding_box_dims=bounding_box,
-        bounding_box_dims_surf=bounding_box_surface,
+        surface_points_sample=567,
+        geom_points_sample=890,
+        model_type="combined",
     )
 
     assert len(dataset) > 0
+    print(f"dataset length: {len(dataset)}")
```
Alexey-Kamenev: Is this for debugging purposes or we need it in the test?

physicsnemo/datapipes/cae/domino_datapipe.py
coreyjadams: The rest of the pipeline uses the if/else syntax, and array_provider can only be these two values:
```
        self.array_provider = cp if self.config.gpu_preprocessing else np
```

Anyways the whole pipeline is ported to torch in the next PR which resolves these ambiguities, so I will leave this for now if you don't mind.

test/datapipes/test_domino_datapipe.py
coreyjadams: No it should be removed!  Thanks.

test/datapipes/test_domino_datapipe.py
coreyjadams: Yes, it's part of the "core" tests.  I didn't do the full matrix of possibilities, it was 1000+ tests.  But I do have it in the suite.

test/datapipes/test_domino_datapipe.py
coreyjadams: Good catch.

test/datapipes/test_domino_datapipe.py
coreyjadams: Done.

test/datapipes/test_domino_datapipe.py
coreyjadams: Debugging, sorry, removed!

test/datapipes/test_domino_datapipe.py
coreyjadams: Yes, good call, thanks!

physicsnemo/datapipes/cae/domino_datapipe.py
```diff
@@ -570,7 +570,7 @@ def preprocess_combined(self, data_dict):
         surf_grid_max_min = xp.stack([s_min, s_max])
 
         return_dict = {
-            "length_scale": length_scale,
+            "length_scale": xp.array(length_scale, dtype=xp.float32),
```
RishikeshRanade: Is this variable used in the training pipe? Maybe we should not return it.

test/datapipes/test_domino_datapipe.py
```diff
@@ -39,92 +41,405 @@ def data_dir(nfs_data_dir):
...
+        gpu_preprocessing=True,
+        normalize_coordinates=normalize_coordinates,
+    )
+
+    sample = dataset[0]
+    validate_sample_structure(sample, model_type, gpu_output=True)
+
+    v_coords = sample["volume_mesh_centers"]
+    s_coords = sample["surface_mesh_centers"]
+    # If normalization is enabled, coordinates should be in [-2, 2] range
```
RishikeshRanade, line 370: Why -2 to 2 and -1 to 1?

test/datapipes/test_domino_datapipe.py
```diff
@@ -39,92 +41,405 @@ def data_dir(nfs_data_dir):
...
+
+    sample = dataset[0]
+    validate_sample_structure(sample, model_type, gpu_output=True)
+
+    v_coords = sample["volume_mesh_centers"]
+    s_coords = sample["surface_mesh_centers"]
+    # If normalization is enabled, coordinates should be in [-1, 1] range if sampling in bbox
+    if sample_in_bbox:
+        assert all(torch.min(v_coords, dim=0).values >= -1.5) and all(
+            torch.max(v_coords, dim=0).values <= 1.5
```
RishikeshRanade: Same comment as above, why not -1 to 1?

test/datapipes/test_domino_datapipe.py
coreyjadams: Well, what I've gathered is this.  Could be unintended but it's how I read the code:

When normalizing, if the user supplies the bounding box, it will normalize the bounding box to -1, 1 and all other points proportionally.  So unless you're asking to sample in the bounding box, AND you're asking to normalize, you can have values show up above -1.

We already have some pipelines to scan the incoming dataset for normalization fields on the output; perhaps we should extend that to infer the bounding box values?

RishikeshRanade: Yes, we should probably do that. We can't guarantee -2 to 2 either if we don't know the bounding box dimensions.

physicsnemo/datapipes/cae/domino_datapipe.py
coreyjadams: I looked, I didn't see it anywhere in the model code or training pipe.  I've removed it.

test/datapipes/test_domino_datapipe.py
coreyjadams: OK; updating the training pipelines to scan datasets while computing scaling factors will come in the next PR.  For now, I've reworked the dataset in the tests;  see comment below.  Thanks!

test/datapipes/test_domino_datapipe.py
coreyjadams: Same rework as below!

coreyjadams:
@RishikeshRanade @Alexey-Kamenev could you take a look at the tests again, quickly?

I reworked the dataset tests in the following ways:
- data is now *exclusively* synthetic for this test.  But that doesn't mean it's all junk: 
  - I generate points on a unit sphere, shifted up +1 in the z direction, get the triangles and normals correctly, and use that as the STL stand-in.
  - I generate surface points on the same sphere, and use synthetic areas and fields.
  - I generate volume points in a rectangle around the sphere, and throw away any points in the sphere.
- With this synthetic data, because I control the data and the bounding boxes supplied, I can have actual informed assertions about where the min / max values should be coming out of the datatype, based on what the min/maxes of the generated data were as well as the bounding boxes supplied.  So the normalization / sampling tests are no longer quasi-junk.
- I also can test .npy, .npz, and .zarr input formats since I generate them all.
- And, the test data is much much smaller, so the tests run much faster and use less CPU memory.  Which was an issue on Blossom.

PR #1107, opened 2025-09-08
Log-uniform sigma sampling in EDMLoss

jleinonen:
# PhysicsNeMo Pull Request

## Description

Adds a log-uniform `sigma` distribution option and a generic mechanism for selecting `sigma` distributions to `EDMLoss`.

Also adds the ability to configure the loss type to the StormCast example.

Further minor changes/additions:

- Option to pass per-channel `sigma_data` as a list in StormCast config.
- Documentation improvements and type hints in `EDMLoss`.

Possible additions to be discussed:

- Currently the dict of available samplers `sigma_samplers` is coded manually, it could also be automatically populated.
- Support for sigma sampler options in other losses.

## Checklist


physicsnemo/metrics/diffusion/loss.py
```diff
@@ -214,29 +299,31 @@ class EDMLoss:
...
-        self, P_mean: float = -1.2, P_std: float = 1.2, sigma_data: float = 0.5
+        self,
+        sigma_data: float | torch.Tensor = 0.5,
+        sigma_sampler: Literal["LogNormalSigma", "LogUniformSigma"] = "LogNormalSigma",
+        **sampler_kwargs,
     ):
-        self.P_mean = P_mean
-        self.P_std = P_std
         self.sigma_data = sigma_data
+        self.sigma_sampler = sigma_samplers[sigma_sampler](**sampler_kwargs)
```
CharlelieLrt: @jleinonen this is BC-breaking (`P_mean` and `P_std` were removed, both as arguments and attributes).

Also, string-typing is not recommended here. If possible, it would be better to avoid passing primitives but instead adopt dependency injection (i.e. pass the instance of the `sigma_sampler` instead of passing a string containing its name)

pzharrington: The passing of `sample_kwargs` here should be able to handle the previous constructor style (`loss = EDMLoss(P_mean=blah)`) as long as the default sigma sampler is lognormal. But ye, the `P_mean` and `P_std` attributes will no longer be defined (maybe we could automatically set the contents of `sampler_kwargs` as attributes?

CharlelieLrt: I think my confusion comes from the string-typing where we pass primitives (here the name of the `sigma_sampler` and the arguments to instantiate it), instead of doing dependency injection (that is passing directly an already instantiated `sigma_sampler`). BTW, we have similar string-typing mechanism in the EDM preconditioner, and the latest code review qualified it as _extremely frustrating_ (difficult to document what are the accepted values, possible names conflict with the splatted arguments  `**sampler_kwargs`, etc.)... so IMO it would be preferable to stay away from the string-typing and go for dependency injection in such cases.

physicsnemo/metrics/diffusion/loss.py
```diff
@@ -300,12 +216,31 @@ class EDMLoss:
...
+        self.P_std = P_std
         self.sigma_data = sigma_data
-        self.sigma_sampler = sigma_samplers[sigma_sampler](**sampler_kwargs)
+
+    def get_noise_params(self, y: torch.Tensor) -> tuple[float]:
+        """Sample the sigma noise parameter for each sample."""
+        shape = (y.shape[0], 1, 1, 1)
+        rnd_normal = torch.randn(shape, device=y.device)
+        sigma = (rnd_normal * self.P_std + self.P_mean).exp()
+        return (sigma,)
```
pzharrington: Any reason to make this a tuple? Or can we just return `sigma`? If so, probably better to rename it to something like `get_sigma` to make it more clear

CharlelieLrt: Yes, I was about to write the same, `get_noise_level` and returns single float instead of tuple

physicsnemo/metrics/diffusion/loss.py
```diff
@@ -214,29 +215,50 @@ class EDMLoss:
...
         Calculate and return the loss corresponding to the EDM formulation.
 
         The method adds random noise to the input images and calculates the loss as the
         square difference between the network's predictions and the input images.
-        The noise level is determined by 'sigma', which is computed as a function of
-        'P_mean' and 'P_std' random values. The calculated loss is weighted as a
-        function of 'sigma' and 'sigma_data'.
+        The noise level is determined by 'sigma', which is drawn from the `sigma_sampler`
+        specified in the constructor. The calculated loss is weighted as a function of
+        'sigma' and 'sigma_data'.
```
CharlelieLrt: I think this is outdated considering the refactors you did

CharlelieLrt (approved):
Just a minor thing to remove from a docstring, otherwise looks good to me!

physicsnemo/metrics/diffusion/loss.py
jleinonen: Resolved in https://github.com/NVIDIA/physicsnemo/pull/1107/commits/e755da1d35de171386d786705ae1b0fd98871258.

physicsnemo/metrics/diffusion/loss.py
jleinonen: Renamed as suggested in https://github.com/NVIDIA/physicsnemo/pull/1107/commits/e755da1d35de171386d786705ae1b0fd98871258.

physicsnemo/metrics/diffusion/loss.py
jleinonen: Resolving this conversation as I think the redesign addresses the comments here.

PR #1113, opened 2025-09-10
Close Domino Performance Gaps

coreyjadams:
# PhysicsNeMo Pull Request

This PR is to close gaps in DoMINO performance.  In includes some level of refactor, as well, with a little more to go.  Opening this PR as a draft while tracking down some final little pieces, which will get listed below.

## Description
This is a large PR.  It covers a lot of pieces which essentially have to be brought in together.  In no particular order, this PR includes:

#### DrivaerML DataSet

This PR includes a separation of IO from pre-processing in the DoMINO datapipe.  There is now one IO interface "drivaer_ml_dataset" (open to a name change, here) that generically will read keys from a dictionary-like object.  The dataset will infer, based on file properties, how to read the data: 
  - `.npy` files are supported (pickled numpy dictionaries) though the entire file has to be read, and then the requested keys are returned as torch cpu tensors.
  - `.npz` files can read directly the tensors requested.
  - `.zarr` files will read with zarr 3.0, unless `tensorstore` is installed.
  - If tensorstore is installed, the zarr files will be read with tensorstore instead.  It is significantly faster than plain zarr, with no extra CPU overhead on the python interpreter.  Previous methods with threading added overhead to the python runtime causing cpu-based stalls in model code.
  - If none of those matches are made, it will check if every directory passed contains only .stl, .vtp, .vtu, or .csv files.  In this case, it will use pyvista to read these.

The dataset will, as part of it's pipeline, pin tensors and move them to the GPU.  Ideally if we're pinning, we would preallocate and write directly to those buffers, but tensorstore does not support that.  If the output device is  GPU, then the data transfer will happen in a separate stream.  Optionally, if this is a multi-stream application, the user can pass a `consumer_stream` (defaults to `torch.cuda.default_stream()`) that will be used to ensure proper stream ordering.

The dataset object has ability to preload indexes in python threads (up to a configurable preload depth) to async load data from disk into torch tensors.  Note that aggressive preloading will accumulate GPU storage usage if the dataset is outputting to GPU (configured by setting the output device in dataset construction).

Construction of these datasets looks like this:
```python 
dataset = DrivaerMLDataset(
                data_dir=self.config.data_path,
                keys_to_read=self.keys_to_read,
                output_device=self.preproc_device,
                pin_memory=pin_memory,
                consumer_stream=torch.cuda.default_stream(),
            )
```

Calling `__getitem__` (aka, `dataset[i]`) will fetch the i'th entry of the dataset.  There is also a `preload` function that will manually queue up the target data.  Further, if you iterate over the dataset (`for i, batch in enumerate(dataset)`) the preloading will happen automatically and transparently.

#### Domino DataPipe

The datapipe for domino has been refactored entirely: the entire datapipe is implemented to operate on torch tensors.  As of opening this PR, the datapipe also contains an instance of the dataset above.  

>However, I think it makes sense to remove that, and use the datapipe purely as a pre-processing type module that has some state (preprocessing configs) and transforms `torch.Tensor` objects from the dataset to model-ready inputs for domino.

The datapipe itself looks very similar to the previous iteration, minus the IO pieces, but has a few additions:
- The datapipe already has built in to it transforms on the output fields.  In inference mode, these are passed by if no output target fields are supplied.
- Because it does the normalization one way, it makes sense to do the normalization the other way in the datapipe too.  The datapipe can now be used to un-normalize or un-standardize outputs, and will pick the right path based on the datapipe configuration.

All logic based on computing scaling factors is removed from the datapipe.  See below.

The datapipe currently contains preprocessing logic too, however, this is going to be removed before merge.  The operations needed from Warp to do this don't interoperate well with pytorch so this is delayed.

The CachedDominoDataset is largely unchanged.

### DoMINO Example ReOrg

Inside the src/ folder of the DoMINO training example, there are the following changes:
- the loss functions from training have moved to `loss.py` but are otherwise unchanged.  They could use a clean up after the physics loss additions.
- A `utils.py` file has been added to contain script-specific logic, like extracting the number of variables for the model from the config and a `dataclass` to contain scaling factor info (see below).
- The test and train_sharded scripts are not yet updated, but need to be.
- The train.py script has been updated only in that pieces have been moved into other scripts and imported, to shorten the script length and complexity.
- A temporary script, `benchmark_dataset.py` is present for development but needs to be removed before merge.
- Inference_on_stl is covered below.

#### Scaling Factors

Currently, in Domino, computing the scaling factors is done implicitly and somewhat slowly, since it uses the pipeline inefficiently.  Instead, there is a compute_scaling_factors function that will use the dataset (not datapipe) object, loop over the data reading only the necessary fields, if possible, and return the mean/std/min/max of each field read.  By default, for drivaerML, this will not only use the output fields but also the surface and volume coordinates.  This can help determine bounding boxes on new datasets, etc.

To facilitate the scaling factor usage more easily, a new script is added to the `src/` dir in the examples, `compute_statistics.py`.  It is meant to be run on a single device, or even possibly the CPU if desired, and it saves the scaling factors into a pickled object that is reloaded in other scripts.  I'm open to changes in this design, but the idea was to make the scale factors more stable, more portable, and computing them stand-alone from the training script.

#### Inference on STL

The entire inference on stl pipeline has been refactored and ported to use GPUs, reusing the dataset and datapipe from above.  Performance is significantly better on GPU.  Two pieces of the script are missing (called out below).

#### RMM by default

The RAPIDS memory manager is now used, by default, to share memory between the kNN in the preprocessing step and the main model code.  This accelerates performance (avoiding cudaFree and cudaMalloc) and also reduces memory usage overall - since the two phases are out of sync with each other, sharing memory makes sense.  It can only be opted out with an environment variable, now documented in the README.

#### SDF is torch only now

The Signed distance function now only accepts torch inputs and returns torch outputs.  The output signature is fixed to the distance itself, and the closest point on the mesh (which is not usually a vertex).  Test have been updated, since they were slightly incorrect.

#### utils/domino/utils.py is torch only

The cupy/numpy backend of those functions is removed and it is torch only, now.  The non-array functionality in utils for vtk reading has been moved to `utils/domino/vtk_file_utils.py` and is unchanged, otherwise.

Missing pieces will be marked below as issues to resolve before merge, however, here's the condensed list as I see it:
- [X] **Datapipe async Preprocessing** is to be removed from the datapipe, and put only in the dataset.  No streams in the datapipe due to warp / torch issues.  The `set_indices` functionality is to be moved to the dataset.
- [X] The dataset constructor should not be used in the datapipe but passed to the datapipe optionally.  the `create_domino_dataset` function can do this transparently to keep it as a "drop in" replacement
- [X] The SDF function in the data preprocessing operates on unscaled data, which might later be scaled.  This should be addressed but isn't quite P0.
- [ ] ShardTensor integration is completely missing.  It shouldn't be difficult to integrate but isn't done yet.
- [ ] ~The SDF function shows up in other places, still.  Some are reimplementations, some are using this.  It needs to be checked if we can accommodate that easily.~ Done, except in re-implementations which are out of scope here.
- [ ] The test.py script must be updated.
- [X] ~The retraining script must be updated.~ It's been deprecated
- [ ] The old versions of `inference_on_stl.py` and `domino_datapipe.py` must be replaced with the new versions.
- [ ] Some sort of check of how this affects the cached versions of the datasets should happen.
- [X] ~I don't know how useful the `openfoam_datapipe.py` is - if it's still in use, we should merge that logic into this and unify it.~ Deprecated

## Checklist

### DIFF
--- a/physicsnemo/datapipes/cae/drivaer_ml_dataset.py
+++ b/physicsnemo/datapipes/cae/drivaer_ml_dataset.py
@@ -669,7 +669,11 @@ class DrivaerMLDataset:
             raise NotADirectoryError(f"Data directory {data_dir} is not a directory")
 
         self._keys_to_read = keys_to_read
-        self._keys_to_read_if_available = keys_to_read_if_available
+
+        # Make sure the optional keys are on the right device:
+        self._keys_to_read_if_available = {
+            k: v.to(output_device) for k, v in keys_to_read_if_available.items()
+        }
 
         self.file_reader, self._filenames = self._infer_file_type_and_filenames(
             data_dir
--- a/physicsnemo/utils/domino/utils.py
+++ b/physicsnemo/utils/domino/utils.py
@@ -442,7 +442,7 @@ def shuffle_array(
 
     # If there are no weights, use uniform weights:
     if weights is None:
-        weights = torch.ones_like(points, device=points.device)
+        weights = torch.ones(points.shape[0], device=points.device)
 
     # Using torch multinomial for this.
     # Multinomial can't work with more than 2^24 input points.
--- a/test/datapipes/test_domino_datapipe.py
+++ b/test/datapipes/test_domino_datapipe.py
@@ -18,7 +18,7 @@ import shutil
 import tempfile
 from dataclasses import dataclass
 from pathlib import Path
-from typing import List, Literal
+from typing import List, Literal, Optional, Sequence
 
 import numpy as np
 import pytest
@@ -32,6 +32,7 @@ from physicsnemo.datapipes.cae.domino_datapipe2 import (
     DoMINODataConfig,
     DoMINODataPipe,
 )
+from physicsnemo.datapipes.cae.drivaer_ml_dataset import DrivaerMLDataset
 
 Tensor = torch.Tensor
 
@@ -242,7 +243,22 @@ def bounding_boxes():
     }
 
 
-def create_basic_dataset(data_dir, model_type, **kwargs):
+def create_basic_dataset(
+    data_dir,
+    model_type,
+    gpu_preprocessing: bool = False,
+    gpu_output: bool = False,
+    normalize_coordinates: bool = False,
+    sample_in_bbox: bool = False,
+    sampling: bool = False,
+    volume_points_sample: int = 1234,
+    surface_points_sample: int = 1234,
+    surface_sampling_algorithm: str = "random",
+    caching: bool = False,
+    scaling_type: Optional[Literal["min_max_scaling", "mean_std_scaling"]] = None,
+    volume_factors: Optional[Sequence] = None,
+    surface_factors: Optional[Sequence] = None,
+):
     """Helper function to create a basic DoMINODataPipe with default settings."""
 
     # assert model_type in ["volume", "surface", "combined"]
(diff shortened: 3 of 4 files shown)


coreyjadams:
This is merging into a `domino` feature branch, not main, so no review required yet.

PR #1119, opened 2025-09-18
Port SDF function to static, torch-only interface

coreyjadams:
Port SDF function to stream-respecting, torch-only interface with static shapes.

This also enables torch.compile over sdf calls, and I've implemented a distributed interface for the SDF function.  Unfortunately, without more sophisticated planning, the distributed algorithm must realize the mesh on all nodes.  But for very large input point vectors, which is the realistic use case, it should be efficient at scale.

This does not, yet, address the proliferation of other sdf implementation in the examples folders.  I have left them be, 
they aren't hurting anyone ...

# PhysicsNeMo Pull Request

## Checklist

### DIFF
--- a/physicsnemo/datapipes/cae/domino_datapipe.py
+++ b/physicsnemo/datapipes/cae/domino_datapipe.py
@@ -64,6 +64,31 @@ from physicsnemo.utils.domino.utils import (
 from physicsnemo.utils.profiling import profile
 from physicsnemo.utils.sdf import signed_distance_field
 
+"""
+These functions, below, are to handle the SDF calculation which only 
+accepts torch tensors.  The entire pipeline is moving to torch, so
+these aren't necessary after that.
+"""
+
+
+def convert_array_to_torch(array: cp.ndarray | np.ndarray) -> torch.Tensor:
+    """
+    TEMPORARY function to convert cupy and numpy arrays to torch tensors.
+    """
+    if isinstance(array, cp.ndarray):
+        return torch.utils.dlpack.from_dlpack(array)
+    elif isinstance(array, np.ndarray):
+        return torch.from_numpy(array)
+    else:
+        raise ValueError(f"Unsupported array type: {type(array)}")
+
+
+def convert_torch_to_array(array: torch.Tensor, provider) -> cp.ndarray | np.ndarray:
+    """
+    TEMPORARY function to convert torch tensors to cupy arrays.
+    """
+    return provider.from_dlpack(array)
+
 
 def domino_collate_fn(batch):
     """
@@ -537,12 +562,14 @@ class DoMINODataPipe(Dataset):
             surf_grid = create_grid(s_max, s_min, [nx, ny, nz])
             surf_grid_reshaped = surf_grid.reshape(nx * ny * nz, 3)
 
-            sdf_surf_grid = signed_distance_field(
-                stl_vertices,
-                mesh_indices_flattened,
-                surf_grid_reshaped,
+            sdf_surf_grid, _ = signed_distance_field(
+                convert_array_to_torch(stl_vertices),
+                convert_array_to_torch(mesh_indices_flattened),
+                convert_array_to_torch(surf_grid_reshaped),
                 use_sign_winding_number=True,
-            ).reshape(nx, ny, nz)
+            )
+            sdf_surf_grid = sdf_surf_grid.reshape(nx, ny, nz)
+            sdf_surf_grid = convert_torch_to_array(sdf_surf_grid, self.array_provider)
 
         else:
             surf_grid = None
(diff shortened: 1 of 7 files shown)


physicsnemo/datapipes/cae/domino_datapipe.py
```diff
@@ -64,6 +64,31 @@
...
 from physicsnemo.utils.sdf import signed_distance_field
 
+"""
+These functions, below, are to handle the SDF calculation which only 
+accepts torch tensors.  The entire pipeline is moving to torch, so
+these aren't necessary after that.
+"""
+
+
+def convert_array_to_torch(array: cp.ndarray | np.ndarray) -> torch.Tensor:
```
peterdsharpe: If this is a temporary function, can we name it starting with a `_` so that end-users aren't using this in the wild? (Not that they should anyway, but this will help communicate that)

physicsnemo/datapipes/cae/domino_datapipe.py
```diff
@@ -64,6 +64,31 @@
...
+    """
+    if isinstance(array, cp.ndarray):
+        return torch.utils.dlpack.from_dlpack(array)
+    elif isinstance(array, np.ndarray):
+        return torch.from_numpy(array)
+    else:
+        raise ValueError(f"Unsupported array type: {type(array)}")
+
+
+def convert_torch_to_array(array: torch.Tensor, provider) -> cp.ndarray | np.ndarray:
```
peterdsharpe: Similar `_` prepending here

physicsnemo/datapipes/cae/domino_datapipe.py
```diff
@@ -879,12 +908,16 @@ def preprocess_volume(
...
+                convert_array_to_torch(mesh_indices_flattened),
+                convert_array_to_torch(volume_coordinates),
                 use_sign_winding_number=True,
             )
+            sdf_nodes = convert_torch_to_array(sdf_nodes, self.array_provider)
+            sdf_node_closest_point = convert_torch_to_array(
+                sdf_node_closest_point, self.array_provider
+            )
+
             # TODO - is this needed?
```
Alexey-Kamenev, line 921: So.. is this question still valid after the changes above?

physicsnemo/utils/sdf.py
```diff
@@ -115,88 +103,103 @@ def signed_distance_field(
...
-            - If `include_hit_points_id` is True, returns a tuple: (sdf,
-              hit_point_ids), where `hit_point_ids` contains the face index of
-              the closest mesh face for each input point.
-            - If both `include_hit_points` and `include_hit_points_id` are True,
-              returns a tuple: (sdf, hit_points, hit_point_ids).
-            - The returned array type (NumPy or CuPy) is determined by the
-            `return_cupy` argument, or inferred from the input arrays.
+        tuple[torch.Tensor, torch.Tensor] of:
+            - signed distance to the mesh, per input point
+            - hith point, per input point. "hit points" are the points on the
```
Alexey-Kamenev: `hith` -> `hit`?

physicsnemo/utils/sdf.py
```diff
@@ -115,88 +103,103 @@ def signed_distance_field(
...
-            arrays_to_return.append(convert(sdf_hit_point))
-        if include_hit_points_id:
-            arrays_to_return.append(convert(sdf_hit_point_id))
 
-        return (
-            arrays_to_return[0]
-            if len(arrays_to_return) == 1
-            else tuple(arrays_to_return)
-        )
+@signed_distance_field.register_fake
```
Alexey-Kamenev, line 184: Just curious, what is the purpose of this?

physicsnemo/datapipes/cae/domino_datapipe.py
coreyjadams: No, probably not!  But, the whole pipeline has been optimized and replaced in this PR: https://github.com/NVIDIA/physicsnemo/pull/1113

This SDF PR was split off the other, it is just a preliminary change to make that domino PR a little more focused.

physicsnemo/utils/sdf.py
coreyjadams: This will let you compile torch code with `torch.compile` even if the SDF function is inside of it.  It can be used in the domino dataloader, or anywhere else we need a signed distance function + torch.compile.

peterdsharpe (approved):
Minor optional changes suggested, but otherwise looks good. Nice work!

physicsnemo/datapipes/cae/domino_datapipe.py
coreyjadams: Sure sounds good.  The temporary functions are only to keep the datapipe stable until it's deleted and replaced  ASAP, so making the functions more obviously private is a good idea.

PR #1120, opened 2025-09-18
Refactor DiTBlock to be more modular

pzharrington:
# PhysicsNeMo Pull Request

## Description
Refactors the DiTBlock to use factories for layernorm/attention modules, and also accept a user-provided attention module if desired. Also adds a new attention module to perform 2D neighborhood attention.

## Checklist


pyproject.toml
```diff
@@ -19,7 +19,7 @@ dependencies = [
     "packaging>=24.2",
     "s3fs>=2023.5.0",
     "setuptools>=77.0.3",
-    "timm>=0.9.12",
+    "timm>=1.0.0",
```
CharlelieLrt, line 22: Do you know if it's okay to bump this (as well as adding the other dependencies below)? Maybe check with @ktangsali ?

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -34,7 +48,395 @@
...
+    NATTEN_AVAILABLE = True
+except ImportError:
+    NATTEN_AVAILABLE = False
+
 from physicsnemo.models.layers import Mlp
+from physicsnemo.distributed import ShardTensor
+from physicsnemo.distributed.shard_utils.natten_patches import partial_na2d
+
+
+def build_layer_norm(
```
CharlelieLrt: We actually already have an almost identical factory for `group_norm` (see [there](https://github.com/NVIDIA/physicsnemo/blob/0183ddd8ab32245739b2fa43fd5f6cc1660de1ef/physicsnemo/models/diffusion/layers.py#L360-L429)). It would be good to be as consistent as possible with this implementation. A few suggestions:
- rename `build_layer_norm` into `get_layer_norm`
- Did you mean "_backend_" instead of "_backbone_"?
- IMO the `layernorm_backbone` that you used is better than the `use_apex_gn` used in `get_group_norm`, because it leaves the possibility to add different backend later on. So let's keep it like this.
- Double-check that state-dicts of the Apex layernorm and the torch layer norm have the same keys (if I recall correctly, there was some trouble with group norm state dict)

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -34,7 +48,395 @@
...
+        Implementation selector.
+    elementwise_affine: bool
+        Whether to learn per-element affine parameters.
+    eps: float
+        Numerical stability epsilon.
+
+    Returns
+    -------
+    nn.Module
+        A configured LayerNorm module from Apex or Torch.
```
CharlelieLrt: Add the expected input/output shapes for the returned `nn.Module`

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -34,7 +48,395 @@
...
+        return FusedLayerNorm(hidden_size, elementwise_affine=elementwise_affine, eps=eps)
+    if layernorm_backbone == "torch":
+        return nn.LayerNorm(hidden_size, elementwise_affine=elementwise_affine, eps=eps)
+    raise ValueError("layernorm_backbone must be one of 'apex' or 'torch'.")
+
+
+def build_attention(
+    hidden_size: int,
+    num_heads: int,
+    attention: Literal["transformer_engine", "timm", "natten2d"],
```
CharlelieLrt: Maybe rename to `attention_backend` to be more explicit?

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -34,7 +48,395 @@
...
+            raise ImportError(
+                "Apex is not available. Please install Apex to use FusedLayerNorm or choose 'torch'."
+            )
+        return FusedLayerNorm(hidden_size, elementwise_affine=elementwise_affine, eps=eps)
+    if layernorm_backbone == "torch":
+        return nn.LayerNorm(hidden_size, elementwise_affine=elementwise_affine, eps=eps)
+    raise ValueError("layernorm_backbone must be one of 'apex' or 'torch'.")
+
+
+def build_attention(
```
CharlelieLrt: 1. We have another attention already defined [there](https://github.com/NVIDIA/physicsnemo/blob/0183ddd8ab32245739b2fa43fd5f6cc1660de1ef/physicsnemo/models/diffusion/layers.py#L619-L733), based on `torch.nn.scaled_dot_product_attention` (SDPA). I just have 2 questions:
- Are the supported backends using torch SPDA?
- If not, do we want to also have an SPDA option in addition of the 3 already there?
- Are these attention backends supposed to be interchangeable?

2. Maybe it would also help to add a few lines in the docstring to explain the difference between these attention backends. For example, I am not very familiar with `natten2d`. Also, I assume the advantage of transformer_engine is performance?
Feel free to add links to papers or other repos in the docstring if you think that might help.

3. Maybe rename to `get_attention`, similarly to the other factories

4. Same comment as for the layer norm factory: could you double-check that the state dicts of all these attention backends are compatible with each other? [only if they are expected to be interchangeable?)

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -34,7 +48,395 @@
...
+    hidden_size: int,
+    num_heads: int,
+    attention: Literal["transformer_engine", "timm", "natten2d"],
+    attn_drop_rate: float = 0.0,
+    proj_drop_rate: float = 0.0,
```
CharlelieLrt, line 104: Missing descriptions in the docstring

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -34,7 +48,395 @@
...
+        -------
+        torch.Tensor
+            Output tensor of shape (Batch, Sequence_Length, Hidden_Size).
+        """
+        pass
+
+
+class TimmSelfAttention(AttentionModuleBase):
+    """Self-attention module that performs self-attention using the timm library implementation.
+    Expects an input tensor of shape (Batch, Sequence_Length, Hidden_Size) and
```
CharlelieLrt: Of shape (B, L, D) if you want to be consistent with the docstring of `build_attention`

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -34,7 +48,395 @@
...
+        Returns
+        -------
+        torch.Tensor
+            Output tensor of shape (Batch, Sequence_Length, Hidden_Size).
+        """
+        if attn_mask is not None and not timm_v1_0_16:
+            raise ValueError("attn_mask in TimmSelfAttention is only supported for timm version 1.0.16 and higher")
+        
+        if not timm_v1_0_16:
+            return self.attn_op(x)
```
CharlelieLrt, line 204: Would be better to avoid `mask` being silently ignored in this case, I think that could have undesirable effects for the user

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -34,7 +48,395 @@
...
+        else:
+            return self.attn_op(x, attn_mask=attn_mask)
+
+
+class TESelfAttention(AttentionModuleBase):
+    """Self-attention module that performs self-attention using the transformer_engine library implementation.
+    Expects an input tensor of shape (Batch, Sequence_Length, Hidden_Size) and
```
CharlelieLrt: Same comment as above for notation consistency

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -34,7 +48,395 @@
...
+            Input tensor of shape (Batch, Sequence_Length, Hidden_Size).
+        attn_mask: Optional[torch.Tensor]
+            The attention mask to apply to the input tensor (passed to timm's Attention module).
+            If None, no mask will be applied. This is only supported for timm version 1.0.16 and higher.
+
+        Returns
+        -------
+        torch.Tensor
+            Output tensor of shape (Batch, Sequence_Length, Hidden_Size).
+        """
```
CharlelieLrt: These docstrings for the `forward` method usually don't appear in the docs, or are not easily discoverable (because we usually never call `forward` directly, but `__call__`). For these reasons I prefer the docstring organization used by pytorch (see [there](https://github.com/pytorch/pytorch/blob/ba56102387ef21a3b04b357e5b183d48f0afefc7/torch/nn/modules/linear.py#L51-L86)), where everything is in the class docstring. For us, the cleanest way to do it is to have sections `Parameters` (for the `__init__`), `Forward` (for the inputs of the forward), and `Outputs` for the outputs of the forward (see [there](https://github.com/NVIDIA/physicsnemo/blob/0183ddd8ab32245739b2fa43fd5f6cc1660de1ef/physicsnemo/models/diffusion/layers.py#L620-L659) for an example).

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -34,7 +48,395 @@
...
+            Input tensor of shape (Batch, Sequence_Length, Hidden_Size).
+        attn_mask: Optional[torch.Tensor]
+            The attention mask to apply to the input tensor (passed to transformer_engine's MultiheadAttention module).
+            If None, no mask will be applied.
+        mask_type: Optional[str]
+            The type of mask to apply to the input tensor (passed to transformer_engine's MultiheadAttention module).
+            If no mask is provided, "no_mask" will be used.
+
+        Returns
+        -------
```
CharlelieLrt: Same comment for the docstring sections and organization

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -34,7 +48,395 @@
...
+        """
+        if attn_mask is not None:
+            mask_type = "arbitrary"
+        return self.attn_op(x, attention_mask=attn_mask, attn_mask_type=mask_type)
+
+
+class Natten2DSelfAttention(AttentionModuleBase):
+    """Self-attention module that performs 2D neighborhood attention using NATTEN.
+    Expects an input tensor of shape (Batch, Sequence_Length, Hidden_Size) and 
+    returns a tensor of the same shape (reshapes sequence to 2D internally).
```
CharlelieLrt, line 266: Would be good to have a little more details about what is NATTEN neighborhood attention

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -148,9 +545,32 @@ def __init__(
...
+        x: torch.Tensor
+            Input tensor of shape (Batch, Sequence_Length, Hidden_Size).
+        c: torch.Tensor
+            Conditioning tensor of shape (Batch, Hidden_Size).
+        attn_kwargs: Optional[Dict[str, Any]]
+            Additional keyword arguments for the attention module.
+        p_dropout: Optional[float | torch.Tensor]
+            The dropout probability for the intermediate dropout module. If None, no dropout will be applied.
+            If a scalar, the same dropout probability will be applied to all samples.
+            Otherwise, it should be a tensor of shape (Batch,) to apply per-sample dropout to each sample in a batch.
```
CharlelieLrt: Should update the existing docstring in the class instead of creating a new one here

CharlelieLrt (changes requested):
I think the PR does a good job at modularizing the attention in the DiT. I just left some comments mostly about documentation, naming conventions/consistency, etc...
Except those, everything looks good to me.

pyproject.toml
ktangsali: Sounds good to me

physicsnemo/experimental/models/dit/layers.py
pzharrington: Pivoted to `get_layer_norm` instead of `build_layer_norm`. The `backbone` name came in from DJ, but yeah considering this is in experimental and there are no released models w/ backward compat concerns, I updated it to `backend` as it makes more sense. The state dicts are compatible at least within the exposed arguments for layernorm

physicsnemo/experimental/models/dit/layers.py
pzharrington: Updated the docstrings with more detail. The attn backends are not interchangeable so I added a note about that. The timm implementation uses torch SDPA under the hood, and TE is a custom implementation.

physicsnemo/experimental/models/dit/layers.py
pzharrington: It's not silently ignored, the clause before this will catch it if the old timm was used and a mask was passed; the only way this section will get run is if there was no mask provided and the timm version is the older one that doesn't support that kwarg

CharlelieLrt: Right, sorry I missed the check just above

PR #1131, opened 2025-09-26
Implement recursive saving/loading of checkpoint for nested Modules

CharlelieLrt:
# PhysicsNeMo Pull Request

## Description

This PR introduces support for saving and loading nested `physicsnemo.Module`s. It is now possible to create nested modules with `m = Module(submodule, ...)`, and save and load them with `Module.save` and `Module.from_checkpoint`.
**⚠️Warning:** - The modules have to be `physicsnemo.Module`s, and not `torch.nn.Module`s.

### Problem

Prior to this PR, creating nested `physicsnemo.Module`s would give an error when trying to save the checkpoint:

```python
class SubModule(Module):
    def __init__(self, a):
        super().__init__()
        # ... submodule implementation ...
        
class MyModel(Module):
    def __init__(self, submodule):
        super().__init__()
        self.submodule = submodule
        # ... model implementation ...
        
subm = SubModule(1.0)
model = MyModel(subm)
model.save("checkpoint.mdlus")
```
`model.save` fails because the argument `submodule` passed to `MyModel.__init__` is not json-seriealzable.

### Solution

We modify the structure of the checkpoint files to account for nested submodules. The structure of `checkpoint.mdlus/args.json` becomes:

```json
{
    "__name__": # Unchanged
    "__module__": # Unchanged
    "__args__": {
        "submodule": "__physicsnemo.Module__.submodule"  # Replaced with a placeholder
    }
    "__physicsnemo.Module__.submodule": {
        "__name__": "SubModule"
        "__module__":  <path_to_module>
        "__args__": {
            "a":  <a_value>
        }       
    } # Nested submodules args are placed at the root of the args.json, with the prefix "__physicsnemo.Module__" to mark their names
}
````

Similarly, the `metadata.json` structure is modified, with information related to submodules placed at the root of the json file with a `__physicsnemo.Module__` prefix:

```json
{
    "physicsnemo_version" : # Unchanged
    "mdlus_file_version":  # Unchanged
    "__physicsnemo.Module__.submodule.mdlus_file_version": <mdlus_file_version for the submodule>
}    
```

The method `Module.from_checkpoint` is adapted to recursively instantiate submodules from the modified json files structures. For cases without nested modules, the structures of the `args.json` and `metadata.json` remain unchanged.
This feature supports multiple levels of nested modules, has well as multiple submodules at each level. The prefix for such submodules is of the form `__physicsnemo.Module__.arg_nameA.arg_nameB.arg_nameC....`.

### Implications

This PR enables common design pattern such as dependency injection or adapters. 

## Checklist

### DIFF
--- a/physicsnemo/models/module.py
+++ b/physicsnemo/models/module.py
@@ -402,8 +402,40 @@ class Module(torch.nn.Module):
 
         # Define some helper functions
         def _save_process(module, args, metadata, mod_prefix="") -> None:
-            """Helper function to recursively populate the dictionaries args
-            and metadata required for saving the module and its submodules."""
+            """Recursively serialize nested physicsnemo.Module instances for checkpoint saving.
+
+            Performs a depth-first search through the module's ``__init__`` arguments. When
+            an argument is a ``physicsnemo.Module`` instance, it is replaced with a
+            placeholder string (prefixed with ``_BASE_CKPT_PREFIX``) and the nested module's
+            information (``__name__``, ``__module__``, ``__args__``) is stored at the root level
+            of the ``args`` dictionary. The nested module metadata (e.g.,
+            ``__model_checkpoint_version__``) is also added at the root level
+            of ``metadata`` dictionary, with keys prefixed with
+            ``_BASE_CKPT_PREFIX``.
+
+            This allows for reconstruction of arbitrarily nested
+            module hierarchies during checkpoint loading.
+
+            Parameters
+            ----------
+            module : physicsnemo.Module
+                The module being processed
+            args : Dict[str, Any]
+                Dictionary to populate with serialized module arguments. Modified in-place.
+                Keys prefixed with ``_BASE_CKPT_PREFIX`` store nested module metadata.
+            metadata : Dict[str, Any]
+                Dictionary to populate with module metadata (e.g., version info).
+                Modified in-place.
+            mod_prefix : str, optional
+                Current module's prefix in the nested hierarchy, by default "". Root module
+                uses empty string; nested modules use format ``_BASE_CKPT_PREFIX.arg_name``.
+
+            Raises
+            ------
+            TypeError
+                If an argument is a ``torch.nn.Module`` instance that has not been converted
+                to a ``physicsnemo.Module`` using ``Module.from_torch``.
+            """
 
             # Pointer to args["__args__"] for submodules
             if mod_prefix == "":
@@ -423,7 +455,7 @@ class Module(torch.nn.Module):
                     )
                     _save_process(arg_value, args, metadata, next_mod_prefix)
                 elif isinstance(arg_value, torch.nn.Module):
-                    warnings.warn(
+                    raise TypeError(
                         f"Submodule {arg_name} of module {module.__class__.__name__} is"
                         f" a PyTorch module, which is not supported by 'Module.save'. Please "
                         f"first convert it to a PhysicsNeMo module using 'Module.from_torch'."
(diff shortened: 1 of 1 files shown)


physicsnemo/models/module.py
```diff
@@ -332,6 +339,43 @@ def save(self, file_name: Union[str, None] = None, verbose: bool = False) -> Non
...
+            else:
+                args_ptr = args[mod_prefix]["__args__"].copy()
+
+            for arg_name, arg_value in args_ptr.items():
+                if isinstance(arg_value, Module):
+                    next_mod_prefix = (
+                        f"{mod_prefix if mod_prefix else _BASE_CKPT_PREFIX}.{arg_name}"
+                    )
+                    args[next_mod_prefix] = arg_value._args.copy()
+                    # TODO: make sure this doesn't modify self._args
```
pzharrington: This `#TODO` is probably taken care of right? Since the initial args passed to this recursive function are a copy of `self._args`?

CharlelieLrt: Yes, this is taken care of now.

physicsnemo/models/module.py
```diff
@@ -552,8 +736,8 @@ class attribute. Attempting to override any other argument will raise
 
     @staticmethod
     def from_torch(
-        torch_model_class: torch.nn.Module, meta: ModelMetaData = None
-    ) -> "Module":
+        torch_model_class: type[torch.nn.Module], meta: ModelMetaData | None = None
```
pzharrington, line 871: Was the lack of `type[...]` the bug you mentioned? Seems like this flat out wouldn't have worked before if that is needed here...

physicsnemo/models/module.py
pzharrington, line 1: We should update the `Module` docstring to describe the updated requirements now that JSON-serializability is not necessarily required, but only in the case of nested `Module`s. Similarly the models readme in the docs folder should be updated...I know there is a concurrent effort to overhaul that page, so regardless on if we update the `.rst` in this PR we should make sure the changes propagate to that update as well.

physicsnemo/models/module.py
CharlelieLrt: No, that was not preventing anything from running, but that was causing lots of warnings in my type checker. So, I changed the type there. I think there are a few more places where the type hint could be improved.

_Note_: There was actually no bug in the model registry. I just did not anticipate that I need to call `from_torch` (which essentially creates a wrapper class around a `torch.nn.Module` subclass and register it in the registry), to be called _**in every test**_. I wrongly assumed that I could call `from_torch` once at the top of the test file, and reuse the wrapped class in every test, but it turns out the register state is cleaned up between tests, so the created class was lost. I have adapted the new CI tests to account for this, so this should be fixed now.

physicsnemo/models/module.py
pzharrington: And also call out that `from_torch` will not work with `from_checkpoint` (it didn't before, and was just not called out in the docs anywhere as far as I can tell). E.g. this fails:
```
class MyModule(Module):
    def __init__(self, submodule):
        super().__init__()
        self.linear = submodule

    def forward(self, x):
        return self.linear(x)

PModule = Module.from_torch(MyModule)
linear = nn.Linear(in_features=10, out_features=10)
m = PModule(linear)
m.save('dummy.mdlus')
m = Module.from_checkpoint('dummy.mdlus') # fails with TypeError
```

CharlelieLrt: It's been updated. LMK if there is still any other correction/improvement that could be made on the `physicsnemo.Module`'s docstring

CharlelieLrt:
@pzharrington sorry I missed your comment about the .rst file in the API docs. I have updated the API docs with instructions to construct nested Modules (both `physicsnemo.Module`s and `torch.nn.Module`s).

physicsnemo/models/module.py
```diff
@@ -332,6 +400,42 @@ def save(self, file_name: Union[str, None] = None, verbose: bool = False) -> Non
...
+                        f"{mod_prefix if mod_prefix else _BASE_CKPT_PREFIX}.{arg_name}"
+                    )
+                    args[next_mod_prefix] = arg_value._args.copy()
+                    args_ptr[arg_name] = next_mod_prefix
+                    metadata[f"{next_mod_prefix}.mdlus_file_version"] = (
+                        arg_value.__model_checkpoint_version__
+                    )
+                    _save_process(arg_value, args, metadata, next_mod_prefix)
+                elif isinstance(arg_value, torch.nn.Module):
+                    warnings.warn(
```
loliverhennigh: I would think instead of a warning you would actually want to raise an error if this is the case

physicsnemo/models/module.py
```diff
@@ -483,8 +594,161 @@ class attribute. Attempting to override any other argument will raise
...
+        # Define some helper functions
+        def _from_checkpoint_process(
+            cls_in,
+            args,
+            metadata,
+            override_args,
+            strict,
+            mod_prefix="",
+        ):
+            """Helper function to recursively instantiate the module and its
```
loliverhennigh: Could you add a doc string that is a bit more informative. Just for future reference that would help a lot. Same with the other checkpoint process helper function.

physicsnemo/models/module.py
CharlelieLrt: Agree, I replaced it.

physicsnemo/models/module.py
CharlelieLrt: Done

docs/api/physicsnemo.models.rst
megnvidia, line 1: line 328 "In that case, one should use" -> "In that case, use"

"**Simple Nesting with PhysicsNeMo Modules**

The simplest case is nesting"
simple is best replaced with 'basic'

megnvidia: but they must first be
converted to ``physicsnemo.Module`` using

but it must be converted to .........

megnvidia, line 1: "without converting it first:"

"without converting it"

PR #1141, opened 2025-10-03
Improve Sharded Convolution Implementation

coreyjadams:
# PhysicsNeMo Pull Request

- Fix some verbose tests
- Fix a docstring in point cloud ops
- Use functional instead of aten in sharded convolution tests.
- Make sure computation of halo output sizes is correct.

## Checklist


physicsnemo/distributed/shard_utils/conv_patches.py
```diff
@@ -278,12 +278,100 @@ def compute_output_shape(
...
+        ctx: torch.autograd.function.FunctionCtx,
+        grad_weight_or_bias: torch.Tensor,
+    ) -> torch.Tensor:
+        # Now, loop over the mesh dims and make sure we sync gradients
+        for mesh_dim in range(ctx.spec.mesh.ndim):
+            if ctx.spec.placements[mesh_dim].is_shard():
+                group = ctx.spec.mesh.get_group(mesh_dim)
+                dist.all_reduce(grad_weight_or_bias, group=group)
+
+        return grad_weight_or_bias, None
```
pzharrington, line 364: IIUC this folds in the model parallel reductions to get triggered during the backward pass; how would this interact with what FSDP automatically triggers for replicated weights after bwd completes? Wouldn't this mean things get reduced twice?

physicsnemo/distributed/shard_utils/conv_patches.py
```diff
@@ -295,208 +383,124 @@ def partial_conv_nd(
...
-        # Convert the local output to a ShardTensor
-        with annotate("partial_conv_nd.from_local"):
-            output = ShardTensor.from_local(
-                local_output,
-                input_spec.mesh,
-                input_spec.placements,
-                sharding_shapes=real_output_shapes,
-            )
+    #####################################################################
+    # Halo computations and metad data
```
pzharrington, line 402: Typo

physicsnemo/distributed/shard_utils/conv_patches.py
coreyjadams: Actually, no, this isn't for model parallel, this is necessary to get the math right in the backwards pass even without FSDP.

The gradient of the weights for a convolution is, itself, a convolution.  Since your gradients are sharded too, the output of the gradient-calculating convolution only has the updates from it's portion of grad_output.  Therefore, in a domain parallel case, you have to reduce the convolution gradients across the domain.

FSDP would then need to reduce-scatter, perhaps, across the "data" dimensions.

If you'd like to test it, if you remove this code all of the convolutional tests that check backwards pass / gradients will start failing in physicsnemo. :)

physicsnemo/distributed/shard_utils/conv_patches.py
```diff
@@ -651,7 +657,7 @@ def repackage_conv_transposed_args(
         "dilation": dilation,
         "output_padding": output_padding,
         "groups": groups,
-        "transposed": True,
+        # "transposed": True,
```
pzharrington, line 660: Can drop these commented out lines

physicsnemo/distributed/shard_utils/conv_patches.py
pzharrington: Ok no you're right I was mixing it up with FSDP, which should effectively sit on top of (& after) the reductions here take place. The reason I thought of it was, in previous projects there's been the trick of folding in this weight grad reduction into the full DDP reduction for performance reasons (mathematically equivalent to reduce over data+model groups as opposed to model then data independently for these grads, so might as well take advantage of the bucketing and infra setup by DDP).

pzharrington (approved):

physicsnemo/distributed/shard_utils/conv_patches.py
coreyjadams: Yes, definitely it would improve performance to take advantage of that bucketing.  I think FSDP is actually *not* bucketing in the last FSDP2 version, though I am not sure.  We were seeing it in the million-little-all-reduces in the optimizer step in profiles, I think.

We should think carefully about how to do this ... it becomes trickier when implementing this generic op-by-op framework instead of bespoke sharded operations.  The convolution implementation doesn't know if it's running with just 2 GPUs, or with 3D parallelism with shard tensor + FSDP ... making it hard to optimize here.  I haven't got the solution yet.

physicsnemo/distributed/shard_utils/conv_patches.py
coreyjadams: Done.

physicsnemo/distributed/shard_utils/conv_patches.py
coreyjadams: Fixed.

PR #1142, opened 2025-10-03
Transolver domain parallel

coreyjadams:
# PhysicsNeMo Pull Request

This PR implements adjustments to transolver to enable better, and more efficient, domain parallelism.  Batch size > 1 locally will not be supported for domain parallelism.

The key detail of these changes are to rearrange the data shapes.  Transolver takes inputs of shape `[Batch, Tokens, Features]`.  Previously, this was being converted to shapes of `[Batch, N_head, N_Tokens, Head_dim]`.  This pr essentialy replaces that with `[Batch, N_tokens, N_Head, Head_dim]` everywhere relevant.  The motivation for this is data contiguousness:  keeping the rearrange on the tensor dims with dim > sharded_dim is a 0-communication op, and everything stays nicely sharded.

Note as well, the attention mechanism is actually *not* parallel here (not that it impacts performance).  Check out the Transolver++ paper for details of their algorithm, but it's actually really easy to implement with ShardTensor - it just works.

https://arxiv.org/pdf/2502.02414v1

The idea is, the slice tokens are computed from a matmul of two pieces: `normed_weights`, and `fx`, and  the matmul contracts over the tokens dimension (which is the sharded dimension).  The output is a `Partial` placement, meaning each rank has a properly-shaped replica of the tokens but need to be all-reduced.

So, right before the attention layer, I implement a redistribute to a `Replicate()` placement which triggers the differentiable all reduce.  After the attention, the multiplication of `out_slice_tokens` and `slice_weights` produces a properly sharded outputs from the PhysicsAttention.  It's equivalent to this section of the upstream code (from the ++ model):
https://github.com/thuml/Transolver_plus/blob/main/models/Transolver_plus.py#L71-L73
but with one reduction, instead of two.

There is a boost to the attention path, too, in ShardTensor: it was computing ring Attention on replicated tensors which is slow (and wrong!) and now it is using a local attention on replicated tensors.

With these adjustments, I can efficiently scale the transolver model to about 3.6M input points on 8xA100.  

### Transolver ++

With these changes, we're actually only a hair's breadth away from having Transolver++.  I have the remaining updates on a separate branch where I want to validate training accuracy.

### Transformer Engine

Is not supported with domain parallelism.  The ops go through a different dispatch path - we need to implement that at some point, but it's not here.

### Tests

I added a new section of multi-gpu tests that are validating transolver end to end with shard tensor inputs.  It checks the outputs have the right shape, and the outputs are sharded properly.

### Changelog

I haven't updated it yet, I would like to bring Transolver ++ in, in it's entirety, and do it in one shot.

## Checklist

### DIFF
--- a/physicsnemo/distributed/shard_utils/attention_patches.py
+++ b/physicsnemo/distributed/shard_utils/attention_patches.py
@@ -642,7 +642,35 @@ def sdpa_wrapper(func: Callable, types: Any, args: tuple, kwargs: dict) -> Shard
     if q._spec.mesh.ndim != 1:
         raise MissingShardPatch("q must be on a 1D mesh")
 
-    return ring_sdpa(q, k, v, attn_mask, **kwargs)
+    # This is to implement sequence-parallel attention.
+    # Make sure the shardings are all the same:
+    if not (q._spec.placements[0] == k._spec.placements[0] == v._spec.placements[0]):
+        raise MissingShardPatch("q, k, and v must all be on the same placement")
+
+    # if the placements are replicatd (which is what we expect in transolver's Physics Attention)
+    # then just run locally and convert the output back to a replicated tensor:
+
+    if v._spec.placements[0].is_replicate():
+        local_q = q.to_local()
+        local_k = k.to_local()
+        local_v = v.to_local()
+        if attn_mask is not None:
+            local_attn_mask = attn_mask.to_local()
+        else:
+            local_attn_mask = None
+        local_output = torch.nn.functional.scaled_dot_product_attention(
+            local_q, local_k, local_v, attn_mask=local_attn_mask, **kwargs
+        )
+
+        output = ShardTensor.from_local(
+            local_output,
+            q._spec.mesh,
+            q._spec.placements,
+            # We don't have to worry about sharding shapes here since it's not sharded ...
+        )
+        return output
+    else:
+        return ring_sdpa(q, k, v, attn_mask, **kwargs)
 
 
 def repackage_sdpa_args(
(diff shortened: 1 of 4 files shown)


physicsnemo/models/transolver/Physics_Attention.py
```diff
@@ -129,28 +133,48 @@ def compute_slices_from_projections(
             - The aggregated features are normalized by the sum of weights for numerical stability.
         """
```
pzharrington, line 139: I think the tensor shapes in the docstring need to be updated to match the inline comments

physicsnemo/distributed/shard_utils/attention_patches.py
```diff
@@ -649,7 +649,42 @@ def sdpa_wrapper(func: Callable, types: Any, args: tuple, kwargs: dict) -> Shard
...
+        raise MissingShardPatch("q, k, and v must all be on the same placement")
+
+    # Make sure the attention mask, if provided, has the same placement as q, k, and v
+    if attn_mask is not None and hasattr(attn_mask, "_spec"):
+        if attn_mask._spec.placements[0] != q._spec.placements[0]:
+            raise MissingShardPatch(
+                "attn_mask must have the same placement as q, k, and v"
+            )
+
+    # if the placements are replicatd (which is what we expect in transolver's Physics Attention)
```
pzharrington: Small typo `replicatd`

physicsnemo/distributed/shard_utils/attention_patches.py
```diff
@@ -649,7 +649,42 @@ def sdpa_wrapper(func: Callable, types: Any, args: tuple, kwargs: dict) -> Shard
...
+        local_output = torch.nn.functional.scaled_dot_product_attention(
+            local_q, local_k, local_v, attn_mask=local_attn_mask, **kwargs
+        )
+
+        output = ShardTensor.from_local(
+            local_output,
+            q._spec.mesh,
+            q._spec.placements,
+            # We don't have to worry about sharding shapes here since it's not sharded ...
+        )
```
pzharrington, line 685: So this won't trigger CPU/GPU sync even though we left out the sharding shapes?

coreyjadams: Correct, CPU/GPU sync is only triggered on sharded dimensions.

PR #1147, opened 2025-10-08
Enable torch.nn.functional.pad with ShardTensor

coreyjadams:
# PhysicsNeMo Pull Request

## Description
This PR enables torch.nn.functional.pad with a sharded implementation.  The motivation for this is two fold: 
- In the DoMINO preprocessing pipeline, there is a padding step that would require this function.
- In UNet, which is now used in some paths of DoMINO, if the padding mode is not zeros, convolutions will dispatch to this function for padding.  The conv-with-non-zero-padding is included as a test.

Therefore, enabling this with domain parallelism is useful for DoMINO in particular, but I made sure the implementation is generic.

### The Implementation

This domain-parallel implementation is straightfoward: 95% of the logic and code is spent deciding, for each rank, how to convert a global padding tuple into a local padding tuple.  To avoid all communication overhead, we have to compute the output shape on each rank + on all the ranks in the same row/column/etc as the local rank.

The actual padding of the local tensor is then dispatched to the underlying torch.Tensor implementation.  Using `ShardTensor.from_local` and `ShardTensor.to_local` makes the whole path differentiable.

## Checklist

### DIFF
--- a/physicsnemo/distributed/shard_utils/attention_patches.py
+++ b/physicsnemo/distributed/shard_utils/attention_patches.py
@@ -401,9 +401,7 @@ class RingSDPABlocking(torch.autograd.Function):
     ) -> torch.Tensor:
         """
         Forward pass for the ring attention implementation.  This implementation
-        will overlap the communication with the computation.  Note that there is an
-        explicit sync in each iteration to prevent the communication stream from getting
-        ahead of the computation stream, by waiting on the all_to_all operation to complete.
+        will NOT overlap the communication with the computation.
         """
 
         ctx.attn_args = attn_args
@@ -569,14 +567,24 @@ def ring_sdpa(
     **kwargs: dict,
 ) -> ShardTensor:
     """
-    High Level, differentiable function to compute neighborhood attention on a sharded tensor.
+    High Level, differentiable function to compute global attention on a sharded tensor.
 
-    Operation works like so:
-    - Figure out the size of halos needed.
-    - Apply the halo padding (differentiable)
-    - Perform the neighborhood attention on the padded tensor. (differentiable)
-    - "UnHalo" the output tensor (different from, say, convolutions)
-    - Return the updated tensor as a ShardTensor.
+    The implementation is a ring communication pattern.  Each rank computes attention
+    locally on it's tensors, and then kv is passed to the next rank while receiving from
+    the previous rank.
+
+    Parameters:
+    ----------
+        q (physicsnemo.distributed.ShardTensor): The attention queries
+        k (physicsnemo.distributed.ShardTensor): The attention keys
+        v (physicsnemo.distributed.ShardTensor): The attention values
+        attn_mask (physicsnemo.distributed.ShardTensor, optional): The attention mask
+        kwargs (dict): key-word arguments to pass to the attention call
+
+    Returns:
+    -------
+        physicsnemo.distributed.ShardTensor: A distributed tensor representing the
+           attention computed on the global context.
 
     """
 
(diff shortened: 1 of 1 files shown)


physicsnemo/distributed/shard_utils/padding.py
```diff
@@ -0,0 +1,248 @@
...
+    # The only way to do that is to loop over the paddings to determine
+    # the tensor axes, and then loop over the mesh / spec to see if that axis is
+    # sharded.
+    #
+    # Further, because we don't want to communicate across GPUs unless it's needed,
+    # We need to compute this for all tensors in the shard spec.
+
+    # Sanity checks
+    if len(pad) % 2 != 0:
+        raise ValueError("Sharded Padding requires len(pag) to be divisible by 2.")
```
Alexey-Kamenev: `len(pag)` -> `len(pad)`?

physicsnemo/distributed/shard_utils/padding.py
```diff
@@ -0,0 +1,248 @@
...
+
+    pad_dims = len(pad) // 2
+
+    if pad_dims > len(inputs.shape):
+        raise ValueError("Sharded Padding specified for {pad_dim}")
```
Alexey-Kamenev: The string _looks_ like it assumes it's an f-string but it's not (no `f` prefix)? Also, `pad_dim` -> `pad_dims`?

physicsnemo/distributed/shard_utils/padding.py
```diff
@@ -0,0 +1,248 @@
...
+        mesh_sizes: The sizes of the mesh
+        tensor_sharding_map: A map from tensor dimension to mesh dimension
+
+    Returns:
+        tuple of the local padding and output shape
+    """
+
+    tensor_rank = len(input_tensor_shape)
+
+    pad_dims = len(pad) // 2
```
Alexey-Kamenev, line 58: This assumes the padding is always specified for both ends e.g. there is no broadcasting etc?

physicsnemo/distributed/shard_utils/padding.py
```diff
@@ -0,0 +1,248 @@
...
+        # If this axis of the tensor is not sharded, we keep these as they are:
+        if tensor_dim not in tensor_sharding_map.keys():
+            output_padding.append(left)
+            output_padding.append(right)
+            local_output_shape[tensor_dim] += left + right
+        else:
+            # The tensor is sharded on this dim.
+            # So, determine if this is an edge or not.
+
+            # four cases here.
```
Alexey-Kamenev, line 80: Nice, thank you for the comment - it helped!

physicsnemo/distributed/shard_utils/padding.py
```diff
@@ -0,0 +1,248 @@
...
+
+@profile
+def repackage_pad_args(
+    inputs: ShardTensor,
+    pad: Union[int, Tuple[int, ...]] = 0,
+    mode: str = "constant",
+    value: float | None = None,
+    *args,
+    **kwargs,
+) -> Tuple[
```
Alexey-Kamenev: Nit: `Tuple` -> `tuple`? And replace `Union` with `|`?

physicsnemo/distributed/shard_utils/padding.py
coreyjadams: Done, thanks!

physicsnemo/distributed/shard_utils/padding.py
coreyjadams: Fixed!

physicsnemo/distributed/shard_utils/padding.py
coreyjadams: Yeah, there is no broadcasting here.  FWIW, there is not in the upstream torch either: https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.pad.html

I do have a check in the entry function that len(pad) % 2 == 0, at least.  

There are some layers like https://docs.pytorch.org/docs/stable/generated/torch.nn.ConstantPad2d.html#torch.nn.ConstantPad2d that will broadcast, but under the hood they are calling this function and the padding happens in the nn.Module level code.

physicsnemo/distributed/shard_utils/padding.py
coreyjadams: Done, got all the other `Tuple` -> `tuple` changes too.  Thanks!

physicsnemo/distributed/shard_utils/padding.py
Alexey-Kamenev: Thank you, makes sense!

PR #1151, opened 2025-10-09
DiT refactor to support custom Module

pzharrington:
# PhysicsNeMo Pull Request

## Description
Refactors `phyiscsnemo.experimental.models.DiT` to support:

- Distinct operations for tokenize, de-tokenize, and self-attention. These can be chosen from a number of available pre-defined backends, or a custom `phyiscsnemo.Module` defined and passed by the user which conforms to the specified API
- ND inputs/outputs, as opposed to just 2D inputs, are enabled by the above if user defines the appropriate tokenizers
- Adds misc features and optional arguments required by some internal projects using DiT; default behavior remains the same as before

## Checklist

### DIFF
--- a/physicsnemo/experimental/models/dit/dit.py
+++ b/physicsnemo/experimental/models/dit/dit.py
@@ -184,7 +184,7 @@ class DiT(Module):
         self.condition_dim = condition_dim
 
         # Input validation
-        if not isinstance(attention_backend, str) and attention_backend not in ["timm", "transformer_engine", "natten2d"]:
+        if attention_backend not in ["timm", "transformer_engine", "natten2d"]:
             raise ValueError("attention_backend must be one of 'timm', 'transformer_engine', 'natten2d'")
 
         if layernorm_backend not in ["apex", "torch"]:


physicsnemo/models/diffusion/layers.py
```diff
@@ -1049,32 +1060,62 @@ def __init__(
         endpoint: bool = False,
         amp_mode: bool = False,
         learnable: bool = False,
+        freq_embed_dim: int | None = None,
+        mlp_hidden_dim: int | None = None,
+        embed_fn: str = "cos_sin",
```
CharlelieLrt: The type hint for `embed_fn` should preferable be `"cos_sin" | "np_sin_cos"`. Also adding some validation in the __init__ would help catching errors when an unsupported option is provided.

physicsnemo/models/diffusion/layers.py
```diff
@@ -1040,6 +1040,17 @@ class PositionalEmbedding(torch.nn.Module):
         A boolean flag indicating whether mixed-precision (AMP) training is enabled. Defaults to False.
     learnable : bool, optional
         A boolean flag indicating whether learnable positional embedding is enabled. Defaults to False.
+    freq_embed_dim: int, optional
+        The dimension of the frequency embedding. Defaults to None, in which case it will be set to num_channels // 2.
```
CharlelieLrt: > Defaults to None, in which case it will be set to num_channels // 2

This default is `num_channels` instead of `num_channels // 2`, right?

physicsnemo/models/diffusion/layers.py
CharlelieLrt, line 1033: I feel like in this new implementation, it's a little bit unclear how the `num_channels` is used. When passing `freq_embed_dim` and `mlp_hidden_dim`, it is totally ignored, right? When passing only one (or none) of these two, it is essentially used to define the default input features and hidden features sizes... At least that's my understanding, but I think it would be worth clarifying in the docstring.

physicsnemo/models/diffusion/layers.py
```diff
@@ -1040,6 +1040,17 @@ class PositionalEmbedding(torch.nn.Module):
...
+    mlp_hidden_dim: int, optional
```
CharlelieLrt, line 1045: It's ignored if `learnable` is set to `False`, correct? If so, maybe better to precise it there too

physicsnemo/models/diffusion/layers.py
```diff
@@ -1049,32 +1060,62 @@ def __init__(
...
             )
 
-    def forward(self, x):
+        if self.embed_fn == "np_sin_cos":
+            half_embed_dim = freq_embed_dim // 2
+            pow = np.arange(half_embed_dim, dtype=np.float32) / half_embed_dim
+            w = np.exp(-np.log(self.max_positions) * pow)
+            self.register_buffer("freqs", torch.from_numpy(w).float())
+
+    def cos_sin_embedding(self, x):
```
CharlelieLrt: Make this one private with `_cos_sin_embedding` (same for the following one), unless we expect to expose them to the user (also the docs build and renders by default all methods that are public, ie not starting with `_`). I'm not sure, but I think we don't want those to be included in the user docs, as they are just some helper functions

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -53,12 +53,24 @@ class DiT(Module):
...
+        Spatial dimensions of the input. If an integer is provided, the input is assumed to be on a square 2D domain.
+        If a tuple is provided, the input is assumed to be on a multi-dimensional domain.
     in_channels (int):
         The number of input channels..
-    patch_size (Union[int, Tuple[int, int]], optional):
-        The size of each image patch along height and width. Defaults to (8,8).
+    patch_size (Union[int, Tuple[int]], optional):
+        The size of each image patch. Defaults to (8,8). If an integer is provided, the patch_size is assumed to be a square 2D patch.
+        If a tuple is provided, the patch_size is assumed to be a multi-dimensional patch.
+    tokenizer (Union[str, Module], optional):
```
CharlelieLrt: This comment applies to the tokenizer and detokenizer (and every time we pass a`Module` to another one). It gives great flexibility to be able to pass `Module`'s but it also significantly increases the chances of errors. Most errors will be because of:
1. **Signature mismatch**: The user passes a `Module` that does not have the required signature
2. **Return mismatch:** The user passes a`Module` that does not return the expect objects
3. **Shapes mismatch:** The user passes a `Module` that has correct signature and return, but the input and/or output tensor don't have the right shape.

I see 2 possible solutions to mitigate these risks:

1. **Documentation:** clearly specify the expected signature and returns of the `Module`. For example,

```python
r"""
...

The tokenizer must a `physicsnemo.Module` that can be called with:

.. code-block:: python

  tokens = tokenizer(x, y, z)

where `tokens` is a Tensor of shape `(A, B, C)`, x is a Tensor of shape `(A, B, C)`, ...

Parameters
-----------
...
"""
```

2. **Strong parameter validation:** catch type and shape errors early in the `forward` before they propagate. For example here we could check that the inputs passed to the tokenizer and detokenizer are always of the right type and expect type shape. We could also do this type/shape checking on the outputs.

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -53,12 +53,24 @@ class DiT(Module):
...
+        The tokenizer to use. Defaults to 'patch_embed_2d'. You may provide:
+        - A string in {"patch_embed_2d"} to select a built-in tokenizer. Built-in tokenizers include:
+            - 'patch_embed_2d': Uses a standard PatchEmbed2D to project the input image to a sequence of tokens.
+        - An instantiated PhysicsNeMo `Module` implementing the tokenizer interface defined in physicsnemo.experimental.models.dit.layers.TokenizerModuleBase.
+    detokenizer (Union[str, Module], optional):
```
CharlelieLrt: Instead of just `str` in the type hint, we could use a literal `"proj_reshape_2d"`. This makes the required type more explicit

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -53,12 +53,24 @@ class DiT(Module):
...
-    patch_size (Union[int, Tuple[int, int]], optional):
-        The size of each image patch along height and width. Defaults to (8,8).
+    patch_size (Union[int, Tuple[int]], optional):
+        The size of each image patch. Defaults to (8,8). If an integer is provided, the patch_size is assumed to be a square 2D patch.
+        If a tuple is provided, the patch_size is assumed to be a multi-dimensional patch.
+    tokenizer (Union[str, Module], optional):
+        The tokenizer to use. Defaults to 'patch_embed_2d'. You may provide:
+        - A string in {"patch_embed_2d"} to select a built-in tokenizer. Built-in tokenizers include:
+            - 'patch_embed_2d': Uses a standard PatchEmbed2D to project the input image to a sequence of tokens.
+        - An instantiated PhysicsNeMo `Module` implementing the tokenizer interface defined in physicsnemo.experimental.models.dit.layers.TokenizerModuleBase.
```
CharlelieLrt: Note: the `physicsnemo.experimental.models.dit.layers.TokenizerModuleBase` does not link properly in the docs. It should be ``:class:`~physicsnemo.experimental.models.dit.layers.TokenizerModuleBase` ``

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -70,29 +82,51 @@ class DiT(Module):
         The number of attention heads. Defaults to 8.
     mlp_ratio (float, optional):
         The ratio of the MLP hidden dimension to the embedding dimension. Defaults to 4.0.
-    attention_backend (str, optional):
-        If 'timm' uses Attention from timm. If 'transformer_engine', uses MultiheadAttention from transformer_engine. Defaults to 'transformer_engine'.
+    attention_backend (Union[str, Module, optional):
```
CharlelieLrt: Same comment for the type hint and the signature documentation

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -70,29 +82,51 @@ class DiT(Module):
...
 
-    Outputs
+    Returns
     -------
-    torch.Tensor: The output tensor of shape (N, out_channels, H, W).
+    torch.Tensor:
+        The output tensor of shape (N, out_channels, *spatial_dims). `spatial_dims` is determined by the input_size/dimensionality.
     
-    Notes
+    Note
```
CharlelieLrt, line 130: I think this should remain `Notes`, the `Note` will not be parsed correctly in the docs build (I assume)

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -70,29 +82,51 @@ class DiT(Module):
...
         If True, applies the DiT specific initialization. Defaults to True.
-
+    tokenizer_kwargs (Dict[str, Any], optional):
+        Additional keyword arguments for the tokenizer module.
+    detokenizer_kwargs (Dict[str, Any], optional):
+        Additional keyword arguments for the detokenizer module.
+    block_kwargs (Dict[str, Any], optional):
+        Additional keyword arguments for the DiTBlock modules.
+    timestep_embed_kwargs (Dict[str, Any], optional):
+        Additional keyword arguments for the timestep embedding module constructor.
```
CharlelieLrt: Hmm did I miss something? The time step embedding module constructor does not appear anywhere in the docstring. What are the accepted keyword arguments that I could pass to it in `timestep_embed_kwargs`?

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -143,52 +185,82 @@ def __init__(
...
 
-        self.x_embedder = PatchEmbed2D(
-            self.input_size,
-            self.patch_size,
-            in_channels,
-            hidden_size,
-        )
-        self.t_embedder = PositionalEmbedding(hidden_size, amp_mode=self.meta.amp_gpu, learnable=True)
-        init_zero = dict(init_mode="kaiming_uniform", init_weight=0, init_bias=0)
+        if not isinstance(attention_backend, str):
```
CharlelieLrt: I think we need stronger validation here: if the string is not in the accepted values, we should raise an error.

This comment also applies to the `tokenizer` and `detokenizer`

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -143,52 +185,82 @@ def __init__(
...
+            if not isinstance(attention_backend, AttentionModuleBase):
```
CharlelieLrt: I understand that this mechanism forces that, in the case of a custom `attention_backend`, the user has to subclass `AttentionModuleBase`. This has pros and cons.

- Pros: strong enforcement of the class APIs. As a result, we can safely assume that the instance passed here as some specific methods and attributes. But do we need such strong assumptions?
- Cons: another abstraction, so more effort for the user.

An alternative, could be to simply pass any `Module` instantiated by the user. This solution also has pros and cons:

- Pros: less abstraction to understand and build, so easier for the user.
- Cons: from our side this requires good documentation + parameter validation to make sure that the `Module` passed by the user satisfies our requirements.

I'm not sure which solution would be best here, so I'll let you make your call. Happy to discuss more if you have questions.

**Note:** these remarks also apply to the `tokenizer` and `detokenizer`

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -143,52 +185,82 @@ def __init__(
...
+                hidden_size=hidden_size,
+                tokenizer=tokenizer,
+                **tokenizer_kwargs,
+            )
+        else:
+            if not isinstance(tokenizer, Module):
+                raise TypeError("tokenizer must be a string or a physicsnemo.models.Module instance")
+            self.tokenizer = tokenizer
+
+        self.t_embedder = PositionalEmbedding(hidden_size, amp_mode=self.meta.amp_gpu, learnable=True, **timestep_embed_kwargs)
```
CharlelieLrt, line 214: Ok now I understand that the time step embedding module I mentioned in [this comment](https://github.com/NVIDIA/physicsnemo/pull/1151/files#r2427459221) is actually the `PositionalEmbedding`. Maybe put a link to this class in the docstring, so users will see what are the accepted values for `timestep_embed_kwargs`.

Another question: is this class sufficient to cover most use-cases to embed time-step information? If not, is it worth allowing custom-defined `Module` to do the time-step embedding?

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -143,52 +185,82 @@ def __init__(
...
+
 
-        self.blocks = nn.ModuleList(
-            [
+        blocks = []
+        for _ in range(depth):
+            if isinstance(attention_backend, str):
+                attn_module = attention_backend
+            else:
+                custom_attn_module_constructor = attention_backend.__class__
```
CharlelieLrt, line 251: That's strange.. Why do you re-instantiate the `attention_backend` here? Why not using the instance directly? When I implemented the nested `Module` I did not envision use case like that, and I'm not sure the feature I implemented would work with such use case. Also this wastes memory, because the user instantiate a module, which is actually never used (only its class is used). Even worse: all custom parameters the the user might have passed to its instantiated `Module` are erased here, so the user ends to using a totally different `Module` instance, without any notice of it.

Unless there is a very strong need for this, I think we should stay away from this pattern, and instead directly use the `Module` instance that is already created.

physicsnemo/experimental/models/dit/layers.py
CharlelieLrt, line 160: Following-up on my [other comment](https://github.com/NVIDIA/physicsnemo/pull/1151/files#r2427504136), the fact that the `AttentionModuleBase` class interface does not require anything else except a `forward` method (which must already be defined in any case), makes me believe that it is not totally necessary to introduce a new abstraction there, and it might just be easier to allow the user to use any `Module` as long as it follows the required signature and output types/shapes.

Maybe your goal is to make sure that the `forward` signature is `x (Tensor) ->y (Tensor)`, but using an abstract class is not a good mechanism for that, since anyone who wants to define a new custom `Module` subclass would have to override the `forward`.

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -664,3 +658,304 @@ def forward(self, x: torch.Tensor, c: torch.Tensor) -> torch.Tensor:
...
+        Token sequence of shape (B, L, D), where L is the length of the token sequence (number of patches for a standard 2D DiT).
+    """
+
+    @abstractmethod
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        pass
+
+    @abstractmethod
+    def initialize_weights(self):
+        pass
```
CharlelieLrt, line 688: Following on my two previous comments ([here](https://github.com/NVIDIA/physicsnemo/pull/1151/files#r2427504136) and [there](https://github.com/NVIDIA/physicsnemo/pull/1151/files#r2427558252)). I see that here we expect the tokenizer module interface to also define a method to initialize the weights, in addition of the `forward`. In that case I think that makes sense to introduce a new abstraction for that. We can't really allow the user to pass any `Module` instance, because we would also need to provide clear instructions about the `initialize_weights` and that becomes too much to document and will lead to unsafe behaviors.

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -664,3 +658,304 @@ def forward(self, x: torch.Tensor, c: torch.Tensor) -> torch.Tensor:
...
+
+
+class PatchEmbed2DTokenizer(TokenizerModuleBase):
```
CharlelieLrt, line 691: Question that applies to all the new classes and factories that you introduced in `layers.py`: do we want to keep them as semi-private within the `layers.py` module, or do we want to expose them to the user? In the latter, we should expose them in the `dit/__init__.py`

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -664,3 +658,304 @@ def forward(self, x: torch.Tensor, c: torch.Tensor) -> torch.Tensor:
...
+    """Default tokenizer using `PatchEmbed2D` followed by a learnable positional embedding.
```
CharlelieLrt: Maybe not "Default". Some users might just import it with a few other layer classes and  use their to build their own custom `Module`

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -664,3 +658,304 @@ def forward(self, x: torch.Tensor, c: torch.Tensor) -> torch.Tensor:
...
+
+    Returns a module whose forward accepts (B, C, *spatial_dims) and returns (B, L, D). 
+    `spatial_dims` is determined by the input_size/dimensionality.
+
+    Parameters
+    ----------
+    input_size: Tuple[int, int]
+        The size of the input image.
+    patch_size: Tuple[int, int]
+        The size of the patch.
```
CharlelieLrt: This seems only applicable to 2D cases?

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -664,3 +658,304 @@ def forward(self, x: torch.Tensor, c: torch.Tensor) -> torch.Tensor:
...
+        return x
+
+
+
+def get_detokenizer(
+    input_size: Tuple[int, int],
+    patch_size: Tuple[int, int],
+    out_channels: int,
+    hidden_size: int,
+    detokenizer: str = "proj_reshape_2d",
```
CharlelieLrt: Literal for the type hint

physicsnemo/experimental/models/dit/dit.py
pzharrington: Other way around :) see e.g. https://github.com/NVIDIA/physicsnemo/blob/a8af5cdaf87372573b6ea4b47ddce36f537a33ce/physicsnemo/models/fno/fno.py#L797

physicsnemo/experimental/models/dit/dit.py
pzharrington: I think this one class should be sufficient for now. There is imo no real need to change the implementation of this part, the only reason I'm adding a new variant is to support the already-trained checkpoints for scv2

physicsnemo/experimental/models/dit/dit.py
pzharrington: Removing this option until we get a better solution (i.e., being able to pass just the custom class, and exempt it from JSON-serializability requirement)

physicsnemo/experimental/models/dit/layers.py
pzharrington: Hmm, this is a good point. Might as well add them to the `__init__` in case people develop custom `Module`s that import them.

pzharrington:
Ok I believe I addressed everything, thanks for the comments. Docstrings / type hints should be improved now, and we will save support for the custom attention module in top-level DiT until later when a need arises.

test/models/test_dit.py
```diff
@@ -15,13 +15,20 @@
...
+
 import pytest
 import torch
 import torch.nn as nn
+import torch.nn.functional as F
 from pytest_utils import import_or_fail
 
 from physicsnemo.experimental.models.dit import DiT
-from physicsnemo.experimental.models.dit.layers import DiTBlock
+from physicsnemo.experimental.models.dit.layers import (
```
CharlelieLrt, line 27: @pzharrington sorry I actually forgot to review the test in my previous pass. It looks good to me, but could you just add a test that saves (with `Module.save`) an instance of DiT using custom submodules, and then tries to reload it with `Module.from_torch` (and maybe some forward validation to make sure they give the same outputs). I've heard some reports of bugs with the nested module feature, so trying to make sure it doesn't break

pzharrington: This test should capture that, no? https://github.com/NVIDIA/physicsnemo/blob/67b9c9e2881b3dc5624e3283cb7bd872f4531157/test/models/test_dit.py#L192 (I assume you mean `Module.from_checkpoint` instead of `Module.from_torch`)

CharlelieLrt: Ah yes it does! I missed this test, thanks

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -143,52 +183,91 @@ def __init__(
...
-        self.x_embedder = PatchEmbed2D(
-            self.input_size,
-            self.patch_size,
-            in_channels,
-            hidden_size,
-        )
-        self.t_embedder = PositionalEmbedding(hidden_size, amp_mode=self.meta.amp_gpu, learnable=True)
-        init_zero = dict(init_mode="kaiming_uniform", init_weight=0, init_bias=0)
+        # Input validation
+        if not isinstance(attention_backend, str) and attention_backend not in ["timm", "transformer_engine", "natten2d"]:
```
CharlelieLrt: @pzharrington this if statement doesn't make sense I think: if `attention_backend` is not an `str`, then there is no way `attention_backend` is in `["timm", "transformer_engine", "natten2d"]`

physicsnemo/experimental/models/dit/dit.py
CharlelieLrt: Oh ok good to know. It seems both are accepted, but they are not rendered the same way in the docs

physicsnemo/experimental/models/dit/dit.py
pzharrington: Ah good catch, I thought I updated this one but missed it. Fixed now.

CharlelieLrt (approved):
LGTM!

PR #1219, opened 2025-11-08
Added coding standards for model implementations as a custom context for greptile

CharlelieLrt:
# PhysicsNeMo Pull Request

## Checklist

## Review Process

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -13,7 +13,7 @@
 - [ ] The documentation is up to date with these changes.
 - [ ] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
 - [ ] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.
-- [ ] If I am implementing a new model or modifying any exusting model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation).
+- [ ] If I am implementing a new model or modifying any existing model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation).
 
 ## Dependencies
 


CODING_STANDARDS/MODELS_IMPLEMENTATION.md
```diff
@@ -0,0 +1,484 @@
...
+complete models that are applicable to a specific domain or specific data
+modality.
+
+---
+
+### MOD-001: Use proper class inheritance for all models
+
+**Description:**
+All model classes must inherit from `physicsnemo.Module`. Subclasses of
+`torch.nn.Module` are not allowed.
```
coreyjadams: We should maybe say _direct_ subclasses are not allowed, but physicsnemo.Module is, itself, a subclass of torch.nn.Module.

CODING_STANDARDS/MODELS_IMPLEMENTATION.md
```diff
@@ -0,0 +1,484 @@
...
+Ensure proper initialization of parent classes using `super().__init__()`. Pass
+the `meta` argument to the `super().__init__()` call if appropriate, otherwise
+set it manually with `self.meta = meta`.
+
+**Rationale:**
+Ensures invariants and functionality of the `physicsnemo.Module` class for all
+models.
+
```
coreyjadams, line 171: I think we should expand on this.  The core motivations for this requirement should be listed here, for users who want to know:

- Enable better serialization of models that use sub-pnm-modules.

...Is there another reason?

CODING_STANDARDS/MODELS_IMPLEMENTATION.md
```diff
@@ -0,0 +1,484 @@
...
+**Example:**
+
+```python
+from physicsnemo.models import Module
```
coreyjadams: This should probably be `physicsnemo.core`

CODING_STANDARDS/MODELS_IMPLEMENTATION.md
```diff
@@ -0,0 +1,484 @@
...
+  folders directly, and backward compatibility is guaranteed. Another exception
+  is when the model class is highly specific to a single example. In this case,
+  it may be acceptable to place it in a module specific to the example code,
+  such as for example `examples/<example_name>/utils/nn.py`.
+
+- Stage 2: Production. After staying in stage 1 for a sufficient amount of time
+  (typically at least 1 release cycle), the model class is promoted to stage 2.
+  It is then moved to the `physicsnemo/nn` or `physicsnemo/models` folders,
+  based on the rule `MOD-000`. During this stage, backward compatibility is
+  guaranteed.
```
coreyjadams, line 218: We should consider what "production" models look like re:checkpoints.  Are they meant to be compatible with pre-trained instances on HF or NGC?  Do we need any versioning?

CODING_STANDARDS/MODELS_IMPLEMENTATION.md
```diff
@@ -0,0 +1,484 @@
...
+  are not recognized by our Sphinx documentation and are prohibited.
+
+- All methods should be documented with a docstring, with at least a `Parameters`
+  section and a `Returns` section. Other sections such as `Notes`, `Examples`,
+  or `..important::` or `..code-block:: python` are possible. Other sections
+  are not recognized by our Sphinx documentation and are prohibited.
+
+- All tensors should be documented with their shape, using LaTeX math notation such
+  as :math:`(N, C, H_{in}, W_{in})` (there is flexibility for naming the
+  dimensions, but te math format should be enforced). Our documentation is
```
coreyjadams: te -> the

CODING_STANDARDS/MODELS_IMPLEMENTATION.md
```diff
@@ -0,0 +1,484 @@
...
+        :math:`(B, D_{latent})`. If ``True``, returns tuple of
+        (latent, hidden) where hidden has shape :math:`(B, D_{hidden})`.
+
+    Examples
+    # Rule: Examples section is allowed and helpful
+    --------
+    >>> model = SimpleEncoder(input_dim=784, latent_dim=64)
+    >>> x = torch.randn(32, 784)
+    >>> latent = model(x)
+    >>> latent.shape
```
coreyjadams, line 465: We should have a comment in the guide about this.  It's excellent to have; we could to note to users that code examples must run successfully since our CI checks it.

CODING_STANDARDS/MODELS_IMPLEMENTATION.md
CharlelieLrt: Edited

CODING_STANDARDS/MODELS_IMPLEMENTATION.md
CharlelieLrt: Added more details: serialization and checkpointing as you said, versioning system for backward compatibility, and can be registered in the registry (not possible with `torch.nn.Module`)

CODING_STANDARDS/MODELS_IMPLEMENTATION.md
CharlelieLrt: Or just `from physicsnemo import Module`? Are we planning to expose `Module` in the top-level package like it is currently the case?

CODING_STANDARDS/MODELS_IMPLEMENTATION.md
CharlelieLrt: Do you mean all models should have a runnable `Examples` section? For some we only have `.. code-block::`, would that be fine or do you think Examples is needed?

CODING_STANDARDS/MODELS_IMPLEMENTATION.md
CharlelieLrt: Good point. Actually the model versioning system is a key feature of `physicsnemo.Module`, and I think it deserves its own rule. I'll add one

CharlelieLrt:
@coreyjadams following up on these 2 things.

> Production Testing structure. What tests are minimally required, what tests are encouraged? We should note that models can not move out of experimental without complete test coverage.

Right, I was planning on adding a rule for that.

> Requirements. For models that introduce extra requirements, we should prescribe a method to note and track that. TBD on managing requirements in general so I'm not sure we need a solution now, but we should update it here when we get there.

What type of mechanism are you thinking of here? For example if something use apex and its not available we use a try/catch:

```python
try:
    import apex
except ImportError:
    OptionalDependencyFailure("apex")
    apex = None
```

It's the pattern used in E2S for optional dependencies. What about cases where there is a fallback (e.g. many of apex layers can just fallback to torch layers in case apex is not available)?

CODING_STANDARDS/MODELS_IMPLEMENTATION.md
coreyjadams: I don't think every model needs a runnable example, what I meant was that this section - if present - will get tested by CI so it's worth a note in the guidelines about it.  Something like ...

```
Your docstring is encouraged to have an example section to demonstrate basic construction and usage.  Note that our CI system will automatically test these sections for correctness, when present.
```

I wasn't trying to force anything complicated here :)

coreyjadams:
re:requirements ...

I'm not sure what the right solution is.  Ideally, I want a system in physicsnemo that can do the following for us:
- For missing required dependencies, fail with an informative message.
- For missing optional dependencies, warn with an informative message without overwhelming.
- For models, track with CI what is used in the model vs. what is stated in the requirements.  If a user or developer adds a new dependency, but doesn't update the requirements, CI should reject / fix it.
- Ideally, avoid too many try/except blocks that then have conditional paths laters.  Can't always be avoided of course.

I just want to get to a state where we have consistent requirements to make installations easier for target models, if that makes sense, without bogging us all down in import paperwork more or less.  I haven't actually got a plan for implementation, this is just daydreaming at the moment.

CODING_STANDARDS/MODELS_IMPLEMENTATION.md
CharlelieLrt: Added rule for versioning models.

CODING_STANDARDS/MODELS_IMPLEMENTATION.md
CharlelieLrt: Changed to `from physicsnemo import Module`

PR #1235, opened 2025-11-19
V2.0 refactor

coreyjadams:
# PhysicsNeMo Pull Request

Rearchitecture of core physicsnemo components.  Highlights:
- Break circular deps in internal imports, and ensure they can't come back via a linting step.
- Organize external dependencies via dependency groups.  Note that `models` and `metrics` have a few "-extra" dependencies that are unprotected still, to be fixed before merge.
- Add linting to ensure we can not add external dependencies *unless* they are accounted for in pyproject.toml
- Update tests to handle new structure.
  - Tests organization mirrors physicsnemo.
- Update examples with changed imports.  Biggest breaking change is the deprecation of `launch`.  Logging utilities moved into `physicsnemo.utils.logging`.
- Main build and install method is now `uv`.

Overall, this should lower the installation challenges and lay the groundwork for the rest of v2.0.

## Checklist

- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.

## Review Process

**All PRs are reviewed by the PhysicsNeMo team before merging.**

Depending on which files are changed, GitHub may automatically assign a maintainer for review.

We are also testing AI-based code review tools (e.g., Greptile), which may add automated comments with a confidence score.
This score reflects the AI’s assessment of merge readiness and is **not** a qualitative judgment of your work, nor is
it an indication that the PR will be accepted / rejected.

AI-generated feedback should be reviewed critically for usefulness.
You are not required to respond to every AI comment, but they are intended to help both authors and reviewers.
Please react to Greptile comments with 👍 or 👎 to provide feedback on their accuracy.

PR #1241, opened 2025-11-20
Adds `CombinedOptimizer`

peterdsharpe:
# PhysicsNeMo Pull Request

Adds a new `CombinedOptimizer` utility, which is useful for the increasingly-popular "architecture-aware optimizers", such as [Muon](github.com/KellerJordan/Muon).

This PR targets the v2.0 refactor branch, so this should only be merged after #1235 .

## Checklist

## Review Process

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -50,6 +50,8 @@ physicsnemo/active_learning @laserkelvin @dallasfoster
 physicsnemo/launch
 physicsnemo/deploy
 
+# Mesh
+physicsnemo/mesh @peterdsharpe
 
 # DATAPIPES
 physicsnemo/datapipes/climate @dallasfoster @nickgeneva @pzharrington
@@ -145,3 +147,6 @@ test/metrics/diffusion @CharlelieLrt
 test/utils/generative @CharlelieLrt
 test/utils/corrdiff @CharlelieLrt
 test/active_learning @laserkelvin @dallasfoster
+
+# Mesh
+test/mesh @peterdsharpe
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -11,6 +11,7 @@ jobs:
     permissions:
       issues: write
       pull-requests: write
+      actions: write
     steps:
       - uses: actions/stale@v10
         with:
--- a/examples/cfd/external_aerodynamics/aero_graph_net/README.md
+++ b/examples/cfd/external_aerodynamics/aero_graph_net/README.md
@@ -126,6 +126,39 @@ of the samples from the test dataset.](../../../../docs/img/drivaernet_results.p
 
 ## Model training
 
+### Prerequisites
+
+This example also requires the `pyvista`, `shapely` and `vtk` libraries. Install with
+
+```bash
+pip install pyvista shapely vtk
+```
+
+BSMS MGN model requires additional dependency:
+
+```bash
+pip install sparse_dot_mkl
+```
+
+Additionally, if you are using the [PhysicsNeMo Docker Container](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/physicsnemo/containers/physicsnemo),
+install a few system-level packages with
+
+```bash
+apt install libosmesa6 libosmesa6-dev
+apt install mesa-utils libgl1-mesa-dev
+```
+
+> [!NOTE]
+> If you are running this example using the
+> [PhysicsNeMo Docker Container](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/physicsnemo/containers/physicsnemo)
+> you may have to pass a few additional flags to your `docker run` command.
+> Specifically, include `--ulimit nofile=65535:65535` and `--shm-size=4g` as
+> additional flags while launching the container. These flags are required
+> because a process pool is used to parallelize dataset creation prior to
+> starting training, which requires additional resources inside the container.
+
+### Running the experiments
+
 The example uses [Hydra](https://hydra.cc/docs/intro/) for experiment configuration.
 Hydra provides a convenient way to change almost any experiment parameter,
 such as dataset configuration, model and optimizer settings and so on.
@@ -142,18 +175,6 @@ environment variable:
 
 ```bash
 HYDRA_FULL_ERROR=1 python train.py ...
-```
-
-This example also requires the `pyvista`, `shapely` and `vtk` libraries. Install with
-
-```bash
-pip install pyvista shapely vtk
-```
-
-BSMS MGN model requires additional dependency:
-
-```bash
-pip install sparse_dot_mkl
 ```
 
 ### Ahmed Body training
--- a/physicsnemo/mesh/README.md
+++ b/physicsnemo/mesh/README.md
@@ -1,0 +1,526 @@
+# PhysicsNeMo-Mesh
+
+**GPU-Accelerated Mesh Processing for Physical Simulation and Scientific Visualization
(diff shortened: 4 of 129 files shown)


physicsnemo/optim/combined_optimizer.py
```diff
@@ -0,0 +1,282 @@
...
+        If ``optimizers`` is empty.
+
+    Notes
+    -----
+    * **Parameter Groups**: The ``param_groups`` attribute aggregates parameter
+      groups from all underlying optimizers, making this wrapper compatible with
+      learning rate schedulers.
+    * **Closure Behavior**: When ``step()`` is called with a closure, the closure
+      is evaluated once (not once per optimizer) and the resulting loss is passed
+      to all optimizers. This differs from calling each optimizer's step separately.
```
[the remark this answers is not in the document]
peterdsharpe: Docs fixed in [c657c5b](https://github.com/NVIDIA/physicsnemo/pull/1241/commits/c657c5b2f95e1557755cdf8b6d27a059f24dc33f)

test/optim/test_combined_optimizer.py
```diff
@@ -0,0 +1,282 @@
...
+            combined_optimizer.zero_grad()
+            loss = model(torch.randn(1, 10)).sum()
+            loss.backward()
+            return loss
+
+        combined_optimizer.step(closure)
+
+        # CombinedOptimizer has 2 optimizers (SGD, Adam), both support closure
+        # So closure should be called at least twice (once per optimizer)
+        assert call_count >= 2, f"Closure called {call_count} times, expected >= 2"
```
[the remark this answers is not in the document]
peterdsharpe: Docs fixed in [c657c5b](https://github.com/NVIDIA/physicsnemo/pull/1241/commits/c657c5b2f95e1557755cdf8b6d27a059f24dc33f)

physicsnemo/optim/combined_optimizer.py
```diff
@@ -0,0 +1,284 @@
...
+
+        ### Setup step functions (optionally compiled)
+        if torch_compile_kwargs is None:
+            self.step_fns: list[Callable] = [opt.step for opt in optimizers]
+        else:
+            self.step_fns: list[Callable] = [
+                torch.compile(opt.step, **torch_compile_kwargs) for opt in optimizers
+            ]
+
+    def zero_grad(self, *args, **kwargs) -> None:
```
coreyjadams: Upstream interface accepts only `set_to_none=True` here, and not other parameters.  I know if we passed it, that would succeed, but unless there are optimizers accepting other vales for `zero_grad` I think we should stick to `set_to_none`.

physicsnemo/optim/combined_optimizer.py
```diff
@@ -0,0 +1,284 @@
...
+    ...     loss = model(torch.randn(32, 10)).sum()
+    ...     loss.backward()
+    ...     combined_opt.step()
+    ...     scheduler.step()
+    """
+
+    def __init__(
+        self,
+        optimizers: Sequence[Optimizer],
+        torch_compile_kwargs: dict[str, Any] | None = None,
```
coreyjadams, line 105: This syntax is not upstream, right?  I assume we can't put `step` into a compile wrapper since that will break the closure behavior?

physicsnemo/optim/combined_optimizer.py
```diff
@@ -0,0 +1,284 @@
...
+        **kwargs
+            Keyword arguments to pass to each optimizer's ``zero_grad()``.
+            Common kwargs include ``set_to_none`` (bool).
+        """
+        for opt in self.optimizers:
+            opt.zero_grad(*args, **kwargs)
+
+    def step(
+        self, closure: Callable[[], torch.Tensor] | None = None
+    ) -> torch.Tensor | None:
```
coreyjadams: Upstream returns float | None, not tensor, FYI, in both the closure and `step`:

https://docs.pytorch.org/docs/stable/generated/torch.optim.Optimizer.step.html#torch.optim.Optimizer.step

physicsnemo/optim/combined_optimizer.py
```diff
@@ -0,0 +1,284 @@
...
+            The loss value returned by the last optimizer, or None if no closure was provided.
+        """
+        loss = None
+        for step_fn in self.step_fns:
+            if closure is None:
+                step_fn()
+            else:
+                res = step_fn(closure)
+                if res is not None:
+                    loss = res
```
coreyjadams, line 201: I don't follow this logic - if there are multiple optimizers and a closure, are they expected to return the same value?  Right now we overwrite with the value of the last non-None `res`.

test/optim/test_combined_optimizer.py
```diff
@@ -0,0 +1,282 @@
...
+    opt2 = Adam(model.layer2.parameters(), lr=0.001)
+    return [opt1, opt2]
+
+
+@pytest.fixture
+def combined_optimizer(optimizers):
+    return CombinedOptimizer(optimizers)
+
+
+class TestInitialization:
```
coreyjadams, line 52: All of our other tests are more or less "functionals" and not class-based tests.  I don't personally have any objection to this style, but I don't know what unforeseen consequences we might see with this.  Any thoughts?

test/optim/test_combined_optimizer.py
```diff
@@ -0,0 +1,282 @@
...
+        new_combined = CombinedOptimizer([opt1, opt2])
+
+        # Load state
+        new_combined.load_state_dict(state)
+
+        # Verify equality (basic check)
+        # Note: strict equality of state dicts might fail due to weak refs or unrelated keys,
+        # so we check structure matches.
+        new_state = new_combined.state_dict()
+        assert len(new_state["optimizers"]) == len(state["optimizers"])
```
coreyjadams, line 274: Can we update this test to check save / restore is getting numerics correct too?  Loading the wrong optimizer state or getting the numbers incorrect would be a pretty frustrating bug in training codes.

physicsnemo/optim/combined_optimizer.py
peterdsharpe: I think the upstream signature is wrong in most cases -- the return type of the model will be a loss represented as a singleton `torch.Tensor`, right? In fact, if you do `loss.item()` to convert `tensor` -> `float`, my hunch is that this would actually break since you can't call `.backward()` on a float (it's detached from the computational graph). I'll test this and do some more research to verify, though.

test/optim/test_combined_optimizer.py
peterdsharpe: This just adds grouping for more helpful printouts / test failure reporting - no functional change here.

physicsnemo/optim/combined_optimizer.py
peterdsharpe: Fixed 7ddf8ab4df5b192a8ea694d45529bd9b5549f918

test/optim/test_combined_optimizer.py
peterdsharpe: fixed

physicsnemo/optim/combined_optimizer.py
peterdsharpe: fixed by adding clarification in docstring; last one takes precedence

physicsnemo/optim/combined_optimizer.py
peterdsharpe: Deliberate design choice to support compilation through the optimizer here. `torch.compile`-ing the step does not break closure behavior.

PR #1264, opened 2025-12-03
UV <---> Pip must stay in sync.

coreyjadams:
This PR is about dependencies and installation.  Goals:
- Ensure uv is first-class
- Ensure pip install works
- Ensure default deps are broadly available and won't hold up other pieces.

To this end, the core deps are listed, and there is linting against them.  We can't import non-core deps "bare" any more.

Further, optional deps are almost* entirely protected with `check_version_spec`, etc., to raise runtime errors at first use only.  Meaning, a missing dep will not break the rest of the code, and won't even crash until a user attempts to use it.  And I've tried to list out the optional deps per package, too, though some could use versioning.

**Everything below is outdated.  I left it to show the original design.**

Ensure that the pip dependency group exclusively matches the resolution of the dependency-groups from Pep 735 that uv enables.

Requirements are to ensure `pip install .` still works, with or without uv, even though we *want* uv to be the sole source of truth.  Rather than maintaining two dep lists, especially since one is nested, I made a pre-commit step to ensure that the pip requirements is always exactly what dependency-groups specifies.

Basically, the workflow is this:
- pre-commit will run `test/ci_tests/sync_deps.py`
- sync_deps.py ingests the pyproject.yaml file.
  - It recurses through the `physicsnemo` dependency group (no extras) and resolves all the groups and dependencies.
  - It generates a flat list of dependencies.
  - De-duplication is automatic, and if the version strings don't agree it's an error.
- Then:
  -  If the list agrees, PASS
  - If the list disagrees, it will overwrite the dependency section of `pyproject.toml` and fail pre-commit.

This should achieve the following:
- if dependency-group dependencies update, we will be unable to commit that code to the repo unless the `pip install .` dependencies also update.
- If the pip dependencies update, that update will be destroyed automatically.

# PhysicsNeMo Pull Request

## Checklist

## Review Process

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,12 +55,14 @@ Changelog = "https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md"
 no-build-isolation-package = [
     "torch_scatter",
     "torch_cluster",
+    "earth2grid",
 ]
 managed = true
 
 # torch-sparse requires us to tell uv about it's dependency on torch.
 [tool.uv.extra-build-dependencies]
 torch-sparse = ["torch"]
+earth2grid = ["setuptools", "torch"] 
 
 # Earth2 grid is not on pypi ....
 [tool.uv.sources]
@@ -126,6 +128,7 @@ nn-extras = [
     "cuml",
     "transformer_engine[pytorch]",
     "scipy",
+    "natten",
     "nvidia-physicsnemo[utils-extras]",
 ]
 model-extras = [
@@ -140,6 +143,8 @@ datapipes-extras = [
     "tensorflow",
     "nvidia-physicsnemo[model-extras]",
 ]
+
+# Use case specific dependency groups.
 gnns = [
     "torch_geometric",
     "torch_scatter",


test/ci_tests/prevent_untracked_imports.py
```diff
@@ -19,14 +19,12 @@
...
-from typing import Dict, List, Set, Union
+from typing import Dict, List, Set
 
 import tomllib
 from importlinter import Contract, ContractCheck, fields, output
 from packaging.requirements import Requirement
 
-Dependency = Union[str, Dict[str, str]]
-
 # For irregular mappings that we don't want to have cause errors:
```
laserkelvin, line 45: Would be nice if you could add a docstring here to describe what the script does at a high level

pyproject.toml
```diff
@@ -71,129 +90,69 @@ exclude = [
...
-    {include-group="datapipes"},
-    {include-group="models-extras"},
-]
-domain_parallel = [
-    {include-group = "nn"}
-]
-physicsnemo = [
-    {include-group = "domain-parallel"},
-    {include-group = "datapipes"},
+    "nvidia-physicsnemo[model-extras]",
```
laserkelvin, line 144: If possible, I would prefer to use `include-groups` instead - I think `pip` should support it too, and just from a glance it feels more sematically distinct

test/ci_tests/prevent_untracked_imports.py
```diff
@@ -59,31 +58,46 @@ def check(self, graph, verbose):
...
+            if not module.startswith(self.container)
         )
 
         upstream_external_modules = remove_standard_library(upstream_modules)
 
         # Now, read the tree from pyproject.toml:
-        dependency_tree = resolve_dependency_group_no_versions(
-            Path("pyproject.toml"), self.dependency_group
-        )
+        dependency_tree = resolve_core_dependencies(Path("pyproject.toml"))
```
laserkelvin, line 84: Dumb question, but does this pathing work regardless of where you are in the folder structure? You don't have to do a relative path?

pyproject.toml
coreyjadams: I don't think this is possible, no.  indluce=groups is specific to dependency-groups, which I pivoted away from since that appears to be development dependency focused.  Shame, though, it's nice interface...

These self-referential packages seem to be the only way to support this.

test/ci_tests/prevent_untracked_imports.py
coreyjadams: Done!

test/ci_tests/prevent_untracked_imports.py
coreyjadams: You can only "lint-imports" from the root dir.  Pre-commit however will run successfully from anywhere in the repo, just like usual.

pyproject.toml
laserkelvin: I am disappoint

pyproject.toml
```diff
@@ -71,129 +90,73 @@ exclude = [
...
-]
-models = [
-    "cftime",
-    "hydra-core",
-    "omegaconf",
-    "xarray>=2023.1.0",
-    "zarr>=2.14.2",
-    "jaxtyping",
-    {include-group="nn"},
+    "nvidia-physicsnemo[utils-extras]",
```
pzharrington, line 132: No `natten`? 😅

coreyjadams: We should add it!  But unless it's easy to install cross-platform, it needs to go into an optional dep.

pyproject.toml
```diff
@@ -71,129 +90,73 @@ exclude = [
...
     "torch_sparse",
     "torch_cluster",
-    "pylibcugraphops",
     "nvfuser",
+    "nvidia-physicsnemo[model-extras]",
 ]
-
 healpix = [
     "earth2grid",
 ]
```
pzharrington, line 159: It seems the `gnns` and `healpix` here are more just general deps groups associated with certain functionality/use-cases but not a part of the base pnm subpackages (`nn`, `models`, `datapipes` ...). Maybe separate them out with a small comment header for clarity? Like
```
# Use-case Specific Dependency Groups
```
or something...

pyproject.toml
```diff
@@ -43,17 +47,32 @@ Changelog = "https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md"
...
 managed = true
-default-groups = ["physicsnemo"]
+
+# torch-sparse requires us to tell uv about it's dependency on torch.
+[tool.uv.extra-build-dependencies]
+torch-sparse = ["torch"]
+
+# Earth2 grid is not on pypi ....
+[tool.uv.sources]
+earth2grid = { url = "https://github.com/NVlabs/earth2grid/archive/main.tar.gz" }
```
pzharrington, line 69: May want to look at the earth2studio `toml` to see the various things Nick added to handle earth2grid...I think there is also a need to include `--no-build-isolation`

coreyjadams: Relevant lines appear to be:
```
no-build-isolation-package = ["earth2grid", "flash-attn", "torch-harmonics"]
```

```
[tool.uv.extra-build-dependencies]
earth2grid = ["setuptools", "torch"]
```

```

[tool.uv.sources]
...
earth2grid = { git = "https://github.com/NVlabs/earth2grid.git", rev = "661445e2c68edc76f52632aa0528af482357f1b8" }
```

I can add the missing pieces, though I suspect we don't need to lock to a revision.

pzharrington: Yeah I need to double check with Nick when he's back why he pinned version, it may have just been out of "abundance of caution" :)

pyproject.toml
pzharrington: `nn-extras` here is optional right? Natten should def be optional regardless, but I think `nn-extras` is the right place for it

PR #1265, opened 2025-12-04
Fea nn refactor

loliverhennigh:
# PhysicsNeMo Pull Request

## Description
Refactor physicsnemo `nn` to more closely align with pytorch.


physicsnemo/core/function.py
```diff
@@ -0,0 +1,26 @@
...
+from __future__ import annotations
+
+import torch
+from torch.autograd import Function as TorchAutogradFunction
+
+from physicsnemo.core.benchmark import BenchmarkMixin
+
+class Function(TorchAutogradFunction, BenchmarkMixin):
+    """Base class for PhysicsNeMo custom autograd functions."""
+    # Placeholder for utilities to bring in warp, fuser, etc.
```
coreyjadams, line 26: This locks us in to the inherit-from-TorchAutogradFunction interface and away from the custom op solution that pytorch seems to prefer here:
https://docs.pytorch.org/tutorials/advanced/python_custom_ops.html

I favor the custom op decorators for several reasons:
- better integration with torch.compile and avoiding compile breaks
- Registration with pytorch dispatch tools, used to enable ShardTensor's distributed dispatch
- built-in testing utilities from pytorch to validate compile support, etc.

physicsnemo/core/module.py
```diff
@@ -67,7 +68,7 @@ def _load_state_dict_with_logging(
     return missing_keys, unexpected_keys
 
 
-class Module(torch.nn.Module):
+class Module(torch.nn.Module, BenchmarkMixin):
```
coreyjadams, line 71: This will automatically expose our models to this BenchmarkMixin functionality.  Not necessarily a _bad_ thing, though I wonder if we'll hit unforeseen consequences in save / load / restore / broken checkpoints, etc.

PR #1267, opened 2025-12-05
PhysicsNeMo-Mesh: adds minimal mesh utilities + manipulation

peterdsharpe:
# PhysicsNeMo Pull Request

## Checklist

## Review Process

### DIFF
--- a/physicsnemo/mesh/utilities/_scatter_ops.py
+++ b/physicsnemo/mesh/utilities/_scatter_ops.py
@@ -59,9 +59,8 @@ def scatter_aggregate(
         If None, uses uniform weights of 1.0.
     aggregation : str
         Aggregation mode:
-        - "mean": Weighted mean
+        - "mean": Weighted mean (uses weights if provided, uniform otherwise)
         - "sum": Weighted sum (no normalization)
-        - "weighted_mean": Same as "mean" (for clarity)
 
     Returns
     -------
@@ -116,7 +115,7 @@ def scatter_aggregate(
     )
 
     ### Handle aggregation mode
-    if aggregation in ("mean", "weighted_mean"):
+    if aggregation == "mean":
         ### Compute sum of weights at each destination
         weight_sums = torch.zeros(n_dst, dtype=dtype, device=device)
         weight_sums.scatter_add_(
@@ -136,8 +135,6 @@ def scatter_aggregate(
         pass
 
     else:
-        raise ValueError(
-            f"Invalid {aggregation=}. Must be 'mean', 'weighted_mean', or 'sum'."
-        )
+        raise ValueError(f"Invalid {aggregation=}. Must be 'mean' or 'sum'.")
 
     return aggregated_data
--- a/test/mesh/mesh/test_init_validation.py
+++ b/test/mesh/mesh/test_init_validation.py
@@ -186,9 +186,13 @@ class TestDimensionValidation:
     def test_manifold_greater_than_spatial_raises(self):
         """Test that manifold_dims > spatial_dims raises ValueError."""
         points = torch.randn(10, 2)  # 2D spatial
-        cells = torch.randint(0, 10, (5, 4))  # 3-manifold (would need at least 3D space)
+        cells = torch.randint(
+            0, 10, (5, 4)
+        )  # 3-manifold (would need at least 3D space)
 
-        with pytest.raises(ValueError, match=r"`n_manifold_dims` must be <= `n_spatial_dims`"):
+        with pytest.raises(
+            ValueError, match=r"`n_manifold_dims` must be <= `n_spatial_dims`"
+        ):
             Mesh(points=points, cells=cells)
 
     def test_0_manifold_in_0_spatial_invalid(self):
@@ -327,7 +331,9 @@ class TestParametrized:
             (1, 1, 3, 0),
         ],
     )
-    def test_valid_combinations(self, n_points, n_cells, n_spatial_dims, n_manifold_dims):
+    def test_valid_combinations(
+        self, n_points, n_cells, n_spatial_dims, n_manifold_dims
+    ):
         """Test various valid dimension combinations."""
         torch.manual_seed(42)
         points = torch.randn(n_points, n_spatial_dims)
@@ -339,4 +345,3 @@ class TestParametrized:
         assert mesh.n_cells == n_cells
         assert mesh.n_spatial_dims == n_spatial_dims
         assert mesh.n_manifold_dims == n_manifold_dims
-
--- a/test/mesh/mesh/test_padding.py
+++ b/test/mesh/mesh/test_padding.py
@@ -32,7 +32,6 @@ import torch
 
 from physicsnemo.mesh import Mesh
 
-
 ### Helper Functions ###
 
 
@@ -205,7 +204,9 @@ class TestPadWithData:
 
         padded = mesh.pad(target_n_cells=15, data_padding_value=-999.0)
 
-        assert torch.allclose(padded.cell_data["pressure"][5:], torch.full((10,), -999.0))
+        assert torch.allclose(
+            padded.cell_data["pressure"][5:], torch.full((10,), -999.0)
+        )
 
     def test_pad_preserves_original_data(self):
         """Test that original data values are preserved."""
@@ -284,7 +285,9 @@ class TestPadToNextPowerBasic:
 
     def test_pad_to_next_power_with_data(self):
         """Test pad_to_next_power with data fields."""
-        mesh = create_simple_mesh(n_points=100, n_cells=50, add_point_data=True, add_cell_data=True)
+        mesh = create_simple_mesh(
+            n_points=100, n_cells=50, add_point_data=True, add_cell_data=True
+        )
 
         padded = mesh.pad_to_next_power(power=2.0)
 
@@ -298,7 +301,10 @@ class TestPadToNextPowerBasic:
         padded = mesh.pad_to_next_power(power=2.0, data_padding_value=-1.0)
 
         # Padded values should be -1.0
-        assert torch.allclose(padded.point_data["temperature"][100:], torch.full((padded.n_points - 100,), -1.0))
+        assert torch.allclose(
+            padded.point_data["temperature"][100:],
+            torch.full((padded.n_points - 100,), -1.0),
+        )
 
 
 class TestPadToNextPowerErrors:
@@ -368,7 +374,11 @@ class TestPadParametrized:
     def test_pad_various_dimensions(self, n_spatial_dims, n_manifold_dims, device):
         """Test padding across various dimension combinations."""
         mesh = create_simple_mesh(
-            n_points=10, n_cells=5, n_spatial_dims=n_spatial_dims, n_manifold_dims=n_manifold_dims, device=device
+            n_points=10,
+            n_cells=5,
+            n_spatial_dims=n_spatial_dims,
+            n_manifold_dims=n_manifold_dims,
+            device=device,
         )
 
         padded = mesh.pad(target_n_points=20, target_n_cells=15)
@@ -399,7 +409,7 @@ class TestPadParametrized:
             # Find n such that floor(base^n) = result
             log_val = math.log(n) / math.log(base)
             n_ceil = math.ceil(log_val)
-            expected = int(base ** n_ceil)
+            expected = int(base**n_ceil)
             # Allow small tolerance for floating point computation
             return n == expected
 
@@ -443,4 +453,3 @@ class TestPadEdgeCases:
         padded = mesh.pad(target_n_points=10)
 
         assert torch.allclose(padded.global_data["time"], torch.tensor(1.5))
-
--- a/test/mesh/mesh/test_point_normals_weighting.py
+++ b/test/mesh/mesh/test_point_normals_weighting.py
@@ -33,7 +33,6 @@ import torch
 
 from physicsnemo.mesh import Mesh
 
-
 ### Helper Functions ###
 
 
@@ -173,7 +172,12 @@ class TestPointNormalsWeightingSchemes:
 
         # For flat surface, all normals should point in +z direction
         expected = torch.tensor([0.0, 0.0, 1.0])
-        for normals in [normals_unweighted, normals_area, normals_angle, normals_angle_area]:
+        for normals in [
+            normals_unweighted,
+            normals_area,
+            normals_angle,
+            normals_angle_area,
+        ]:
             for i in range(mesh.n_points):
                 # Allow for sign flip
                 assert torch.allclose(normals[i].abs(), expected.abs(), atol=1e-5)
@@ -247,14 +251,18 @@ class TestPointNormalsErrors:
 
         assert mesh.n_manifold_dims == 1
 
-        with pytest.raises(ValueError, match="Angle-based weighting requires n_manifold_dims >= 2"):
+        with pytest.raises(
+            ValueError, match="Angle-based weighting requires n_manifold_dims >= 2"
+        ):
             mesh.compute_point_normals(weighting="angle")
 
     def test_angle_area_weighting_for_edges_raises(self):
         """Test that angle_area weighting for edges raises ValueError."""
         mesh = create_edge_mesh_2d()
 
-        with pytest.raises(ValueError, match="Angle-based weighting requires n_manifold_dims >= 2"):
+        with pytest.raises(
+            ValueError, match="Angle-based weighting requires n_manifold_dims >= 2"
+        ):
             mesh.compute_point_normals(weighting="angle_area")
 
     def test_area_weighting_for_edges_works(self):
@@ -295,9 +303,7 @@ class TestEdgeNormals2D:
     def test_corner_edge_normal(self):
         """Test normal at corner where two edges meet at 90 degrees."""
         # L-shaped path
-        points = torch.tensor(
-            [[0.0, 0.0], [1.0, 0.0], [1.0, 1.0]], dtype=torch.float32
-        )
+        points = torch.tensor([[0.0, 0.0], [1.0, 0.0], [1.0, 1.0]], dtype=torch.float32)
         cells = torch.tensor([[0, 1], [1, 2]], dtype=torch.long)
         mesh = Mesh(points=points, cells=cells)
 
@@ -405,4 +411,3 @@ class TestCaching:
         # Second access uses cache
         normals2 = mesh.point_normals
         assert torch.equal(normals1, normals2)
-
--- a/test/mesh/utilities/test_cache.py
+++ b/test/mesh/utilities/test_cache.py
@@ -236,7 +236,9 @@ class TestCacheDevices:
 class TestCacheDtypes:
     """Tests for dtype handling in cache operations."""
 
-    @pytest.mark.parametrize("dtype", [torch.float32, torch.float64, torch.int64, torch.int32])
+    @pytest.mark.parametrize(
+        "dtype", [torch.float32, torch.float64, torch.int64, torch.int32]
+    )
     def test_cache_various_dtypes(self, dtype):
         """Test caching with various dtypes."""
         data = TensorDict({}, batch_size=[10])
@@ -300,4 +302,3 @@ class TestCacheIntegrationWithMesh:
         assert get_cached(cell_data, "centroids") is not None
         assert get_cached(cell_data, "areas") is not None
         assert get_cached(cell_data, "normals") is not None
-
--- a/test/mesh/utilities/test_scatter_ops.py
+++ b/test/mesh/utilities/test_scatter_ops.py
@@ -84,31 +84,31 @@ class TestScatterAggregateMean:
         assert torch.allclose(result, expected)
 
 
-class TestScatterAggregateWeightedMean:
-    """Tests for weighted_mean aggregation mode."""
+class TestScatterAggregateMeanWithWeights:
+    """Tests for mean aggregation with explicit weights."""
 
-    def test_weighted_mean_simple(self):
+    def test_mean_with_weights_simple(self):
         """Test simple weighted mean aggregation."""
         src_data = torch.tensor([1.0, 3.0])
         src_to_dst = torch.tensor([0, 0])
         weights = torch.tensor([1.0, 3.0])
 
         result = scatter_aggregate(
-            src_data, src_to_dst, n_dst=1, weights=weights, aggregation="weighted_mean"
+            src_data, src_to_dst, n_dst=1, weights=weights, aggregation="mean"
         )
 
         # Weighted mean: (1*1 + 3*3) / (1 + 3) = 10/4 = 2.5
         expected = torch.tensor([2.5])
         assert torch.allclose(result, expected)
 
-    def test_weighted_mean_equal_weights_equals_mean(self):
+    def test_mean_with_equal_weights_equals_unweighted(self):
         """Test that equal weights produce same result as unweighted mean."""
         src_data = torch.tensor([1.0, 2.0, 3.0])
         src_to_dst = torch.tensor([0, 0, 0])
         weights = torch.tensor([1.0, 1.0, 1.0])
 
         result_weighted = scatter_aggregate(
-            src_data, src_to_dst, n_dst=1, weights=weights, aggregation="weighted_mean"
+            src_data, src_to_dst, n_dst=1, weights=weights, aggregation="mean"
         )
         result_unweighted = scatter_aggregate(
             src_data, src_to_dst, n_dst=1, aggregation="mean"
@@ -116,34 +116,19 @@ class TestScatterAggregateWeightedMean:
 
         assert torch.allclose(result_weighted, result_unweighted)
 
-    def test_weighted_mean_zero_weight_ignored(self):
+    def test_mean_with_zero_weight_ignored(self):
         """Test that zero-weighted sources are effectively ignored."""
         src_data = torch.tensor([1.0, 2.0, 100.0])  # 100 should be ignored
         src_to_dst = torch.tensor([0, 0, 0])
         weights = torch.tensor([1.0, 1.0, 0.0])  # Zero weight for 100
 
         result = scatter_aggregate(
-            src_data, src_to_dst, n_dst=1, weights=weights, aggregation="weighted_mean"
+            src_data, src_to_dst, n_dst=1, weights=weights, aggregation="mean"
         )
 
         # Weighted mean: (1*1 + 2*1 + 100*0) / (1 + 1 + 0) = 3/2 = 1.5
         expected = torch.tensor([1.5])
         assert torch.allclose(result, expected)
-
-    def test_weighted_mean_mode_alias(self):
-        """Test that 'weighted_mean' and 'mean' behave same with weights."""
-        src_data = torch.tensor([1.0, 3.0])
-        src_to_dst = torch.tensor([0, 0])
-        weights = torch.tensor([1.0, 2.0])
-
-        result_mean = scatter_aggregate(
-            src_data, src_to_dst, n_dst=1, weights=weights, aggregation="mean"
-        )
-        result_weighted = scatter_aggregate(
-            src_data, src_to_dst, n_dst=1, weights=weights, aggregation="weighted_mean"
-        )
-
-        assert torch.allclose(result_mean, result_weighted)
 
 
 class TestScatterAggregateSum:
@@ -206,10 +191,12 @@ class TestScatterAggregateMultiDimensional:
 
     def test_mean_3d_data(self):
         """Test mean aggregation with 3D source data (e.g., tensors)."""
-        src_data = torch.tensor([
-            [[1.0, 2.0], [3.0, 4.0]],
-            [[5.0, 6.0], [7.0, 8.0]],
-        ])
+        src_data = torch.tensor(
+            [
+                [[1.0, 2.0], [3.0, 4.0]],
+                [[5.0, 6.0], [7.0, 8.0]],
+            ]
+        )
         src_to_dst = torch.tensor([0, 0])
 
         result = scatter_aggregate(src_data, src_to_dst, n_dst=1, aggregation="mean")
@@ -307,13 +294,15 @@ class TestScatterAggregateDevices:
 class TestScatterAggregateParametrized:
     """Parametrized tests for scatter_aggregate."""
 
-    @pytest.mark.parametrize("aggregation", ["mean", "weighted_mean", "sum"])
+    @pytest.mark.parametrize("aggregation", ["mean", "sum"])
     def test_all_aggregation_modes(self, aggregation):
         """Test that all aggregation modes work."""
         src_data = torch.tensor([1.0, 2.0, 3.0])
         src_to_dst = torch.tensor([0, 0, 1])
 
-        result = scatter_aggregate(src_data, src_to_dst, n_dst=2, aggregation=aggregation)
+        result = scatter_aggregate(
+            src_data, src_to_dst, n_dst=2, aggregation=aggregation
+        )
 
         assert result.shape == (2,)
 
@@ -323,7 +312,9 @@ class TestScatterAggregateParametrized:
         src_data = torch.randn(20)
         src_to_dst = torch.randint(0, n_dst, (20,))
 
-        result = scatter_aggregate(src_data, src_to_dst, n_dst=n_dst, aggregation="mean")
+        result = scatter_aggregate(
+            src_data, src_to_dst, n_dst=n_dst, aggregation="mean"
+        )
 
         assert result.shape == (n_dst,)
 
@@ -346,4 +337,3 @@ class TestScatterAggregateParametrized:
 
         expected_shape = (3,) + data_shape[1:]
         assert result.shape == expected_shape
-


physicsnemo/mesh/README.md
```diff
@@ -0,0 +1,521 @@
...
+  - Both [DEC](https://en.wikipedia.org/wiki/Discrete_exterior_calculus) (Discrete
+    Exterior Calculus) and LSQ (Least-Squares) methods
+  - Intrinsic (tangent space) and extrinsic (ambient space) derivatives
+- **Differential Geometry**: [Gaussian curvature](https://en.wikipedia.org/wiki/Gaussian_curvature),
+  [mean curvature](https://en.wikipedia.org/wiki/Mean_curvature),
+  [normals](https://en.wikipedia.org/wiki/Normal_(geometry)),
+  [tangent spaces](https://en.wikipedia.org/wiki/Tangent_space)
+- **Curvature Analysis**: [Angle defect](https://en.wikipedia.org/wiki/Angular_defect)
+  (intrinsic) and [cotangent Laplacian](https://en.wikipedia.org/wiki/Discrete_Laplace_operator)
+  (extrinsic) methods
```
coreyjadams, line 83: Is this all in this PR?

physicsnemo/mesh/README.md
```diff
@@ -0,0 +1,521 @@
+# PhysicsNeMo-Mesh
```
coreyjadams, line 1: This README.md needs to migrate to the physicsnemo docs.

physicsnemo/mesh/README.md
```diff
@@ -0,0 +1,521 @@
...
+| Mesh statistics | ✅ | |
+| **I/O** | | |
+| PyVista integration | ✅ | All PyVista-supported formats |
+| **Visualization** | | |
+| Matplotlib backend | ✅ | 2D/3D plotting |
+| PyVista backend | ✅ | Interactive 3D |
+| Scalar colormapping on points or cells | ✅ | Auto L2-norm for vectors |
+<!-- markdownlint-enable MD013 -->
+
+**Legend:** ✅ Complete | ❌ Not Implemented, but planned
```
coreyjadams, line 350: This is a big matrix of features.  This is all going into several PRs, right?  Perhaps this could be a good docs page on it's own.

physicsnemo/mesh/README.md
```diff
@@ -0,0 +1,521 @@
...
+```
+
+### Spatial Queries
+
+```python
+from physicsnemo.mesh.spatial import BVH
+
+bvh = BVH.from_mesh(mesh)
+cell_candidates = bvh.find_candidate_cells(torch.rand(1000, 3))
+sampled = mesh.sample_data_at_points(query_points, data_source="points")
```
coreyjadams, line 413: Pros and cons of this over warp in terms of speed, perf, broader usability?

physicsnemo/mesh/primitives/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+    - procedural: Procedurally generated mesh variations
+    - pyvista_datasets: Wrappers for PyVista example datasets
+
+Usage:
+    >>> from physicsnemo.mesh import examples
+    >>> mesh = examples.surfaces.sphere_icosahedral.load(radius=1.0, subdivisions=2)
+    >>> mesh = examples.pyvista_datasets.bunny.load()
+"""
+
+from physicsnemo.mesh.primitives import (
```
NickGeneva, line 45: Not trying to steer towards OOP too much but thinking about defining some simple interface / protocol for a primitive object that has this load function as a part of it and then make the primative objects a class with an option init that exposes some controllable params.


This would then allow you to do something like instantiate a circle but provide simple controls for like the radius or center.

Like:

```python
class Circle:
   radius: float = 1
  def __init__(self, radius: int = 1):
      self. radius = radius

  def load(self, device: str):
      #... do something with self.radius ...
      return mesh


# Could add some static load here:
def load(str):
   return Circle().load(device)
```

The main thing would be to have some formal definition of what the primative function needs to expose via the interface (not parent class) and also would make the load functions all the same, customization is done by means of instantiation which users could then keep to create different clones of a primitive.

The load function technically does everything the same, but I guess the potential for changing function signatures is just not super ideal maybe :shrug:

physicsnemo/mesh/README.md
```diff
@@ -0,0 +1,521 @@
...
+Note that these use an efficient sparse (`indices`, `offsets`) encoding of the adjacency
+relationships, which is used internally for all computations. (See the dedicated
+[`physicsnemo.mesh.neighbors._adjacency.py`](physicsnemo/mesh/neighbors/_adjacency.py)
+module.) You can convert these to a typical ragged list-of-lists representation with
+`.to_list()`, which is useful for debugging or interoperability, at the cost of
+performance:
+
+```python
+point_neighbors = mesh.get_point_to_points_adjacency().to_list()
+cell_neighbors = mesh.get_cell_to_cells_adjacency().to_list()
```
coreyjadams, line 432: Careful to distinguish what is this vs physicsnemo.nn.neighbors?

physicsnemo/mesh/primitives/__init__.py
NickGeneva: Just for some discussion or thinking

physicsnemo/mesh/README.md
```diff
@@ -0,0 +1,521 @@
...
+with five components:
+
+```python
+Mesh(
+    points: torch.Tensor,      # (n_points, n_spatial_dims)
+    cells: torch.Tensor,       # (n_cells, n_manifold_dims + 1), integer dtype
+    point_data: TensorDict,    # Per-vertex data
+    cell_data: TensorDict,     # Per-cell data
+    global_data: TensorDict,   # Mesh-level data
+)
```
coreyjadams, line 449: If I recall, the cell/point relationship is intrinsic to the structure.  But the components you've listed here are _nearly_ the pieces of a graph description, just that you have quite tightly constrained the connectivity.

What is to stop us from actually generalizing this into an arbritrary graph  node /edge framework?

physicsnemo/mesh/io/io_pyvista.py
```diff
@@ -0,0 +1,306 @@
+from typing import Literal
+
+import numpy as np
+import pyvista as pv
+import torch
+from pyvista import CellType
```
coreyjadams: We need to guard pyvista imports in v2.0.

physicsnemo/mesh/utilities/_padding.py
```diff
@@ -0,0 +1,25 @@
...
+            tensor,
+            torch.full(
+                (size - tensor.shape[0], *tensor.shape[1:]),
+                fill_value=value,
+                dtype=tensor.dtype,
+                device=tensor.device,
+            ),
+        ],
+        dim=0,
+    )
```
coreyjadams: Pytorch has padding functionals build in, why are we redoing them here?

physicsnemo/mesh/utilities/_scatter_ops.py
```diff
@@ -0,0 +1,118 @@
...
+    elif aggregation == "sum":
+        # Already computed weighted sum, no normalization needed
+        pass
+
+    else:
+        raise ValueError(
+            f"Invalid {aggregation=}. Must be 'mean', 'weighted_mean', or 'sum'."
+        )
+
+    return aggregated_data
```
coreyjadams, line 140: Couple thoughts here.  Please consider them for discussion and less about needed changes...

- You have scatter_aggregate, with several kernel launches, though it might be interesting to consider writing this into one or two warp kernels for better performance.  Warp is a core, mandatory dep of physicsnemo, so it adds nothing to our dependency list.
- We have some deps on torch_scatter already, I can't remember exactly where, but a lot of it is just calling `scatter` or `gather`.  Optimized scatter / gather utilies built in to physicsnemo could be a nice value add.
- If we have the weights built in, that is probably easy to optimize with a kernel that doesn't need a second launch for the denominator.  Food for thought.
- How might this evolve if we come along in 6 months and ask for a mesh to be distributed across GPUs?  Would tensor->shardTensor work in functions like this?

physicsnemo/mesh/mesh.py
```diff
@@ -0,0 +1,1047 @@
...
+    Expensive geometric computations (centroids, areas, normals) are automatically
+    cached in the data dictionaries under keys prefixed with ``_cache``. The cache
+    is invalidated when creating new ``Mesh`` instances but persists across
+    repeated property accesses on the same instance.
+
+    **Pure Simplicial Complexes**
+
+    This class assumes a *pure* simplicial complex where all cells have the same
+    manifold dimension. Mixed-dimension meshes (e.g., triangles and tetrahedra
+    together) are not supported.
```
coreyjadams: For the sake of our users who aren't familiar with this math and, uh, definitely _not_ me, some documentation about what this means, and what it implies for using mesh with their data (and examples of data that would or wouldn't work) could be _really_ useful.

physicsnemo/mesh/mesh.py
```diff
@@ -0,0 +1,1047 @@
...
+    ) -> None:
+        ### Assign tensorclass fields
+        self.points = points
+        self.cells = cells
+
+        # For data fields, convert inputs to TensorDicts if needed
+        if not isinstance(point_data, TensorDict):
+            point_data = TensorDict(
+                {} if point_data is None else dict(point_data),
+                batch_size=torch.Size([self.n_points]),
```
coreyjadams, line 222: I assume batch_size here has a meaning to tensordict and not what we usually think?

physicsnemo/mesh/mesh.py
```diff
@@ -0,0 +1,1047 @@
...
+    def codimension(self) -> int:
+        """Compute the codimension of the mesh.
+
+        The codimension is the difference between the spatial dimension and the
+        manifold dimension: codimension = n_spatial_dims - n_manifold_dims.
+
+        Examples:
+            - Edges (1-simplices) in 2D: codimension = 2 - 1 = 1 (codimension-1)
+            - Triangles (2-simplices) in 3D: codimension = 3 - 2 = 1 (codimension-1)
+            - Edges in 3D: codimension = 3 - 1 = 2 (codimension-2)
```
coreyjadams, line 334: OK, edges in 3D space sure sounds like a GNN.  Did you accidentally write a new GNN framework?

physicsnemo/mesh/mesh.py
```diff
@@ -0,0 +1,1047 @@
...
+        if not overwrite_keys:
+            for key in self.cell_data.exclude("_cache").keys():
+                if key in self.point_data.keys():
+                    raise ValueError(
+                        f"Key {key!r} already exists in point_data. "
+                        f"Set overwrite_keys=True to overwrite."
+                    )
+
+        ### Convert each cell data field to point data
+        from physicsnemo.mesh.utilities import scatter_aggregate
```
coreyjadams: Not a huge fan of secret delayed imports :)

physicsnemo/mesh/mesh.py
```diff
@@ -0,0 +1,1047 @@
...
+        target_n_points = next_power_size(self.n_points, power)
+        target_n_cells = next_power_size(self.n_cells, power)
+
+        return self.pad(
+            target_n_points=target_n_points,
+            target_n_cells=target_n_cells,
+            data_padding_value=data_padding_value,
+        )
+
+    def draw(
```
coreyjadams, line 1279: I would prefer to see this functionality not as part of the core data object.

physicsnemo/mesh/mesh.py
```diff
@@ -0,0 +1,1047 @@
...
+
+        Convenience wrapper for physicsnemo.mesh.transformations.translate().
+
+        Args:
+            offset: Translation vector, shape (n_spatial_dims,)
+
+        Returns:
+            New Mesh with translated geometry
+        """
+        from physicsnemo.mesh.transformations import translate
```
coreyjadams: I know the utility you're going for here.  I think there could be a better implementation or way to register this mapping.  But either way let's avoid these one off imports in functions?

physicsnemo/mesh/primitives/README.md
```diff
@@ -0,0 +1,261 @@
+# PhysicsNeMo-Mesh Examples
+
+This module provides a comprehensive collection of canonical meshes for tutorials,
+testing, and experimentation. All meshes are generated at runtime and organized by
+category and dimensional configuration.
```
coreyjadams, line 5: I think you could move this entire thing to examples.  It would be worth having there.  Is this required for your test suite?   I am still trying to figure out how to test our examples, but having all of these in the core is the wrong place I think.

physicsnemo/mesh/README.md
peterdsharpe: Ah, a good catch! No, this is not in this PR - though it is coming in the subsequent PRs. (The code's already written and tested; I've just excised it from this branch to make the PRs incremental.)

I'll modify this README.md to cut out the not-yet-implemented changes, and then we can pull them back in with the subsequent PRs.

physicsnemo/mesh/README.md
peterdsharpe: Good call - I'll add a note here that this is about topological neighbors (e.g., based on mesh connectivity), while nn.neighbors has tooling for spatial neighbors (e.g., radius_search)

physicsnemo/mesh/mesh.py
peterdsharpe: Yes! For TensorDict, assigning a `batch_size` forces every leaf tensor to start with those dimensions.

So if I have `batch_size=[5, 3]`, then every tensor in the TensorDict must be of shape `[5, 3, ...]`. (If I try assigning a tensor with an incompatible shape, it will error.)

This makes it so that `point_data` and `cell_data` *must* have data that is shape-compatible with the number of points/cells in the mesh.

physicsnemo/mesh/mesh.py
peterdsharpe: Hahaha... effectively yes! Undirected graphs are a special case of the "arbitrary-dimensional pure-simplicial meshes" that `physicsnemo.mesh` supports.

1D manifolds (i.e., each "cell" is a line segment / edge) embedded in higher-dimensional space essentially give you all graph capabilities you would need for geometric deep learning, assuming that those graphs are undirected. And you naturally get the ability to compute things like graph Laplacians just with a generalized manifold Laplacian operation (and same for gradients, and so on).

This is definitely something we should strongly keep exploring in the future!

physicsnemo/mesh/mesh.py
peterdsharpe: Fair enough! This was intended to prevent circular dependency import with submodules, since those would import `Mesh` in order to do type-hinting. However, I can instead break this circular dependency by doing `from typing import TYPE_CHECKING` in submodules, and guarding `Mesh` imports behind `if TYPE_CHECKING`. This way, imports in `mesh.py` can be top-level.

physicsnemo/mesh/mesh.py
peterdsharpe: Fair enough - let's discuss options for achieving this later today. The method-chaining syntax (i.e., "[fluent interface](https://en.wikipedia.org/wiki/Fluent_interface)") is very ergonomic for manipulation, e.g.

```python
my_mesh.subdivide().translate().rotate()
```

I'd like to keep this method-chaining, but I can also see how this creates two sources of truth for the function signature, which is less-than-ideal.

The way I see it, our options are:

1. only do the functional interface (i.e., no methods on `Mesh`)
1. only do the fluent interface (i.e., all of the functions become private `_`-prefixed functions, and the user-facing API is purely via `Mesh` methods)
1. do both (current implementation)
    - possible drawback: two sources of truth
1. do both, but register this mapping dynamically
    - possible drawback: we should make sure that IDE autocomplete tools pick up this dynamic registration

physicsnemo/mesh/mesh.py
peterdsharpe: Fixed in 31418590a7e60cb264a00a5dcab99a82f5c5e4b0

physicsnemo/mesh/mesh.py
peterdsharpe: Linking to discussion here: https://github.com/NVIDIA/physicsnemo/pull/1267#discussion_r2600421370

coreyjadams: Right, I think there is a difference on this particular function.  The chaining is one thing, below, I was suggesting the visualization method(s) itself should live elsewhere to keep the core object clean of matplotlib, etc.

physicsnemo/mesh/utilities/_padding.py
peterdsharpe: It's a good question. For `_pad_with_value`, we could use `torch.nn.functional.pad`, though the syntax is a bit more cumbersome. However, for `_pad_by_tiling_last`, we cannot use PyTorch's padding functionals:

From `F.pad` [docs](https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.pad.html):

> "Circular, replicate and reflection padding are implemented for padding the last 3 dimensions of a 4D or 5D input tensor, the last 2 dimensions of a 3D or 4D input tensor, or the last dimension of a 2D or 3D input tensor."

So, you can't do `'replicate'` padding with the first dimension of a 2D input tensor, which is what we need to do here. This seems like quite a silly limitation on PyTorch's part here, but it is what it is. Furthermore, leaf tensors can be arbitrary-dimensional, not just 2D, which is also not supported by `F.pad` if you want to pad the first dimension.

Because of that, and the fact that this code is only ~20 lines, I think I'll keep this file.

physicsnemo/mesh/README.md
peterdsharpe: Agreed; TODO in subsequent PR after the full -Mesh merge.

physicsnemo/mesh/README.md
peterdsharpe: Fixed 90d081e2b97c36dee8e86470e7326f35fd348de2

physicsnemo/mesh/mesh.py
peterdsharpe: Resolved from our discussion: keep 3 (do both).

physicsnemo/mesh/primitives/README.md
peterdsharpe: Hmmm - this is a tricky one.

The primitives are required for the test suite, and they're a useful set of examples for various meshes and how to manipulate them.

I see two options:

1. Keep it as-is, with `primitives` in the `/physicsnemo/mesh/` directory: this has the downside that we're shipping these primitives with the library code itself, rather than in `/examples/mesh` or similar.
2. Move the `primitives` into `/examples/mesh/` directory: this has the downside that the package's tests depends on the `examples` folder - is that something we're ok with?

peterdsharpe: To me, both options are equally preferable - it's a judgment call. What do you think?

peterdsharpe: RESOLVED after Slack discussion: keeping as-is.

physicsnemo/mesh/primitives/__init__.py
NickGeneva: Okay thought this over more and yeah the oop side isnt ideal here since its just creating a Mesh object. Its better the way you have it.

I guess are these just for testing mostly? Or do you see more CSG usage of these... may they arent really needed to be documented in the docs page.

physicsnemo/mesh/primitives/procedural/noise.py
```diff
@@ -0,0 +1,194 @@
...
+
+    Generates smooth pseudo-random noise values using gradient interpolation
+    on an n-dimensional lattice. Works for any number of spatial dimensions.
+
+    The implementation uses:
+    - Smoothstep interpolation for C² continuity
+    - Hash-based pseudo-random gradients
+    - n-linear interpolation in hypercube
+
+    Args:
```
NickGeneva: Shouldnt these be numpy doc strings?

physicsnemo/mesh/primitives/procedural/noise.py
```diff
@@ -0,0 +1,194 @@
...
+        >>>
+        >>> # Works on GPU
+        >>> points_gpu = points.cuda()
+        >>> noise_gpu = perlin_noise_nd(points_gpu, scale=1.0, seed=42)
+    """
+    device = points.device
+    n_dims = points.shape[-1]
+
+    ### Create permutation table from seed
+    torch.manual_seed(seed)
```
NickGeneva: instead of setting the global manual seed here, perhaps a generator is better so this does not accidentally influence other parts of the code based when used?

peterdsharpe: Good call, will do!

physicsnemo/mesh/primitives/planar/regular_polygon.py
```diff
@@ -0,0 +1,48 @@
...
+
+Dimensional: 2D manifold in 2D space.
+"""
+
+import torch
+
+from physicsnemo.mesh.mesh import Mesh
+
+
+def load(n_sides: int = 6, radius: float = 1.0, device: str = "cpu") -> Mesh:
```
NickGeneva: Nit: Technically these should be

```
device: torch.device | str = "cpu"
```

but I typically just default to

```
device: torch.device = "cpu"
```

The torch.device only would align with pyt docs

physicsnemo/mesh/primitives/__init__.py
peterdsharpe: Yeah, these are mostly just for testing + providing user examples of how to instantiate your own `Mesh` objects.

I think it's important to provide info to users on how to instantiate meshes directly from tensors, because we *also* provide `physicsnemo.mesh.io` which lets you transfer `Mesh` objects back and forth with PyVista. This runs the risk that users develop the bad habit of *only* instantiating via PyVista - something we don't want, since PyVista/VTK is CPU-bound and has very slow save/load. So, by providing these example primitives, we show the user "hey, I know there's that comfortable PyVista interface for compatibility, but you *really* shouldn't be afraid to directly construct your own `Mesh` objects instead."

I would move these primitives to the `/examples/` directory, except that the test suite uses these primitives, and we don't want the tests to depend on the examples directory.

As far as CSG goes - currently, no plans to support CSG operations (e.g., booleans). This gets really tricky since you need to ensure the resulting mesh is pure-simplicial (i.e., consists only of [simplexes](https://en.wikipedia.org/wiki/Simplex)), which is not straightforward for general boolean ops since you need to do some re-triangulation. Would definitely be nice to have, but yeah - a tricky one to implement 😅

physicsnemo/mesh/mesh.py
```diff
@@ -0,0 +1,1044 @@
...
+
+from physicsnemo.mesh.transformations import rotate, scale, transform, translate
+from physicsnemo.mesh.utilities import get_cached, scatter_aggregate, set_cached
+from physicsnemo.mesh.utilities._padding import _pad_by_tiling_last, _pad_with_value
+from physicsnemo.mesh.utilities.mesh_repr import format_mesh_repr
+from physicsnemo.mesh.visualization import draw_mesh
+
+
+@tensorclass(tensor_only=True)
+class Mesh:
```
NickGeneva, line 37: does the `tensorclass` decorator give Mesh the `.to(device)` method?

NickGeneva: NVM read the doc string....

physicsnemo/mesh/mesh.py
peterdsharpe: Resolved per our conversation: plan is to keep the `.draw()` method, but rigorously ensure that there are no import-time errors if `pyvista` or `matplotlib` are not installed (similar to how we've done this in other parts of the refactored library)

physicsnemo/mesh/mesh.py
```diff
@@ -0,0 +1,1044 @@
...
+    -----
+    **Simplex Definition**
+
+    An n-simplex is the convex hull of n+1 affinely independent points. The vertices
+    are stored as indices into the ``points`` array, so a triangle (2-simplex) is
+    represented by 3 indices, a tetrahedron (3-simplex) by 4 indices, etc.
+
+    **Caching**
+
+    Expensive geometric computations (centroids, areas, normals) are automatically
```
NickGeneva: While caching is nice, I'm wondering if there needs to be some controls since these could be fairly large tensors for big meshes.

Like I'm wondering if these should have setters (maybe this data could be manually loaded), getters (property) and deleter to handle the cache interface under the hood.

physicsnemo/mesh/mesh.py
```diff
@@ -0,0 +1,1047 @@
...
+            ### Normalize to get unit normals
+            # For isolated points (zero weighted sum), F.normalize returns zero vector
+            cached = F.normalize(weighted_normals, dim=-1)
+            set_cached(self.point_data, "normals", cached)
+
+        return cached
+
+    @classmethod
+    def merge(
+        cls, meshes: Sequence["Mesh"], global_data_strategy: Literal["stack"] = "stack"
```
NickGeneva, line 792: do all meshes need to be on the same device here?

physicsnemo/mesh/mesh.py
```diff
@@ -0,0 +1,1047 @@
...
+            )
+
+            new_point_data[key] = point_values
+
+        ### Return new mesh with updated point data
+        return Mesh(
+            points=self.points,
+            cells=self.cells,
+            point_data=new_point_data,
+            cell_data=self.cell_data,
```
NickGeneva, line 1077: General question about Tensor dicts, are these things passed by reference here like normal dicts?

So if I convert, then mutate some cell data will it also change in the original mesh? Is there any concern about that?

physicsnemo/mesh/mesh.py
```diff
@@ -0,0 +1,1047 @@
...
+            cells=torch.cat(
+                [m.cells + offset for m, offset in zip(meshes, cell_index_offsets)],
+                dim=0,
+            ),
+            point_data=TensorDict.cat([m.point_data for m in meshes], dim=0),
+            cell_data=TensorDict.cat([m.cell_data for m in meshes], dim=0),
+            global_data=global_data,
+        )
+
+    def slice_points(self, indices: int | slice | torch.Tensor) -> "Mesh":
```
NickGeneva: Question, if im slicing a mesh... doesnt both cell and points need updating?
Here I slice some points, but not some of cells will point to points no longer in the mesh.

Am I thinking about whats happening incorrectly?

physicsnemo/mesh/primitives/procedural/noise.py
peterdsharpe: Good call - all docstrings converted to NumPy style here: 6092b4cb63e1a4fbba5666a71de461a575eeddb0

(This is a holdover from when it was the standalone TorchMesh project, where I used Google-style docstrings)

physicsnemo/mesh/README.md
peterdsharpe: This is already generalized into an arbitrary graph node / edge framework.

Specifically, consider the case of a 1D manifold (i.e., line segments) embedded into 3D space. Then:
- `points` has shape `(n_points, 3)`, and are effectively node positions
- `cell` has shape (`n_cells, 2)`, and are effectively undirected edges between nodes

physicsnemo/mesh/io/io_pyvista.py
peterdsharpe: Good catch, will fix!

physicsnemo/mesh/mesh.py
```diff
@@ -0,0 +1,1175 @@
...
+                batch_size=torch.Size([target_n_points]),
+            ),
+            cell_data=self.cell_data.apply(  # type: ignore
+                lambda x: _pad_with_value(x, target_n_cells, data_padding_value),
+                batch_size=torch.Size([target_n_cells]),
+            ),
+            global_data=self.global_data,
+        )
+
+    def pad_to_next_power(
```
NickGeneva, line 1215: How does this work in practice when loading meshes with different data?
Do I need to manually set the properties of the current mesh object with appropriate indices?

I see the point of this, but having trouble envisioning its use in a training loop of sorts. Hopefully the question makes sense.

physicsnemo/mesh/mesh.py
```diff
@@ -0,0 +1,1175 @@
...
+            n = (torch.tensor(current_size).log() / torch.tensor(base).log()).ceil()
+            return int(torch.tensor(base) ** n)
+
+        target_n_points = next_power_size(self.n_points, power)
+        target_n_cells = next_power_size(self.n_cells, power)
+
+        return self.pad(
+            target_n_points=target_n_points,
+            target_n_cells=target_n_cells,
+            data_padding_value=data_padding_value,
```
NickGeneva, line 1276: Should the default padding value be like a NaN or something else here versus a 0?

physicsnemo/mesh/mesh.py
NickGeneva: Because I'm also not super familiar with some of these calcs without opening a wikipedia page... can you point me to the location of the area / volume unit tests and also normal calculation unit tests?

I see some references with these in other transform tests, but looking for some small analytical confirmation these produce the right results and theres a lot of tests :smile:

physicsnemo/mesh/utilities/_scatter_ops.py
```diff
@@ -0,0 +1,127 @@
...
+    torch.Tensor
+        Aggregated data at destinations, shape (n_dst, *data_shape).
+        For "mean" mode, values are weighted averages.
+        For "sum" mode, values are weighted sums.
+
+    Examples
+    --------
+    >>> # Aggregate cell data to points
+    >>> src_data = torch.tensor([[1.0], [2.0], [3.0]])  # 3 cells
+    >>> src_to_dst = torch.tensor([0, 0, 1])  # map to 2 points
```
NickGeneva, line 76: I've gotten myself confused, why is `src_to_dst` 1D here? I'm not following how this maps to 2 points.
I guess this makes sense on a 1D mesh, but what if I'm scattering data and need to aggregate 3 points into 1 new location.

physicsnemo/mesh/utilities/_scatter_ops.py
NickGeneva: Would scatter_aggregate be something that would go into the functional submodule? I think maybe your second point here is the key, I like this, but perhaps out of scope of this PR?

Namely, some optimized general scatter ops should probably be more rigorously checked unrelated to the mesh object. :thinking:

physicsnemo/mesh/README.md
NickGeneva: Do you know if these will get packed into the wheel currently?
I dont think they will since we dont have any manifest.in or project data files... but wanted to confirm by default python will just grab the python files.

peterdsharpe: It's a good question. My *understanding* (@coreyjadams @ktangsali to confirm) is that different tools treat this differently. For example, `twine` wheel-builds do require a manifest to pick up anything other than the `.py` files - but `uv` by default builds everything inside the source directory. So, it comes down to how we're building it when we ship it.

physicsnemo/mesh/README.md
peterdsharpe: I haven't yet compared this against a warp implementation, but this would definitely be worth doing down the road! Going to deem this beyond scope of this PR, but marking this on my TODO backburner list.

physicsnemo/mesh/mesh.py
peterdsharpe: Overhauled the Mesh docstring to include more definitions and make it more accessible in general: f0ace32ed81952cef68bb06842348e3a6d181da0

peterdsharpe: @NickGeneva on tests:

Good call - these methods are tested throughout the `primitives` test suite, but it's probably helpful to pull them out into a separately-defined test suite associated with the Mesh methods itself. Now added at `./test/mesh/mesh/test_geometry_properties.py` in commit 762b5c9d34cf01776cc8b9962a3073153ed21c64.

physicsnemo/mesh/README.md
NickGeneva: Good enough for me lol. Thats my understanding as well, it will just pick up the .py files :+1:

physicsnemo/mesh/mesh.py
peterdsharpe: Ah, good catch! Yes, this will error in the `torch.cat()` if these aren't on the same device.

However, this is a bit complicated, because `Mesh.device` is actually `None` by default. This essentially indicates that the constituent tensors are "uncontrolled" - we're not mandating that all tensor live on any given device. It's only after saying `mesh.device = "cuda"` or `mesh.to("cuda")` or similar that we actually have a non-null assigned value for `mesh.device`.

So, it's tricky, since two Meshes could have `Mesh.device == None` and be either a) device-compatible or b) device-incompatible. So, we can't easily catch this potential hiccup early.

I think the most natural thing to do here is to just allow `torch.cat` to raise its RuntimeError in a mixed-device scenario, but to add a note to the docstring of `Mesh.merge()` to indicate that it can raise a RuntimeError.

For posterity, that RuntimeError looks like:
```
RuntimeError: Expected all tensors to be on the same device, but got tensors is on cuda:0, different from other tensors on cpu (when checking argument in method wrapper_CUDA_cat)
```

physicsnemo/mesh/primitives/procedural/noise.py
peterdsharpe: Done in a8a19d45091559e2fa053857ac85a6bf06282c88

physicsnemo/mesh/mesh.py
peterdsharpe: Oh yeah - not sure why it slipped my mind, but padding to NaN is a better choice here. Yes, I'll implement it :)

physicsnemo/mesh/primitives/procedural/noise.py
peterdsharpe: Fixed

physicsnemo/mesh/mesh.py
peterdsharpe: Fixed

physicsnemo/mesh/mesh.py
peterdsharpe: The goal of `pad_to_next_power` really comes down to `torch.compile` compatibility.

Specifically: a lot of models operate on data with different shapes - for example, consider a GNN-based architecture where each sample of the training data is a graph with a different number of nodes or edges. In such a case, you have two options for making `torch.compile` work:
1. You can use `torch.compile(dynamic=True)`, which will compile a less-specialized kernel with flexible array shapes. However, you can't use CUDAGraphs acceleration, since this requires static shapes.
2. You can use `torch.compile(dynamic=False)`, which uses static shapes, and you can then use CUDAGraphs to strip out all of the Python-side kernel launch overhead. But the downside is that you need to re-compile the kernel for *every* sample in your dataset, which is super costly and a non-starter.

This presents a 3rd option:
- Whenever you receive a sample, pad it to the next power of some given factor. So, if I get a graph with 900 nodes, and the factor is 2, pad this to 1024 nodes.
- This makes it so that your samples tend to have common shapes - so you might have a specialized compiled kernel for 512 nodes, 1024 nodes, and 2048 nodes, but you won't have a kernel for every single sample size seen in your dataset.

In short, it basically enables you to use CUDAGraphs on datasets with non-static shapes, practically.

physicsnemo/mesh/mesh.py
peterdsharpe: Resolved per Slack discussion.

physicsnemo/mesh/primitives/planar/regular_polygon.py
peterdsharpe: This is a good catch, and we should do this - this is important if set device to something like `my_other_tensor.device`. Implementing...

peterdsharpe: Done 67259ac9968978e8f5ab865d963dad664891e195

physicsnemo/mesh/mesh.py
peterdsharpe: This is a good point, and a true bug! Good catch. Fixing now!

physicsnemo/mesh/mesh.py
peterdsharpe: Done: ae83356bb75328712b19e9ea0a2c7a1955d65b99

physicsnemo/mesh/mesh.py
peterdsharpe: Done: c4f93e3240a2c6903fdd7ff0d1a29241670e54fa

physicsnemo/mesh/mesh.py
peterdsharpe: Good question. Yes, that would be the case. I don't think it's clear-cut how to handle this - since that cell_data could potentially be quite large.

I think I'll just add a notice in this function that unmodified fields are passed by reference.

physicsnemo/mesh/utilities/_scatter_ops.py
peterdsharpe: `src_to_dst` is a mapping of indices in the destination array.

It maps to two points, since `n_dst=2` later. And specifically, we're aggregating to destination indices 0 and 1 (i.e., 2 points).

If you need to aggregate 3 points onto one new location, you'd probably have something like:

`src_to_dist = torch.tensor([0, 0, 0])`
and
`n_dst = 1`

(Though at this point, you could just torch.mean and it'd be faster.)

physicsnemo/mesh/mesh.py
peterdsharpe: To be deferred to subsequent PR

physicsnemo/mesh/utilities/_scatter_ops.py
peterdsharpe: I agree with generalizing this, but plan to defer that to a future PR

physicsnemo/mesh/io/io_pyvista.py
peterdsharpe: Fixed

PR #1295, opened 2025-12-20
Fixes Non-Deterministic Tests

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

On rare occasion, hitting sporadic CI failure due to unseeded randomness in tests.

## Checklist

## Review Process

### DIFF
--- a/.github/workflows/install-ci.yml
+++ b/.github/workflows/install-ci.yml
@@ -8,6 +8,7 @@ on:
   schedule:
     - cron: '0 6 * * *'  # 6 AM UTC daily
   workflow_dispatch:  # Allows manual triggering from GitHub UI
+  merge_group:
 
 jobs:
   test-pip:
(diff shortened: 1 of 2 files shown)


test/models/meshgraphnet/test_meshgraphnet.py
```diff
@@ -174,6 +174,11 @@ def test_meshgraphnet_checkpoint(device, pytestconfig, set_physicsnemo_force_te)
 
     from physicsnemo.models.meshgraphnet import MeshGraphNet
 
+    # Seed for reproducibility
+    torch.manual_seed(0)
+    np.random.seed(0)
+    random.seed(0)
+
     # Construct MGN model
```
coreyjadams, line 177: Any idea why we weren't hitting sporadic failures with this, in the past?  Is it cuda or cpu or both that is failing now?

coreyjadams (approved):

test/models/meshgraphnet/test_meshgraphnet.py
peterdsharpe: We were, in fact, hitting sporadic failures with this on Blossom - very spooky! 😅 (This was the impetus for this PR.)

laserkelvin:
Actually a dumb question: would it work if we just had a top level `conftest.py` that had a function-level scope, so we don't have to think about adding it to every test?

```python
# not sure if function, module, or session would be better
@pytest.fixture(autouse=True, scope="module")
def set_random_seed():
    SEED = 3519581
    torch.manual_seed(SEED)
    random.seed(SEED)
    np.random.seed(SEED)  # use legacy global state
    yield
```
 
IMO it does lift the burden from making sure individual (and new) tests are seeded so this doesn't become an issue

That being said though, tests that are failing because of unset seeds could be an issue with the test itself...

peterdsharpe:
@laserkelvin Never a dumb question! I actually like your proposal a lot - seems like a better way to prevent this mistake in the future.

I'll have to do some digging to see if it's technically feasible, but if it is, I'll implement it instead.

laserkelvin:
> @laserkelvin Never a dumb question! I actually like your proposal a lot - seems like a better way to prevent this mistake in the future.
> 
> I'll have to do some digging to see if it's technically feasible, but if it is, I'll implement it instead.

I'm prepared to delete my comment if it doesn't work 😆

dallasfoster (approved):

PR #1304, opened 2026-01-05
PhysicsNemo Datapipes

coreyjadams:
# PhysicsNeMo Pull Request

## Description

This PR implements Physicsnemo Datapipes as configurable pipelines with reusable components.  It is meant to provide a foundation for training and inference pipelines, though full scale replacement / updating / deprecation of existing targeted pipelines is for the future and not implemented here.

I will defer a description of the datapipe design to the README and tutorials - located in `examples/minimal/datapipes/`. 

The core components of the datapipe are all present:
- reader abstraction, for getting data from disk into CPU.  Implementations for zarr and .npz are here.
- transforms, for manipulating data on the GPU.  Some transforms are already here, but it's easy to extend.
- datasets, for preloading and staging data to the GPU
- dataloader, for drop-in replacement with pytorch.

## Checklist

## Review Process

### DIFF
--- a/physicsnemo/datapipes/readers/zarr.py
+++ b/physicsnemo/datapipes/readers/zarr.py
@@ -54,31 +54,31 @@ class ZarrReader(Reader):
 
     >>> # Directory with sample_0.zarr, sample_1.zarr, ...
     >>> # Each contains arrays like "positions", "features", etc.
-    >>> reader = ZarrReader("data_dir/", group_pattern="sample_*.zarr")
-    >>> data, metadata = reader[0]  # Returns (TensorDict, dict) tuple
+    >>> reader = ZarrReader("data_dir/", group_pattern="sample_*.zarr")  # doctest: +SKIP
+    >>> data, metadata = reader[0]  # Returns (TensorDict, dict) tuple  # doctest: +SKIP
 
     Load only specific fields:
 
-    >>> reader = ZarrReader("data_dir/", fields=["positions", "velocity"])
-    >>> data, metadata = reader[0]
+    >>> reader = ZarrReader("data_dir/", fields=["positions", "velocity"])  # doctest: +SKIP
+    >>> data, metadata = reader[0]  # doctest: +SKIP
 
     Load attributes from Zarr groups:
 
     >>> # If the Zarr group has attributes like "timestep" or "scale_factor",
     >>> # you can request them as fields:
-    >>> reader = ZarrReader("data_dir/", fields=["positions", "timestep", "scale_factor"])
-    >>> data, metadata = reader[0]  # data["timestep"] contains the attribute value
+    >>> reader = ZarrReader("data_dir/", fields=["positions", "timestep", "scale_factor"])  # doctest: +SKIP
+    >>> data, metadata = reader[0]  # data["timestep"] contains the attribute value  # doctest: +SKIP
 
     With coordinated subsampling for large arrays:
 
-    >>> reader = ZarrReader(
+    >>> reader = ZarrReader(  # doctest: +SKIP
     ...     "data_dir/",
     ...     coordinated_subsampling={
     ...         "n_points": 50000,
     ...         "target_keys": ["volume_coords", "volume_fields"],
     ...     }
     ... )
-    >>> data, metadata = reader[0]
+    >>> data, metadata = reader[0]  # doctest: +SKIP
     """
 
     def __init__(
--- a/physicsnemo/datapipes/transforms/normalize.py
+++ b/physicsnemo/datapipes/transforms/normalize.py
@@ -44,31 +44,42 @@ class Normalize(Transform):
 
     Parameters can be provided directly or loaded from a ``.npz`` file.
 
-    Example (mean-std scaling):
-        >>> norm = Normalize(
-        ...     input_keys=["pressure", "velocity"],
-        ...     method="mean_std",
-        ...     means={"pressure": 101325.0, "velocity": 0.0},
-        ...     stds={"pressure": 1000.0, "velocity": 10.0},
-        ... )
-        >>> normalized = norm(sample)
+    Examples
+    --------
+    Mean-std scaling:
 
-    Example (min-max scaling):
-        >>> norm = Normalize(
-        ...     input_keys=["pressure", "velocity"],
-        ...     method="min_max",
-        ...     mins={"pressure": 100000.0, "velocity": -50.0},
-        ...     maxs={"pressure": 110000.0, "velocity": 50.0},
-        ... )
-        >>> normalized = norm(sample)
+    >>> import torch
+    >>> from tensordict import TensorDict
+    >>> sample = TensorDict({
+    ...     "pressure": torch.tensor([101325.0, 102325.0, 100325.0]),
+    ...     "velocity": torch.tensor([10.0, -10.0, 0.0]),
+    ... })
+    >>> norm = Normalize(
+    ...     input_keys=["pressure", "velocity"],
+    ...     method="mean_std",
+    ...     means={"pressure": 101325.0, "velocity": 0.0},
+    ...     stds={"pressure": 1000.0, "velocity": 10.0},
+    ... )
+    >>> normalized = norm(sample)
+    >>> normalized["pressure"]
+    tensor([ 0.,  1., -1.])
+    >>> normalized["velocity"]
+    tensor([ 1., -1.,  0.])
 
-    Example (loading from file):
-        >>> norm = Normalize(
-        ...     input_keys=["pressure", "velocity"],
-        ...     method="mean_std",
-        ...     stats_file="normalization_stats.npz",
-        ... )
-        >>> normalized = norm(sample)
+    Min-max scaling (normalizes to [-1, 1]):
+
+    >>> sample = TensorDict({
+    ...     "pressure": torch.tensor([100000.0, 105000.0, 110000.0]),
+    ... })
+    >>> norm = Normalize(
+    ...     input_keys=["pressure"],
+    ...     method="min_max",
+    ...     mins={"pressure": 100000.0},
+    ...     maxs={"pressure": 110000.0},
+    ... )
+    >>> normalized = norm(sample)
+    >>> normalized["pressure"]
+    tensor([-1.,  0.,  1.])
     """
 
     def __init__(
--- a/physicsnemo/datapipes/__init__.py
+++ b/physicsnemo/datapipes/__init__.py
@@ -25,26 +25,6 @@ Designed for clean separation of concerns:
 - **Dataset**: Reader + transforms pipeline with optional auto device transfer
 - **DataLoader**: Batched iteration with optional prefetching
 
-Example:
-    >>> import physicsnemo.datapipes as dp
-    >>> from tensordict import TensorDict
-    >>>
-    >>> # Create a dataset with transforms and automatic device transfer
-    >>> dataset = dp.Dataset(
-    ...     reader=dp.HDF5Reader("data.h5", fields=["pressure", "velocity"]),
-    ...     transforms=[
-    ...         dp.Normalize(input_keys=["pressure"], means={"pressure": 0.0}, stds={"pressure": 1.0}),
-    ...         dp.SubsamplePoints(input_keys=["pressure", "velocity"], n=10000),
-    ...     ],
-    ...     device="cuda",  # Automatic GPU transfer!
-    ... )
-    >>>
-    >>> # Create a dataloader
-    >>> loader = dp.DataLoader(dataset, batch_size=16, shuffle=True)
-    >>>
-    >>> # Iterate over batches
-    >>> for data, metadata in loader:
-    ...     output = model(data["pressure"])
 """
 
 from tensordict import TensorDict
--- a/physicsnemo/datapipes/collate.py
+++ b/physicsnemo/datapipes/collate.py
@@ -122,13 +122,15 @@ class DefaultCollator(Collator):
     ... ]
     >>> collator = DefaultCollator()
     >>> batched_data = collator(samples)
-    >>> batched_data["x"].shape  # torch.Size([2, 10, 3])
+    >>> batched_data["x"].shape
+    torch.Size([2, 10, 3])
 
     With metadata collation enabled:
 
     >>> collator = DefaultCollator(collate_metadata=True)
     >>> batched_data, metadata_list = collator(samples)
-    >>> metadata_list  # [{"file": "a.h5"}, {"file": "b.h5"}]
+    >>> metadata_list
+    [{'file': 'a.h5'}, {'file': 'b.h5'}]
     """
 
     def __init__(
@@ -221,14 +223,17 @@ class ConcatCollator(Collator):
     ... ]
     >>> collator = ConcatCollator(dim=0, add_batch_idx=True)
     >>> batched_data = collator(samples)
-    >>> batched_data["points"].shape  # torch.Size([250, 3])
-    >>> batched_data["batch_idx"].shape  # torch.Size([250])
+    >>> batched_data["points"].shape
+    torch.Size([250, 3])
+    >>> batched_data["batch_idx"].shape
+    torch.Size([250])
 
     With metadata collation enabled:
 
     >>> collator = ConcatCollator(dim=0, add_batch_idx=True, collate_metadata=True)
     >>> batched_data, metadata_list = collator(samples)
-    >>> metadata_list  # [{"file": "a.h5"}, {"file": "b.h5"}]
+    >>> metadata_list
+    [{'file': 'a.h5'}, {'file': 'b.h5'}]
     """
 
     def __init__(
--- a/physicsnemo/datapipes/dataloader.py
+++ b/physicsnemo/datapipes/dataloader.py
@@ -60,21 +60,21 @@ class DataLoader:
     --------
     >>> from physicsnemo.datapipes import DataLoader, Dataset, HDF5Reader, Normalize
     >>>
-    >>> dataset = Dataset(
+    >>> dataset = Dataset(  # doctest: +SKIP
     ...     HDF5Reader("data.h5", fields=["input", "target"]),
     ...     transforms=Normalize(["input"], method="mean_std", means={"input": 0.0}, stds={"input": 1.0}),
     ...     device="cuda",  # Automatic GPU transfer
     ... )
-    >>> loader = DataLoader(dataset, batch_size=16, shuffle=True)
+    >>> loader = DataLoader(dataset, batch_size=16, shuffle=True)  # doctest: +SKIP
     >>>
-    >>> for batch in loader:
+    >>> for batch in loader:  # doctest: +SKIP
     ...     output = model(batch["input"])
 
     With DistributedSampler:
 
     >>> from torch.utils.data.distributed import DistributedSampler
-    >>> sampler = DistributedSampler(dataset)
-    >>> loader = DataLoader(dataset, batch_size=16, sampler=sampler)
+    >>> sampler = DistributedSampler(dataset)  # doctest: +SKIP
+    >>> loader = DataLoader(dataset, batch_size=16, sampler=sampler)  # doctest: +SKIP
     """
 
     def __init__(
--- a/physicsnemo/datapipes/dataset.py
+++ b/physicsnemo/datapipes/dataset.py
@@ -72,26 +72,26 @@ class Dataset:
     other computation.
 
     >>> # Start prefetching
-    >>> dataset.prefetch(0, stream=stream0)
-    >>> dataset.prefetch(1, stream=stream1)
+    >>> dataset.prefetch(0, stream=stream0)  # doctest: +SKIP
+    >>> dataset.prefetch(1, stream=stream1)  # doctest: +SKIP
     >>>
     >>> # Retrieve results (waits if not ready)
-    >>> sample_0 = dataset[0]  # Uses prefetched result
+    >>> sample_0 = dataset[0]  # Uses prefetched result  # doctest: +SKIP
 
     Examples
     --------
     >>> from physicsnemo.datapipes import Dataset, HDF5Reader, Normalize
     >>>
-    >>> reader = HDF5Reader("data.h5", fields=["pressure", "velocity"])
-    >>> transforms = Normalize(
+    >>> reader = HDF5Reader("data.h5", fields=["pressure", "velocity"])  # doctest: +SKIP
+    >>> transforms = Normalize(  # doctest: +SKIP
     ...     ["pressure"],
     ...     method="mean_std",
-    ...     means={"pressure": 0.0},
-    ...     stds={"pressure": 1.0},
+    ...     means={"pressure": 0.0},  # doctest: +SKIP
+    ...     stds={"pressure": 1.0},  # doctest: +SKIP
     ... )
     >>>
-    >>> dataset = Dataset(reader, transforms=transforms, device="cuda")
-    >>> sample, metadata = dataset[0]
+    >>> dataset = Dataset(reader, transforms=transforms, device="cuda")  # doctest: +SKIP
+    >>> sample, metadata = dataset[0]  # doctest: +SKIP
     """
 
     def __init__(
--- a/physicsnemo/datapipes/readers/base.py
+++ b/physicsnemo/datapipes/readers/base.py
@@ -56,7 +56,7 @@ class Reader(ABC):
     --------
     Custom reader implementation:
 
-    >>> class MyReader(Reader):
+    >>> class MyReader(Reader):  # doctest: +SKIP
     ...     def __init__(self, path: str, **kwargs):
     ...         super().__init__(**kwargs)
     ...         self.data = load_my_data(path)
--- a/physicsnemo/datapipes/readers/hdf5.py
+++ b/physicsnemo/datapipes/readers/hdf5.py
@@ -53,15 +53,15 @@ class HDF5Reader(Reader):
     Single file mode:
 
     >>> # File structure: data.h5 with datasets "pressure" (N, 100), "velocity" (N, 100, 3)
-    >>> reader = HDF5Reader("data.h5", fields=["pressure", "velocity"])
-    >>> data, metadata = reader[0]  # Returns (TensorDict, dict) tuple
-    >>> data["pressure"].shape  # torch.Size([100])
+    >>> reader = HDF5Reader("data.h5", fields=["pressure", "velocity"])  # doctest: +SKIP
+    >>> data, metadata = reader[0]  # Returns (TensorDict, dict) tuple  # doctest: +SKIP
+    >>> data["pressure"].shape  # torch.Size([100])  # doctest: +SKIP
 
     Directory mode:
 
     >>> # Directory with sample_0.h5, sample_1.h5, ...
-    >>> reader = HDF5Reader("data_dir/", file_pattern="sample_*.h5")
-    >>> data, metadata = reader[0]  # Loads all datasets from sample_0.h5
+    >>> reader = HDF5Reader("data_dir/", file_pattern="sample_*.h5")  # doctest: +SKIP
+    >>> data, metadata = reader[0]  # Loads all datasets from sample_0.h5  # doctest: +SKIP
     """
 
     def __init__(
--- a/physicsnemo/datapipes/readers/numpy.py
+++ b/physicsnemo/datapipes/readers/numpy.py
@@ -43,15 +43,15 @@ class NumpyReader(Reader):
 
     Example (single .npz):
         >>> # data.npz with arrays "positions" (N, 100, 3), "features" (N, 100)
-        >>> reader = NumpyReader("data.npz", fields=["positions", "features"])
-        >>> data, metadata = reader[0]  # Returns (TensorDict, dict) tuple
+        >>> reader = NumpyReader("data.npz", fields=["positions", "features"])  # doctest: +SKIP
+        >>> data, metadata = reader[0]  # Returns (TensorDict, dict) tuple  # doctest: +SKIP
         >>> # Or load all arrays:
-        >>> reader = NumpyReader("data.npz")  # fields=None loads all
+        >>> reader = NumpyReader("data.npz")  # fields=None loads all  # doctest: +SKIP
 
     Example (directory):
         >>> # Directory with sample_0.npz, sample_1.npz, ...
-        >>> reader = NumpyReader("data_dir/", file_pattern="sample_*.npz")
-        >>> data, metadata = reader[0]  # Returns (TensorDict, dict) tuple
+        >>> reader = NumpyReader("data_dir/", file_pattern="sample_*.npz")  # doctest: +SKIP
+        >>> data, metadata = reader[0]  # Returns (TensorDict, dict) tuple  # doctest: +SKIP
     """
 
     def __init__(
--- a/physicsnemo/datapipes/readers/tensorstore_zarr.py
+++ b/physicsnemo/datapipes/readers/tensorstore_zarr.py
@@ -59,24 +59,24 @@ class TensorStoreZarrReader(Reader):
     Basic usage:
 
     >>> # Directory with sample_0.zarr, sample_1.zarr, ...
-    >>> reader = TensorStoreZarrReader("data_dir/", group_pattern="sample_*.zarr")
-    >>> data, metadata = reader[0]  # Returns (TensorDict, dict) tuple
+    >>> reader = TensorStoreZarrReader("data_dir/", group_pattern="sample_*.zarr")  # doctest: +SKIP
+    >>> data, metadata = reader[0]  # Returns (TensorDict, dict) tuple  # doctest: +SKIP
 
     Load only specific fields:
 
-    >>> reader = TensorStoreZarrReader("data_dir/", fields=["positions", "velocity"])
-    >>> data, metadata = reader[0]
+    >>> reader = TensorStoreZarrReader("data_dir/", fields=["positions", "velocity"])  # doctest: +SKIP
+    >>> data, metadata = reader[0]  # doctest: +SKIP
 
     With coordinated subsampling for large arrays:
 
-    >>> reader = TensorStoreZarrReader(
+    >>> reader = TensorStoreZarrReader(  # doctest: +SKIP
     ...     "data_dir/",
     ...     coordinated_subsampling={
     ...         "n_points": 50000,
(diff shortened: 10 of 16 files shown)


examples/minimal/datapipes/README.md
```diff
@@ -0,0 +1,209 @@
+# PhysicsNeMo DataPipes
+
+Dataloading is critical to SciML applications, both for training and inference,
+and the physicsnemo datapipe infrastructure aims to deliver a flexible and configurable
```
peterdsharpe, line 4: Somewhat minor, but I notice that `python` / `physicsnemo` / `pytorch` are consistently not capitalized throughout this *.md and others. My preference would be to capitalize them in any plaintext docs ("Python", "PhysicsNeMo", "PyTorch").

examples/minimal/datapipes/README.md
```diff
@@ -0,0 +1,209 @@
...
+2. **Threading over Multiprocessing** - In python, true concurrency is typically only
+available via multiprocessing or when offloading to compiled libraries or GPU kernels.
+For this reason, many data loaders leverage multiprocessing for data concurrency:
+load images in separate processes, and collate a batch on the main thread.
+For simplicity, with a GPU-first paradigm, the physicsnemo datapipe focuses on GPU
+concurrency via ascynchronous execution and stream-based parallelism.  IO is coordinated
+in multiple threads, instead of multiple processes, and streams enable multiple
+preprocessing pipelines to execute concurrently on the GPU.
+
+3. **Unambiguous Configuration and Serialization** - Datapipes can be a particularly
```
peterdsharpe, line 37: Would be worth defining "datapipes" here for new readers

physicsnemo/datapipes/transforms/base.py
```diff
@@ -0,0 +1,122 @@
...
+        ...         # Apply transformation to all tensors
+        ...         return data.apply(lambda x: x * self.scale)
+    """
+
+    def __init__(self) -> None:
+        """Initialize the transform."""
+        self._device: Optional[torch.device] = None
+
+    @abstractmethod
+    def __call__(self, data: TensorDict) -> TensorDict:
```
peterdsharpe, line 65: Are re-implementations allowed to have optional kwargs, or is it really just `data` allowed and that's it?

physicsnemo/datapipes/core/transforms/base.py
```diff
@@ -0,0 +1,122 @@
...
+        Returns:
+            Transformed TensorDict.
+        """
+        raise NotImplementedError
+
+    def to(self, device: torch.device | str) -> Transform:
+        """
+        Move any internal tensors to the specified device.
+
+        Override this method if your transform has internal tensor state.
```
peterdsharpe: Could we provide a default implementation that checks `self.__dict__` for any tensor leaves, and transfers them if so?

physicsnemo/datapipes/core/transforms/base.py
```diff
@@ -0,0 +1,122 @@
...
+            device: Target device.
+
+        Returns:
+            Self for chaining.
+        """
+        self._device = torch.device(device) if isinstance(device, str) else device
+        return self
+
+    @property
+    def device(self) -> Optional[torch.device]:
```
peterdsharpe:
```suggestion
    def device(self) -> torch.device | None:
```
Minor, but preferred in latest Python versions

physicsnemo/datapipes/core/transforms/base.py
```diff
@@ -0,0 +1,122 @@
...
+        Load state from a state dictionary.
+
+        Override this to restore transform state.
+        """
+        pass
+
+    def __repr__(self) -> str:
+        extra = self.extra_repr()
+        if extra:
+            return f"{self.__class__.__name__}({extra})"
```
peterdsharpe: Super minor, but can't this `if` statement be removed for readability, since it defaults to `""`?

```python
   def __repr__(self) -> str:
       return f"{self.__class__.__name__}({self.extra_repr()})"
```

examples/minimal/datapipes/tutorial_01_getting_started.py
```diff
@@ -0,0 +1,447 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
```
peterdsharpe: Would these tutorials be perhaps clearer as Jupyter notebooks?

Seems like there's a *lot* of `print()` going on that might be better presented as a Markdown block. (Plus other benefits, like being able to convert this into a docs page later, etc.)

physicsnemo/datapipes/core/readers/base.py
```diff
@@ -0,0 +1,226 @@
...
+    Device transfers are handled automatically by Dataset (if device parameter set).
+    Threading/prefetching is handled by the DataLoader.
+
+    Example custom reader:
+        >>> class MyReader(Reader):
+        ...     def __init__(self, path: str, **kwargs):
+        ...         super().__init__(**kwargs)
+        ...         self.data = load_my_data(path)
+        ...
+        ...     def _load_sample(self, index: int) -> dict[str, torch.Tensor]:
```
peterdsharpe: The docs above say "Return (TensorDict, metadata_dict)", but here it looks like everything's returning `dict[str, torch.Tensor]` rather than TensorDict?

physicsnemo/datapipes/core/readers/hdf5.py
```diff
@@ -0,0 +1,211 @@
...
+
+    Example (directory):
+        >>> # Directory with sample_0.h5, sample_1.h5, ...
+        >>> reader = HDF5Reader("data_dir/", file_pattern="sample_*.h5")
+        >>> sample = reader[0]  # Loads all datasets from sample_0.h5
+    """
+
+    def __init__(
+        self,
+        path: Union[str, Path],
```
peterdsharpe:
```suggestion
        path: Path | str,
```

Union types should be `|` operator in 3.10+

examples/minimal/datapipes/README.md
coreyjadams: Good point, I'll fix it.  At least you know that if an LLM had written this, it would have capitalized it :)

examples/minimal/datapipes/README.md
coreyjadams: Good idea, on it.

physicsnemo/datapipes/transforms/base.py
coreyjadams: The design is really "just data and that's it".  The call to all transforms is mostly automatic and transparent to the user, it happens in the dataset: 

```
# Apply transforms (data is now on target device if specified)
if self.transforms is not None:
    if stream is not None:
        with torch.cuda.stream(stream):
            data = self.transforms(data)
        # Record event for synchronization
        result.event = torch.cuda.Event()
        result.event.record(stream)
    else:
        data = self.transforms(data)
```

My intention is that configuration of transforms happens at initialization - and you can have multiple of the same kind with different arguments, that's fine. Is there some dynamic behavior we need that isn't supported?

physicsnemo/datapipes/core/transforms/base.py
coreyjadams: Sure good idea, makes it easier to extend in user space.

examples/minimal/datapipes/tutorial_01_getting_started.py
coreyjadams: Sure!  I didn't do that because we don't have many notebooks floating around in the examples section.  But it is cleaner in a notebook, yeah.

physicsnemo/datapipes/core/readers/base.py
coreyjadams: The docs are referring to the __getitem__ call, which does the actual tensordict wrapping:

```
def __getitem__(self, index: int) -> tuple[TensorDict, dict[str, Any]]:
```

The idea here was it's conceptually simple to implement the `def _load_sample(self, index: int) -> dict[str, torch.Tensor]` function (and `__len__` is also required) and the base class will handle the metadata + tensordict conversion for you.

physicsnemo/datapipes/core/transforms/spatial.py
```diff
@@ -0,0 +1,369 @@
...
+        self.bbox_min = self.bbox_min.to(device)
+        self.bbox_max = self.bbox_max.to(device)
+        return self
+
+    def __repr__(self) -> str:
+        return f"CreateGrid(output_key={self.output_key}, resolution={self.resolution})"
+
+
+@register_transform()
+class KNNNeighbors(Transform):
```
laserkelvin: This is a dumb nitpick but `NNN` feels very typo prone

examples/minimal/datapipes/tutorial_03_custom_gnn_datapipe.py
```diff
@@ -0,0 +1,469 @@
...
+    section_2_pyg_collator()
+
+    # Section 3: Complete pipeline
+    section_3_complete_pipeline()
+
+    print("=" * 70)
+    print("Tutorial 3 Complete!")
+    print()
+    print("Key takeaways:")
+    print("  1. KNNNeighbors transform: Computes graph structure from point clouds")
```
Alexey-Kamenev: I think Kelvin already pointed that out, but one N is redundant. I would use something like `KNearestNeighbors` instead.

Alexey-Kamenev (approved):
LGTM!

physicsnemo/datapipes/core/__init__.py
```diff
@@ -0,0 +1,114 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
```
peterdsharpe: We talked over Slack about the pros/cons of the `./datapipes/core/` vs `./datapipes/` dir structure -- either is fine with me and I'll leave this to your discretion, just putting this comment as a reminder.

physicsnemo/datapipes/core/transforms/utility.py
```diff
@@ -0,0 +1,351 @@
...
+
+        Returns:
+            TensorDict with renamed keys.
+
+        Raises:
+            KeyError: If strict=True and a key in the mapping is not found.
+            ValueError: If a new key name already exists in the data.
+        """
+        # Check for missing keys if strict mode
+        data_keys = set(str(k) for k in data.keys())
```
peterdsharpe: Is casting `str(k)` on a `k` from `data.keys()` type-safe in the case of nested TensorDicts? Or what if you want to rename keys in a nested TensorDict - wouldn't the input and output types be `tuple[str, ...] | str`, depending on whether they're nested?

physicsnemo/datapipes/core/transforms/utility.py
```diff
@@ -0,0 +1,351 @@
...
+        Args:
+            data: Input TensorDict.
+
+        Returns:
+            TensorDict with keys removed according to the configuration.
+
+        Raises:
+            KeyError: If strict=True and a specified key is not found.
+        """
+        available_keys = set(str(k) for k in data.keys())
```
peterdsharpe: I think this is more robustly implemented with `TensorDict.select` and `TensorDict.exclude`, which work in the case of nested TensorDicts: https://docs.pytorch.org/tensordict/stable/tutorials/tensordict_keys.html#selecting-and-excluding-keys

I think this implementation breaks in the nested case due to the direct `.keys()` iteration?

physicsnemo/datapipes/core/transforms/utility.py
coreyjadams: Ah hmmm I didn't consider nested dictionaries here, only top level ones.

At the highest level of a dict, casting to str should be fine on all the keys.  I am inclined to say "everything we build and ship in the core repo is single layer" while users are welcome to inherit and replace functionality as needed.  The goal is to be extensible and configurable, not necessarily perfectly generic.

physicsnemo/datapipes/transforms/__init__.py
```diff
@@ -0,0 +1,92 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
```
peterdsharpe, line 1: In general in the transforms in this folder:

Be careful about direct iteration over `tensordict.keys()` or `.items()` or `.values()`, as opposed to using TensorDict-specific interfaces. This:
- Can sometimes result in unintended behavior in the case of nested TensorDicts, as by-default, this will only get the top-level keys rather than all leaves.
- In the case where you do get the leaves (e.g., `.keys(include_nested=True)`), the resulting keys are not necessarily strings - they are `tuple[str, ...] | str`, depending on whether they're nested or not.
- As an alternative to specifying whether nested keys are included, you can also use the `.flatten()` and `.unflatten()` interface to deal with this (but it may be less memory-efficient? Not sure).

Overall these transforms look good - just want to be sure they don't break in the case of nested keys.

More info on this:
https://docs.pytorch.org/tensordict/stable/tutorials/tensordict_keys.html#nested-values

physicsnemo/datapipes/core/transforms/geometric.py
```diff
@@ -0,0 +1,360 @@
...
+
+        # Apply translation to all keys
+        updates = {}
+        for key in self.input_keys:
+            if key in data:
+                updates[key] = data[key] - center
+
+        return data.update(updates)
+
+    def to(self, device: Union[torch.device, str]) -> "Translate":
```
peterdsharpe: Perhaps we want these Transforms to be `tensorclass` objects (part of TensorDict; basically just dataclasses) instead of regular classes so that we don't need to repeatedly provide `.to()` methods in each one?

physicsnemo/datapipes/core/transforms/geometric.py
```diff
@@ -0,0 +1,360 @@
...
+
+    def __repr__(self) -> str:
+        return (
+            f"TranslationInvariance(input_keys={self.input_keys}, "
+            f"center={self.center_key_or_value})"
+        )
+
+
+@register()
+class ReScale(Transform):
```
peterdsharpe:
```suggestion
class Scale(Transform):
```
Would recommend `Scale`, or barring that, `Rescale` (lowercase `s`)

physicsnemo/datapipes/core/transforms/geometric.py
```diff
@@ -0,0 +1,360 @@
...
+
+        # Apply scaling to all keys
+        updates = {}
+        for key in self.input_keys:
+            if key in data:
+                updates[key] = data[key] / scale
+
+        return data.update(updates)
+
+    def to(self, device: Union[torch.device, str]) -> "ReScale":
```
peterdsharpe: See suggestion above about making all Transforms `tensorclasses` to avoid this boilerplate

physicsnemo/datapipes/transforms/normalize.py
```diff
@@ -0,0 +1,390 @@
...
+        >>> normalized = norm(sample)
+    """
+
+    def __init__(
+        self,
+        input_keys: list[str],
+        means: Optional[dict[str, float | torch.Tensor] | float | torch.Tensor] = None,
+        stds: Optional[dict[str, float | torch.Tensor] | float | torch.Tensor] = None,
+        *,
+        method: Optional[Literal["mean_std", "min_max"]] = None,
```
peterdsharpe, line 91: A few issues:

1. `Optional` is out of favor after 3.10+ (e.g., Ruff usually complains here), and would prefer `Literal["mean_std", "min_max", None]` instead.
2. However, in the `None` cases (which is in fact the default case): doesn't this trip the ValueError? Seems like you can get to the `method` validation without this being overwritten.

physicsnemo/datapipes/transforms/normalize.py
```diff
@@ -0,0 +1,390 @@
...
+        self.input_keys = list(input_keys)
+        self.eps = eps
+
+        # Handle backward compatibility: if means/stds provided without method
+        if means is not None and stds is not None and method is None:
+            warnings.warn(
+                "Providing 'means' and 'stds' without 'method' parameter is deprecated. "
+                "Please specify method='mean_std' explicitly. "
+                "This will become an error in a future version.",
+                DeprecationWarning,
```
peterdsharpe, line 143: Do we need to worry about deprecations on brand-new code? As this is new, seems that there shouldn't be any backwards-compatibility to break, right?

physicsnemo/datapipes/core/transforms/normalize.py
```diff
@@ -0,0 +1,390 @@
...
+            tensor = data[key]
+
+            if self.method == "mean_std":
+                mean = self._means[key]
+                std = self._stds[key]
+
+                # Normalize: (x - mean) / std
+                updates[key] = (tensor - mean) / (std + self.eps)
+
+            else:  # min_max
```
peterdsharpe:
```suggestion
            elif self.method == "min_max"  # min_max
```
Helps readability and type checkers to explicitly check - can do an `else: raise ValueError` at the end.

physicsnemo/datapipes/transforms/normalize.py
```diff
@@ -0,0 +1,390 @@
...
+            "method", "mean_std"
+        )  # Default for backward compat
+        self.eps = state_dict["eps"]
+
+        if self.method == "mean_std":
+            self._means = {k: v.clone() for k, v in state_dict["means"].items()}
+            self._stds = {k: v.clone() for k, v in state_dict["stds"].items()}
+            self._mins = None
+            self._maxs = None
+        else:  # min_max
```
peterdsharpe, line 424:
```suggestion
        elif self.method == "min_max":
```
+ `else: raise ValueError` block (see related comment earlier in this file)

physicsnemo/datapipes/core/transforms/spatial.py
peterdsharpe: Agreed - also, isn't it redundant? "K Nearest Neighbors Neighbors"?

physicsnemo/datapipes/transforms/base.py
```diff
@@ -0,0 +1,122 @@
...
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import Any, Optional
+
+import torch
+from tensordict import TensorDict
+
+
+class Transform(ABC):
```
peterdsharpe, line 30: Can this be a `@tensorclass`? This would solve a lot of challenges that happen in later implementations of this ABC, like repeated `.to()` and `.device()` methods, as well as giving a nice default `repr` base.

https://docs.pytorch.org/tensordict/stable/reference/generated/tensordict.tensorclass.html

It should be possible for something to be *both* a tensorclass and an `ABC`, so hopefully no issues there. Maybe a pattern to consider?

peterdsharpe (changes requested):
Overall looks good, excited to get this in! Readers look great.

Some areas in the transforms that should be fixed/addressed before we pull this in - might be worth considering making Transform a tensorclass?

Tutorials look great - the suggestion to convert to ipynb is there, but up to you if you want to do it just yet.

physicsnemo/datapipes/core/transforms/utility.py
peterdsharpe: Fair enough! That said, the easy solution would be to just flatten and unflatten before/after:

```python
# `data` is a TensorDict, possibly with nested TensorDict keys

data_f = data.flatten()  # now all keys are "parent.child.grandchild" strings
# <do all the transforms>
data = data_f.unflatten()
```

which opens up the possibility of using nesting without errors at the cost of just 2 added lines of code.

I can imagine nested TensorDicts could be quite useful as we do more multiphysics simulations in the future - e.g., coupling a DoMINO fluids model with a MeshGraphNet structural model. But it's your call here if you want to implement it now or save it for later.

physicsnemo/datapipes/core/transforms/spatial.py
coreyjadams: Thannnk you for poinnnting this out.  It nnneeded to be fixed. Donnne.

physicsnemo/datapipes/core/transforms/utility.py
coreyjadams: Done, I went with the flatten/unflatten route.  I also added some tests on this.  You're right that we're maybe going to want to implement nested dictionaries with complicated data pipes.

physicsnemo/datapipes/core/transforms/geometric.py
coreyjadams: I think implementing a generic `to` function that will move all `torch.Tensor` instances in the __dict__ object should be sufficient.  I'm hesitant to inherit from tensorclass just because it seems like an unnecessary step, since this generic `to` class should do enough.  Are there other benefits?

examples/minimal/datapipes/README.md
```diff
@@ -0,0 +1,209 @@
+# PhysicsNeMo DataPipes
+
+Dataloading is critical to SciML applications, both for training and inference,
+and the physicsnemo datapipe infrastructure aims to deliver a flexible and configurable
+set of tools to enable your application.
+
+There are plenty of tools in the python eco system for loading, preprocessing, and
```
laserkelvin: Python ecosystem

examples/minimal/datapipes/README.md
```diff
@@ -0,0 +1,209 @@
...
+preparing your data for training or inference.  To compare / contrast some of these
+tools with the ecosystem available, and see if the physicsnemo datapipe interface
+might be valuable to your workload, consider the following design principles
+we followed when building the physicsnemo datapipes:
+
+1. **GPU-first** - Many scientific datasets are *large* for even a single example:
+the data is high resolution and the preprocessing needs benefit from GPU acceleration.
+Compare this to other methods where the data preprocessing is predominantly CPU-based,
+such as the pytorch Dataloader: whereas CPU-based preprocessing may introduce GPU
```
laserkelvin: PyTorch `DataLoader`

examples/minimal/datapipes/README.md
```diff
@@ -0,0 +1,209 @@
...
+pipeline stalls on high resolution data, GPU-based preprocessing will maximize
+throughput.
+
+2. **Threading over Multiprocessing** - In python, true concurrency is typically only
```
laserkelvin: Python

physicsnemo/datapipes/core/transforms/geometric.py
coreyjadams: Went with the generic `to` option that mostly works, with a few overrides:
- Normalize can hold a dictionary of means/stds and mins/maxs.  So it overrides `to` to call it recursively.
- Compose needs to call `to` on all it's sub-transforms.  There is not a good solution to implement that with tensorclass IMO.

examples/minimal/datapipes/README.md
```diff
@@ -0,0 +1,209 @@
...
+available via multiprocessing or when offloading to compiled libraries or GPU kernels.
```
laserkelvin: I think this section could be clarified a bit at the beginning since both "threading" **and** "multiprocessing" are pretty overloaded terms. Maybe like:

" In Python, true concurrency is difficult due to the global interpreter lock, and is typically only achieved by spawning subprocesses or by offloading..."

examples/minimal/datapipes/README.md
```diff
@@ -0,0 +1,209 @@
+# PhysicsNeMo DataPipes
+
+Dataloading is critical to SciML applications, both for training and inference,
+and the physicsnemo datapipe infrastructure aims to deliver a flexible and configurable
+set of tools to enable your application.
+
+There are plenty of tools in the python eco system for loading, preprocessing, and
+preparing your data for training or inference.  To compare / contrast some of these
+tools with the ecosystem available, and see if the physicsnemo datapipe interface
+might be valuable to your workload, consider the following design principles
```
laserkelvin, line 16: Could we define the scope of a datapipe here, since it's the introductory passage?

examples/minimal/datapipes/README.md
```diff
@@ -0,0 +1,209 @@
...
+3. **Unambiguous Configuration and Serialization** - Datapipes can be a particularly
+frustrating component in reproducibility of AI results - the preprocessing, sampling,
+batching and other parameters can be hard to infer from training scripts.  Here,
+we make a deliberate design choice to enable datapipe configuration serialization
+as a first-class citizen.  PhysicsNeMo Datapipes can be built directly in Python,
+but also instantiated from hydra yaml files for version control and distribution.
+
+4. **Familiar Interfaces** - We built our tools from scratch, but they are meant
+to look familiar and inter-operate with the tools you already know.  Use
+physicsnemo DataLoaders as a replacement for PyTorch's Dataloader; tools like
```
laserkelvin: PhysicsNeMo

examples/minimal/datapipes/README.md
```diff
@@ -0,0 +1,209 @@
...
+concurrency via asynchronous execution and stream-based parallelism.  IO is coordinated
+in multiple threads, instead of multiple processes, and streams enable multiple
+preprocessing pipelines to execute concurrently on the GPU.
+
+3. **Unambiguous Configuration and Serialization** - Datapipes can be a particularly
+frustrating component in reproducibility of AI results - the preprocessing, sampling,
+batching and other parameters can be hard to infer from training scripts.  Here,
+we make a deliberate design choice to enable datapipe configuration serialization
+as a first-class citizen.  PhysicsNeMo Datapipes can be built directly in Python,
+but also instantiated from hydra yaml files for version control and distribution.
```
laserkelvin: `hydra` YAML

examples/minimal/datapipes/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+4. **Familiar Interfaces** - We built our tools from scratch, but they are meant
+to look familiar and inter-operate with the tools you already know.  Use
+physicsnemo DataLoaders as a replacement for PyTorch's Dataloader; tools like
+DistributedSampler will still work. Users of `torchvision` will be familiar
```
laserkelvin: `DistributedSampler`

examples/minimal/datapipes/README.md
```diff
@@ -0,0 +1,209 @@
...
+physicsnemo datapipe is extensible: you can build custom data readers for
+new dataformats, and plug them in to datasets; you can build new transforms
+for your data that we might not have, and simply plug them into a transformation
+pipeline. You can even package all of this up as a pip-installable extension: Using
+the built in registry enables you to still instantiate and version control datapipes,
+when the components are not even part of PhysicsNeMo.
+
+## When should I use PhysicsNeMo datapipes over X/Y/Z data utility?
+
+In general, the physicsnemo datapipe utility is built to deliver good performance
```
laserkelvin: PhysicsNeMo

examples/minimal/datapipes/README.md
```diff
@@ -0,0 +1,209 @@
...
+consists of independent **examples**.  Each example has one or more **tensor components**:
+image data may have input images and target labels; CFD data may have positions,
+target pressures, a mesh object, boundary conditions, etc.; weather data may contain
+sensor readings as a function of time.  Each example may be the same size as the others,
+or each example may be a unique size.  Even the components of an example can be variable,
+though this can require extra care in reading and using the dataset.
+
+The PhysicsNeMo datapipe consists of the following components:
+
+- `reader` objects contain the logic to understand a **dataset** on disk, and
```
laserkelvin: To be a bit pedantic/clarifying, here objects are referring to classes right? i.e. `Reader`/`Dataset`/`Transform`

Objects would be the concrete, but here the components being described are the abstract right?

examples/minimal/datapipes/README.md
```diff
@@ -0,0 +1,209 @@
...
+### Tutorial 2: Transforms and Data Preprocessing
+
+**File:** `tutorial_02_transforms.py`
+
+Build preprocessing pipelines with transforms:
+
+- Apply a single transform (Normalize)
+- Compose multiple transforms together
+- Subsample point clouds with SubsamplePoints
+- Use geometric transforms (Translate, ReScale)
```
laserkelvin: Rename to `Rescale`

examples/minimal/datapipes/conf/reader/npz.yaml
laserkelvin, line 1: This is a license header only file?

examples/minimal/datapipes/generate_regular_data.py
laserkelvin, line 1: Given that `uv` is the main environment manager now, how about making this script standalone? Then we wouldn't have to have the `try/except` for catching if `zarr` isn't installed.

```python
#!/usr/bin/env -S uv run --script
#
# /// script
# dependencies = ["numpy", "zarr"]
# ///
```

We could choose to also pin the versions as well to make sure the examples will work out-of-the-box

examples/minimal/datapipes/tutorial_01_getting_started.py
laserkelvin: How about as a `marimo` notebook/script? Best of both worlds :)

physicsnemo/datapipes/core/dataset.py
```diff
@@ -0,0 +1,384 @@
...
+        >>>
+        >>> dataset = Dataset(reader, transforms=transforms)
+        >>> sample = dataset[0]
+    """
+
+    def __init__(
+        self,
+        reader: Reader,
+        *,
+        transforms: Optional[Union[Transform, Sequence[Transform]]] = None,
```
laserkelvin: Prefer `Union` or `|`?

physicsnemo/datapipes/core/readers/base.py
```diff
@@ -0,0 +1,226 @@
...
+
+        # Create TensorDict
+        data = TensorDict(data_dict, device=torch.device("cpu"))
+
+        return data, metadata
+
+    def __iter__(self) -> Iterator[tuple[TensorDict, dict[str, Any]]]:
+        """Iterate over all samples."""
+        for i in range(len(self)):
+            yield self[i]
```
laserkelvin: I don't know if this will incur a big performance hit, but IMO it might be good to add exception handling here, even if it's just "Sample failed at index {i} for reason:" to help with debugging since we don't expect users/developers to override this method

physicsnemo/datapipes/core/readers/zarr.py
```diff
@@ -0,0 +1,416 @@
...
+            field_name: Name of the field (for error messages).
+
+        Returns:
+            A torch.Tensor containing the attribute value.
+
+        Raises:
+            TypeError: If the attribute value cannot be converted to a tensor.
+        """
+        try:
+            if isinstance(value, np.ndarray):
```
laserkelvin: Could we write this as `match` instead?

physicsnemo/datapipes/readers/zarr.py
```diff
@@ -0,0 +1,416 @@
...
+        """
+        total_points = slice_stop - slice_start
+
+        if total_points < n_points:
+            raise ValueError(
+                f"Slice size {total_points} is less than the number of points "
+                f"{n_points} requested for subsampling"
+            )
+
+        start = np.random.randint(slice_start, slice_stop - n_points + 1)
```
laserkelvin, line 283: Random state?

v2.0-MIGRATION-GUIDE.md
```diff
@@ -53,7 +55,13 @@ is still a viable installation method.
 
 ## PhysicsNeMo Datapipes
 
-[TBD]
+PhysicsNeMo DataPipes is a GPU-first, high-performance data loading
```
laserkelvin, line 58: So as part of the migration guide, I feel like this block should be written more as a "I'm in 2.0 now, how should I use `datapipes` and where should I look first"

test/datapipes/core/conftest.py
```diff
@@ -0,0 +1,187 @@
...
+    return path
+
+
+# ============================================================================
+# Device Fixtures
+# ============================================================================
+
+
+@pytest.fixture
+def device():
```
laserkelvin: You don't want to test on both?

```python
@pytest.fixture(params=["cpu", "cuda"], ids=["cpu", "gpu"])
def device():
    device_name = request.param
    if device_name == "cuda" and not torch.cuda.is_available():
        pytest.skip("CUDA  not available")
    return device_name
````

laserkelvin (changes requested):
I've left some general comments, particularly on the write ups.

The main point I want to bring up in my review is I feel like the unit tests for readers in particular could do with some negative cases, just to mark what is and what isn't supported (e.g. deviations from the expected schema for each reader). I would have also preferred fuzz testing for data in general, as well, and not rely on testing against specific shapes and keys.

physicsnemo/datapipes/core/transforms/geometric.py
coreyjadams: Done.

physicsnemo/datapipes/transforms/normalize.py
coreyjadams: Done.

examples/minimal/datapipes/README.md
coreyjadams: Yeah I dropped the "object" word to pacify you :)

examples/minimal/datapipes/README.md
coreyjadams: It's changed to `Scale`

examples/minimal/datapipes/conf/reader/npz.yaml
coreyjadams: No it's a missing implementation, good catch!

physicsnemo/datapipes/core/dataset.py
coreyjadams: Yeah done.

physicsnemo/datapipes/core/readers/base.py
coreyjadams: Ok, I wrapped in in a try / except to share info about the exception.  We'll make it configurable if it's too slow.

test/datapipes/core/conftest.py
coreyjadams: Thanks for catching this, this shouldn't be here.

physicsnemo/datapipes/transforms/base.py
peterdsharpe: Yeah, actually the more I think about it I think this current pattern of "configure at init rather than at call" is probably the right abstraction here - this makes arbitrary composition easier.

laserkelvin (approved):

coreyjadams:
For posterity:

I have flattened the structure (but made no changes) and the new code is at physicsnemo/datapipes instead of .../core.  Examples updated, tests updated.

I also had to update zarr to 3.0 for CI, however, that will get moved to an optional dep in a subsequent PR shortly.

PR #1306, opened 2026-01-07
Refactor rnn models for the new standards

ktangsali:
# PhysicsNeMo Pull Request

## Description
Refactors the RNN models to the new coding standards. 

## Checklist

## Review Process


ktangsali:
> ### Additional Comments (3)
> 1. `physicsnemo/models/rnn/rnn_one2many.py`, line 100-103 ([link](/nvidia/physicsnemo/blob/e5adc09e73950fbc717d87f32471eb1e6f49500e/physicsnemo/models/rnn/rnn_one2many.py#L100-L103))
>    **syntax:** output shape mismatch in docstring example - the example shows input `torch.randn(4, 6, 1, 16, 16)` which is [N=4, C=6, T=1, H=16, W=16], but the output shows `torch.Size([4, 6, 16, 16, 16])` which is [N=4, C=6, T=16, H=16, W=16]. the spatial dimensions should remain [16, 16], not become [16, 16, 16]
> 2. `physicsnemo/nn/conv_layers.py`, line 258-263 ([link](/nvidia/physicsnemo/blob/e5adc09e73950fbc717d87f32471eb1e6f49500e/physicsnemo/nn/conv_layers.py#L258-L263))
>    **logic:** padding order is incorrect for 2D input - the current code applies padding as `[pad_h // 2, pad_h - pad_h // 2, pad_w // 2, pad_w - pad_w // 2]` but F.pad expects padding in reverse dimension order: `[left, right, top, bottom]` which should be `[pad_w // 2, pad_w - pad_w // 2, pad_h // 2, pad_h - pad_h // 2]`
> 3. `physicsnemo/nn/conv_layers.py`, line 432-437 ([link](/nvidia/physicsnemo/blob/e5adc09e73950fbc717d87f32471eb1e6f49500e/physicsnemo/nn/conv_layers.py#L432-L437))
>    **logic:** padding/cropping order is incorrect for 2D input - torch.nn.functional.pad expects padding in reverse dimension order. current code uses `[pad_h // 2 : ..., pad_w // 2 : ...]` but the dimensions should be reversed to match width then height
>    can you verify the expected padding order matches your framework's conventions for transpose convolutions?
> 
> 7 files reviewed, 3 comments
> 
> [Edit Code Review Agent Settings](https://app.greptile.com/review/github) | [Greptile](https://greptile.com?utm_source=greptile_expert&utm_medium=github&utm_campaign=code_reviews&utm_content=physicsnemo_1306)

Fixed point 2 and 3. For point 1, the output is correct. [4, 6, 16, 16, 16] means batch=4, channels=6, timesteps=16, height=16, width=16

coreyjadams:
I think this looks good to me.

When UNet is updated (Which is currently assigned to @peterdsharpe ...) we might consider consolidating some of the "ConvBlock" and "ConvLayer" etc implementation.  But this one came first so no need to hold it up :)

test/models/rnn/test_rnn.py
CharlelieLrt, line 1: IMO there are a few things missing in the current tests:

- The tests for the constructor should check the values of the public attributes (e.g. for `model = One2ManyRNN(...)`, the values of `model.nr_tsteps`, `model.nr_residual_blocks`, etc...)

- There should be a test that loads a checkpoint file. Unlike existing tests using checkpoints, the checkpoint file *should be part of the test suite*. It **should not** be created at runtime, otherwise we are not testing the ability to load checkpoints generated with older version of the codebase, which is the principal source of failure for loading checkpoints.

physicsnemo/nn/conv_layers.py
```diff
@@ -282,3 +310,671 @@ def forward(self, x):
...
+                self.kernel_size,
+                self.stride,
+                bias=True,
+            )
+        else:
+            raise ValueError("Only 1D, 2D and 3D dimensions are supported")
+
+        self.reset_parameters()
+
+    def exec_activation_fn(
```
CharlelieLrt: Is this method expected to be externally exposed or is it for internal use? In the latter case it should be semi-private. This applies to all semi-private methods and attributes in these layers (e.g. `reset_parameters` below, etc.)

test/models/rnn/test_rnn_layers.py
CharlelieLrt, line 1: The tests here only check the shape of the outputs. Would it be possible to have some non-regression tests that also check the values with `validate_accuracy`, the same way as it is done for the models?

ktangsali: Addressed in [e2cf412](https://github.com/NVIDIA/physicsnemo/pull/1306/commits/e2cf4122618520ef3284e27bfdd4efdedd0021e3)

physicsnemo/nn/conv_layers.py
ktangsali: Addressed in [7541105](https://github.com/NVIDIA/physicsnemo/pull/1306/commits/7541105265b97122099a7908eec135a4e332d7d0)

test/models/rnn/test_rnn.py
ktangsali: Addressed in [7541105](https://github.com/NVIDIA/physicsnemo/pull/1306/commits/7541105265b97122099a7908eec135a4e332d7d0)

coreyjadams:
> There should be a test that loads a checkpoint file. Unlike existing tests using checkpoints, the checkpoint file should be part of the test suite.

@CharlelieLrt I like this idea a lot, but I'm thinking about having [~15 models] x [ checkpoint or two each] x [several MB per checkpoint or more, per model] leading to repo bloat.  Do we want to keep these checkpoint files with `git lfs` or perhaps a separate repo we can use as a `git submodule` during testing?

CharlelieLrt:
@coreyjadams agree the size of the combined .mdlus checkpoints could be a problem. When I did that for other tests I made sure to use very tiny checkpoints that were all around 10KB to 100KB. I'm not sure if that would more acceptable (or even possible for some models). 

> a separate repo we can use as a git submodule during testing?

Agree, that would be a good solution

CharlelieLrt (approved):
LGTM!

ktangsali:
> @coreyjadams agree the size of the combined .mdlus checkpoints could be a problem. When I did that for other tests I made sure to use very tiny checkpoints that were all around 10KB to 100KB. I'm not sure if that would more acceptable (or even possible for some models).
> 
> > a separate repo we can use as a git submodule during testing?
> 
> Agree, that would be a good solution

I like the idea of a separate repo for assets too. Let's address creation of a different repo for assets in a separate PR though! For now I have modified the checkpoints to be in the order of 10s of KBs.

coreyjadams:
@ktangsali @CharlelieLrt  sounds good - we'll aim for a test-data repo or similar we can use for testing (I know there is already a private one) but since this isn't really a "functionality" issue it might have to wait a bit.

PR #1316, opened 2026-01-08
Update transolver to comply with model standards

coreyjadams:
# PhysicsNeMo Pull Request

## Checklist

## Review Process

### DIFF
--- a/physicsnemo/diffusion/preconditioners/preconditioners.py
+++ b/physicsnemo/diffusion/preconditioners/preconditioners.py
@@ -321,6 +321,33 @@ class BaseAffinePreconditioner(Module, ABC):
     >>> out = wrapped_dit(x, t, condition)
     >>> out.shape
     torch.Size([1, 2, 8, 8])
+
+    **Example: Using ConcatConditionWrapper with a preconditioner**
+
+    The pattern in the previous example, where (spatially-varying) conditioning
+    is concatenated to the noised latent state (and possibly vector conditioning
+    is also passed as a separate argument) is common across several diffusion
+    use-cases.
+    Thus for convenience, we provide the wrapper class :class:`~physicsnemo.diffusion.utils.ConcatConditionWrapper`
+    to save you the trouble of writing your own wrapper for this common pattern:
+
+    >>> from physicsnemo.diffusion.preconditioners import EDMPreconditioner
+    >>> from physicsnemo.diffusion.utils import ConcatConditionWrapper
+    >>> base_model = SongUNet(img_resolution=8, in_channels=4, out_channels=3, label_dim=4)
+    >>> wrapped_model = ConcatConditionWrapper(base_model)
+    >>> precond = EDMPreconditioner(wrapped_model, sigma_data=0.5)
+    >>> x = torch.rand(1, 3, 8, 8)
+    >>> t = torch.rand(1)
+    >>> condition = TensorDict({
+    ...     "cond_concat": torch.rand(1, 1, 8, 8),  # image condition
+    ...     "cond_vec": torch.rand(1, 4),           # vector condition
+    ... }, batch_size=[1])
+    >>> out = precond(x, t, condition)
+    >>> out.shape
+    torch.Size([1, 3, 8, 8])
+
+    The same wrapper can be used with :class:`~physicsnemo.experimental.models.dit.DiT`
+    backbones as well, with ``cond_vec`` passed to the model's ``condition`` argument.
     """
 
     def __init__(
--- a/physicsnemo/diffusion/utils/__init__.py
+++ b/physicsnemo/diffusion/utils/__init__.py
@@ -14,4 +14,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from .model_wrappers import ConcatConditionWrapper
 from .utils import InfiniteSampler, StackedRandomGenerator
--- a/physicsnemo/diffusion/utils/model_wrappers.py
+++ b/physicsnemo/diffusion/utils/model_wrappers.py
@@ -1,0 +1,259 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import warnings
+from typing import Any
+
+import torch
+from jaxtyping import Float
+from tensordict import TensorDict
+
+from physicsnemo.core import Module
+from physicsnemo.models.diffusion_unets import (
+    DhariwalUNet,
+    SongUNet,
+    SongUNetPosEmbd,
+    SongUNetPosLtEmbd,
+)
+
+
+class ConcatConditionWrapper(Module):
+    r"""
+    Wrapper that handles channel-concatenated conditioning in addition to
+    optional vector conditioning.
+
+    This wrapper adapts backbones with different conditioning signatures to the
+    :class:`~physicsnemo.diffusion.DiffusionModel` interface by concatenating
+    image-like conditions to ``x`` and routing vector conditions to the
+    appropriate argument. It is intended for cases where image-like conditioning
+    (i.e., conditioning with the same spatial dimensions as ``x``) should be
+    concatenated along the channel dimension to the noised latent state.
+
+    Externally, wrapping with this wrapper will allow a backbone with an
+    incompatible conditioning signature to satisfy the :class:`~physicsnemo.diffusion.DiffusionModel`
+    interface. Currently, the wrapper supports the following backbones:
+
+     - :class:`~physicsnemo.models.diffusion_unets.SongUNet`,
+     - :class:`~physicsnemo.models.diffusion_unets.SongUNetPosEmbd`,
+     - :class:`~physicsnemo.models.diffusion_unets.SongUNetPosLtEmbd`,
+     - :class:`~physicsnemo.models.diffusion_unets.DhariwalUNet`,
+     - :class:`~physicsnemo.experimental.models.dit.DiT`.
+     - Any backbone with an forward signatude matching that of the `DiT` model:
+       ``model(x, t, condition=None, **model_kwargs)`` where ``condition`` is a
+       tensor of shape :math:`(B, d)` and ``**model_kwargs`` are additional
+       keyword arguments forwarded to the backbone.
+
+    The wrapper supports conditioning passed as either a ``TensorDict`` or a
+    ``torch.Tensor``. If a ``TensorDict`` is passed, it must use the keys
+    ``cond_concat`` and ``cond_vec`` to identify the image and vector
+    conditioning tensors, respectively, or the user may specify the expected
+    keys using the ``image_cond_key`` and ``vector_cond_key`` arguments when
+    instantiating the wrapper.
+
+    If a ``torch.Tensor`` is passed, it is treated as the image conditioning
+    input to be concatenated (i.e., treated like the value of ``cond_concat``
+    when a ``TensorDict`` is passed).
+
+    The wrapper will route arguments approporiately depending on the backbone
+    type. The default behavior for unknown backbones is to concatenate the value
+    of ``cond_concat`` (if provided) and pass the value of ``cond_vec`` as the
+    ``condition`` keyword argument.
+
+    Parameters
+    ----------
+    model : physicsnemo.Module
+        Backbone model to wrap.
+    image_cond_key : str, optional
+        TensorDict key for the image conditioning tensor to concatenate to
+        ``x`` along channels, by default ``"cond_concat"``.
+    vector_cond_key : str, optional
+        TensorDict key for the vector conditioning tensor to pass through to
+        the backbone, by default ``"cond_vec"``.
+
+    Forward
+    -------
+    x : torch.Tensor
+        Noisy latent state of shape :math:`(B, *)` where :math:`B` is the
+        batch size.
+    t : torch.Tensor
+        Diffusion time tensor of shape :math:`(B,)`.
+    condition : torch.Tensor, TensorDict, or None, optional, default=None
+        Conditioning data. Use a ``TensorDict`` to explicitly supply
+        concatenated conditioning and vector conditioning as keys
+        ``cond_concat`` and ``cond_vec`` (or the custom keys
+        specified by ``image_cond_key`` and ``vector_cond_key``).
+        Alternately, supply a plain ``Tensor`` input to be concatenated.
+    **model_kwargs : Any
+        Additional keyword arguments forwarded to the backbone.
+
+    Outputs
+    -------
+    torch.Tensor
+        Model output with the same shape as ``x``.
+
+    Examples
+    --------
+    Wrap :class:`~physicsnemo.models.diffusion_unets.SongUNet` with channel
+    concatenation and vector conditioning:
+
+    >>> import torch
+    >>> from tensordict import TensorDict
+    >>> from physicsnemo.models.diffusion_unets import SongUNet
+    >>> net = SongUNet(in_channels=4, out_channels=3, label_dim=4, img_resolution=8)
+    >>> wrapper = ConcatConditionWrapper(net)
+    >>> x = torch.randn(2, 3, 8, 8)
+    >>> t = torch.rand(2)
+    >>> condition = TensorDict(
+    ...     {
+    ...         "cond_concat": torch.randn(2, 1, 8, 8),
+    ...         "cond_vec": torch.randn(2, 4),
+    ...     },
+    ...     batch_size=[2],
+    ... )
+    >>> out = wrapper(x, t, condition)
+    >>> out.shape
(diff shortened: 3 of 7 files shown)


physicsnemo/models/transolver/__init__.py
```diff
@@ -28,6 +36,49 @@
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
+
+References
+----------
+- `Transolver paper <https://arxiv.org/pdf/2402.02366>`_
+- `Transolver++ paper <https://arxiv.org/pdf/2502.02414>`_
```
CharlelieLrt: AFAIK this "References" section will not be recognized and rendered properly in the online docs. Also, IMO this big docstring with examples and so on, should not be `__init__.py` but either in the class docstring, or in the `docs/` directory (at leats it's what other models do)

physicsnemo/models/transolver/Embedding.py
CharlelieLrt, line 1: Why is this file (and a few others) named with upper case (e.g. `Embedding.py`, and `Physics_Attention.py` looks even worse)?

CharlelieLrt, line 1: We already have some embeddings in `physicsnemo/nn`, so would this makes sense to move the content of this module to there?

physicsnemo/models/transolver/Embedding.py
```diff
@@ -35,89 +40,304 @@
...
 
+        # Compute frequencies via outer product
+        freqs = torch.einsum("... i , j -> ... i j", t, self.inv_freq)  # (B, N, D//2)
 
-def rotate_half(x):
+        # Concatenate to get full dimension
+        return torch.cat((freqs, freqs), dim=-1)  # (B, N, D)
+
+
+def rotate_half(x: Float[torch.Tensor, "... dim"]) -> Float[torch.Tensor, "... dim"]:
```
CharlelieLrt: Question about this `rotate_half`, and a few other functions defined in this module: are they meant to be exposed or are they just some internal helper functions for this module? If the former case, would it make sense to make them semi-private?

physicsnemo/models/transolver/Embedding.py
```diff
@@ -35,89 +37,304 @@
...
         self.register_buffer("inv_freq", inv_freq)
 
-    def forward(self, coordinates, device):
-        # coordinates [b, n]
-        t = coordinates.to(device).type_as(self.inv_freq)
+    def forward(
+        self,
+        coordinates: Float[torch.Tensor, "batch seq"],
+        device: torch.device,
+    ) -> Float[torch.Tensor, "batch seq dim"]:
```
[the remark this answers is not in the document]
CharlelieLrt: That's true, but shape validation for every single layer might be an overkill. Shape validation as detailed in MOD-005 is more critical for larger models comprising multiple layers.

physicsnemo/nn/module/physics_attention.py
CharlelieLrt, line 1: Same question as for the embeddings: we have a specific module in `physicsnemo/nn`. Would it make sense to move the content of this module to there?

physicsnemo/models/transolver/Physics_Attention.py
```diff
@@ -45,52 +55,97 @@
 else:
     te = None
 
-from einops import rearrange
-from torch.autograd.profiler import record_function
-from torch.distributed.tensor.placement_types import Replicate
-
-from physicsnemo.domain_parallel import ShardTensor
 
+def gumbel_softmax(
```
CharlelieLrt: I used this gumbel_softmax at least twice in other projects and I had to reimplement it every time... Might be worth moving to `physicsnemo.nn`

physicsnemo/models/transolver/Physics_Attention.py
```diff
@@ -45,52 +55,97 @@
...
-        torch.Tensor: The Gumblel Softmax of the logits.
+    Parameters
+    ----------
+    logits : torch.Tensor
+        Input logits tensor of shape :math:`(*, K)` where :math:`K` is the
+        number of categories.
+    tau : torch.Tensor | float, optional, default=1.0
+        Temperature parameter. Can be a scalar float or a tensor for
+        per-element temperature. Lower values make the distribution more
+        concentrated.
```
CharlelieLrt: `or a tensor foror a tensor for per-element temperature`
This is not super clear to me. Does that mean it should have same shape as `logits`?

physicsnemo/nn/module/physics_attention.py
```diff
@@ -152,63 +214,84 @@ def __init__(
...
+
+        Returns
+        -------
+        torch.Tensor | tuple[torch.Tensor, torch.Tensor]
+            For Transolver++: single projected tensor of shape
+            :math:`(B, N, H, D_h)`.
+            For standard Transolver: tuple of (x_mid, fx_mid) both of shape
+            :math:`(B, N, H, D_h)`.
         """
         raise NotImplementedError("Subclasses must implement this method")
```
CharlelieLrt: This seems like an anti-pattern to me. This is already an `abstractmethod` in an `ABC` class, so there is no point in this `NotImplementedError`. Also, if it is a true ABC class then it should not even be possible to instantiate it at all

physicsnemo/nn/module/physics_attention.py
```diff
@@ -152,63 +214,84 @@ def __init__(
...
-
-        Notes:
-            - The function first computes a temperature-scaled softmax over the slice projections to obtain slice weights.
-            - It then aggregates the latent features (fx) for each slice using these weights.
-            - The aggregated features are normalized by the sum of weights for numerical stability.
+        self,
+        slice_projections: Float[torch.Tensor, "batch tokens heads slice_num"],
+        fx: Float[torch.Tensor, "batch tokens heads head_dim"],
+    ) -> tuple[
+        Float[torch.Tensor, "batch tokens heads slice_num"],
```
CharlelieLrt: I think we agreed on a convention that batch size should be B (?) At least I have seen multiple models using B in the jaxtyping annotation. I think it's better if we avoid any kind awkward mixup here.
I would also make sure that all annotations in this PR are consistent. If I'm not mistaken I have seen both tokens and N for the sequence length.
Ideally it would be great if we could in addition be consistent beween jaxtyping annotations and docstrings.

physicsnemo/nn/module/physics_attention.py
```diff
@@ -152,63 +214,84 @@ def __init__(
         self.out_dropout = nn.Dropout(dropout)
 
     @abstractmethod
-    def project_input_onto_slices(self, x) -> tuple[torch.Tensor, torch.Tensor]:
-        """
-        Project the input onto the slice space.
+    def project_input_onto_slices(
```
CharlelieLrt, line 183: This method and the few others that follow look like they should be semi-private

physicsnemo/nn/module/physics_attention.py
```diff
@@ -45,52 +55,97 @@
...
+    heads : int
+        Number of attention heads.
+    dim_head : int
+        Dimension per attention head.
+    dropout : float
+        Dropout rate.
+    slice_num : int
+        Number of physics slices.
+    use_te : bool
+        Whether to use transformer engine.
```
CharlelieLrt, line 92: And what happens if I pass `use_te=True` but TE is not available?

physicsnemo/nn/module/physics_attention.py
```diff
@@ -45,52 +55,97 @@
...
+    # Add noise and apply temperature-scaled softmax
     y = logits + gumbel_noise
     y = y / tau
-
     y = torch.nn.functional.softmax(y, dim=-1)
 
     return y
 
 
 class PhysicsAttentionBase(nn.Module, ABC):
```
CharlelieLrt, line 63: Note: I am getting a lot of F722 errors with jaxtyping. Is there a way to silence those? I thought we should add a blank space at the begiining: `" batch tokens heads head_dim"`, but I haven't trie that yet.

physicsnemo/nn/module/physics_attention.py
```diff
@@ -283,107 +384,151 @@ def compute_slice_attention_sdpa(self, slice_tokens: torch.Tensor) -> torch.Tens
...
         """
-        Forward pass of the Physics Attention module.
-
-        Input x should have shape of [Batch, N_tokens, N_Channels] ([B, N, C])
-        """
-
-        # Project the inputs onto learned spaces:
+        # Input validation (skip during torch.compile)
+        if not torch.compiler.is_compiling():
+            if x.ndim != 3:
```
CharlelieLrt, line 412: Is this the only requirement on the shape? I thought `x.shape[2]` should match the `dim` provided in the `__init__`, is that correct?

physicsnemo/models/transolver/transolver.py
```diff
@@ -71,75 +70,195 @@
 
 
 class MLP(nn.Module):
```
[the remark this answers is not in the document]
CharlelieLrt: It's for internal use in the module, so it's fine if it does not inherit from `physicsnemo.Module`. MOD-001 is for models that are directly exposed, it's not critical to impose this requirement for those that are not exposed like helpers and so on.

physicsnemo/models/transolver/transolver.py
```diff
@@ -71,75 +75,195 @@
...
```
CharlelieLrt: Another MLP? Can it be replaced with `FullyConnected` or one that we already have? (maybe add a TE option in the existing FullyConnected?)

physicsnemo/models/transolver/transolver.py
```diff
@@ -71,75 +70,195 @@
...
-                nn.Sequential(linear_layer(n_hidden, n_hidden), act())
+                nn.Sequential(linear_layer(n_hidden, n_hidden), act_fn())
                 for _ in range(n_layers)
             ]
         )
 
-    def forward(self, x):
+    def forward(
+        self, x: Float[torch.Tensor, "... d_in"]
+    ) -> Float[torch.Tensor, "... d_out"]:
```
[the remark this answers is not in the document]
CharlelieLrt: That's true, but not critical to have shape validation for internal models that are not exposed

physicsnemo/models/transolver/transolver.py
```diff
@@ -71,75 +75,195 @@
...
                 x = self.linears[i](x) + x
             else:
                 x = self.linears[i](x)
+
+        # Project to output dimension
         x = self.linear_post(x)
         return x
 
 
 class Transolver_block(nn.Module):
```
CharlelieLrt: Should be `TransolverBlock`. The mixture of Pascal case and snake case is not allowed

physicsnemo/models/transolver/transolver.py
```diff
@@ -334,53 +515,56 @@ def __init__(
         self.__name__ = "Transolver"
```
CharlelieLrt: Do you know what is this for?

physicsnemo/models/transolver/Embedding.py
```diff
@@ -35,89 +40,304 @@
...
     x = rearrange(x, "... (j d) -> ... j d", j=2)
     x1, x2 = x.unbind(dim=-2)
     return torch.cat((-x2, x1), dim=-1)
 
 
-def apply_rotary_pos_emb(t, freqs):
+def apply_rotary_pos_emb(
+    t: Float[torch.Tensor, "... dim"],
+    freqs: Float[torch.Tensor, "... dim"],
+) -> Float[torch.Tensor, "... dim"]:
```
CharlelieLrt: I am not sure the ellipsis `...` is 100% correct in the jaxtyping there? The dimensions should be the same between all 3 tensors, right?

physicsnemo/models/transolver/Embedding.py
coreyjadams: Historical reasons, the files came in like this externally.  Now is a good time to fix.

physicsnemo/models/transolver/Embedding.py
coreyjadams: I honestly think they are not much used.  I will make them semi-private, good idea.

physicsnemo/models/transolver/Physics_Attention.py
coreyjadams: Yeah it's probably worth a fusion with warp or triton too.  I'll move it.

physicsnemo/models/transolver/Physics_Attention.py
coreyjadams: It means it can have the same shape as logits, OR it can be a scalar.  To be honest I don't know if there are use cases where it is a tensor ... I think going "scalar only" in the docstring is the right choice.  Probably no need to change the code.

physicsnemo/nn/module/physics_attention.py
coreyjadams: This is meant to kill those: https://github.com/NVIDIA/physicsnemo/blob/main/pyproject.toml#L209

But, I think that merged after this PR, it might not be here?  I'll update against main and check locally.

physicsnemo/nn/module/physics_attention.py
coreyjadams: Good catch.  It's protected higher up in teh model but not here.  Since PhysicsAttention will move to modules, I'll add that check here too.

https://github.com/NVIDIA/physicsnemo/blob/main/physicsnemo/models/transolver/transolver.py#L132-L135

physicsnemo/models/transolver/transolver.py
coreyjadams: Removed!

physicsnemo/models/transolver/transolver.py
coreyjadams: I've added an option to the MLP we have in `physicsnemo.nn.module.mlp_layers` to use transformer_engine.  Unfortunately, I have to keep a thin wrapper of `_TransformerMLP` which inherits from `Mlp` - this is needed to map statedict keys and keep checkpoint compatibility.  But, the implementation is the NN one.  Maybe someday we can deprecate it.

physicsnemo/models/transolver/Embedding.py
coreyjadams: Embedding has been removed anyways.  So, not an issue.

physicsnemo/models/transolver/transolver.py
coreyjadams: Fixed.

physicsnemo/models/transolver/Embedding.py
coreyjadams: I ended up removing them.  they are not used anywhere in the implementations or examples as far as I can tell.

physicsnemo/nn/module/physics_attention.py
coreyjadams: I was able to get the same functionality as the TimeEmbedding in transolver out of PositionalEmbedding from physicsnemo.  So I removed the embeddings.py file entirely.

physicsnemo/models/transolver/Physics_Attention.py
coreyjadams: I moved it and added a Layer for it too.  So, we have `gumbel_softmax` and `GumbelSoftmax`.  I didn't move the implementation to the functional code yet, let me know if you think that is necessary.

physicsnemo/nn/module/physics_attention.py
coreyjadams: I left this one not semi-private, since it's meant to be inherited.  I can make the others semi private.

physicsnemo/nn/module/physics_attention.py
coreyjadams: Fixed.

physicsnemo/nn/module/physics_attention.py
coreyjadams: Added.

physicsnemo/models/transolver/Physics_Attention.py
CharlelieLrt: Okay, got it! I think my initial confusion was about its shape, but adding the shape requirement (must match that of `logits`) in the docstring and/or in the jaxtyping annotation would be a sufficient clarification.

> To be honest I don't know if there are use cases where it is a tensor


Maybe there are. Let's keep this functionality

coreyjadams: In the end I changed the docstring and didn't mention tensor, but type hints and functionality allow it just fine.

physicsnemo/nn/module/gumbel_softmax.py
```diff
@@ -0,0 +1,123 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+import torch.nn as nn
+from jaxtyping import Float
+
+
+def gumbel_softmax(
+    logits: Float[torch.Tensor, "... num_categories"],
```
CharlelieLrt, line 23: If I want to be really _**rigid**_, the jaxtyping should be:

```python
logits: Float[torch.Tensor, "*dims num_categories"]
tau: Float[torch.Tensor, "*dims num_categories"] | float = 1.0
 return ->  Float[torch.Tensor, "*dims num_categories"]
```

physicsnemo/nn/module/gumbel_softmax.py
```diff
@@ -0,0 +1,123 @@
...
+
+    # Add noise and apply temperature-scaled softmax
+    y = logits + gumbel_noise
+    y = y / tau
+    y = torch.nn.functional.softmax(y, dim=-1)
+
+    return y
+
+
+class GumbelSoftmax(nn.Module):
```
CharlelieLrt, line 60: Disclaimer: I am getting more and more confused with this

But, did we explicitly forbid direct subclass of `torch.nn.Module`, and decided that everything should subclass `physicsnemo.nn.Module`

coreyjadams: I have no idea what we settled on anymore.  I personall think for things in module its fine to use torch.nn.Module.  For models its mandatory to use physicsnemo.Module.

... at this point I don't think it makes a huge difference in modules. 🤷

PR #1317, opened 2026-01-08
Diffusion preconditioners refactor

### LINKED ISSUE #796: 🚀[FEA]: Allow passing models as to EDMPrecond

Improvement

Medium

Currently models are passed as names (strings) and splatted arguments. This makes it confusing to determine which arguments get passed to the model and which go to the preconditioner. It would be nice to also allow just passing an initialized model itself. The downside is that the user would be required to determine the correct input and output channels, but I don't think this is too onerous (and it could also be checked when the preconditioner is initialized.

_No response_

CharlelieLrt:
# PhysicsNeMo Pull Request

## Overview

This PR refactors the diffusion model preconditioners to introduce a clean, extensible architecture. The goal is to standardize how preconditioners wrap neural network models and apply the preconditioning formula, making it easier to implement new preconditioning schemes and maintain existing ones.

- **New `BasePreconditioner` abstract base class** in `physicsnemo/diffusion/preconditioners/preconditioners.py` that defines a standardized interface for wrapping diffusion models preconditioning. The base class handles the common forward pass logic while subclasses implement `compute_coefficients()` to define their specific preconditioning scheme. Subclasses can optionally override `sigma()` to implement custom noise schedules (time-to-noise mappings). A key improvement of this refactor is to enable dependency-injection design patterns by wrapping arbitrary `physicsnemo.Module` instance with a preconditioner.

- **Four preconditioner reimplementations** based on the new `BasePreconditioner`: `VPPreconditioner`, `VEPreconditioner`, `IDDPMPreconditioner`, and `EDMPreconditioner`. These are cleaner, standalone versions of the existing legacy preconditioners with comprehensive docstrings including mathematical formulas for the preconditioning coefficients and noise schedules.

- **Migrated legacy preconditioners** in `legacy.py` to inherit from the new base classes. This eliminates code duplication while maintaining full backward compatibility—all existing method signatures, attributes, and behaviors are preserved. Users of the legacy API do not need to change their code.

- **Comprehensive CI tests** in `test/diffusion/test_preconditioners.py` covering:
  - Constructor instantiation and attribute verification
  - `sigma()` and `compute_coefficients()` methods
  - Forward pass with non-regression testing against saved reference data
  - Checkpoint save/load roundtrip via `physicsnemo.Module.from_checkpoint()`

- Closes #796 .

## Checklist

## Review Process

### DIFF
--- a/physicsnemo/diffusion/preconditioners/preconditioners.py
+++ b/physicsnemo/diffusion/preconditioners/preconditioners.py
@@ -159,6 +159,7 @@ class BaseAffinePreconditioner(Module, ABC):
     We first define a simple model to wrap:
 
     >>> import torch
+    >>> from tensordict import TensorDict
     >>> from physicsnemo.nn import Module
     >>> class SimpleModel(Module):
     ...     def __init__(self, channels: int):
@@ -174,7 +175,7 @@ class BaseAffinePreconditioner(Module, ABC):
     >>> from physicsnemo.diffusion.preconditioners import (
     ...     BaseAffinePreconditioner,
     ... )
-    >>> class EDMPreconditioner(BaseAffinePreconditioner):
+    >>> class SimpleEDMPreconditioner(BaseAffinePreconditioner):
     ...     def __init__(self, model, sigma_data: float = 0.5):
     ...         super().__init__(model)
     ...         self.sigma_data = sigma_data
@@ -190,10 +191,11 @@ class BaseAffinePreconditioner(Module, ABC):
     ...         return c_in, c_noise, c_out, c_skip
     ...
     >>> model = SimpleModel(channels=3)
-    >>> precond = EDMPreconditioner(model, sigma_data=0.5)
+    >>> precond = SimpleEDMPreconditioner(model, sigma_data=0.5)
     >>> x = torch.randn(2, 3, 16, 16)
     >>> t = torch.rand(2)
-    >>> out = precond(x, t, condition=None)
+    >>> condition = TensorDict({}, batch_size=[2])
+    >>> out = precond(x, t, condition)
     >>> out.shape
     torch.Size([2, 3, 16, 16])
 
@@ -220,52 +222,8 @@ class BaseAffinePreconditioner(Module, ABC):
     ...         return c_in, c_noise, c_out, c_skip
     ...
     >>> precond_ve = VEPreconditioner(model)
-    >>> out_ve = precond_ve(x, t, condition=None)
+    >>> out_ve = precond_ve(x, t, condition)
     >>> out_ve.shape
-    torch.Size([2, 3, 16, 16])
-
-    The following example shows how to use the ``condition`` argument with a
-    conditional model. The model receives a TensorDict with keys ``"y"`` and
-    ``"z"`` as conditioning information.
-
-    >>> from tensordict import TensorDict
-    >>> class ConditionalModel(Module):
-    ...     def __init__(self, channels: int):
-    ...         super().__init__()
-    ...         self.channels = channels
-    ...         self.net = torch.nn.Conv2d(channels, channels, 1)
-    ...
-    ...     def forward(self, x, t, condition=None):
-    ...         # Use conditioning tensors from TensorDict
-    ...         if condition is not None:
-    ...             y = condition["y"]  # shape: (B, C, H, W)
-    ...             z = condition["z"]  # shape: (B, C, H, W)
-    ...             x = x + y + z
-    ...         return self.net(x)
-    ...
-    >>> class EDMPreconditionerCond(BaseAffinePreconditioner):
-    ...     def __init__(self, model, sigma_data: float = 0.5):
-    ...         super().__init__(model)
-    ...         self.sigma_data = sigma_data
-    ...
-    ...     def compute_coefficients(self, t: torch.Tensor):
-    ...         sigma_data = self.sigma_data
-    ...         c_skip = sigma_data**2 / (t**2 + sigma_data**2)
-    ...         c_out = t * sigma_data / (t**2 + sigma_data**2).sqrt()
-    ...         c_in = 1 / (sigma_data**2 + t**2).sqrt()
-    ...         c_noise = t.log() / 4
-    ...         return c_in, c_noise, c_out, c_skip
-    ...
-    >>> cond_model = ConditionalModel(channels=3)
-    >>> precond_cond = EDMPreconditionerCond(cond_model, sigma_data=0.5)
-    >>> x = torch.randn(2, 3, 16, 16)
-    >>> t = torch.rand(2)
-    >>> condition = TensorDict({
-    ...     "y": torch.randn(2, 3, 16, 16),
-    ...     "z": torch.randn(2, 3, 16, 16),
-    ... }, batch_size=[2])
-    >>> out_cond = precond_cond(x, t, condition)
-    >>> out_cond.shape
     torch.Size([2, 3, 16, 16])
 
     **Wrapping existing models to satisfy the DiffusionModel interface**
@@ -273,68 +231,96 @@ class BaseAffinePreconditioner(Module, ABC):
     Some models in PhysicsNeMo have signatures that differ from the
     :class:`~physicsnemo.diffusion.DiffusionModel` interface. Below are
     examples showing how to write thin wrappers to make them compatible
-    with preconditioners (and other diffusion utilities).
+    with preconditioners, including image-based conditioning via channel
+    concatenation.
 
-    **Example: Wrapping a diffusion UNet model**
+    **Example: Wrapping SongUNet**
 
     The :class:`~physicsnemo.models.diffusion_unets.SongUNet` model has
     the signature ``forward(x, noise_labels, class_labels, augment_labels)``.
-    We wrap it to match ``forward(x, t, condition)``:
+    We wrap it to match ``forward(x, t, condition)``, where ``condition``
+    contains both class labels (1D vector) and an image to concatenate
+    channel-wise:
 
     >>> from physicsnemo.models.diffusion_unets import SongUNet
     >>> from physicsnemo.diffusion import DiffusionModel
+    >>> from tensordict import TensorDict
     >>> class SongUNetWrapper(Module):
-    ...     def __init__(self, **song_unet_kwargs):
+    ...     def __init__(self, img_channels, cond_channels, label_dim, **kwargs):
     ...         super().__init__()
-    ...         self.net = SongUNet(**song_unet_kwargs)
+    ...         # in_channels = img_channels + cond_channels for concatenation
+    ...         self.net = SongUNet(
+    ...             in_channels=img_channels + cond_channels,
+    ...             out_channels=img_channels,
+    ...             label_dim=label_dim,
+    ...             **kwargs,
+    ...         )
     ...
-    ...     def forward(self, x, t, condition=None):
-    ...         # t -> noise_labels (diffusion time/noise level)
-    ...         # condition["class_labels"] -> class_labels
-    ...         class_labels = condition["class_labels"] if condition else None
-    ...         return self.net(x, noise_labels=t, class_labels=class_labels)
+    ...     def forward(self, x, t, condition):
+    ...         # Concatenate image condition "y" channel-wise to input
+    ...         y = condition["y"]  # shape: (B, C_cond, H, W)
+    ...         x_cat = torch.cat([x, y], dim=1)
+    ...         # Extract 1D vector condition for class_labels
+    ...         class_labels = condition["class_labels"]  # shape: (B, label_dim)
+    ...         return self.net(x_cat, noise_labels=t, class_labels=class_labels)
     ...
-    >>> wrapped = SongUNetWrapper(img_resolution=8, in_channels=2, out_channels=2)
+    >>> wrapped = SongUNetWrapper(
+    ...     img_channels=2, cond_channels=1, label_dim=4, img_resolution=8
+    ... )
     >>> isinstance(wrapped, DiffusionModel)
     True
     >>> x = torch.rand(1, 2, 8, 8)
     >>> t = torch.rand(1)
-    >>> out = wrapped(x, t, condition=None)
+    >>> condition = TensorDict({
+    ...     "y": torch.rand(1, 1, 8, 8),           # image condition
+    ...     "class_labels": torch.rand(1, 4),      # 1D vector condition
+    ... }, batch_size=[1])
+    >>> out = wrapped(x, t, condition)
     >>> out.shape
     torch.Size([1, 2, 8, 8])
 
     **Example: Wrapping DiT**
 
     The :class:`~physicsnemo.experimental.models.dit.DiT` model has
-    the signature ``forward(x, t, condition, p_dropout, attn_kwargs)``.
-    Its ``condition`` expects a Tensor, not a TensorDict. We wrap it
-    to extract the tensor from a TensorDict if needed:
+    the signature ``forward(x, t, condition, ...)``. We wrap it to support
+    both image conditioning (via channel concatenation) and vector
+    conditioning:
 
     >>> from physicsnemo.experimental.models.dit import DiT
     >>> class DiTWrapper(Module):
-    ...     def __init__(self, **dit_kwargs):
+    ...     def __init__(self, img_channels, cond_channels, cond_dim, **kwargs):
     ...         super().__init__()
-    ...         self.net = DiT(**dit_kwargs)
+    ...         # in_channels = img_channels + cond_channels for concatenation
+    ...         self.net = DiT(
+    ...             in_channels=img_channels + cond_channels,
+    ...             out_channels=img_channels,
+    ...             condition_dim=cond_dim,
+    ...             **kwargs,
+    ...         )
     ...
-    ...     def forward(self, x, t, condition=None, **kwargs):
-    ...         # Extract tensor from TensorDict if provided
-    ...         if isinstance(condition, TensorDict):
-    ...             cond_tensor = condition["embedding"]  # or relevant key
-    ...         else:
-    ...             cond_tensor = condition  # already a Tensor or None
-    ...         return self.net(x, t, condition=cond_tensor, **kwargs)
+    ...     def forward(self, x, t, condition):
+    ...         # Concatenate image condition "y" channel-wise to input
+    ...         y = condition["y"]  # shape: (B, C_cond, H, W)
+    ...         x_cat = torch.cat([x, y], dim=1)
+    ...         # Extract 1D vector condition
+    ...         vec = condition["vec"]  # shape: (B, cond_dim)
+    ...         return self.net(x_cat, t, condition=vec)
     ...
     >>> wrapped_dit = DiTWrapper(
-    ...     input_size=8, patch_size=4, in_channels=2, attention_backend="timm"
+    ...     img_channels=2, cond_channels=1, cond_dim=4,
+    ...     input_size=8, patch_size=4, attention_backend="timm",
     ... )
     >>> isinstance(wrapped_dit, DiffusionModel)
     True
     >>> x = torch.rand(1, 2, 8, 8)
     >>> t = torch.rand(1)
-    >>> out = wrapped_dit(x, t, condition=None)
+    >>> condition = TensorDict({
+    ...     "y": torch.rand(1, 1, 8, 8),  # image condition
+    ...     "vec": torch.rand(1, 4),       # 1D vector condition
+    ... }, batch_size=[1])
+    >>> out = wrapped_dit(x, t, condition)
     >>> out.shape
     torch.Size([1, 2, 8, 8])
-
     """
 
     def __init__(


physicsnemo/diffusion/preconditioners/preconditioners.py
```diff
@@ -14,14 +14,701 @@
...
+    """
+
+    def __init__(
+        self,
+        model: Module,
+        meta: ModelMetaData | None = None,
+    ) -> None:
+        super().__init__()
+        self.meta = meta
+        self.model = model
```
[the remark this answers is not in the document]
CharlelieLrt: Wrong, the `BasePreconditioner.__init__` is definitely calling `super().__init__()`

physicsnemo/diffusion/preconditioners/legacy.py
```diff
@@ -112,25 +118,34 @@ def __init__(
...
+
+        # Initialize parent class with model and VP parameters
+        super().__init__(
+            model=model,
+            beta_d=beta_d,
+            beta_min=beta_min,
+            M=M,
+        )
+        # Override meta from parent
+        self.meta = VPPrecondMetaData
```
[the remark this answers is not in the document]
CharlelieLrt: No, it's pretty standard to do that IMO

physicsnemo/diffusion/preconditioners/legacy.py
```diff
@@ -288,21 +301,28 @@ def __init__(
...
             out_channels=img_channels,
             label_dim=label_dim,
             **model_kwargs,
-        )  # TODO needs better handling
+        )
+
+        # Initialize parent class with model
+        super().__init__(model=model)
+        # Override meta from parent
+        self.meta = VEPrecondMetaData
```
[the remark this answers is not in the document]
CharlelieLrt: The PR is anyways not focusing on these legacy implementation. I'll ignore this comment.

physicsnemo/diffusion/preconditioners/legacy.py
```diff
@@ -419,33 +437,34 @@ def __init__(
...
+
+        # Initialize parent class with model and iDDPM parameters
+        super().__init__(
+            model=model,
+            C_1=C_1,
+            C_2=C_2,
+            M=M,
+        )
+        # Override meta from parent
+        self.meta = iDDPMPrecondMetaData
```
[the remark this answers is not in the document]
CharlelieLrt: Same comment.

physicsnemo/diffusion/preconditioners/legacy.py
```diff
@@ -605,33 +620,38 @@ def __init__(
...
+        )
 
-    def forward(
+        # Initialize parent class with model and sigma_data
+        super().__init__(
+            model=model,
+            sigma_data=sigma_data,
+        )
+        # Override meta from parent
+        self.meta = EDMPrecondMetaData
```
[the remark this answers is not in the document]
CharlelieLrt: Same comment: these legacy implementations are not the focus of this PR.

physicsnemo/diffusion/preconditioners/preconditioners.py
```diff
@@ -14,14 +14,701 @@
...
+    The wrapped model :math:`F` must have the following signature:
+
+    .. code-block::
+
+        model(
+            x: torch.Tensor,
+            t: torch.Tensor,
+            condition: Dict[str, torch.Tensor],
+            **model_kwargs: Any,
+        ) -> torch.Tensor
```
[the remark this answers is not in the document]
CharlelieLrt: Good suggestion. Change implemented.

physicsnemo/diffusion/preconditioners/preconditioners.py
```diff
@@ -14,14 +14,701 @@
...
+                raise ValueError(
+                    f"Expected t to have shape ({B},) matching batch size of "
+                    f"x, but got {t.shape}."
+                )
+            for k, v in condition.items():
+                if v.shape[0] != B:
+                    raise ValueError(
+                        f"Condition tensor '{k}' has batch size {v.shape[0]} "
+                        f"but expected {B} to match x."
+                    )
```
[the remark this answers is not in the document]
CharlelieLrt: The preconditioners do not have strong requirements on the input tensors shapes. Thay are supposed to be shape-agnostic and work for any number of spatial dimension. The only requirement is to have batched inputs, which is already verified.

physicsnemo/diffusion/preconditioners/legacy.py
```diff
@@ -419,33 +437,34 @@ def __init__(
...
-        self.sigma_min = float(u[M - 1])
-        self.sigma_max = float(u[0])
+        # Store legacy-specific attributes
+        self.img_resolution = img_resolution
+        self.img_channels = img_channels
+        self.label_dim = label_dim
+        self.use_fp16 = use_fp16
+        # Use the u buffer from parent to compute sigma_min and sigma_max
+        self.sigma_min = float(self.u[M - 1])
+        self.sigma_max = float(self.u[0])
```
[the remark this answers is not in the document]
CharlelieLrt: Already verified in tests.

test/diffusion/test_preconditioners.py
```diff
@@ -0,0 +1,598 @@
...
+    ),
+    (
+        VEPreconditioner,
+        {},
+        "ve_precond",
+    ),
+    (
+        IDDPMPreconditioner,
+        {"C_1": 0.001, "C_2": 0.008, "M": 100},
+        "iddpm_precond",
```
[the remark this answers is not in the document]
CharlelieLrt: Right. However `M=100` is also a value that would make sense for this precondtioner, even though it is not the default one. Ideally we would like to test non-regression for bother default and non-default parameters, but to limit the number of tests, we will ignore this comment for now.

physicsnemo/diffusion/preconditioners/preconditioners.py
```diff
@@ -14,14 +14,701 @@
...
+        meta: ModelMetaData | None = None,
+    ) -> None:
+        super().__init__()
+        self.meta = meta
+        self.model = model
+
+    @abstractmethod
+    def compute_coefficients(
+        self, t: torch.Tensor, /
+    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
```
NickGeneva, line 338: Are you positive it will always be 4 coefficients?

physicsnemo/diffusion/preconditioners/preconditioners.py
```diff
@@ -14,14 +14,701 @@
...
+            Noise conditioning value of shape :math:`(B, 1, ..., 1)`.
+        c_out : torch.Tensor
+            Output scaling coefficient of shape :math:`(B, 1, ..., 1)`.
+        c_skip : torch.Tensor
+            Skip connection scaling coefficient of shape
+            :math:`(B, 1, ..., 1)`.
+        """
+        ...
+
+    def sigma(self, t: torch.Tensor) -> torch.Tensor:
```
NickGeneva, line 369: outside of the forward function, are there other components that need this sigma functional exposed? Like the loss?

Wondering if the abstract class could just require the forward function signature and then let the subclasses figure out the details.

physicsnemo/diffusion/preconditioners/preconditioners.py
```diff
@@ -14,14 +14,701 @@
...
+                    f"x, but got {t.shape}."
+                )
+            for k, v in condition.items():
+                if v.shape[0] != B:
+                    raise ValueError(
+                        f"Condition tensor '{k}' has batch size {v.shape[0]} "
+                        f"but expected {B} to match x."
+                    )
+
+        # Map time step to noise level via sigma method
```
NickGeneva, line 420: Something to maybe consider here is the readability of these preconditioners with the questions in the paper.

IMO it is much much easier to understand whats going on with the following in the subclass:

```
sigma = t.view(B, *((1,) * (x.ndim - 1)))

        c_skip = self.sigma_data**2 / (sigma**2 + self.sigma_data**2)
        c_out = sigma * self.sigma_data / (sigma**2 + self.sigma_data**2).sqrt()
        c_in = 1 / (self.sigma_data**2 + sigma**2).sqrt()
        c_noise = (sigma.log() / 4).flatten()

        return c_skip * x + c_out * self.model(
            c_in * x, c_noise, condition, **model_kwargs
        )
   ```
   
   that you made for me. I could map it directly to the equations in the paper (having a comment with question number would be even cooler).
   
   Versus this which now splits up the code.
   
   Repeat code yeah, but imo its worth.
   
You then cut off any coupling you have between preconditioners which will help with maintenance int the future.

File: physicsnemo/diffusion/preconditioners/preconditioners.py [hunk 13]
```diff
@@ -14,14 +14,701 @@
...
+    The preconditioning coefficients are:
+
+    .. math::
+
+        c_{\text{skip}} &= 1 \\
+        c_{\text{out}} &= -\sigma \\
+        c_{\text{in}} &= \frac{1}{\sqrt{\sigma^2 + 1}} \\
+        c_{\text{noise}} &= (M - 1) \cdot \sigma^{-1}(\sigma)
+
+    Parameters
```
NickGeneva, line 469: Thanks for using Numpy doc strings! :+1:

physicsnemo/diffusion/preconditioners/preconditioners.py
```diff
@@ -14,14 +14,701 @@
...
+
+    The preconditioner is agnostic to the prediction target of the wrapped
+    model :math:`F`. The same preconditioning formula is applied regardless of
+    whether the model is an :math:`\mathbf{x}_0`-predictor, an
+    :math:`\epsilon`-predictor, a score predictor, or a
+    :math:`\mathbf{v}`-predictor.
+
+    Parameters
+    ----------
+    model : physicsnemo.Module
```
NickGeneva, line 92: Any module or does this need the diffusion forward function interface implemented?

physicsnemo/diffusion/preconditioners/preconditioners.py
```diff
@@ -14,14 +14,701 @@
...
+
+# TODO: once noise schedulers are implemeneted, some of the methods they define
+# can be reused here, e.g. preconditioner.sigma = noise_scheduler.sigma for the
+# noise schedule. This would allow to avoid duplicate code between the
+# preconditioners and the noise schedulers. Particularly for the iDDPM
+# preconditioner, that requires more computations than the other
+# preconditioners.
+
+
+class BasePreconditioner(Module, ABC):
```
NickGeneva: General question, whats the expected behavior when using these pre-conditioners and checkpointing?

How does this work when I need to save the training checkpoint but then I want to using the same checkpoint to then run inference (so no more pre-conditioner)

NickGeneva: I wonder if theres merit in having a training / eval mode, but I also do not like the idea of hiding people networks under a preconditioner class.

wish there was a clean way to inject these objects as sort of object that could be used inside a diffusion model network class rather than wrapping... dont know enough about this though

NickGeneva: NVM, I miss understood.

physicsnemo/diffusion/preconditioners/preconditioners.py
CharlelieLrt: Yes, the sigma method is required at multiple places throughout both training and sampling. I am working on another abstraction called `NoiseScheduler` that implements the sigma method. Once this is done, creating a new preconditioner should just mirror the sigma method from the associated `NoiseScheduler`. For example to re-implement the EDM preconditioner:

```python
from noise_schedulers import EDMNoiseScheduler

class MockEDMPreconditioner(...):
    ...
    def sigma(self, t):
        return EDMNoiseScheduler.sigma(t, self.sigma_data)
```

> Wondering if the abstract class could just require the forward function signature

It theory it could. But it is also preferable to have the checkpointing capabilities of `physicsnemo.Module` for the preconditioner. We don't strictly need these capabilities, but being more opiniated and enforcing that it is a `physicsnemo.Module` removes ambiguity and possible confusion.

physicsnemo/diffusion/preconditioners/preconditioners.py
CharlelieLrt: I think we need to strike the right balance between hackability and extensibility. The snippet above has all logic inlined in one place, which is great for hackability. On the opposite it's not so good for extensibility because every time a new preconditioner subclass is defined, the entire foward pass logic needs to be reimplemented.

IMO these preconditioners are relatively easy and simple, so they are intended for users that are not diffusion experts. It is unlikely that these users would go to see the code details further than the docstring (whoch already contains the math)

physicsnemo/diffusion/preconditioners/preconditioners.py
CharlelieLrt: The preconditioner works seamlessly with checkpointing. The saved checkpoint is essentially then preconditioner itself.

> How does this work when I need to save the training checkpoint but then I want to using the same checkpoint to then run inference (so no more pre-conditioner)

Preconditioner behavior is the same during training and inference, so this does not matter.

physicsnemo/diffusion/preconditioners/preconditioners.py
CharlelieLrt: Good point!

I renamed the base class to `BaseAffinePreconditioner`. This covers most existing preconditioning schemes in the literature, while keeping the door open for different schemes.

physicsnemo/diffusion/preconditioners/preconditioners.py
CharlelieLrt: Same as above, it is not strictly required but preferable to have a `physicsnemo.Module` to maintain checkpointing capabilities. And yes, this Module must implement the signature specified in the docstring. I added a `Protocol` in `diffusion/base.py` as well as a link in the doctstring to emphasize this point.

physicsnemo/diffusion/preconditioners/preconditioners.py
```diff
@@ -14,14 +14,701 @@
...
+    **model_kwargs : Any
+        Additional keyword arguments passed to the underlying model.
+
+    Outputs
+    -------
+    torch.Tensor
+        Preconditioned model output with the same shape as the original model
+        output.
+
+    .. note::
```
pzharrington, line 120: Just curious, is this alternate syntax for the below?
```
Note
-------
```

physicsnemo/diffusion/preconditioners/preconditioners.py
```diff
@@ -14,14 +14,701 @@
...
+          keeping the same preconditioning interface, in particular for
+          preconditioning schemes based on noise level (that is
+          :math:`c_{\text{in}}(\sigma)`,
+          :math:`c_{\text{noise}}(\sigma)`, :math:`c_{\text{out}}(\sigma)`,
+          :math:`c_{\text{skip}}(\sigma)` instead of :math:`c_{\text{in}}(t)`,
+          :math:`c_{\text{noise}}(t)`, :math:`c_{\text{out}}(t)`,
+          :math:`c_{\text{skip}}(t)`).
+
+        - The ``forward`` method of the preconditioner *should not* be
+          overriden.
```
pzharrington, line 143: Why do we need this requirement?

pzharrington: I guess you said "should not" instead of "can not" so maybe this isn't an absolute requirement per se

physicsnemo/diffusion/preconditioners/preconditioners.py
pzharrington: Yeah I think it's better to have it as `compute_coefficients` -- the EDM paper has been around for awhile and people can always go to the original NVLabs repo if they want to see the equations written out in all their glory and hack away. I think our goal in this package should be to provide an easy-to-use interface/wrapper that abstracts away the details that are not typically tuned (e.g. `c_*` coeffs) while conveniently exposing hyperparameters that are tuned (e.g. `sigma_data` in `EDMPreconditioner`).

pzharrington: I'd expect 99% of users will never need to write their own subclass anyways

physicsnemo/diffusion/preconditioners/preconditioners.py
```diff
@@ -14,14 +14,701 @@
...
+    -------
+    x : torch.Tensor
+        Noisy latent state of shape :math:`(B, *)` where :math:`B` is the
+        batch size and :math:`*` denotes any number of additional dimensions.
+    t : torch.Tensor
+        Diffusion time tensor of shape :math:`(B,)`.
+    condition : Dict[str, torch.Tensor]
+        Dictionary of conditioning tensors. Each tensor must have shape
+        :math:`(B, *)` where the batch size :math:`B` matches that of ``x``.
+        These are passed to the wrapped ``model`` without modification.
```
pzharrington: This is should be marked optional, no? What about unconditional diffusion models? Similarly the underlying models themselves should not have to require a `condition` argument

physicsnemo/diffusion/preconditioners/preconditioners.py
```diff
@@ -14,14 +14,701 @@
...
+            Preconditioning coefficients (:math:`c_{\text{in}}`,
+            :math:`c_{\text{noise}}`, :math:`c_{\text{out}}`,
+            :math:`c_{\text{skip}}`) of shape :math:`(B, 1, ..., 1)`.
+        """
+        sigma_data = self.sigma_data
+        c_skip = sigma_data**2 / (t**2 + sigma_data**2)
+        c_out = t * sigma_data / (t**2 + sigma_data**2).sqrt()
+        c_in = 1 / (sigma_data**2 + t**2).sqrt()
+        c_noise = t.log() / 4
+        return c_in, c_noise, c_out, c_skip
```
pzharrington, line 921: A general comment about these coefficients. It's relatively minor, but we noticed in domain parallel training w/ `ShardTensor` that there was an outsize amount of kernel launch latency / GPU idling in the preconditioning phase, arising from creating a bunch of these tensors from the scalar class attributes (`self.sigma_data`, etc) -- since stored in CPU memory there is some CPU-GPU syncing that occurs to create the tensor and it leads to jitter in the model parallel training. The problem is also there for non-domain-parallel training but still exists. Not worth holding up this PR over, but I think a better solution is to register the small constants as buffers so they get allocated and moved to GPU easily via `.to(device)` at the start of training -- the same goes for small constants in custom noise schedules.

physicsnemo/diffusion/preconditioners/preconditioners.py
CharlelieLrt: Yes, and if I recall it renders slightly differently in the online docs. In any case I am planning to have a PR for the diffusion API docs once the entire package is in place. At that moment I'll make sure all docstrings are properly rendered in the online docs

physicsnemo/diffusion/preconditioners/preconditioners.py
CharlelieLrt: Yeah I said should not, because overriding the `forward` would require to make the right calls to `self.sigma` and `self.compute_coefficients`. Nothing very difficult, but to avoid problems lambda users should not attempt that IMO

physicsnemo/diffusion/preconditioners/preconditioners.py
CharlelieLrt: Yeah I was thinking about making `condition` optional, but if we make it an optional argument, how do we separate it from those in `**model_kwargs`? There might be conflicts and mix-up between the two

> Similarly the underlying models themselves should not have to require a condition argument

Right, but then similar problem: how do we differentiate between models that require one and others that don't?
I found making `condition` required was the cleanest solution to remove all type of confusion and ambiguity, even though this argument is not needed for many cases.

But if you have a better ideaI I am open to it

physicsnemo/diffusion/preconditioners/preconditioners.py
CharlelieLrt: Very good point, I was not aware of this, thanks for letting me know!

I refactored all these scalar attributes such that they are all buffers now

physicsnemo/diffusion/preconditioners/preconditioners.py
pzharrington: Ugh, yeah good points the model kwargs make this annoying. It seems hard to avoid some sort of awkwardness here, apart from flat out defining separate preconditioners for the conditional and unconditional cases. To me it just feels wrong to have to wrap underlying unconditional backbones with something to pass a dummy `condition` arg, and similarly for the top-level preconditioner, but I'm having trouble coming up with alternate solutions.

pzharrington: What do you think of a capability flag passed to preconditioner init `conditional` which will specify whether or not the preconditioner (and underlying model) are expected to use the `condition` arg? Then we could have
```
forward(x, condition: TensorDict | None = None, **model_kwargs)
```
where within the forward pass we call things based on the value of `self.conditional`.

pzharrington: Alternate, and possibly spicier suggestion, we drop the mention of `condition` entirely from the `forward` signature. It is absorbed into `model_kwargs`, may or may not be passed, and is up to the user to do the input validation (which is the only operation on `condition` within the `forward` of the preconditioner`). This would also allow for flexible nomenclature of the conditioning in principle, i.e. some bespoke model wants to call it `era5_condition` (+ optionally add others like `time_of_day_condition`, etc), then it is welcome to.

NickGeneva: @pzharrington I would be interested in how this condition idea maps to DiTs.

With the current HRRR surface model, the condition fields get channel concatenated onto the input with added noise... so both the input and conditions need to get applied the patching operation in multi-diffusion which is easier to do in the loss (thus needing the model to have the condition parameter).

Now if the patching op could get re-designed to work well inside the model and loss then maybe I could see it being removed from the function signature.

NickGeneva: (That is unless in the samplers theres some specific operation that needs conditional fields explicitly)

CharlelieLrt: > What do you think of a capability flag passed to preconditioner init conditional which will specify whether or not the preconditioner (and underlying model) are expected to use the condition arg?

@pzharrington if it were only for the precondtioner, I would say okay. But we would need this conditional flag everywhere the model is passed as a callback (loss, samplers, etc). That will make things very heavy IMO. Also loss object is supposed to be purely functional, so the flag would need to be passed to the `__call__` and not the `__init__`.

I prefer the second alternative that you proposed. Another option I just thought of would be to make the `condition` argument **keyword-only**, with a signature `model(x, t, *, condition=None, **model_kwargs)`. That removes some possible ambiguity and mix up because by forcing to always pass `condition` by name, we forbid calls such as `model(x, t, condition={...}, kwargs1=..., condition={...}, other_kwarg=...)`, which are anyways invalid.

CharlelieLrt: @NickGeneva 's comment above reminded another reason why I wanted to have an explicit argument for condition. In multi-diffusion (whose implementation is still at the stage of philosophical reflection as of now), the model needs to know which argument is condition, because it needs to apply specific operations on the conditioning tensors (patching, sometimes interpolation). So, in multi-diffusion, `condition` cannot be considered the same as *any* `model_kwargs`.

Roughly, it should like:

```python
multi_diffusion_model = MultiDiffusionModel(model, patching_options)
x0 = multi_diffusion_model(x, t, condition={"y": y, "z": z}, some_kwarg=some_val)
```

Under the hood the forward pass to `multi_diffusion_model` applies patching (+ optionally interpolation) and concatenation to the items in `condition`, but it leaves all other kwargs untouched.

pzharrington: > In multi-diffusion, the model needs to know which argument is condition

And it would, right? The underlying model would define `condition` in its forward kwargs and handle accordingly within its forward pass. Under my second suggestion, the preconditioner wrapping that would simply pass it through as part of the `**model_kwargs`, it doesn't need to know about or explicitly do anything with the condition (there is no condition-dependent preconditioning and if there was, that's crazy 😅).

E.g., in your snippet, we'd have
```
multi_diffusion_model = MultiDiffusionModel(model, patching_options) # Wrap a base backbone with multidiffusion
model_precond = EDMPreconditioner(multi_diffusion_model) # Wrapped by preconditioner, ready for training
x0 = model_precond(x, t, condition={"y": y, "z": z}, some_kwarg=some_val)
```
Preconditioner doesn't care about the conditioning or the patching/interpolation applied to it, it delegates that to the underlying models. This is fine, no? Or am I missing something?

CharlelieLrt: Hmm I'm trying to understand that...

In your snippet above, what would be the signatures of `model`, `model_precond`, and `multi_diffusion_model `? (And I mean the actual signatures, not just the way they are called, because `model(x, t, condition={}, kwarg=val)` may look the same whether `condition` is a separate keyword argument or part of the kwargs, but the signature signals intent)

pzharrington: Not sure I understand the latter part "signature signals intent" but I was thinking, for the conditional case, the signatures would be:
```
model(x, t, condition: torch.Tensor | None = None, **other_model_kwargs)
multi_diffusion_model(x, t, condition: torch.Tensor | None = None, **other_model_kwargs)
model_precond(x, t, **model_kwargs)
```
`model_kwargs` are passed through to the underlying models in the preconditioner forward, and the underlying models will pick out the conditioning and deal with it as required in their respective forward passes.

The unconditional case would be the same but the underlying models would not have a `condition` arg. But my point is that regardless, the preconditioner does not need to know about that (and thus would not need to specify a requirement on the presence/absence/type of condition args) for the expected interface.

CharlelieLrt: Ok, thanks for the details. One problem with these signatures is that `model_precond ` has a signature that differs from the two others. Ideally we would like the signature of the model to be invariant throughout the entire diffusion package. More precisely, all diffusion models have to implement the `DiffusionModel` protocol defined in `diffusion/base.py`. The signature could be `(x: Tensor, t: Tensor, condition: TensorDict, **kwargs)`, or `(x, t, condition: TensorDict | None = None,  **kwargs)`, or even `(x, t, **kwargs)`, but what really matters is that it is *the same everywhere*, wether we are dealing with a preconditioned model or not.
At least it's the general philosophy, and I do think it has the benefit of being very clear.

That being said, I agree that the current protocol that enforces `(x: Tensor, t: Tensor, condition: TensorDict, **kwargs)` is too constraining, in particular for unconditional diffusion where one would need to pass an empty TensorDict everywhere for the `condition` argument.

One the other hand `(x, t, **model_kwargs)` is not sufficient for multi-diffusion, because there is no explicit information about conditioning.

So, the best choice might just be `(x, t, condition: TensorDict | None = None, **model_kwargs)` (or to be even more precise: `(x, t, *, condition: TensorDict | None = None, **model_kwargs)`), which is I think what you proposed in your initial comment?

jleinonen: I found this confusing too when I saw it in the code. Another issue is that the models have a very different idea of what condition means. It seems the `condition` argument in DiT is not spatially variable so it's necessary to concat the condition to the noisy input instead of passing it to the `condition` parameter. This line from the current `EDMPrecond` should work:
https://github.com/NVIDIA/physicsnemo/blob/6c7b51165a5c915abcc1c04e29ae99dee5470976/physicsnemo/diffusion/preconditioners/legacy.py#L665-L666
But this possibility is missing from the new formulation of the preconditioner. It seems to be the most robust way to pass spatially variable conditions, though.

jleinonen: Here's the `forward` I came up with in case you find it useful. It could probably benefit some dimensions checking but with this I can use `DiT`.
```python
    def forward(
        self,
        x: torch.Tensor,
        t: torch.Tensor,
        condition: TensorDict | torch.Tensor | None = None,
        **model_kwargs: Any,
    ) -> torch.Tensor:
        if not torch.compiler.is_compiling():
            B = x.shape[0]
            if t.shape != (B,):
                raise ValueError(
                    f"Expected t to have shape ({B},) matching batch size of "
                    f"x, but got {t.shape}."
                )
            if isinstance(condition, TensorDict) and condition.batch_size and condition.batch_size[0] != B:
                cond_B = condition.batch_size[0]
                raise ValueError(
                    f"Condition TensorDict has batch size {cond_B} "
                    f"but expected {B} to match x."
                )

        # Map time step to noise level via sigma method
        sigma_t = self.sigma(t).reshape(-1, *([1] * (x.ndim - 1)))

        # Compute preconditioning coefficients
        c_in, c_noise, c_out, c_skip = self.compute_coefficients(sigma_t)

        x_in = c_in * x
        if isinstance(condition, torch.Tensor):
            x_in = torch.concat([x_in, condition], dim=1)
            condition = None

        # Forward through the underlying model
        F_x = self.model(
            x_in,
            c_noise.flatten(),
            condition,
            **model_kwargs,
        )

        D_x = c_skip * x + c_out * F_x

        return D_x
```

CharlelieLrt: @jleinonen this is not ideal, the preconditioner should only care about the interface, i.e. it should not need to do anything with respect to the `condition` or `x`, except multiplying `x`. Also concatenation is just one way to condition a model. There is an infinity of methods to add conditioning (embedding + concatenation, addition, cross-attention, adaptive normalization, etc...). If the preconditioner needs to be aware of all these mechanisms to handle conditioning, then we end up with an infinity of preconditioner.

> It seems the condition argument in DiT is not spatially variable

That's precisely what the current implementation does: it does not make any assumption whether condition is a spatial variable or not. Actually there is not even a notion of spatial here, even on `x`. The only assumption is that both `x` and `condition` are batched.

> so it's necessary to concat the condition to the noisy input


That's the job of `model` that is wrapped, but not of the preconditioner.

jleinonen: >>so it's necessary to concat the condition to the noisy input
>>
>That's the job of model that is wrapped, but not of the preconditioner.

I actually thought attaching the condition to the input was why it was called "preconditioner".

Anyway, I get the motivation for supporting more complex conditions on the model side. But this also requires all models to be redesigned to conform to the preconditioners. Wouldn't it be reasonable to support the very common access pattern of concatenating the condition to the input so we can continue using older models with the new preconditioner, especially since that can be achieved with a pretty small amount of added code?

The `isinstance` test I have in the above example is probably not the ideal way to do it but could there be an option to `__init__`, something like
```python
condition_handling: Literal["concat", "pass_to_model"]
```
to control what to do with the condition?

Edit: To be more specific, I'm trying to avoid a situation where neither the old nor the new flavor of preconditioner works with DiT, which is the situation with the current state of the PR.

CharlelieLrt: > I actually thought attaching the condition to the input was why it was called "preconditioner".

No, preconditioning is just computing the coefficients `c_out, c_noise, c_skip, c_in` and do the affine scaling operation on the input and output of the model. It's true that the terms are a little confusing, but "preconditioning" has nothing to do with the `condition` argument. Actually preconditioning even applies to unconditional models.

> so we can continue using older models with the new preconditioner

The old preconditioners are marked as legacy, but they will not be deprecated. So it will always be possible to use old models without any adaptation or redesign necessary.

> Wouldn't it be reasonable to support the very common access pattern of concatenating the condition to the input

To support something like this, we would need another abstraction like maybe `ConditionalModel`, and the hierarchy would be for example `preconditioner > ConditionalModel > SongUNet/DiT/...`. But is there really a need to introduce a new abstraction for 2 lines of pytorch? In the concatenation case you described, IMO it's easier to just write your own operation in your model to handle it rather than introduce some complicated abstraction.

In any case, even if there is really a need for it, the condition handling is not done by the preconditioner, and it should be delegated to another abstraction.

pzharrington: Yeah unfortunately the word "condition" has gotten way overloaded in the land of diffusion 😆 I agree with Charlelie that we shouldn't support yet another abstraction for this relatively simple thing, the majority of use-cases it should be possible to delegate this to the model itself, which should know details about the conditioning input(s) and handle them accordingly while the preconditioner should not care. @jleinonen To support this with both the older models and newer DiTs in the unified stormcast/stormscope recipe, I think we can just make some wrappers around the `SongUNets` for that example, right? Or otherwise modify the data prep methods in the training loop?

CharlelieLrt: > Yeah unfortunately the word "condition" has gotten way overloaded in the land of diffusion

100%, and we are not even talking about condition used with DPS-style guidance in test-time generation.

> I think we can just make some wrappers around the SongUNets for that example

Definitely agree with @pzharrington here. Writing some simple adapter/wrapper is the way to go. I think this is one of these cases where good documentation in terms of expected interface is better than another abstraction.

This does raise the issue of saving/loading checkpoints. If we have a hierarchy `Preconditioner > Custom wrapper > SongUNet (or other)`, then to be able to load checkpoints (e.g. in E2S) we need the custom wrapper class to be discoverable by `importlib`. If the custom wrapper lives outside of the physicsnemo core library (e.g. in the examples), I am not sure how to handle this in a clean way.

Maybe we should have some directory in physicsnemo specifically for these type of application-specific wrappers that we need for shipping checkpoints. That would be easier if training code and inference code were living in the same package, but that's not the case...

@pzharrington what are your thoughts on this?

pzharrington: I'm not opposed to dropping something like a `ConcatImageConditionWrap` somewhere in `physicsnemo.models` to facilitate making this easier, seems like low effort for pretty good payoff. Maybe into `physicsnemo.models.diffusion_unets` since those are the ones that expect the legacy pattern of the 2D conditioning being concatenated to the noised/preconditioned `x`. Then the policy would generally be:

- Base model architectures that support conditioning should do it themselves in their `forward`, accepting conditioning as one (or more) kwargs
- Any existing backbones like these more legacy ones could be wrapped by the lightweight `ConcatImageConditionWrap` to ensure API compatibility
- `BaseAffinePreconditioner` API (introduced here) would not do anything with conditioning, it just passes everything through as `**model_kwargs` so it is on the caller (training loop / inference workflow) to pass conditioning.

CharlelieLrt: > Base model architectures that support conditioning should do it themselves in their forward, accepting conditioning as one (or more) kwargs

This I definitely agree.

> Any existing backbones like these more legacy ones could be wrapped by the lightweight ConcatImageConditionWrap to ensure API compatibility

Nothing against this, but it will be difficult to make something that is truly general. Right now we already need to cover these cases:
- conditioning on single tensor `y`
- conditioning on multiple conditioning tensors `y1`, `y2`, ... which all have the same shape ("spatial tensors")
- conditioning on multiple spatial tensors, plus some 1D vectors (e.g. condition on time embeddings)

So, the wrapper needs to know what to do with every tensor in each case.

My feeling is that these conditioning ops are so application-specific that it will be difficult to make an abstraction that can handle more than 2-3 use cases... But after all, maybe we don't need to care about this, if the only goal is to be able to ship checkpoints while preserving loading capability. We don't even need to expose the abstraction if it's an internal-only thing.

Note: for existing checkpoints, we don't need this wrapper, because they rely on the legacy preconditioners, which are **not** going to be deprecated.

CharlelieLrt: > BaseAffinePreconditioner API (introduced here) would not do anything with conditioning, it just passes everything through as **model_kwargs so it is on the caller (training loop / inference workflow) to pass conditioning.

This however I agree less, mostly [because of this](https://github.com/NVIDIA/physicsnemo/pull/1317#discussion_r2678106104). I think we really need to have this separate optional `condition` argument, not for the preconditioner itself, but to be consistent with other abstraction like multi-diffusion which are going to require it.

pzharrington: > But after all, maybe we don't need to care about this, if the only goal is to be able to ship checkpoints while preserving loading capability. We don't even need to expose the abstraction if it's an internal-only thing.
Note: for existing checkpoints, we don't need this wrapper, because they rely on the legacy preconditioners, which are not going to be deprecated.

Yeah I think this is a good perspective, let's not overthink it. We're not as of now needing to support new/non-legacy checkpoints under this setup, and if we were, it wouldn't be hard to put in some wrappers, not necessarily even public, to take care of things on an as-needed basis.

CharlelieLrt: Agree!

jleinonen: Good to see this discussion, I was also going to raise the concern that having a wrapper implemented in the example was going to make delivering checkpoints difficult. I'm not sure from the above what your decision was, though. It seemed to me that Peter's suggestion to implement a wrapper in PNM for the most common concat use case would solve this, but did you decide to do something else in the end?

pzharrington: > It seemed to me that Peter's suggestion to implement a wrapper in PNM for the most common concat use case would solve this, but did you decide to do something else in the end?

Decision is we will implement wrappers like this as needed

physicsnemo/diffusion/base.py
```diff
@@ -0,0 +1,100 @@
...
+    True
+    """
+
+    def __call__(
+        self,
+        x: Float[torch.Tensor, "B *dims"],  # noqa: F821
+        t: Float[torch.Tensor, "B"],  # noqa: F821
+        condition: Float[torch.Tensor, "B *cond_dims"] | TensorDict | None = None,  # noqa: F821
+        **model_kwargs: Any,
+    ) -> Float[torch.Tensor, "B *dims"]:  # noqa: F821
```
pzharrington: nit: these `# noqa: F821` can be dropped, see https://nvidia.slack.com/archives/C09QAS52AKV/p1768397257037909?thread_ts=1768355832.954499&cid=C09QAS52AKV

CharlelieLrt: For some reason, I still have linter errors if I remove them. For example `Float[torch.Tensor, "B *dims"` gives me `Undefined name 'B'`

pzharrington: Try the space in front of B, like `Float[torch.Tensor, " B *dims"]`. I think that's the fix jaxtyping recommends, though a bit ugly

pzharrington (approved):
LGTM pending fix for the `noqa` flags!

physicsnemo/diffusion/base.py
CharlelieLrt: Yep it worked! Fixed

PR #1318, opened 2026-01-09
Fea nn funcitonal

loliverhennigh:
# PhysicsNeMo Pull Request

## Description
proposal for functional change.


coreyjadams:
Hey @loliverhennigh is this ready for review?

physicsnemo/domain_parallel/shard_utils/knn.py
```diff
@@ -28,7 +28,7 @@
     RingPassingConfig,
     perform_ring_iteration,
 )
-from physicsnemo.nn.neighbors._knn._cuml_impl import knn_impl
+from physicsnemo.nn.functional.knn._cuml_impl import knn_impl
```
loliverhennigh, line 32: When functionals collide with distributed stuff this raises some questions. For right now lets not worry about it in this PR.

physicsnemo/nn/functional/knn/knn.py
```diff
@@ -0,0 +1,199 @@
...
+        raise NotImplementedError
+
+    @classmethod
+    def dispatch(
+        cls,
+        points: torch.Tensor,
+        queries: torch.Tensor,
+        k: int,
+        backend: Literal["cuml", "torch", "scipy", "auto"] = "auto",
+        implementation: str | None = None,
```
loliverhennigh: whoops, still need to clean this up a bit...

docs/api/physicsnemo.utils.rst
megnvidia, line 1: nvidia style guide strongly suggests Title Case for all titles

physicsnemo/core/function_spec.py
megnvidia, line 1: 2025 should be 2026

physicsnemo/nn/functional/knn/knn.py
megnvidia, line 1: 2026

physicsnemo/nn/functional/radius_search/utils.py
```diff
@@ -0,0 +1,35 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
```
megnvidia, line 1: 2026

physicsnemo/nn/module/__init__.py
```diff
@@ -0,0 +1,71 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
```
megnvidia, line 1: 2026

test/core/test_function_spec.py
```diff
@@ -0,0 +1,302 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
```
megnvidia, line 1:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
```

physicsnemo/nn/functional/knn/knn.py
loliverhennigh: Done

physicsnemo/core/function_spec.py
```diff
@@ -0,0 +1,611 @@
...
+@dataclass(frozen=True)
+class Implementation:
+    """Stores all needed information for an implementation."""
+
+    name: str
+    func: Callable
+    required_imports: Tuple[str, ...]
+    rank: int
+    baseline: bool
+    available: bool = True
```
coreyjadams, line 57: Can we add some comments here to describe the role of each data item?

physicsnemo/core/function_spec.py
```diff
@@ -0,0 +1,611 @@
...
+    ---------------------------
+    1. Subclass ``FunctionSpec``.
+    2. Register one or more backend implementations with the decorator
+       ``@FunctionSpec.register``. Provide a ``name`` and a ``rank`` (lower
+       wins). Optionally set ``baseline=True`` for the reference implementation
+       used in benchmarking. The decorator must be used inside the class body.
+    3. Implement :meth:`make_inputs` and :meth:`compare` for benchmarking and
+       correctness checks. ``make_inputs`` should be a generator that yields
+       representative input tuples for the function, and ``compare`` should
+       validate that outputs from two implementations match.
```
coreyjadams, line 95: It's not clear from the docs here - is make_inputs and compare mandatory for an implementation? Not necessarily objecting if it is.

physicsnemo/core/function_spec.py
```diff
@@ -0,0 +1,611 @@
...
+            Mapping of implementation names to Implementation objects.
+
+        Raises
+        ------
+        KeyError
+            If no implementations are registered for the class.
+        """
+        impls = cls._impl_registry.get(cls._class_key(), {})
+        if not impls:
+            raise KeyError(f"No implementations registered for {cls.__name__}")
```
coreyjadams: Realistically, is this a case we can ever trigger?  Presumably a functional wouldn't be written without at least one implementation, though I suppose the implementation details technically allow this?

physicsnemo/domain_parallel/shard_utils/knn.py
coreyjadams: I can give some context here for why the distributed implementation is doing this.

- ShardTensor uses pytorch dispatch resolution to match an implementation to a function: so only functions registered with pytorch can be easily overloaded and dispatched properly from shard tensor.
- In the original knn implementation, I registered the cuml version since the fall back was pure pytorch.  In principle, we can use the generic version (`knn` instead of `knn_impl`) with the exact same distributed algorithm.  But it's not possible to easily trigger that code path with shard tensor since `knn` is not registered with pytorch.
- The "default" path for knn actually succeeds because of DTensor support, but, that's neither here nor there honestly.

Looks like in the functional tools you've made here put us in the same situation: we have to intercept things registered with torch, but it comes at the implementation level.  I think, though, if implementations are "standardized" to accept the same inputs / give the same outputs, we can basically register all of them through the ShardTensor dispatch mechanism: so, register `knn_impl_cuml` and `knn_impl_torch`, etc.  Then, we implement one distributed alg which will call the higher-level interface `knn`.  Since the distributed alg will call that with torch.Tensor and not ShardTensor, we'll avoid the dispatch issue in the distributed alg and it should work.

But yeah, let's catch this in the next PR.

physicsnemo/nn/__init__.py
```diff
@@ -14,25 +14,25 @@
...
 from physicsnemo.core import Module
 
-from .activations import (
+from .module.activations import (
     CappedGELU,
     CappedLeakyReLU,
     Identity,
     SquarePlus,
     Stan,
     get_activation,
```
coreyjadams, line 35: Do I understand correctly that the plan is to expose to users code like `from physicsnemo.nn import CappedGELU` while we're storing the actual code in `physicsnemo.nn.module.activations`?  If so, I agree and like it :).

physicsnemo/nn/functional/knn/knn.py
coreyjadams: @megnvidia this is true in all files in the repo.  We should update them all in a separate PR.

physicsnemo/nn/functional/sdf.py
```diff
@@ -222,10 +218,53 @@ def _(
...
+        max_dist: float = 1e8,
+        use_sign_winding_number: bool = False,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        return signed_distance_field_impl(
+            mesh_vertices,
+            mesh_indices,
+            input_points,
+            max_dist=max_dist,
+            use_sign_winding_number=use_sign_winding_number,
+        )
```
coreyjadams, line 250: @peterdsharpe does physicsnemo.mesh provide a pure-torch impl of SDF?  Perhaps it's worth adding in a follow.

physicsnemo/domain_parallel/shard_utils/knn.py
loliverhennigh: Oh this is very interesting. I will think this through a bit more.

physicsnemo/nn/__init__.py
loliverhennigh: Yup, thats the plan.

physicsnemo/nn/functional/sdf.py
loliverhennigh: No but +1 on adding it. It would allows us to give a nice benchmark to compare too. I am sure this will be 10x faster then a pytorch version.

docs/api/physicsnemo.utils.rst
loliverhennigh: Done

physicsnemo/core/function_spec.py
loliverhennigh: Added a doc string

physicsnemo/core/function_spec.py
loliverhennigh: I added a doc string. Said its not mandatory but highly recommended and needed for benchmarking.

physicsnemo/core/function_spec.py
loliverhennigh: Removed it.

loliverhennigh:
Blaa, wasn't able to rebase. Still fixing but will close this as Ill need to make a new PR.

PR #1322, opened 2026-01-14
`pyproject.toml` changes for robustness & housekeeping

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Adds a few minor but very important small changes:

- Ruff now ignores F722 in order to play nice with jaxtyping. See: https://docs.kidger.site/jaxtyping/faq/
- We now explicitly indicate Linux support, which is consistent with the README.md.
  - Note: installation on Windows & Mac already fail before this PR. (In the Mac case, this is because there are no wheels for nvtx, which is a core dependency.) The difference now is that when users on Win/Mac now try to install, they will now get a helpful error message rather than a cryptic one.
- Fixes a bug where we would fetch all packages from `pypi.nvidia.com`, rather than just the ones we opt-into. This was causing dependency failures for packages that NVIDIA does not mirror.
- Commits `uv.lock`, which:
  - [Is considered best-practice](https://old.reddit.com/r/learnpython/comments/1jbo88t/uv_for_python_project_and_version_management/)
  - Will increase reproducibility across installs, particularly in CI. Now, if a new version of a dependency comes out that would break PhysicsNeMo, we don't immediately fail - this is deferred until we `uv sync --upgrade`, which we can do once-per-release-cycle. This prevents really annoying and show-stopping bugs where we're trying to merge an unrelated change, but a dependency change forces us to "contaminate" the PR with unrelated changes.
  - Also allows us to more-reliably reproduce CI failures locally.

## Checklist

## Review Process

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,12 +334,15 @@ way is to start with a reference sample and then update it for your own use case
 ## Installation
 
 The following instructions help you install the base PhysicsNeMo modules to get
-started. In addition to this, optional dependencies can be installed to provide additional functionality. A complete list of optional dependencies is available in the [`pyproject.toml`](./pyproject.toml) file.
+started. In addition to this, optional dependencies can be installed to provide
+additional functionality. A complete list of optional dependencies is available
+in the [`pyproject.toml`](./pyproject.toml) file.
 
-The [training recipes](./examples) are not packaged into the pip wheels or the container to keep the
-footprint low. We recommend users clone the appropriate training recipes and use them
-as a starting point. These training recipes may require additional example-specific dependencies,
-as indicated through an associated `requirements.txt` file in such cases.
+The [training recipes](./examples) are not packaged into the pip wheels or the
+container to keep the footprint low. We recommend users clone the appropriate
+training recipes and use them as a starting point. These training recipes may
+require additional example-specific dependencies, as indicated through an
+associated `requirements.txt` file in such cases.
 
 ### PyPI
 


coreyjadams:
Hey @peterdsharpe thanks for this.  Couple questions:

- Do we want to commit uv.lock?
- We actually DO have mac install and CPU testing, it's part of CI now every night on github.  I can't speak to windows but we shouldn't specify linux only.
- Ruff R722 is good, I think we have a couple PRs adding this, first one past the post wins!

peterdsharpe:
I *think* we want to commit `uv.lock` such that resolution is consistent between all environments. This solves:
- Both the "I want to merge Feature A but underlying dependency B released an update that breaks something unrelated, so now CI is failing and I have to introduce unrelated fixes to my PR." and "CI is failing on the CI server but I can't easily reproduce locally, since we have different dependencies"

[And Astral says on `uv.lock`](https://docs.astral.sh/uv/guides/projects/#python-version): 
> This file should be checked into version control, allowing for consistent and reproducible installations across machines.

pyproject.toml
```diff
@@ -60,6 +60,10 @@ Changelog = "https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md"
 
 [tool.uv]
 managed = true
+environments = [
```
ktangsali: We should be platform agnostic. Any libraries that are OS dependent should not be in the required dependencies.

peterdsharpe:
Re: Linux-only:

Interesting! I guess `nvtx` is building from source on Mac, then?

In any case, the main problem is that right now we have a conflict:
* Currently in `README.md` we state that we're Linux-only, and in fact `linux/amd64`-only and ARM may fail.
  > Currently, only `linux/amd64` and `linux/arm64` platforms are supported. If using
`linux/arm64`, some dependencies like `warp-lang` might not install correctly.
* However, in `pyproject.toml` we state we're OS-independent.

I'm down to resolve this either way, but we should unify this story.

ktangsali:
> Re: Linux-only:
> 
> Interesting! I guess `nvtx` is building from source on Mac, then?
> 
> In any case, the main problem is that right now we have a conflict:
> 
> * Currently in `README.md` we state that we're Linux-only, and in fact `linux/amd64`-only and ARM may fail.
> * However, in `pyproject.toml` we state we're OS-independent.
> 
> I'm down to resolve this either way, but we should unify this story.

Yes, I think we need to address the Readme. From my understanding we test the wheels on Mac and native Windows presently for compatability.

peterdsharpe:
Sounds good, thanks @ktangsali ! If we have Windows and Mac CI working, let's go with modifying the `README.md` to reflect this - I'll fix it and revert the Linux-spec changes.

ktangsali:
> > Re: Linux-only:
> > Interesting! I guess `nvtx` is building from source on Mac, then?
> > In any case, the main problem is that right now we have a conflict:
> > 
> > * Currently in `README.md` we state that we're Linux-only, and in fact `linux/amd64`-only and ARM may fail.
> > * However, in `pyproject.toml` we state we're OS-independent.
> > 
> > I'm down to resolve this either way, but we should unify this story.
> 
> Yes, I think we need to address the Readme. From my understanding we test the wheels on Mac and native Windows presently for compatability.

nvtx used to be absent from the required dependencies, but seems like it got added during the refactor merge. We might have to test this again. The last time I checked, it was not working very well on other platforms.

peterdsharpe:
> nvtx used to be absent from the required dependencies, but seems like it got added during the refactor merge. We might have to test this again. The last time I checked, it was not working very well on other platforms.

Got it - yeah. Will we have a chance to test core install on Mac (and Windows, for that matter) before next release?

Flagging this as well: `warp-lang` (also a core dep) also has no wheels on Intel Mac, from what I can see. Are we supporting Intel Mac too, or only ARM Mac?

coreyjadams:
```
physicsnemo-temp on  main [!?] via 🐍 v3.14.0
❯ uv run python
warning: The `extra-build-dependencies` option is experimental and may change without warning. Pass `--preview-features extra-build-dependencies` to disable this warning.
import nvtcPython 3.12.8 (v3.12.8:2dc476bcb91, Dec  3 2024, 14:43:19) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import nvtx
>>>
```

This is on my (arm) mac, so nvtx is not impossible.  We test it in CI already too.

IMO Warp-lang needs to be a core dependency of physicsnemo, and if we're missing support for things like intel-mac we should raise that with the warp team.

ktangsali:
Also seems like `nvtx` works on Windows too now. 

image

ktangsali:
Tagging @abokov-nv - Hi Aleksei, can we verify what all platforms we can support with the latest `main` branch?

peterdsharpe:
> Tagging @abokov-nv - Hi Aleksei, can we verify what all platforms we can support with the latest `main` branch?

^ including whether we're supporting Intel Mac, ARM Mac, or both

pyproject.toml
peterdsharpe: Addressed

ktangsali (approved):

PR #1324, opened 2026-01-14
MGN Refactor

mnabian:
# PhysicsNeMo Pull Request

## Description

This MR refactors the meshgraphnet model and test according to the new model standards.

## Checklist

## Review Process

### DIFF
--- a/physicsnemo/models/meshgraphnet/meshgraphnet.py
+++ b/physicsnemo/models/meshgraphnet/meshgraphnet.py
@@ -17,7 +17,7 @@
 from contextlib import nullcontext
 from dataclasses import dataclass
 from itertools import chain
-from typing import Callable, List, Tuple, Union
+from typing import Callable, Literal, Tuple, Union
 from warnings import warn
 
 import torch
@@ -64,7 +64,7 @@ class MeshGraphNet(Module):
         Number of outputs.
     processor_size : int, optional, default=15
         Number of message passing blocks.
-    mlp_activation_fn : Union[str, List[str]], optional, default="relu"
+    mlp_activation_fn : str, optional, default="relu"
         Activation function to use.
     num_layers_node_processor : int, optional, default=2
         Number of MLP layers for processing nodes in each message passing block.
@@ -87,16 +87,16 @@ class MeshGraphNet(Module):
     num_layers_node_decoder : Union[int, None], optional, default=2
         Number of MLP layers for the node feature decoder. If ``None`` is provided,
         the MLP collapses to an identity function, i.e. no decoder.
-    aggregation : str, optional, default="sum"
-        Message aggregation type.
+    aggregation : Literal["sum", "mean"], optional, default="sum"
+        Message aggregation type. Allowed values are ``"sum"`` and ``"mean"``.
     do_concat_trick : bool, optional, default=False
         Whether to replace concat+MLP with MLP+idx+sum.
     num_processor_checkpoint_segments : int, optional, default=0
         Number of processor segments for gradient checkpointing (0 disables checkpointing).
     checkpoint_offloading : bool, optional, default=False
         Whether to offload the checkpointing to the CPU.
-    norm_type : str, optional, default="LayerNorm"
-        Normalization type [``"LayerNorm"``, ``"TELayerNorm"``].
+    norm_type : Literal["LayerNorm", "TELayerNorm"], optional, default="LayerNorm"
+        Normalization type. Allowed values are ``"LayerNorm"`` and ``"TELayerNorm"``.
 
     Forward
     -------
@@ -105,7 +105,15 @@ class MeshGraphNet(Module):
     edge_features : torch.Tensor
         Input edge features of shape :math:`(N_{edges}, D_{in}^{edge})`.
     graph : :class:`~physicsnemo.nn.gnn_layers.utils.GraphType`
-        Graph container compatible with the mesh graph blocks.
+        Graph connectivity/topology container (PyG).
+        Connectivity/topology only. Do not duplicate node or edge features on the graph;
+        pass them via ``node_features`` and ``edge_features``. If present on
+        the graph, they will be ignored by the model.
+        ``node_features.shape[0]`` must equal the number of nodes in the graph ``graph.num_nodes``.
+        ``edge_features.shape[0]`` must equal the number of edges in the graph ``graph.num_edges``.
+        The current :class:`~physicsnemo.nn.gnn_layers.graph_types.GraphType` resolves to
+        PyTorch Geometric objects (``torch_geometric.data.Data`` or ``torch_geometric.data.HeteroData``). See
+        :mod:`physicsnemo.nn.gnn_layers.graph_types` for the exact alias and requirements.
 
     Outputs
     -------
@@ -137,8 +145,7 @@ class MeshGraphNet(Module):
 
     Note
     ----
-    Reference: Pfaff, Tobias, et al. "Learning mesh-based simulation with graph networks."
-    arXiv preprint arXiv:2010.03409 (2020).
+    Reference: `Learning Mesh-Based Simulation with Graph Networks <https://arxiv.org/pdf/2010.03409>`.
 
     See also :class:`~physicsnemo.nn.gnn_layers.mesh_graph_mlp.MeshGraphMLP`,
     :class:`~physicsnemo.nn.gnn_layers.mesh_edge_block.MeshEdgeBlock`,
@@ -151,7 +158,7 @@ class MeshGraphNet(Module):
         input_dim_edges: int,
         output_dim: int,
         processor_size: int = 15,
-        mlp_activation_fn: Union[str, List[str]] = "relu",
+        mlp_activation_fn: str = "relu",
         num_layers_node_processor: int = 2,
         num_layers_edge_processor: int = 2,
         hidden_dim_processor: int = 128,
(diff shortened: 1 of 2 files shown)


physicsnemo/models/meshgraphnet/meshgraphnet.py
```diff
@@ -51,55 +52,71 @@ class MetaData(ModelMetaData):
...
+        Input edge features of shape :math:`(N_{edges}, D_{in}^{edge})`.
+    graph : :class:`~physicsnemo.nn.gnn_layers.utils.GraphType`
+        Graph container compatible with the mesh graph blocks.
+
+    Outputs
+    -------
+    torch.Tensor
+        Output node features of shape :math:`(N_{nodes}, D_{out})`.
 
     Example
```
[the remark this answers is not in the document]
mnabian: Fixed

physicsnemo/models/meshgraphnet/meshgraphnet.py
```diff
@@ -204,20 +230,86 @@ def __init__(
...
+    edge_features : torch.Tensor
+        Edge features of shape :math:`(N_{edges}, D_{edge})`.
+    graph : :class:`~physicsnemo.nn.gnn_layers.utils.GraphType`
+        Graph container.
+
+    Outputs
+    -------
+    torch.Tensor
+        Updated node features of shape :math:`(N_{nodes}, D_{node})`.
+    """
```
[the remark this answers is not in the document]
mnabian: Removed the parameters from the `__init__` method.

coreyjadams (approved):
LGTM @mnabian !

physicsnemo/models/meshgraphnet/meshgraphnet.py
```diff
@@ -293,19 +397,23 @@ def set_checkpoint_offload_ctx(self, enabled: bool):
             self.checkpoint_offload_ctx = nullcontext()
 
     def set_checkpoint_segments(self, checkpoint_segments: int):
```
CharlelieLrt: I'm just wondering if this `set_checkpoint_segments` (and other methods as well) is supposed for internal use or if it's meant to be exposed?

- If it's internal only, could it be possible to make it semi-private (i.e. rename to `_set_checkpoint_segments`?
- If it's external, it should be tested in CI.

Note: if it's a public method, then it will automatically appear in the online docs for the class. If it's private, it won't, so the choice matters.

test/models/meshgraphnet/test_meshgraphnet.py
CharlelieLrt, line 1: I think some stuff is missing in tests:

- In the constructor test, we need to check that the values of public attributes of the object are correctly set (protect against unexpected disapearance/modification of public attributes). However, not necessary to do a forward pass in this test

- All public methods (if there are any), need to be tested the same way forward is

physicsnemo/models/meshgraphnet/meshgraphnet.py
```diff
@@ -51,55 +52,71 @@ class MetaData(ModelMetaData):
...
-    num_processor_checkpoint_segments: int, optional
-        Number of processor segments for gradient checkpointing, by default 0 (checkpointing disabled)
-    checkpoint_offloading: bool, optional
-        Whether to offload the checkpointing to the CPU, by default False
-
-    Example
+        Number of outputs.
+    processor_size : int, optional, default=15
+        Number of message passing blocks.
+    mlp_activation_fn : Union[str, List[str]], optional, default="relu"
```
CharlelieLrt: What happens if you pass a `List[str]`? Does the length need to match the number of blocks? I think this should be documented + validated in the `__init__`

physicsnemo/models/meshgraphnet/meshgraphnet.py
```diff
@@ -51,55 +52,71 @@ class MetaData(ModelMetaData):
...
+        Hidden layer size for the edge feature encoder.
+    num_layers_edge_encoder : Union[int, None], optional, default=2
+        Number of MLP layers for the edge feature encoder. If ``None`` is provided,
+        the MLP collapses to an identity function, i.e. no edge encoder.
+    hidden_dim_node_decoder : int, optional, default=128
+        Hidden layer size for the node feature decoder.
+    num_layers_node_decoder : Union[int, None], optional, default=2
+        Number of MLP layers for the node feature decoder. If ``None`` is provided,
+        the MLP collapses to an identity function, i.e. no decoder.
+    aggregation : str, optional, default="sum"
```
CharlelieLrt: I think for all these `str` optional arguments it's better to use the literal type: `Literal["sum", ..
]`, as it provides stronger type checking and clarity on the accepted values. For example, here by reading the docstring I am not sure what are the accepted values beyond `"sum"`

physicsnemo/models/meshgraphnet/meshgraphnet.py
```diff
@@ -122,6 +139,10 @@ class MeshGraphNet(Module):
     ----
     Reference: Pfaff, Tobias, et al. "Learning mesh-based simulation with graph networks."
     arXiv preprint arXiv:2010.03409 (2020).
```
CharlelieLrt: Prefer direct link to arxiv paper

physicsnemo/models/meshgraphnet/meshgraphnet.py
```diff
@@ -51,55 +52,71 @@ class MetaData(ModelMetaData):
...
+
+    Forward
     -------
-    >>> # `norm_type` in MeshGraphNet is deprecated,
+    node_features : torch.Tensor
+        Input node features of shape :math:`(N_{nodes}, D_{in}^{node})`.
+    edge_features : torch.Tensor
+        Input edge features of shape :math:`(N_{edges}, D_{in}^{edge})`.
+    graph : :class:`~physicsnemo.nn.gnn_layers.utils.GraphType`
+        Graph container compatible with the mesh graph blocks.
```
CharlelieLrt: Maybe unrelated to this PR, but I was a little confused by this `graph` argument when working with the MGN in another project, and I found this difficult to use. What is this object expected to contain: only the connectivity? Or  connectivity + some duplication of `node_features` and `edge_features`? Also, how to create this argument to pass it top the model?
I think the MGN would benefit from a bit more documentation here, to make it easier for the user to get started using it without having to dive too deep into another abstraction (the `GraphType ` class)

Alexey-Kamenev: This type alias was introduced to simplify migrating from DGL to PyG. [Previously](https://github.com/NVIDIA/physicsnemo/blob/ee4a52651f56bdd6ee5d7940fee33c1490e0ff05/physicsnemo/models/meshgraphnet/meshgraphnet.py#L210) `graph` parameter type hint was something like:
```py
    def forward(
        self,
        node_features: Tensor,
        edge_features: Tensor,
        graph: Union[DGLGraph, List[DGLGraph], CuGraphCSC],

```
so instead of adding PyG graph type (`torch_geometric.data.Data`) to the union, I added a type alias `GraphType`.
Considering that DGL is gone and so is cugraphops, this `GraphType` alias can be potentially replaced with above-mentioned PyG graph type. This won't solve the confusion though as PyG graph types (and DGL for that matter) are somewhat complicated (they can hold connectivity, attributes, all of the above, and none of the above).

CharlelieLrt: Makes sense, thanks for the details @Alexey-Kamenev!

> This won't solve the confusion though as PyG graph types are somewhat complicated

Yeah, definitely. But at least, if the type hint is just PyG graph type, then it's a little more immediate to figure out that it can contain "whatever PyG graph type supports, no matter how complicated".

Or maybe just one line in the docstring like "Instances of `torch_geometric.data.Data` graph are accepted"... just to inform the user that there is no need to convert the graph object to another abstraction internal to physicsnemo (that's what I thought at the beginning).

But there is still something unclear to me: nodes and edge features are already passed in `node_features` and `edge_features `, so what happens to the nodes and edge features contained in `graph`? Are we not passing twice the same inputs? Or are they concatenated? Or just ignored?

mnabian: @CharlelieLrt this has been clarified in the docstrings. Graph only represent connectivity, and should not contain `node_features` and `edge_features`. If those are specified, they will be simply ignored.

physicsnemo/models/meshgraphnet/meshgraphnet.py
mnabian: Changed to direct link.

physicsnemo/models/meshgraphnet/meshgraphnet.py
mnabian: Switched from `str` to `Literal`.

physicsnemo/models/meshgraphnet/meshgraphnet.py
CharlelieLrt: Great, thanks for the clarification!

physicsnemo/models/meshgraphnet/meshgraphnet.py
mnabian: The implementation actually presently does not support list of activations. I fixed the docstrings.

test/models/meshgraphnet/test_meshgraphnet.py
mnabian: The constructor test is modified to check public attributes. Some of the MGN methods are made private.

physicsnemo/models/meshgraphnet/meshgraphnet.py
mnabian: Yes, we don;t need to have those as public methods. I've changed them to private methods now.

CharlelieLrt (approved):
LGTM, thanks for addressing the comments!

PR #1330, opened 2026-01-16
Refactors for `FullyConnected` and `UNet` models

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Brings `FullyConnected` (both model and PN.nn variants) and `UNet` into compliance with new refactor standards.

## Checklist

## Review Process

### DIFF
--- a/physicsnemo/nn/conv_layers.py
+++ b/physicsnemo/nn/conv_layers.py
@@ -190,7 +190,9 @@ class Conv2d(torch.nn.Module):
         down: bool = False,
         resample_filter: List[int] = [1, 1],
         fused_resample: bool = False,
-        init_mode: Literal["xavier_uniform", "xavier_normal", "kaiming_uniform", "kaiming_normal"] = "kaiming_normal",
+        init_mode: Literal[
+            "xavier_uniform", "xavier_normal", "kaiming_uniform", "kaiming_normal"
+        ] = "kaiming_normal",
         init_weight: float = 1.0,
         init_bias: float = 0.0,
         fused_conv_bias: bool = False,
--- a/physicsnemo/nn/fully_connected_layers.py
+++ b/physicsnemo/nn/fully_connected_layers.py
@@ -495,7 +495,9 @@ class Linear(Module):
         in_features: int,
         out_features: int,
         bias: bool = True,
-        init_mode: Literal["xavier_uniform", "xavier_normal", "kaiming_uniform", "kaiming_normal"] = "kaiming_normal",
+        init_mode: Literal[
+            "xavier_uniform", "xavier_normal", "kaiming_uniform", "kaiming_normal"
+        ] = "kaiming_normal",
         init_weight: int = 1,
         init_bias: int = 0,
         amp_mode: bool = False,


peterdsharpe:
Blocked by possible fixes in https://github.com/NVIDIA/physicsnemo/pull/1332

physicsnemo/models/unet/unet.py
```diff
@@ -36,23 +37,32 @@
...
+    channel dimension while preserving the spatial structure.
 
-    Parameters:
+    Parameters
     ----------
-        normalized_shape (int or list/tuple of ints): Input shape from an expected input of size. If a single integer is used,
-            it is treated as a singleton list.
-        eps (float, optional): A value added to the denominator for numerical stability. Default is 1e-5.
-        elementwise_affine (bool, optional): Whether to learn affine parameters (scale and shift). Default is True.
+    normalized_shape : int or tuple of ints
```
CharlelieLrt: `int | Tuple[int]`

Note: Since yesterday or so Cursor has started replacing all my `List[int]` or `Tuple[int]` with `tuple of ints`... not sure why. Maybe the same happened there

physicsnemo/models/unet/unet.py
```diff
@@ -36,23 +37,32 @@
 
 
 class ReshapedLayerNorm(te.LayerNorm if te else nn.LayerNorm):
```
CharlelieLrt, line 39: Genuine question: this is implicit as the user will never know if TE is being used or not. Are we okay with handling the optional dependencies this way? I thought there was a validation decorator that would do at least some checks on those, but I haven't followed this conversation closely enough

physicsnemo/models/unet/unet.py
```diff
@@ -61,35 +71,56 @@ def __init__(
...
         x = x.view(shape[0], shape[1], -1).transpose(1, 2).contiguous()
         x = super().forward(x)
         x = x.transpose(1, 2).contiguous().view(shape)
         return x
 
 
-class ConvBlock(nn.Module):
-    """
-    A convolutional block, followed by an optional normalization and activation.
+class ConvBlock(Module):
```
CharlelieLrt: Can we rename it `Conv3DBlock` to make more explicit the fact that it is a 3D convolution?

Same comments for all modules here that are 3D

physicsnemo/models/unet/unet.py
```diff
@@ -61,35 +71,56 @@ def __init__(
...
+        Number of input channels :math:`C_{in}`.
+    out_channels : int
+        Number of output channels :math:`C_{out}`.
+    kernel_size : int or tuple, optional, default=3
+        Size of the convolving kernel.
+    stride : int or tuple, optional, default=1
+        Stride of the convolution.
+    padding : int or tuple, optional, default=1
+        Padding added to all sides of the input.
+    dilation : int or tuple, optional, default=1
```
CharlelieLrt: `int | Tuple[???]`
Not sure why, but all types descriptions got kind of messed up in this file...

physicsnemo/models/unet/unet.py
```diff
@@ -61,35 +71,56 @@ def __init__(
...
+    groups : int, optional, default=1
+        Number of blocked connections from input to output channels.
+    bias : bool, optional, default=True
+        If ``True``, adds a learnable bias to the output.
+    padding_mode : str, optional, default="zeros"
+        Padding mode: ``"zeros"``, ``"reflect"``, ``"replicate"``, or ``"circular"``.
+    activation : Optional[str], optional, default="relu"
+        Activation function name from :mod:`torch.nn.functional`.
+    normalization : Optional[str], optional, default="groupnorm"
+        Normalization type: ``"groupnorm"``, ``"batchnorm"``, or ``"layernorm"``.
```
CharlelieLrt: I'm trying to wrap me head around this: `Optional[str]` means that it's possible to pass `None`; what happens if `normalization` is None? There is just no normalization applied?

physicsnemo/models/unet/unet.py
```diff
@@ -151,35 +182,57 @@ def __init__(
...
         x = self.conv3d(x)
         x = self.norm(x)
         x = self.activation(x)
         return x
 
 
-class ConvTranspose(nn.Module):
-    """
-    A transposed convolutional block, followed by an optional normalization and activation.
+class ConvTranspose(Module):
```
CharlelieLrt: Possibility to rename with "3D" in the class name?

physicsnemo/models/unet/unet.py
```diff
@@ -243,29 +296,44 @@ def __init__(
...
         x = self.conv3d_transpose(x)
         x = self.norm(x)
         x = self.activation(x)
         return x
 
 
-class Pool3d(nn.Module):
-    """
-    A pooling block that applies a specified 3D pooling operation over an input signal.
+class Pool3d(Module):
```
CharlelieLrt: Should use PascalCase, so `Pool3D`

physicsnemo/models/unet/unet.py
```diff
@@ -243,29 +296,44 @@ def __init__(
...
-        pooling_type (str): Type of pooling operation ('AvgPool3d', 'MaxPool3d', or custom types if supported).
-        kernel_size (int, tuple): Size of the window to take a pool over.
-        stride (int, tuple, None): Stride of the pooling operation. Default is None (same as kernel_size).
-        padding (int, tuple): Implicit zero padding to be added on both sides of the input. Default is 0.
-        dilation (int, tuple): Control the spacing between the kernel points; useful for dilated pooling. Default is 1.
-        ceil_mode (bool): When True, will use ceil instead of floor to compute the output shape. Default is False.
-        count_include_pad (bool): Only used for AvgPool3d. If True, will include the zero-padding in the averaging calculation.
-
-    Returns:
+    pooling_type : str, optional, default="AvgPool3d"
```
CharlelieLrt: Maybe some of these `str` arguments would benefit from being `Literal["AvgPool3d", "MaxPool3d"]` (at least when the number of possible values is finite, small, and not expected to significantly expand in the future)

physicsnemo/models/unet/unet.py
```diff
@@ -305,37 +373,53 @@ def __init__(
...
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """Forward pass through the pooling layer."""
         return self.pooling(x)
 
 
-class AttentionBlock(nn.Module):
-    """
-    Attention block for the skip connections using LayerNorm instead of BatchNorm.
+class AttentionBlock(Module):
```
CharlelieLrt: We already have a few attention layers in `physicsnemo/nn`, including one for a UNet. Would it be possible to reuse any of them here?

physicsnemo/models/unet/unet.py
```diff
@@ -344,33 +428,61 @@ def __init__(self, F_g, F_l, F_int):
 
         self.relu = nn.ReLU(inplace=True)
 
-    def forward(self, g, x):
-        # g is the decoder's upsampled features (query)
-        # x is the encoder's skip connection features (key/value)
-        g1 = self.W_g(g)
-        x1 = self.W_x(x)
+    def forward(self, g: torch.Tensor, x: torch.Tensor) -> torch.Tensor:
```
CharlelieLrt: No jaxtyping here?

physicsnemo/models/unet/unet.py
```diff
@@ -36,23 +37,32 @@
...
 
-    Returns:
+    Outputs
     -------
-        torch.Tensor: The input tensor after applying reshaped layer normalization.
+    torch.Tensor
+        Normalized tensor with same shape as input.
     """
 
     def __init__(
```
CharlelieLrt, line 73: type hints missing in the `__init__`

physicsnemo/models/unet/unet.py
```diff
@@ -61,35 +71,56 @@ def __init__(
         super().__init__(normalized_shape, eps, elementwise_affine)
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
```
CharlelieLrt: No jaxtyping here?

physicsnemo/models/unet/unet.py
```diff
@@ -61,35 +71,56 @@ def __init__(
...
+    x : torch.Tensor
+        Input tensor of shape :math:`(B, C_{in}, D, H, W)`.
+
+    Outputs
+    -------
+    torch.Tensor
+        Output tensor of shape :math:`(B, C_{out}, D', H', W')`.
     """
 
     def __init__(
```
CharlelieLrt, line 136: The `Union[tuple]` types in the `__init__` might be better if they are replaced with `int | List[int]`. I think this occurs not only in this `__init__` but also for multiple other modules in this file

physicsnemo/models/unet/unet.py
```diff
@@ -151,35 +182,57 @@ def __init__(
             self.norm = nn.Identity()
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """Forward pass through the convolutional block."""
```
CharlelieLrt, line 198: There is no jaxtyping for this forward for the tensor argument, I am not sure if this is on purpose or not. I have seen this here and also in the forward method of other modules defined in this file.

physicsnemo/models/unet/unet.py
```diff
@@ -650,13 +855,42 @@ def __init__(
                 ]
             )
 
-    def checkpointed_forward(self, layer, x):
-        """Wrapper to apply gradient checkpointing if enabled."""
+    def checkpointed_forward(self, layer: nn.Module, x: torch.Tensor) -> torch.Tensor:
```
CharlelieLrt: Should this be a private method?

physicsnemo/models/unet/unet.py
```diff
@@ -539,25 +681,86 @@ class MetaData(ModelMetaData):
...
+    padding_mode : str, optional, default="zeros"
+        Padding mode.
+    pooling_type : str, optional, default="MaxPool3d"
+        Pooling type: ``"AvgPool3d"`` or ``"MaxPool3d"``.
+    pool_size : int, optional, default=2
+        Pooling window size.
+    normalization : Optional[str], optional, default="groupnorm"
+        Normalization type.
+    normalization_args : Optional[dict], optional, default=None
+        Additional normalization arguments.
```
CharlelieLrt: This is not ideal, because the normalization module is not even in the docs, so the user if left on its own to figure out what can be passed here. Actually, it's most likely impossible to do so without opening and reading the source code of the `UNet`.

Not sure you can do anything about it though?

physicsnemo/models/unet/unet.py
CharlelieLrt, line 1: One question about the very last block of code in this file (I can't directly comment on it, since it was not edited in this PR):

Do we want to keep this?

```python
if __name__ == "__main__":
    inputs = torch.randn(1, 1, 96, 96, 96).cuda()
    print("The shape of inputs: ", inputs.shape)
    model = UNet(
        in_channels=1,
        out_channels=1,
        model_depth=5,
        feature_map_channels=[64, 64, 128, 128, 256, 256, 512, 512, 1024, 1024],
        num_conv_blocks=2,
    ).cuda()
    x = model(inputs)
    print("model: ", model)
    print("The shape of output: ", x.shape)
```

test/models/mlp/test_fully_connected.py
```diff
@@ -41,79 +41,70 @@ def test_fully_connected_forward(device):
...
-            bsize = random.randint(1, 16)
-            invar = torch.randn(bsize, kw_args["in_features"]).to(device)
-            outvar = model(invar)
-            assert outvar.shape == (bsize, kw_args["out_features"])
+@pytest.mark.parametrize(
+    "config",
+    ["default", "custom"],
+    ids=["with_defaults", "with_custom_args"],
+)
+def test_fully_connected_constructor(device, config):
```
CharlelieLrt, line 49: Ideally the constructor test should check that the values of all public attributes are correctly set as expected. However, it does not need to run a forward and check the shape.

test/models/unet/test_unet.py
```diff
@@ -47,104 +43,141 @@ def test_unet_forward(device, pytestconfig):
...
 @requires_module(["transformer_engine"])
-@pytest.mark.parametrize("device", ["cuda:0", "cpu"])
-def test_unet_constructor(device, pytestconfig):
-    """Test unet constructor options"""
+@pytest.mark.parametrize(
+    "config",
+    ["default", "custom"],
+    ids=["with_defaults", "with_custom_args"],
+)
+def test_unet_constructor(device, config, pytestconfig):
```
CharlelieLrt: Same comment here for the constructor that should validate the values of public attribute, if there is any to validate

physicsnemo/models/unet/unet.py
peterdsharpe: It's a good question! This is what we landed on in one of our refactor meetings awhile back - we thought that the silent fallback was preferable to failure. That said, I'll add a comment in the ReshapedLayerNorm docstring so that this is more visible.

physicsnemo/models/unet/unet.py
peterdsharpe: fixed

physicsnemo/models/unet/unet.py
peterdsharpe: fixed

physicsnemo/models/unet/unet.py
peterdsharpe: Fixed

physicsnemo/models/unet/unet.py
peterdsharpe: Fixed

physicsnemo/models/unet/unet.py
peterdsharpe: Yep, exactly! Added more documentation to clarify.

physicsnemo/models/unet/unet.py
peterdsharpe: Fixed

physicsnemo/models/unet/unet.py
peterdsharpe: Fixed, and in fact using `Sequence` where possible for maximum generality

physicsnemo/models/unet/unet.py
peterdsharpe: Fixed! This was not on purpose, these were just not jaxtyped since they were inner classes rather than user-facing ones. That said, it definitely doesn't hurt to jaxtype here too - added!

physicsnemo/models/unet/unet.py
peterdsharpe: Yeah - my preferred API design pattern would have been:

Rather than having `normalization: str` and `normalization_args` as two separate input arguments, have the end-user pass in just a single argument `norm_layer: nn.Module | physicsnemo.core.module.Module`, as an object. That way the user can define the normalization layer, along with any kwarg tweaks, on their end.

That said, this would break back-compatibility. What do you think? I'll think about this more!

physicsnemo/models/unet/unet.py
peterdsharpe: Good call, eliminated

physicsnemo/models/unet/unet.py
peterdsharpe: It's a good question! So, there are some subtle differences between this AttentionBlock (now Attention3DBlock after renaming) and the UNetAttention one.

Specifically, the challenging ones to abstract would be:
- This UNet is attention gating, while the UNetAttention is self-attention - so fundamentally different mechanisms here.
- This is 3D while UNetAttention is 2D. It's possible to abstract this out, but not trivial to rewrite in a spatially-dimensionally-agnostic way.

There are some smaller differences (e.g. UNetAttention has a residual mechanism; this does not) too, but overall I think that at least in this PR, it's probably best to keep these separate. We can definitely revisit this after the refactor though!

physicsnemo/models/unet/unet.py
CharlelieLrt: After offline discussion, it appears this is more tricky than it looks. Will defer refactor of this pattern to future PR

test/common/optimization.py
```diff
@@ -315,7 +315,7 @@ def validate_combo_optims(
 
     # Torch script, need to save it as seperate model since TS model doesnt have meta
     if model.meta.jit:
-        fwd_model = torch.jit.script(model)
+        fwd_model = torch.compile(model)
```
mnabian, line 318: I think this change could be confusing. Why are we replacing `torch.jit.script` with `torch.compile` under the same jit metadata flag? If the goal is to deprecate TorchScript support, it might be clearer to rename the metadata (e.g., `compile` or `torch_compile`) rather than reusing jit.

Also, not all models that are TorchScript-compatible are necessarily compatible with torch.compile, so I’m curious whether this change fails any tests.

peterdsharpe: Good feedback here! This work is refactored into https://github.com/NVIDIA/physicsnemo/pull/1332, which addresses the TorchScript -> Compile switch.

On the `model.meta.jit` naming - I'll play with switching this over to `model.meta.compile` on the new PR, and see if anything breaks.

To the latter point on cross-compatibility - fortunately on the new branch, it seems that almost everything is playing nicely with a simple hot-swap of TorchScript to torch.compile! Some issues with `physicsnemo.nn.interpolation` in two functions, but that's it.

PR #1333, opened 2026-01-16
Adds `physicsnemo.mesh`, Part 2/2

peterdsharpe:
# PhysicsNeMo Pull Request

## Checklist

## Review Process

### DIFF
--- a/v2.0-MIGRATION-GUIDE.md
+++ b/v2.0-MIGRATION-GUIDE.md
@@ -89,15 +89,8 @@ mesh = Mesh(
 `points` is a float tensor of vertex coordinates; `cells` is an integer tensor
 of indices into `points` that define each simplex. All geometry and field data
 moves together under `.to("cuda")` / `.to("cpu")`, and most operations are
-differentiable through PyTorch autograd.
-
-Capabilities include discrete calculus (gradient, divergence, curl,
-Laplace-Beltrami via DEC and least-squares methods, with intrinsic and
-extrinsic variants), Gaussian and mean curvature, subdivision (linear, Loop,
-Butterfly), mesh repair, BVH spatial queries, topology analysis, visualization
-(`mesh.draw()`), and a library of ready-made primitive meshes. The only core
-dependencies are PyTorch and TensorDict; PyVista is optional for I/O and 3D
-visualization. See `physicsnemo/mesh/README.md` for the full feature matrix.
+differentiable through PyTorch autograd. See `physicsnemo/mesh/README.md` for
+the full feature matrix.
 
 ## Updating your code
 


physicsnemo/mesh/boundaries/_boundary_extraction.py
```diff
@@ -0,0 +1,172 @@
...
+        >>> boundary_cells, boundary_data = extract_boundary_mesh_data(vol_mesh)
+        >>> boundary_mesh = Mesh(points=vol_mesh.points, cells=boundary_cells, cell_data=boundary_data)
+        >>> assert boundary_mesh.n_manifold_dims == 2  # Surface triangles
+    """
+    from physicsnemo.mesh.boundaries._facet_extraction import (
+        _aggregate_point_data_to_facets,
+        categorize_facets_by_count,
+        compute_aggregation_weights,
+        extract_candidate_facets,
+    )
```
coreyjadams: I'm surprised this passes pre-commit, usually it complains if imports come after code definitions?

physicsnemo/mesh/sampling/sample_data.py
```diff
@@ -0,0 +1,633 @@
...
+    # Shape: (n_cells, n_vertices_per_cell - 1, n_spatial_dims)
+    v0 = cell_vertices[:, 0:1, :]  # (n_cells, 1, n_spatial_dims)
+    relative_vectors = (
+        cell_vertices[:, 1:, :] - v0
+    )  # (n_cells, n_manifold_dims, n_spatial_dims)
+
+    ### Compute query points relative to v0
+    # Broadcast query_points and v0 for all combinations
+    # Shape: (n_queries, n_cells, n_spatial_dims)
+    query_relative = query_points.unsqueeze(1) - v0.squeeze(1).unsqueeze(0)
```
coreyjadams, line 155: Any issues with memory explosion here?

physicsnemo/mesh/sampling/sample_data.py
```diff
@@ -0,0 +1,633 @@
...
+    bary_coords = compute_barycentric_coordinates(query_points, cell_vertices)
+
+    ### Determine which query-cell pairs have the point inside
+    is_inside = (bary_coords >= -tolerance).all(dim=-1)
+
+    ### For each query, collect all containing cells
+    containing_cells = []
+    for i in range(len(query_points)):
+        containing = torch.where(is_inside[i])[0]
+        containing_cells.append(containing)
```
coreyjadams: Is there a way to vectorize this?

physicsnemo/mesh/sampling/sample_data.py
```diff
@@ -0,0 +1,633 @@
...
+        b = query_relative.unsqueeze(-1)
+        weights_1_to_n = torch.linalg.lstsq(A, b).solution.squeeze(-1)
+
+    ### Compute w_0 = 1 - sum(w_i for i=1..n)
+    w_0 = 1.0 - weights_1_to_n.sum(dim=-1, keepdim=True)
+
+    ### Concatenate to get all barycentric coordinates
+    barycentric_coords = torch.cat([w_0, weights_1_to_n], dim=-1)
+
+    return barycentric_coords
```
coreyjadams: I think it will be valuable to return not just the barycentric coords found by the solve but also the residual error: using the vertices + coords, compute residual = | point - barycentric_solution_point|.  This will let algorithms apply a tolerance for skin depth more correctly too.

physicsnemo/mesh/sampling/sample_data.py
```diff
@@ -0,0 +1,633 @@
...
+        return projected, dist_sq
+
+    ### Otherwise, find the closest face
+    # For simplicity, check all faces (subsets of vertices)
+    n_vertices = cell_vertices.shape[0]
+    best_projected = None
+    best_dist_sq = float("inf")
+
+    # Try all (n-1)-dimensional faces
+    for i in range(n_vertices):
```
coreyjadams: With the addition of a residual calculation in the barycentric calculation, you ought to be able to remove this loop and do a `min` of `arg_min`, etc, on the residuals.

physicsnemo/mesh/sampling/sample_data.py
```diff
@@ -0,0 +1,633 @@
...
+    """
+    ### Compute all cell centroids
+    cell_centroids = mesh.cell_centroids  # (n_cells, n_spatial_dims)
+
+    ### Compute distances from all queries to all cell centroids
+    # query_points: (n_queries, n_spatial_dims)
+    # cell_centroids: (n_cells, n_spatial_dims)
+    # Broadcast to (n_queries, n_cells, n_spatial_dims)
+    diffs = query_points.unsqueeze(1) - cell_centroids.unsqueeze(0)
+    distances_sq = (diffs**2).sum(dim=-1)  # (n_queries, n_cells)
```
coreyjadams: Why do this with N^2 and not a kNN / tree search with k=1?

physicsnemo/mesh/sampling/sample_data.py
```diff
@@ -0,0 +1,633 @@
...
+
+
+def sample_data_at_points(
+    mesh: "Mesh",
+    query_points: torch.Tensor,
+    data_source: Literal["cells", "points"] = "cells",
+    multiple_cells_strategy: Literal["mean", "nan"] = "mean",
+    project_onto_nearest_cell: bool = False,
+    tolerance: float = 1e-6,
+) -> TensorDict:
```
coreyjadams, line 739: Can we improve the name or docstring of this function?  It confused me at first: sample I suppose does not mean "random sample" here.  Perhaps `extract_mesh_data_at_points` is more informative?

Or, improve the docstring to be a little more descriptive of the task?

physicsnemo/mesh/sampling/sample_data_hierarchical.py
```diff
@@ -0,0 +1,313 @@
...
+from physicsnemo.mesh.sampling.sample_data import (
+    compute_barycentric_coordinates_pairwise,
+)
+from physicsnemo.mesh.spatial import BVH
+
+if TYPE_CHECKING:
+    from physicsnemo.mesh.mesh import Mesh
+
+
+def sample_data_at_points(
```
coreyjadams: I haven't reviewed this implementation here closely.  But it begs the question: why not use something like Warp for the BVH?  And, with the new functional implementation in physicsnemo, is it worth deploying this as a functional?

physicsnemo/mesh/smoothing/laplacian.py
```diff
@@ -0,0 +1,487 @@
...
+        convergence: Convergence criterion relative to bounding box diagonal.
+            Stops early if max vertex displacement < convergence * bbox_diagonal.
+            Set to 0.0 to disable early stopping. Default: 0.0
+        feature_angle: Angle threshold (degrees) for sharp edge detection.
+            Edges with dihedral angle > feature_angle are considered sharp features.
+            Only used for codimension-1 manifolds. Default: 45.0
+        boundary_smoothing: If True, boundary vertices remain fixed during smoothing.
+            If False, boundary vertices are smoothed like interior vertices. Default: True
+        feature_smoothing: If True, vertices on sharp features remain fixed.
+            If False, feature vertices are smoothed. Default: False
```
coreyjadams: I must be misreading these: `boundary_smoothing=False` causes boundary vertices to be smoothed like interior vertices?

physicsnemo/mesh/smoothing/laplacian.py
```diff
@@ -0,0 +1,487 @@
...
+    ### Save original positions for constrained vertices
+    original_points = mesh.points.clone()
+
+    ### Identify constrained vertices (boundaries and features)
+    constrained_vertices = torch.zeros(n_points, dtype=torch.bool, device=device)
+
+    if boundary_smoothing:
+        # Boundary vertices should not move
+        boundary_vertex_mask = _get_boundary_vertices(mesh, edges)
+        constrained_vertices |= boundary_vertex_mask
```
coreyjadams, line 157: Didn't know this was a possibility on tensors!

physicsnemo/mesh/smoothing/laplacian.py
```diff
@@ -0,0 +1,487 @@
...
+            -weighted_vectors,
+        )
+
+        # Accumulate weight sums
+        weight_sum.scatter_add_(0, edges[:, 0], edge_weights)
+        weight_sum.scatter_add_(0, edges[:, 1], edge_weights)
+
+        ### Normalize by total weight per vertex
+        # Avoid division by zero for isolated vertices
+        weight_sum = weight_sum.clamp(min=1e-10)
```
coreyjadams: I do not know if you have any constraints on precision for `mesh` data - in fact I assume you don't.  In fp16, 1e-10 rounds to 0 and this clamp is useless.

I think it's fine in fp32 and bf16, but `if dtype==float16` needs a different clamp min.

physicsnemo/mesh/spatial/bvh.py
```diff
@@ -0,0 +1,380 @@
...
+
+import torch
+from tensordict import tensorclass
+
+if TYPE_CHECKING:
+    from physicsnemo.mesh.mesh import Mesh
+
+
+@tensorclass
+class BVH:
```
coreyjadams, line 236: I haven't looked closely at the implementation here.  But I think we need a good reason to not just be wrapping this: https://nvidia.github.io/warp/api_reference/_generated/warp.Bvh.html

physicsnemo/mesh/validation/quality.py
```diff
@@ -0,0 +1,161 @@
...
+    n_cells = mesh.n_cells
+    n_verts_per_cell = mesh.n_manifold_dims + 1
+
+    # Compute all pairwise edge lengths within each cell
+    edge_lengths_list = []
+    for i in range(n_verts_per_cell):
+        for j in range(i + 1, n_verts_per_cell):
+            edge = cell_vertices[:, j] - cell_vertices[:, i]
+            length = torch.norm(edge, dim=-1)
+            edge_lengths_list.append(length)
```
coreyjadams: vectorize?

physicsnemo/mesh/validation/statistics.py
```diff
@@ -0,0 +1,140 @@
...
+    ### Compute edge length statistics
+    cell_vertices = mesh.points[mesh.cells]  # (n_cells, n_verts, n_dims)
+    n_verts_per_cell = mesh.n_manifold_dims + 1
+
+    edge_lengths_list = []
+    for i in range(n_verts_per_cell):
+        for j in range(i + 1, n_verts_per_cell):
+            edge = cell_vertices[:, j] - cell_vertices[:, i]
+            length = torch.norm(edge, dim=-1)
+            edge_lengths_list.append(length)
```
coreyjadams: also vectorize?

physicsnemo/mesh/validation/validate.py
```diff
@@ -0,0 +1,319 @@
...
+
+                if raise_on_error:
+                    raise ValueError(
+                        f"Found {n_duplicates} pairs of duplicate vertices "
+                        f"(within tolerance={tolerance}).\n"
+                        f"First few pairs: {results['duplicate_vertex_pairs'][:5].tolist()}"
+                    )
+        else:
+            # For large meshes, skip exact check (too expensive)
+            # Could implement approximate duplicate detection with spatial hashing
```
coreyjadams: Using a kNN might be appropriate here?

coreyjadams: In fact it's probably appropriate for both implementations, small and large.

physicsnemo/mesh/validation/validate.py
```diff
@@ -0,0 +1,319 @@
...
+            results["n_duplicate_vertices"] = -1  # Not checked
+
+    ### Check for degenerate cells
+    if check_degenerate_cells and mesh.n_cells > 0:
+        # Compute cell areas
+        areas = mesh.cell_areas
+
+        # Find cells with area below tolerance
+        degenerate_mask = areas < tolerance
```
coreyjadams: Is it correct to use the same tolerance for distance-like displacements and area-like units?

physicsnemo/mesh/mesh.py
```diff
@@ -998,6 +1107,105 @@ def slice_cells(
...
+        """
+        from physicsnemo.mesh.sampling import sample_random_points_on_cells
+
+        return sample_random_points_on_cells(
+            mesh=self,
+            cell_indices=cell_indices,
+            alpha=alpha,
+        )
+
+    def sample_data_at_points(
```
coreyjadams, line 1127: Same comment about naming as in sampling.

physicsnemo/mesh/mesh.py
```diff
@@ -1138,6 +1346,317 @@ def point_data_to_cell_data(self, overwrite_keys: bool = False) -> "Mesh":
...
+            point_data=filtered_point_data,  # User data only, no cached properties
+            cell_data=facet_cell_data,  # Aggregated cell data
+            global_data=self.global_data,  # Share global data
+        )
+
+    def get_boundary_mesh(
+        self,
+        data_source: Literal["points", "cells"] = "cells",
+        data_aggregation: Literal["mean", "area_weighted", "inverse_distance"] = "mean",
+    ) -> "Mesh":
```
coreyjadams, line 1450: In something like drivaerML volume data, would this produce the exterior boundary, the interior boundary, or both at once?

physicsnemo/mesh/sampling/sample_data.py
peterdsharpe: Fixed!

physicsnemo/mesh/validation/statistics.py
peterdsharpe: Fixed

physicsnemo/mesh/validation/quality.py
peterdsharpe: Fixed

physicsnemo/mesh/smoothing/laplacian.py
peterdsharpe: Renamed the API here!

physicsnemo/mesh/projections/_embed.py
```diff
@@ -0,0 +1,154 @@
...
+    Key behaviors:
+        - Manifold dimension (n_manifold_dims) is preserved
+        - Topology (cells connectivity) is preserved
+        - Point/cell/global data are preserved as-is
+        - Cached geometric properties are cleared (they depend on spatial embedding)
+
+    Examples of use cases:
+        - [2, 2] → [2, 3]: Embed 2D surface in 2D space into 3D space
+        - [1, 3] → [1, 2]: Project 3D curve down to 2D plane
+        - [2, 3] → [2, 4]: Embed 3D surface into 4D space
```
coreyjadams: I like this functionality but I think the docstring is somewhat misleading, at least when I read it.

Couple notes:
- why limit to the trailing dimensions?  I can forsee easy examples where a user wants to project or expand their mesh _not_ along the last dimension.
- I think it's worth considering separate entry points for embedding vs. projecting.  One is destructive in information, the other is not.

It'd be cool to do this in arbitrary directions (like, along a plane, or line, whatever) but I think that's probably better supported with rotation + project/embed.

physicsnemo/mesh/projections/_extrude.py
```diff
@@ -0,0 +1,301 @@
...
+) -> Mesh:
+    """Extrude an N-dimensional mesh into an (N+1)-dimensional mesh.
+
+    This function sweeps an N-dimensional manifold along a vector to create an
+    (N+1)-dimensional manifold. Each N-simplex is extruded into a prism-like shape,
+    which is then tessellated into (N+1) child (N+1)-simplices.
+
+    The extrusion creates new vertices by offsetting all original vertices by the
+    extrusion vector. Each parent N-simplex generates (N+1) child (N+1)-simplices
+    connecting the original and extruded vertices.
```
coreyjadams: I do not think I understand this or its purpose.  What's the usecase?

physicsnemo/mesh/remeshing/_remeshing.py
```diff
@@ -0,0 +1,86 @@
...
+
+    from physicsnemo.mesh.io.io_pyvista import from_pyvista, to_pyvista
+    from physicsnemo.mesh.repair import repair_mesh
+
+    pyacvd = importlib.import_module("pyacvd")
+    clustering = pyacvd.Clustering(to_pyvista(mesh))
+    clustering.cluster(n_clusters)
+    new_mesh = from_pyvista(clustering.create_mesh())
+    new_mesh, stats = repair_mesh(new_mesh)
+    return new_mesh
```
coreyjadams, line 95: Where did we settle on this?  For v2.0 release, I'm in favor of making this functionality into an example to show how to do it, since this is adding an optional dependency that runs on CPU only.  We should look at GPU accelerating this for v2.1.

physicsnemo/mesh/repair/degenerate_removal.py
```diff
@@ -0,0 +1,128 @@
...
+            "n_zero_area_cells": 0,
+            "n_duplicate_vertex_cells": 0,
+            "n_cells_original": 0,
+            "n_cells_final": 0,
+        }
+
+    ### Check 1: Zero area cells
+    cell_areas = mesh.cell_areas
+    non_degenerate_by_area = cell_areas >= area_tolerance
+    n_zero_area = (~non_degenerate_by_area).sum().item()
```
coreyjadams, line 78: If this will be used in a training or inference loop, consider

```
tensor.to("cpu", non_blocking=True)
```

instead of `item()` right away, since `item` will be a synchronous transfer.  Collect everything with `item` at the end.

physicsnemo/mesh/repair/degenerate_removal.py
```diff
@@ -0,0 +1,128 @@
...
+    # For each cell, check if all vertices are unique
+    # Sort vertices in each cell and check for adjacent duplicates
+    cells_sorted = torch.sort(mesh.cells, dim=1).values  # (n_cells, n_verts)
+
+    # Check if any adjacent sorted vertices are equal
+    n_verts_per_cell = mesh.n_manifold_dims + 1
+    has_duplicates = torch.zeros(n_original, dtype=torch.bool, device=device)
+
+    for i in range(n_verts_per_cell - 1):
+        has_duplicates |= cells_sorted[:, i] == cells_sorted[:, i + 1]
```
coreyjadams: Can you vectorize this for higher dim meshes?  Maybe not worth it, up to you.

```
has_duplicates = ((cells_sorted[:,1:] - cells_sorted[:, :-1]) == 0).any(dim=-1)
```

Not sure about that any(dim=-1) syntax though.

physicsnemo/mesh/repair/orientation.py
```diff
@@ -0,0 +1,223 @@
...
+        is_oriented[start_face] = True
+        component_id[start_face] = n_components
+        current_front = torch.tensor([start_face], device=device, dtype=torch.long)
+
+        component_size = 1
+
+        # Iteratively expand front until no more neighbors (fully vectorized)
+        for iteration in range(n_cells):  # Max iterations = n_cells (diameter of graph)
+            if len(current_front) == 0:
+                break
```
coreyjadams: I shudder at this while loop with a for loop inside it. Yuck.  But I understand this algorithm is not conducive to avoiding that.

Have you benchmarked this at all?

coreyjadams (commented):
OK, review 2 of ???.  I've gotten through projections, remeshing, and repair here.  I am not planning to look closely at primitives, since they seem to drive examples and testing rather than core functionality.  Meaning - a bug in primitives doesn't bring the whole thing down.

To Go:
- boundaries
- calculus
- curvature
- geometry
- neighbors
- examples

physicsnemo/mesh/sampling/sample_data.py
peterdsharpe: Vectorized!

physicsnemo/mesh/sampling/sample_data_hierarchical.py
peterdsharpe: In a nutshell, just simplicity - having a torch-only version of this rather than depending on Warp for this. Preserves the ability to do things like torch.compile, run on CPU if you wanted to, do things at bf16 or fp64, etc.

physicsnemo/mesh/smoothing/laplacian.py
peterdsharpe: Fixed with a new utilities/_tolerance.py, which gives a small helper for setting default eps based on dtype.

physicsnemo/mesh/mesh.py
peterdsharpe: Both - any external faces. I'll document this!

physicsnemo/mesh/boundaries/_boundary_extraction.py
peterdsharpe: In-function imports for modules within the same package (e.g., `physicsnemo`) are permitted, as this is often used to avoid circular dependencies and incurs zero runtime cost. If memory serves right, pre-commit (via Ruff) will raise a flag for external imports inline.

physicsnemo/mesh/sampling/sample_data.py
peterdsharpe: Fixed!

physicsnemo/mesh/sampling/sample_data.py
peterdsharpe: Good call - done!

physicsnemo/mesh/sampling/sample_data.py
peterdsharpe: Ah, I can see the confusion. I think the "sample" nomenclature is nice for parallel to `pyvista.Dataset.sample()` (which users may be coming from) - but I'll update the docstring to clarify what this does.

physicsnemo/mesh/spatial/bvh.py
peterdsharpe: In a nutshell, just simplicity - having a torch-only version of this rather than depending on Warp for this. Preserves the ability to do things like torch.compile, run on CPU if you wanted to, do things at bf16 or fp64, etc.

Also, Warp BVH is fundamentally 3D-only, so it breaks the dimensional genericity of PhysicsNeMo-Mesh. You also get things like @tensorclass serialization, no Warp JIT cost (potentially relevant in preprocessing cases), etc.

physicsnemo/mesh/sampling/sample_data.py
peterdsharpe: Done! Now this reuses BVH.

physicsnemo/mesh/projections/_extrude.py
peterdsharpe: Ah! The classic use case would be something like AirFRANS: you have a 2D dataset, but you want to use a model that is inherently 3D (e.g., DoMINO, which is inherently 3D since the ball query uses a Warp backend based on their `vec3` type.) This can be a surprisingly painful barrier to cross, as Kaustubh and I experienced recently - I was using the original 2D version of AirFRANS, and he was using a PyVista-extruded 3D version. So, this allows you to eliminate that barrier in a one-liner.

physicsnemo/mesh/repair/degenerate_removal.py
peterdsharpe: Yep - removed the .item() call.

physicsnemo/mesh/repair/degenerate_removal.py
peterdsharpe: Vectorized: e031e3bf8205334d42b87e74349806219320d4f7

physicsnemo/mesh/projections/_extrude.py
peterdsharpe: To add: I'll update the docstring here to reflect this!

physicsnemo/mesh/repair/orientation.py
peterdsharpe: Yeah... I was looking around for more efficient implementations, but didn't find anything. That said, it's not as bad as it looks, in the sense that the max depth of the while loop is only sqrt(N).

physicsnemo/mesh/projections/_embed.py
peterdsharpe: Fixed; added an argument to the embed function to allow choice of which dim gets added!

And, split embedding and projection into two separate files here.

9497bb1601664ca5d6d5f2ab0f9bc2d37da10843

physicsnemo/mesh/validation/validate.py
peterdsharpe: Done, uses BVH

physicsnemo/mesh/validation/validate.py
peterdsharpe: Fixed, good call! Now accepts a separate area_tolerance.

physicsnemo/mesh/mesh.py
peterdsharpe: Fixed with docstring

physicsnemo/mesh/remeshing/_remeshing.py
peterdsharpe: Noting discussion on Slack: going to keep this for v2.0, but making it less visible by not having a Mesh.remesh() top-level method.

coreyjadams:
It's challenging to make review comments on the .ipynb notebooks so I'm going to do it here.

# Example 1

- Section 4, discussing global data in a mesh.  If you attach a vector like global data, and then rotate the mesh, does that rotate too?  Imaging embedding `inlet_velocity` and then spinning the mesh or something.
- Section 5, vis.  It's not a PR request but a docs request: I think Pyvista lets you interact with 3D visualizations, would be cool to have that in our docs or blog or similar!
- Section 7, differentiabily.  Autograd requires leaf tensors to find the gradients.  If a user creates a mesh, can they then `detach()` it into a leaf tensor that they can get grads on?

# Example 2

- What happens when you rotate a non 2D or 3D mesh?
- In the projection of a sphere from 3D to 2D via the 3x2 transform matrix, the visualization seems to either be missing half the points or perfectly aligned so we don't see them.  Which is it doing?
-  The interpolation section and tool is cool, nice job.
- Section 4, merging meshes.  I am imaging a DrivaerML style mesh with volumetric data: you might want to keep the surface mesh and volume field mesh "together" in an abstract way: let you apply the same scaling, rotation, transformation, etc.  I think that's incompatible to merge as a single mesh since the manifold dim is different (right?).  Is there a more abstract Mesh "container" that might do this?  Not saying it is needed for release; but it would be useful to have that generic container that only requires the spatial dimension is consistent to operate on one "environment". 
- Section 5 on Boundary and Facet Extraction - the graphic is challenging to interpret, it's very busy.  Is there another way to show this?  Perhaps you can show 2x2 images: original, and original sliced (to show it's filled) + edge extracted, and edge extracted twice (to show the interior more cleanly)
- when describing facet extraction, can you be a little more verbose about what it is doing and the role of `manifold_codimension`?  I don't think it's clear from "extract ALL lower-dimensional ..." vs. specifying some dimension.  Pretty sure *I* understand it but I've been staring at this for awhile.  And even then it's not assured 😅 .
- Section 6, data conversion: can you select which data to move between points and cells or is it always all the data?
- section 7, topology checks.  An example or two of **non-manifold** meshes would be a nice addition here.
- Drop "Next steps" I think it's obvious from the ordering, and then you don't have to update if you add a tutorial somewhere.
 

I will make additional comments for additional tutorials, so that's it's not all-at-once.

coreyjadams:
## Tutorial 3

- At the top, describing the operations.  Are there any restrictions beyond dimensionality to these statements?  Like, "you can only compute curvature on manifolds" or something?
- section 1: Can you plot the 2D square mesh gradients?
- When you print the gradient [:5], why is the first value 0?
- What are the other sources of error there?  the first point is not enough to produce 0.4% error on 1089 points.  Is it the boundaries?
- I don't think you have any examples of the `dec` method for gradient computations. 
- Can the numerical lsq method populate .grad attributes and work with autograd?  If you run a model on a mesh, and then compute the gradients and at the end like this, can it be done in a backwards pass?
- Love the shout out to 7D cross products.  Where's the 7D mesh example with a cross product?! (this is sarcasm, no need to implement this)
- The curvature calculation plot on a bunny is not rendering.
- I love Differential Geometry as much as the next person, but .... what is the expected use case of the Gauss-Bonnet Theorem section?  Not advocating you remove it, just suggesting maybe it's good to add "Here's why this matters for your application"
- Can you add a visualization of intrinsic vs extrinsic gradients on a sphere?
- Section 6: render those vector calc identities in symbolic form?
- Again, drop next steps.


Overall, for tutorial 3 on calculus, I think you have two things to address:
1. What is the role of the non-least-squares methods?
2. What is the origin of errors in gradient calculations?  Or flipping that on it's head, from a user perspective: When can my lsq calcutions of gradients be relied upon?

coreyjadams:
## Tutorial 4 (neighbors)

- Section 1: Topological neighbors.  It's annoying to ask but I think it would make a big difference in understandability if you could plot each of these.  Make the plot of the sphere, and then highlight a point/cell (just one or two max) and show the objects returned by the adjacency.
- re:pyg - we support PyG in a lot of models.  It'd be a boon to data augmentation and preprocessing to natively support that export operation.  I know it's only a few lines but it'd be user-friendly.
- Section 3: BVH needs visualizations.  
- Section 4: You need visualizations here too, but also some context about use cases.  "This is useful for upsampling an AI model prediction on points that might be outside a mesh" etc.
- Section 5: Message Passing.  This is a cool and valuable biproduct of mesh.  But it speaks to a lot of under-the-hood complexity of dealing with the repeat_interleave / CSR representation, scatter ops, etc.  I think it'd be better standalone where you can build a real MPNN example from the ground up and differentiate it.  Propose you cut this section and do it differently, if you have time.
- 
Overall, this neighbors tutorial is **begging** for better (any!) visualizations to really get it's point across.

coreyjadams:
## Tutorial 6:

- I think it's worth being more upfront about what this notebook is doing: You're not *really* doing ML on a mesh, you're using mesh objects for accelerated pre and post processing, inference acceleration, etc.
- It would be cool though maybe not useful to build an ML Model that actually takes mesh as input and output and works that way end to end.  Not for this PR though.
- Can you make these perf results into easy to digest tables?
- Except for one point, the curvature is slower or the same on GPU.  Why?
- Is the neighbor computation _really_ a single kernel launch?  I am doubtful...perhaps use more exact language like "In a single GPU-accelerated pass over the mesh"?
- Random point sampling: your measurements are fishy.  CPU time consistently increases with level / mesh size.  PNM Mesh time increases and then decreases again.
- Looking at your perf plots at the end of the day, I think it is a little unfair to not note the small mesh / large mesh difference.  PyVista clearly wins when the mesh is very small.  PNM wins when the mesh is really big.  Maybe we can make this a little more fair to pyvista and give good guidance about "this is useful on BIG meshes"

Batching:
- I don't actually know the reasoning behind 1.5 power padding.  What is it? 

Boundary conditions:
- There are some weird errors printed that could be cleared.
- Can you make what you're doing here a little more clear?  I don't understand how you're sorting the points.

End-to-End CAE Workflow
- This feels out of place after all the other components.  And, it might be nice to integrate this type of workflow with the datapipe structure?
- Could probably cut this and focus tutorial 6 on performance, and an eventual end-to-end workflow would be standalone.

I think you can drop the Summary, Key Takeaway, and Conclusion sections, they feel very chatGPT.  That type of next steps makes sense in the README.md anyways.

coreyjadams:
## Tutorial 5 (repair)

This is a lighter tutorial, and my comments are more superficial.

- I think you are missing some plots in the first section.
- Section 4, repair, could use some images if you have time to make them.
- Section 5, topology checks, is duplicate of a previous tutorial (operations?) but I do like it here too.  Can you also duplicate some images? 
- Is it possible to use an invalid mesh for the practical workflow?  And, include pre- and post-images?

As usual, it's my view that you should drop the next steps.

coreyjadams (approved):
Ok, I have skimmed the remaining sections.  I have so little recent expertise on this math that my review can't really provide benefit there.  I think I have left sufficient comments to what I would have picked at (for loops, etc) but really my remaining comments will be at the following:
- making sure the public facing documentation is all incorporated into the docs.  And, for some of the math heavy sections, making sure it renders properly.
- Avoiding for loops and brute-force algs wherever possible.
- Getting good test coverage.  If you haven't measured, might be worth running and looking for any significant gaps?


I know I've left a huge pile of comments everywhere, and the PR is 40k LOC.  This really could stand on it's own as it's own product, but here we are.  Overall I think you've done good work and this will be a really powerful tool in our toolbox moving forward.

peterdsharpe:
Depends on https://github.com/NVIDIA/physicsnemo/pull/1389

PR #1341, opened 2026-01-21
Fixes Torch 2.10 <-> CuML <-> earth2grid Dependency Issues; Removes disallowed `pytz` imports.

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Explanation:

> If you're seeing weird CI failures related to dependency resolution today (Jan 21), here's what's going on. PyTorch 2.10.0 was released just hours ago, and it introduces a new dependency on cuda-bindings==12.9.4 for Linux x86_64. cuml-cu13 requires cuda-bindings>=13.0.1 through its cuda-python dependency. These two version requirements are mutually exclusive, so the resolver can't find a valid solution.
> This didn't happen yesterday because torch 2.9.x doesn't depend on cuda-bindings at all.
> earth2grid is also involved because it dynamically pins its torch requirement at build time based on whichever latest torch version is available. When uv builds earth2grid from source, it now picks torch 2.10.0 (the latest), which creates a torch>=2.10,<2.11 requirement dynamically at install time. So ordinarily the resolver would fall back to 2.9.x, but earth2grid's dynamic dependencies are locking in torch 2.10.0.
> For our options, some possibilities would be to either:
> a) pin torch<2.10 temporarily in our dependencies until the cuda-bindings versions align (hopefully in the next cuml-cu13 release?), or
> b) we could remove earth2grid from our optional dependencies since it's only used in one place in the repo, and this instance has a pure-torch fallback.
> Thoughts?

Also pins `earth2grid` version to a specific commit instead of a moving target, which will prevent unintentional `uv` re-resolutions in the future.

## Checklist

## Review Process


pyproject.toml
```diff
@@ -72,9 +73,19 @@ name = "nvidia"
...
+# with cuml-cu13's requirement for cuda-bindings>=13.0.1 via cuda-python.
+# The CUDA 13 build of torch pins cuda-bindings==13.0.3, which is compatible.
+[[tool.uv.index]]
+name = "pytorch-cu130"
+url = "https://download.pytorch.org/whl/cu130"
+explicit = true
 
 [tool.uv.sources]
-earth2grid = { url = "https://github.com/NVlabs/earth2grid/archive/main.tar.gz" }
+torch = { index = "pytorch-cu130" }
```
coreyjadams: With this, I can no longer `uv sync` on my mac.  We might be breaking all CPU compatibility with this.

peterdsharpe: Agh, good catch. Will follow up on Slack

peterdsharpe: Resolved per our chat! Now we follow PyTorch's guidance: Mac pulls from PyPI, Win/Linux pull from the PyTorch-hosted indices.

PR #1357, opened 2026-01-28
Update dependencies and enforce dependency checking in physicsnemo

coreyjadams:
# PhysicsNeMo Pull Request

This PR has grown a little bit bigger than anticipated, so let me summarize:

1. First goal was to move zarr and xarray to optional deps instead of core deps.
2. Along the way, I realized how many optional deps in datapipes aren't protected against spurious imports.  It's not a breaking issue, since we don't import those, but I added protection layers against all of them.
3. This also let me update some dependencies in pyproject.toml.

For pyproject.toml, a new optional dependency group is created for "performance" centric items.  Since, in reality, the best performance is from nvidia's libraries, this section is entirely nvidia packages with some sort of cuda binding.

I updated uv.lock as well.

## Checklist

## Review Process

### DIFF
--- a/physicsnemo/experimental/nn/__init__.py
+++ b/physicsnemo/experimental/nn/__init__.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
 # SPDX-FileCopyrightText: All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
--- a/test/experimental/__init__.py
+++ b/test/experimental/__init__.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
 # SPDX-FileCopyrightText: All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #


coreyjadams:
This PR also removes `fcn_mip_plugin` from models.  And makes sure that all core packages are part of the import scan.

physicsnemo/datapipes/climate/utils/invariant.py
```diff
@@ -117,6 +126,11 @@ def __init__(
         normalize=False,
         interp_method="linear",
     ):
+        if not XARRAY_AVAILABLE:
+            raise ImportError(
+                "FileInvariant requires xarray to be installed. "
+                "Install with: pip install xarray"
```
pzharrington: Do we want these error messages to mention the optional deps groups specified in the toml? May be something for another PR, but I feel like discoverability of those groups is not that good right now

pzharrington: For ex. Nick set up some import handlers/checkers that are based on deps groups and raise errors accordingly -- might clean up a lot of the dedicated check methods that have been added in this PR. See https://github.com/NVIDIA/earth2studio/blob/main/earth2studio/utils/imports.py

physicsnemo/datapipes/healpix/data_modules.py
```diff
@@ -26,21 +27,41 @@
...
 
+# External modules
+# Check for optional dependencies
+DASK_AVAILABLE = check_version_spec("dask", hard_fail=False)
+
+if DASK_AVAILABLE:
+    _dask_diagnostics = importlib.import_module("dask.diagnostics")
+    ProgressBar = _dask_diagnostics.ProgressBar
+else:
+    ProgressBar = None
```
pzharrington: Ugh we should just remove this stupid progress bar

pzharrington: Not worth the extra dep

coreyjadams: Yes please, don't have to tell me twice.  I was trying not to break functionality, but it's pretty much the only dask dependency I think.

coreyjadams: Dask is gone.

physicsnemo/datapipes/climate/utils/invariant.py
coreyjadams: It would be nice to have something like that.  I was thinking somehow about how we could reduce this boiler plate... I haven't got a solution yet, I'll look at this too.  Do you want this in this PR or just in some nebulous future?

pzharrington: In this PR there are a lot of things like the below, which in some cases led to having to create extra files or just some potentially unnecessary bloat:
```
def _raise_missing_xarray():
    raise ImportError(
        "xarray is required for physicsnemo.datapipes.healpix but is not available. "
        "Please install xarray with `pip install xarray`."
    )
```
I think they could be cleaned up nicely by instead having some sort of e.g. `physicsnemo.core.raise_missing_optional_dep` method or decorator that could print some ImportError message like
```
<package> is required for <caller/object/source of import error> but is not available. Please install <package> directly, or install the optional dependency <deps-group>
```
Where the caller provides the bracketed items in the src code, something like
```
raise_missing_optional_dep("xarray", "datapipes-extras")
```
Would also standardize the messaging and herd people towards paying attention to deps groups

coreyjadams: We have, in my mind, two problems to solve:
1. When dependencies are missing, we need to be informative about what is missing and how to get it.
2. When dependencies are missing, we need to turn `ImportError` into `RunTimeError` so that no one ever crashes out with `import physicsnemo` when they don't even need [healpix | transformer_engine | dali | ...].

Thinking about your suggestion, maybe we can have some sort of fake dependency factory:

```
class MissingImport():
  def __init__(self, library_name):
    self.lib_name = library_name

  def __getattr__(self, key):
    raise RuntimeError("Missing {self.lib_name}, please blah blah blah")
```

then, we could go a step further with 

```
def maybe_import(package_name: str):
  # importlib logic to check and import
  if not available:
    return MissingImport(package_name)
  else:
    return package
```

It might consolidate the boiler plate and let us automate error raising when it's in python code.  It won't help, though, when we do something else like use `package.decorator` or inherit from something.  Those might not be realistic use cases.

pzharrington: Yeah I like that, sounds like it would get the two most important birds. Not sure, but for this:

> It won't help, though, when we do something else like use package.decorator or inherit from something

We might also be able to use your idea in some decorator form that would raise the error on the `__init__` of an inherited class. Not sure if it would be doable to double-decorate something successfully though

coreyjadams:
I blew up this PR, a bit, based on some of the discussions with @pzharrington.  It looks bigger than it is.  Major changes:
- New `OptionalImport` object from core.version_spec.  This can be used as a lazy import, but also will raise a runtime error at first access and not import time.
- Allow type checking ifs in import linting.  So, we can use `if TYPE_CHECKING` for static checking without breaking the optional import paradigm.  This I don't reallllly love since it's duplicating imports, but it's not necessary except in a couple places and it's not complicated.
- I made a registry of useful install hints in OptionalImport.  So, we don't need to worry much about writing good error messages, since it's consolidated.  It specifies the uv dependency group, where to pip install, other instruction locations, etc.  We can elaborate on this - I haven't gone through closely, yet, let's make sure it's the right model first.

about 5000 lines of the +5k / -5k here are just replacing boilerplate with better syntax and un-indenting blocks of code that no longer need to be in an if/else.

physicsnemo/core/version_check.py
```diff
@@ -16,30 +16,374 @@
...
+    "torch_cluster": _format_install_hint(
+        "torch_cluster",
+        group="gnns",
+        docs_url="https://pytorch-geometric.readthedocs.io/en/latest/install/installation.html",
+    ),
+    "dgl": _format_install_hint(
+        "dgl",
+        group="gnns",
+        docs_url="https://www.dgl.ai/pages/start.html",
+    ),
```
pzharrington: `dgl` is deprecated so we can drop this

physicsnemo/datapipes/climate/era5_hdf5.py
```diff
@@ -14,31 +14,27 @@
...
-        + "The package can be installed at:\n"
-        + "https://docs.nvidia.com/deeplearning/dali/user-guide/docs/installation.html"
-    )
-
 from dataclasses import dataclass
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
 from typing import Dict, Iterable, List, Tuple, Union
 
+import h5py
```
dallasfoster, line 22: While we are at it, I would argue that h5py could also be made an optional dependency.

pyproject.toml
```diff
@@ -35,16 +35,15 @@ dependencies = [
     "s3fs>=2023.5.0",
     "packaging>=24.2",
     "timm>=1.0.22",
-    "xarray>=2025.6.1",
     "einops>=0.8.1",
     "h5py>=3.15.1",
     "cftime>=1.6.5",
```
dallasfoster: I would target cftime as another dependency that appears optional, it only appears to be used in examples.

pyproject.toml
```diff
@@ -35,16 +35,15 @@ dependencies = [
     "s3fs>=2023.5.0",
```
dallasfoster, line 35: @NickGeneva do we really need to keep s3fs?

pyproject.toml
dallasfoster, line 26: What do we think about pandas? @pzharrington it appears that it might be used in healpix utilities which could be made optional-model dependent?

pzharrington: Hmm, it's also used by the insolation function and the drivaernet datapipe, so not totally clear which optional deps group it would belong in. Since it's not causing install problems (AFAIK) and not too bulky/slow to build, I wouldn't mind just leaving it in core.

pyproject.toml
coreyjadams: Good call.  Used to be used in some functionality but it's been removed since.

coreyjadams:
Something happened in this merge to blow this up to 1400+ files.  I"m going to rebase this work onto a fresh PR.

PR #1366, opened 2026-02-03
Add concatenation wrapper for legacy diffusion models

pzharrington:
# PhysicsNeMo Pull Request

## Description
Adds a wrapper that will align legacy diffusion backbones (`SongUNet*`, `DhariwalUNet`) and `DiT`, which expect image-shaped conditioning to be concatenated to the noised `x` input, with the new diffusion model protocol.

- Added ConcatConditionWrapper in physicsnemo/diffusion/utils/wrappers.py to handle channel-concat conditioning outside preconditioners, with SongUNet/DiT routing via isinstance and a warning fallback for unknown backbones.
- Exported the wrapper from physicsnemo/diffusion/utils/__init__.py.
- Extended BaseAffinePreconditioner docstring with a concrete example using ConcatConditionWrapper + EDMPreconditioner.
- Fix a bug in ShardTensor that prevents importing `DiT` caused by PyTorch shuffling things around in v2.10.0

## Checklist

## Review Process

### DIFF
--- a/physicsnemo/diffusion/utils/model_wrappers.py
+++ b/physicsnemo/diffusion/utils/model_wrappers.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
 # SPDX-FileCopyrightText: All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
--- a/physicsnemo/experimental/nn/__init__.py
+++ b/physicsnemo/experimental/nn/__init__.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
 # SPDX-FileCopyrightText: All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
--- a/physicsnemo/experimental/nn/symmetry/__init__.py
+++ b/physicsnemo/experimental/nn/symmetry/__init__.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
 # SPDX-FileCopyrightText: All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
--- a/physicsnemo/experimental/nn/symmetry/activation.py
+++ b/physicsnemo/experimental/nn/symmetry/activation.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
 # SPDX-FileCopyrightText: All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
--- a/physicsnemo/experimental/nn/symmetry/grid.py
+++ b/physicsnemo/experimental/nn/symmetry/grid.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
 # SPDX-FileCopyrightText: All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
--- a/physicsnemo/experimental/nn/symmetry/so2_conv.py
+++ b/physicsnemo/experimental/nn/symmetry/so2_conv.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
 # SPDX-FileCopyrightText: All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
--- a/test/diffusion/test_wrappers.py
+++ b/test/diffusion/test_wrappers.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
 # SPDX-FileCopyrightText: All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
--- a/test/experimental/__init__.py
+++ b/test/experimental/__init__.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
 # SPDX-FileCopyrightText: All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
--- a/test/experimental/nn/__init__.py
+++ b/test/experimental/nn/__init__.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
 # SPDX-FileCopyrightText: All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
--- a/test/experimental/nn/symmetry/__init__.py
+++ b/test/experimental/nn/symmetry/__init__.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
 # SPDX-FileCopyrightText: All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
--- a/test/experimental/nn/symmetry/test_activation.py
+++ b/test/experimental/nn/symmetry/test_activation.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
 # SPDX-FileCopyrightText: All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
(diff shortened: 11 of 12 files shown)


physicsnemo/domain_parallel/custom_ops/_tensor_ops.py
```diff
@@ -28,9 +28,16 @@
...
 )
-from torch.distributed.tensor._ops.utils import (
-    register_prop_rule,
-)
+
+try:
+    from torch.distributed.tensor._ops.utils import (
+        register_prop_rule,
+    )
+except ImportError:
```
coreyjadams: I've been bitten before with this: an unrelated typo falls under the ImportError category and everything breaks.  We could use some of the utility functions to have some branching logic?

with https://github.com/NVIDIA/physicsnemo/blob/main/physicsnemo/core/version_check.py#L58

```
if check_version_spec(torch, "2.10.0"):
  # do new import
else
  # do old import, torch is mandatory so no checks needed.
```

pzharrington: Yeah makes sense, updated

physicsnemo/diffusion/preconditioners/preconditioners.py
```diff
@@ -321,6 +321,30 @@ class BaseAffinePreconditioner(Module, ABC):
     >>> out = wrapped_dit(x, t, condition)
     >>> out.shape
     torch.Size([1, 2, 8, 8])
+
+    **Example: Using ConcatConditionWrapper with a preconditioner**
+
+    The :class:`~physicsnemo.diffusion.utils.ConcatConditionWrapper` lets you
+    perform channel-wise concatenation outside of the preconditioner while still
+    using the standard diffusion model interface:
```
CharlelieLrt: This example is good, but maybe just add a few lines of explanations to relate it to the previous examples. Something like "Hey, this is the DIY way of writing a wrapper that matches the `DiffusionModel` interface. Now if your use-case is simply a conditioning based on channel-wise concatenation, you can use the `ConcatConditionWrapper` to save you the trouble of writing your own wrapper."

physicsnemo/diffusion/utils/wrappers.py
```diff
@@ -0,0 +1,209 @@
...
+    SongUNet,
+    SongUNetPosEmbd,
+    SongUNetPosLtEmbd,
+)
+
+
+class ConcatConditionWrapper(Module):
+    r"""
+    Wrapper that handles channel-concatenated conditioning outside diffusion
+    preconditioners.
```
CharlelieLrt: I don't think there should be any mention of preconditioner here. The `DiffusionModel` interface is meant to be required by all diffusion utilities for training (e.g., the losses, multi-diffusion). So, this wrapper will be necessary (or at least useful) not only for the preconditioner, but also everywhere else a `DiffusionModel` is required.

physicsnemo/diffusion/utils/model_wrappers.py
CharlelieLrt, line 1: nit: rename into `model_wrappers.py`?

physicsnemo/diffusion/utils/wrappers.py
```diff
@@ -0,0 +1,209 @@
...
+
+    This wrapper adapts backbones with different conditioning signatures to the
+    :class:`~physicsnemo.diffusion.DiffusionModel` interface by concatenating
+    image-like conditions to ``x`` and routing vector conditions to the
+    appropriate argument. It is intended for cases where image-like conditioning
+    (i.e., conditioning with the same spatial dimensions as ``x``) should be
+    concatenated along the channel dimension, which should not be handled by
+    preconditioners.
```
CharlelieLrt: Same here: useful for preconditioner, but not limited to it, since the `DiffusionModel` interface is a requirement of the entire diffusion training stack.

physicsnemo/diffusion/utils/model_wrappers.py
```diff
@@ -0,0 +1,209 @@
...
+    Default behavior for unknown backbones is to concatenate ``cond_concat`` (if
+    provided) and pass ``cond_vec`` as the ``condition`` keyword argument.
+
+    Parameters
+    ----------
+    model : physicsnemo.Module
+        Backbone model to wrap.
+    image_cond_key : str, optional
+        TensorDict key for the image conditioning tensor to concatenate to
+        ``x`` along channels, by default ``"cond_concat"``.
```
CharlelieLrt, line 80: I am not a big fan of having the default to some hard-coded key like `"cond_concat"`. Intuitively, as a user, I would expect the default behavior to be `image_cond_key = None`, in which case there is nothing that is used for image conditioning. If I want to use channel-wise concatenation, then I need to explicitly pass an str key to `image_cond_key`. Same comment for the `vector_cond_key` argument.

Another example in the same line: if I explicitly require some specific keys with `image_cond_key= "my_key"`, but then inadvertently passes `condition: Tensor` to the forward, then the key I passed to the constructor is silently ignored. My feeling is that making making behavior more explicit here would help to mitigate confusion with these edge cases.

Note: if we need to condition on multiple image-like tensors (e.g. in corrdiff), we cant't do it here. We could extend by doing `image_cond_key: str| List[str]`, but not sure if we want to support it right now?

physicsnemo/diffusion/utils/model_wrappers.py
```diff
@@ -0,0 +1,209 @@
...
+    >>> out = wrapper(x, t, condition)
+    >>> out.shape
+    torch.Size([2, 3, 8, 8])
+
+    Wrap :class:`~physicsnemo.experimental.models.dit.DiT` similarly:
+
+    >>> from physicsnemo.experimental.models.dit import DiT
+    >>> dit = DiT(in_channels=3, out_channels=3, condition_dim=4, input_size=8, patch_size=4)
+    >>> dit_wrapper = ConcatConditionWrapper(dit)
+    >>> out = dit_wrapper(x, t, condition)
```
CharlelieLrt, line 134: nit: missing expected output `torch.Size(...)` for doctest

physicsnemo/diffusion/utils/wrappers.py
```diff
@@ -0,0 +1,209 @@
...
+        if isinstance(condition, TensorDict):
+            if self.image_cond_key in condition:
+                cond_concat = condition[self.image_cond_key]
+            if self.vector_cond_key in condition:
+                cond_vec = condition[self.vector_cond_key]
+            if cond_concat is None and cond_vec is None:
+                raise ValueError(
+                    "Condition TensorDict must include at least one of "
+                    f"'{self.image_cond_key}' or '{self.vector_cond_key}'."
+                )
```
CharlelieLrt: Not ideal behavior IMO: if I pass both arguments `image_cond_key` and `vec_cond_key`, but somehow mess up the way I construct `condition`, then this will not raise any error and will silently fail to grab the right tensor from `condition`

It should be `or` instead of `and`

physicsnemo/diffusion/utils/model_wrappers.py
```diff
@@ -0,0 +1,209 @@
...
+        ----------
+        x : torch.Tensor
+            Noisy latent state of shape :math:`(B, *)` where :math:`B` is the
+            batch size.
+        t : torch.Tensor
+            Diffusion time tensor of shape :math:`(B,)`.
+        condition : torch.Tensor, TensorDict, or None, optional, default=None
+            Conditioning data. Use a ``TensorDict`` to supply ``cond_concat`` and
+            ``cond_vec`` tensors. A plain tensor is treated as the concatenated
+            conditioning input (equivalent to ``cond_concat``).
```
CharlelieLrt, line 170: "A plain tensor is treated as the concatenated conditioning input": I am not sure if I understand this sentence correctly: why is the conditioning already concatenated? the concatenation should be done within the forward of the wrapper, right?

physicsnemo/diffusion/utils/wrappers.py
```diff
@@ -0,0 +1,209 @@
...
+
+        if not torch.compiler.is_compiling():
+            batch_size = x.shape[0]
+            if t.shape != (batch_size,):
+                raise ValueError(
+                    f"Expected t to have shape ({batch_size},) matching batch size of "
+                    f"x, but got {t.shape}."
+                )
+            if cond_vec is not None and cond_vec.shape[0] != batch_size:
+                raise ValueError(
```
CharlelieLrt: nit: could add another check on cond_vec to make sure it has ndim=2

physicsnemo/diffusion/utils/model_wrappers.py
pzharrington: > I would expect the default behavior to be image_cond_key = None, in which case there is nothing that is used for image conditioning.

Why would you use this wrapper at all though, in the case of not having any concatenated conditioning? My thinking was that if you are using this, you want to concatenate something. If you pass a `TensorDict`, then you are required to either abide by the expected key names or specify the keys to use. If you pass a `Tensor`, it is assumed you want to concatenate that. So, addressing

> if I explicitly require some specific keys with image_cond_key= "my_key", but then inadvertently passes condition: Tensor to the forward, then the key I passed to the constructor is silently ignored.

If you specify a custom key here in the init but then just pass a plain tensor, it will still get concatenated (though I agree it should probably raise a warning in that case). I'm not opposed to supporting multiple concatenated images though, can add that relatively easily

CharlelieLrt: > Why would you use this wrapper at all though, in the case of not having any concatenated conditioning?

Yeah true, I meant for example concatenate only a vector conditioning (without the image conditioning), and vice versa. 

> If you specify a custom key here in the init but then just pass a plain tensor, it will still get concatenated (though I agree it should probably raise a warning in that case)

Ah yeah, right. The warning is probably a good idea though.

physicsnemo/diffusion/utils/model_wrappers.py
pzharrington: Yeah clearly the wording here is causing confusion. It is not already concatenated, but if the condition is passed as a raw tensor then this wrapper will do the concatenation. How about:
```
        condition : torch.Tensor, TensorDict, or None, optional, default=None
            Conditioning data. Use a ``TensorDict`` to explicitly supply 
            concatenated conditioning and vector conditioning as keys 
            ``cond_concat`` and ``cond_vec`` (or the custom keys
            specified by ``image_cond_key`` and  ``vector_cond_key``).
            Alternately, supply a plain ``Tensor`` input to be concatenated.
```

physicsnemo/diffusion/utils/model_wrappers.py
pzharrington: > I meant for example concatenate only a vector conditioning

Ah you mean like for a vector of dimension `D`, this could expand/copy it across the spatial dims and concat?

CharlelieLrt: No, sorry for the wording. I was thinking of cases where one would want to only pass a vector conditioning (the class labels in the `SongUNet`, without concatenating any image). I know that the wrapper is called `ConcatConditionWrapper` so this is kinda strange to use it to do that, but there are definitely use-cases where ones would like to use only vector-based conditioning.

pzharrington: OK fair. In that case I think I'd like to keep the behavior aligned with the wrapper name, i.e. if someone just wants to pass vector-only conditioning they either need no wrapper, or maybe should set up a custom one for more complex (multiple vectors, etc), but not necessarily use this one which is specifically targeting the concat use-case.

physicsnemo/diffusion/utils/model_wrappers.py
```diff
@@ -0,0 +1,209 @@
...
+        ):
+            augment_labels = model_kwargs.pop("augment_labels", None)
+            return self.model(
+                x,
+                noise_labels=t,
+                class_labels=cond_vec,
+                augment_labels=augment_labels,
+            )
+
+        return self.model(x, t, condition=cond_vec, **model_kwargs)
```
CharlelieLrt, line 259: This is only for DiT, right? Would it make sense to add an `isinstance(model, DiT)`?

physicsnemo/diffusion/utils/model_wrappers.py
CharlelieLrt: Works for me!

physicsnemo/diffusion/utils/model_wrappers.py
pzharrington: I actually had an `isinstance(model, DiT)` but then realized it was falling back to the same behavior as the default/"any backbone" case:
```
Default behavior for unknown backbones is to concatenate ``cond_concat`` and pass ``cond_vec`` as the ``condition`` keyword argument.
```
So just dropped it

CharlelieLrt: Right

physicsnemo/diffusion/preconditioners/preconditioners.py
pzharrington: Polished up the writing more for this one

physicsnemo/diffusion/utils/wrappers.py
pzharrington: Done, I polished the docstring in general here

PR #1371, opened 2026-02-04
HealDA layers and DiT patches for PNM Core

aayushg55:
# PhysicsNeMo Pull Request

## Description

Adds necessary layers and patches the existing DiT to enable HealDA integration into PhysicsNemo.

**DiT**
- Makes the DiT conditioning module for timestep and condition modular, allowing for custom `conditioning_embedder` modules. The existing embedding method embedded the timestep and condition separately and then added the two (following the original DiT implementation), as opposed to the EDM/SongUNet style approach, where the two are embedded jointly.
- Adds `DropPath` to the DiT
- Adds a `final_dropout` toggle to the `Mlp`
- Adds options for qk normalization in DiT attention with `timm` and TE (Note: TE does not support `qk_norm_affine` unlike ``timm``)

**HealDA**
- Scatter aggregation method for tokenizing sparse data onto a dense grid

**Breaking Changes/Bug fixes:**
- TE `MultiHeadAttention` defaults to sequence first `sbhd` qkv_format format matching PyTorch. This does not match the PNM documentation of `TESelfAttention`, suggesting batch-first `(B, L, D) `or the `timm` default of batch-first, so the `qkv_format` argument is exposed and the default is changed to `bshd` (batch-first) for consistency.
- TE  `MultiHeadAttention` does not support projection dropout, and so the appropriate dropout layer was skipped/not applied previously, making the timm and TE backends incompatible. Added a dropout layer following the `attn_op` for consistency. Breaking change for any model trained using TE backend with `proj_drop_rate > 0`.
- Previously in the DiTBlock, the dropout param in the Mlp was hardcoded to 0, so `mlp_drop_rate` was ignored. Changed to propagate this correctly. Breaking change to any model  trained with `mlp_drop_rate>0`
 

## Checklist

## Review Process

### DIFF
--- a/physicsnemo/nn/module/mlp_layers.py
+++ b/physicsnemo/nn/module/mlp_layers.py
@@ -16,21 +16,15 @@
 
 """Multi-layer perceptron (MLP) module with optional Transformer Engine support."""
 
-import importlib
-
 import torch
 from torch import nn
 
-from physicsnemo.core.version_check import check_version_spec
+from physicsnemo.core.version_check import OptionalImport
 
 from .activations import get_activation
 
 # Check for Transformer Engine availability
-TE_AVAILABLE = check_version_spec("transformer_engine", hard_fail=False)
-if TE_AVAILABLE:
-    te = importlib.import_module("transformer_engine.pytorch")
-else:
-    te = None
+te = OptionalImport("transformer_engine.pytorch")
 
 
 class Mlp(nn.Module):
@@ -86,8 +80,6 @@ class Mlp(nn.Module):
     >>> out.shape
     torch.Size([2, 32])
 
-    >>> # MLP with Transformer Engine (if available)
-    >>> mlp = Mlp(in_features=64, hidden_features=128, out_features=32, use_te=True)
     """
 
     def __init__(
@@ -101,13 +93,6 @@ class Mlp(nn.Module):
         use_te: bool = False,
     ):
         super().__init__()
-
-        # Validate Transformer Engine availability
-        if use_te and not TE_AVAILABLE:
-            raise RuntimeError(
-                "Transformer Engine is not available. "
-                "Install it with `pip install transformer-engine` or set use_te=False."
-            )
 
         self.use_te = use_te
 
--- a/.importlinter
+++ b/.importlinter
@@ -1,9 +1,9 @@
 [importlinter]
 root_package = physicsnemo
 include_external_packages = True
+exclude_type_checking_imports = True
 contract_types =
     forbidden_import: test.ci_tests.prevent_untracked_imports.ForbiddenImportContract
-exclude_type_checking_imports = true
 
 [importlinter:contract:physicsnemo-modules]
 name = Prevent Upward Imports in the PhysicsNemo Structure
--- a/physicsnemo/core/version_check.py
+++ b/physicsnemo/core/version_check.py
@@ -32,7 +32,7 @@ Example usage::
     _pyg = OptionalImport("torch_geometric.data")
 
     def my_gnn_function(data):
-        # Import happens here; raises RuntimeError with install hint if missing
+        # Import happens here; raises ImportError with install hint if missing
         Data = _pyg.Data
         ...
 
@@ -524,12 +524,8 @@ def require_version_spec(package_name: str, spec: str = "0.0.0"):
 
     Raises
     ------
-    RuntimeError
+    ImportError
         If the package is missing or does not satisfy the version requirement.
-        Always raises ``RuntimeError`` for consistency
-        with :class:`OptionalImport`, which uses ``RuntimeError`` to avoid
-        breaking ``import physicsnemo`` for users who don't need the optional
-        dependency.
 
     Example
     -------
@@ -539,7 +535,7 @@ def require_version_spec(package_name: str, spec: str = "0.0.0"):
     ...     return pv.examples.download_bunny()
 
     If pyvista is not installed or the version is too low, calling the function
-    will raise ``RuntimeError`` with installation instructions.
+    will raise ``ImportError`` with installation instructions.
     """
 
     def decorator(func):
@@ -548,17 +544,12 @@ def require_version_spec(package_name: str, spec: str = "0.0.0"):
             # Use OptionalImport to get the nice error message with install hints
             opt = OptionalImport(package_name)
             if not opt.available:
-                opt._get_module()  # This will raise RuntimeError with hint
+                opt._get_module()  # This will raise ImportError with hint
             # If version specified, also check version.
-            # Catch ImportError from check_version_spec and re-raise as
-            # RuntimeError for consistency — require_version_spec always raises
-            # RuntimeError for optional-dependency issues (both missing and
-            # version-too-low), matching OptionalImport's design rationale.
+            # check_version_spec raises ImportError with hard_fail=True,
+            # which we let propagate directly.
             if spec != "0.0.0":
-                try:
-                    check_version_spec(package_name, spec, hard_fail=True)
-                except ImportError as exc:
-                    raise RuntimeError(str(exc)) from exc
+                check_version_spec(package_name, spec, hard_fail=True)
             return func(*args, **kwargs)
 
         return wrapper
@@ -576,10 +567,11 @@ class OptionalImport:
     Lazy import wrapper for optional dependencies.
 
     Delays the import of an optional module until it is actually accessed.
-    If the module is not available, raises ``RuntimeError`` with a helpful
-    message on first access. Using ``RuntimeError`` (not ``ImportError``)
-    ensures that ``import physicsnemo`` doesn't crash for users who don't
-    need specific optional dependencies.
+    If the module is not available, raises ``ImportError`` with a helpful
+    message on first access. This is safe because the import is lazy —
+    it only raises on attribute access, not at import time, so
+    ``import physicsnemo`` won't crash for users who don't need specific
+    optional dependencies.
 
     Instances are cached by module name - calling ``OptionalImport("foo")``
     multiple times returns the same instance.
@@ -604,9 +596,9 @@ class OptionalImport:
     ...     return torch_scatter.scatter(data, ...)
 
     If torch_scatter is not installed, accessing ``torch_scatter.scatter``
-    will raise ``RuntimeError`` with install instructions::
+    will raise ``ImportError`` with install instructions::
 
-        RuntimeError: Missing optional dependency: torch_scatter
+        ImportError: Missing optional dependency: torch_scatter
 
         torch_scatter is part of the [graph] optional dependency group.
         Install with:
@@ -648,7 +640,7 @@ class OptionalImport:
         object.__setattr__(self, "_module", None)
 
     def _get_module(self) -> ModuleType:
-        """Import the module, raising RuntimeError with helpful message if unavailable."""
+        """Import the module, raising ImportError with helpful message if unavailable."""
         # Use object.__getattribute__ to avoid triggering __getattr__ recursion
         module = object.__getattribute__(self, "_module")
         if module is not None:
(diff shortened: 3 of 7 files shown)


physicsnemo/experimental/models/dit/layers.py
```diff
@@ -239,24 +268,46 @@ class TESelfAttention(AttentionModuleBase):
...
             warnings.warn(
-                "Transformer Engine MultiheadAttention does not support projection dropout (proj_drop_rate > 0). "
-                "The specified proj_drop_rate will be ignored."
+                "Transformer Engine does not support disabling affine parameters for QK norm. "
+                "Ignoring qk_norm_affine=False and using affine parameters.",
+                UserWarning,
+                stacklevel=2,
             )
-        self.attn_op = MultiheadAttention(hidden_size=hidden_size, num_attention_heads=num_heads, attention_dropout=attn_drop_rate, **kwargs)
+        kwargs.pop("qk_norm_affine", None)
```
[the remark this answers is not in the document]
aayushg55: TE's MultiheadAttention accepts qk_norm_type

NickGeneva:
Moved HPX layers into module folder, can update this one now to use the embedding layers there

https://github.com/NVIDIA/physicsnemo/pull/1377

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -251,8 +261,11 @@ def __init__(
             self.detokenizer = detokenizer
 
 
+        # Linear drop_path schedule: 0 -> drop_path
+        drop_path_rates = [drop_path * i / max(1, depth - 1) for i in range(depth)]
```
NickGeneva: In stead of hard coding the drop-path values to be this linear approach, perhaps its a bit more robust to make the entire parameter a list/array of floats and this gets replaced with some check that it is of the right length.

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -299,31 +314,20 @@ def forward(
         condition: Optional[torch.Tensor] = None,
         p_dropout: Optional[float | torch.Tensor] = None,
         attn_kwargs: Optional[Dict[str, Any]] = None,
+        tokenizer_kwargs: Optional[Dict[str, Any]] = None,
```
NickGeneva: Can we just make all of these ` Dict[str, Any] = {}`, like why support None... maybe some legacy thing @pzharrington

physicsnemo/experimental/models/dit/dit.py
NickGeneva, line 59: No a problem from you Aayush... but generally this whole doc string needs an update to the correct format others are in

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -239,24 +268,46 @@ class TESelfAttention(AttentionModuleBase):
...
+        **kwargs: Any,
+    ):
         super().__init__()
         if not TE_AVAILABLE:
             raise ImportError(
                 "Transformer Engine is not installed. Please install it with `pip install transformer-engine`."
             )
 
-        if proj_drop_rate > 0:
+        if "qk_norm_affine" in kwargs and not kwargs["qk_norm_affine"]:
```
NickGeneva, line 286: Does the proj drop rate check still need to be here?

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -592,14 +651,14 @@ def forward(
             modulated_attn_input,
             **(attn_kwargs or {}),
         )
-        x = x + attention_gate.unsqueeze(1) * attention_output
+        x = torch.addcmul(x, self.drop_path(attention_gate.unsqueeze(1)), attention_output)
```
NickGeneva, line 659: Will this impact the loading of previous DiTs (stromscope) since this is a unconditional module additon? @pzharrington

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -982,3 +1042,279 @@ def get_detokenizer(
             **detokenizer_kwargs,
         )
     raise ValueError("detokenizer must be 'proj_reshape_2d', no other supported detokenizers are available yet.")
+
+class ConditioningEmbedderBase(Module, ABC):
```
NickGeneva: Can we move all these into a separate file.

Also can I suggest a protocol here instead of an abstract class... theres nothing getting inherited from `ConditioningEmbedderBase` so no need to force people to explicitly use this

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -982,3 +1042,279 @@ def get_detokenizer(
...
+                emb = emb + self.map_label(tmp * math.sqrt(self.map_label.in_features))
+
+        # MLP 
+        emb = torch.nn.functional.silu(self.map_layer0(emb))
+        emb = self.map_layer1(emb)
+
+        return emb
+
+
+def get_conditioning_embedder(
```
NickGeneva: I would sugest moving this into the embedders file too

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -982,3 +1042,279 @@ def get_detokenizer(
...
+    hidden_size: int,
+    conditioning_embedder: Literal["dit", "edm", "zero"] = "dit",
```
NickGeneva: Maybe an enum here is more appropriate where the values are the class objects of each?

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -982,3 +1042,279 @@ def get_detokenizer(
...
+    if conditioning_embedder == "dit":
+        return DiTConditionEmbedder(
+            hidden_size=hidden_size,
+            condition_dim=condition_dim,
+            **embedder_kwargs,
+        )
+    if conditioning_embedder == "edm":
+        return EDMConditionEmbedder(
+            emb_channels=condition_dim,
+            noise_channels=embedder_kwargs.pop("noise_channels", hidden_size),
```
NickGeneva: whats up with the weird pop here....

shame all of these can have similar init signatures since we already pass `kwargs`

physicsnemo/experimental/models/dit/layers.py
```diff
@@ -191,8 +199,26 @@ class TimmSelfAttention(AttentionModuleBase):
...
         Output tensor of shape (B, L, D).
     """
-    def __init__(self, hidden_size: int, num_heads: int, attn_drop_rate: float = 0.0, proj_drop_rate: float = 0.0, **kwargs: Any):
+    def __init__(
+        self,
+        hidden_size: int,
+        num_heads: int,
+        attn_drop_rate: float = 0.0,
+        proj_drop_rate: float = 0.0,
+        qk_norm_type: Optional[Literal["RMSNorm", "LayerNorm"]] = None,
```
NickGeneva: Nit: Switch this to `Literal["RMSNorm", "LayerNorm"] | None`

physicsnemo/experimental/models/dit/dit.py
pzharrington: Yeah no need to support `None`, defaulting to `{}` sounds fine

physicsnemo/experimental/models/dit/layers.py
aayushg55: Moved the conditioning embedder classes into a separate file, using a protocol instead.

physicsnemo/experimental/models/dit/layers.py
pzharrington: Given that it's just a dropout module (no state dict params)  I think it should be fine. Tests should verify

aayushg55:
Thanks @NickGeneva, updated to use the new hpx embedding layers.

physicsnemo/experimental/models/dit/layers.py
aayushg55: Previously, the `proj_drop_rate` was unused since TE does not natively support this dropout within its attention block. As we used this in HealDA, I added a dropout layer corresponding to this (and to match timm attention).
```
        # TE doesn't support proj_drop natively, so we add it manually after the attention output
        self.proj_drop = nn.Dropout(proj_drop_rate)
```

physicsnemo/experimental/models/dit/dit.py
aayushg55: Changed to take in a list of floats with length validation

physicsnemo/experimental/models/dit/dit.py
aayushg55: Changed the default to `{}`

physicsnemo/experimental/models/dit/layers.py
aayushg55: Changed

physicsnemo/experimental/models/dit/dit.py
```diff
@@ -164,12 +177,14 @@ def __init__(
         attention_backend: Literal["timm", "transformer_engine", "natten2d"] = "timm",
         layernorm_backend: Literal["apex", "torch"] = "torch",
         condition_dim: Optional[int] = None,
+        conditioning_embedder: Union[Literal["dit", "edm"], Module] = "dit",
         dit_initialization: Optional[int] = True,
+        conditioning_embedder_kwargs: Dict[str, Any] = {},
         tokenizer_kwargs: Dict[str, Any] = {},
         detokenizer_kwargs: Dict[str, Any] = {},
         block_kwargs: Dict[str, Any] = {},
-        timestep_embed_kwargs: Dict[str, Any] = {},
```
pzharrington, line 171: Removal of this will break the existing StormScope checkpoints, which pass `timestep_embed_kwargs`, specifically for the `np_sin_cos` setting of `embed_fn`. I'm not opposed to the more flexible `Protocol`-based approach in this PR for condition embedding, and the more sensible name `conditioning_embedder_kwargs`, but we should use the backward compat checkpoint handling to make sure those legacy checkpoints can be loaded without error. @aayushg55 I can do that part in a follow-up PR if you don't want to bother with it. However can you update `DiTConditionEmbedder` to properly accept and route all the kwarg options for `PositionalEmbedding`?

physicsnemo/experimental/models/dit/layers.py
NickGeneva: awesome!

physicsnemo/experimental/models/dit/layers.py
NickGeneva: Will leave this unresolved, until we confirm stormscope doesnt break... what would be the best way to verify that checkpoint runs fine?

something like, with the [hf files](https://huggingface.co/nvidia/stormscope-goes-mrms)?
```python
registry = json.load("registry.json")
pkg = registry[model_name]

model_spec = []
for m in pkg["checkpoints"]:
  model = Module.from_checkpoint(m["path"])
  model_spec.append(
      {
          "model": model_wrap(model),
          "sigma_min": float(m["sigma_min"]),
          "sigma_max": float(m["sigma_max"]),
      }
  )
```

e2s doesnt support psn 2.x yet so it will break

physicsnemo/experimental/models/dit/dit.py
NickGeneva: thanks!

physicsnemo/experimental/models/dit/layers.py
NickGeneva: Ah I see. Thanks for clarifying

physicsnemo/experimental/models/dit/layers.py
aayushg55: Changed to use enums:
```
class ConditioningEmbedderType(Enum):
    """Conditioning embedder types for DiT models."""

    DIT = DiTConditionEmbedder
    EDM = EDMConditionEmbedder
    ZERO = ZeroConditioningEmbedder

```

physicsnemo/experimental/models/dit/layers.py
aayushg55: Cleaned up the initialization to not require popping and instead just propagate the `kwargs`

physicsnemo/experimental/models/dit/dit.py
aayushg55: I updated it to propagate all the kwargs to `PositionalEmbedding` as before. Would be great if you could help with the backward compatibility checkpoint handling.

NickGeneva:
Over all I think this looks good, only big pending item left  imois verifying that stormscope checkpoints remain operational. Other than than the changes reasonable extensions that dont have major api breaks.

physicsnemo/experimental/models/dit/dit.py
pzharrington: I added some checkpoint compat handling here: https://github.com/pzharrington/physicsnemo/commit/ac42b55d330d2d1df6b19b8c10134ec14f363a8e

Will try and test it tomorrow in E2S, if it works then I'll just push it to this branch/PR. @NickGeneva for viz

pzharrington: Pushed the updates to this branch, test of the stormscope model with it looks good:
image

physicsnemo/experimental/models/dit/dit.py
pzharrington: Will address this in un-migrating DiT from experimental

physicsnemo/nn/module/mlp_layers.py
```diff
@@ -95,6 +95,7 @@ def __init__(
         out_features: int | None = None,
         act_layer: nn.Module | type[nn.Module] | str = nn.GELU,
         drop: float = 0.0,
+        final_dropout: bool = True,
```
pzharrington, line 92: @aayushg55 can you add description of this to the docstring?

aayushg55: Just added

PR #1367, opened 2026-02-03
Add experimental SO2 equivariant convolution and activation

laserkelvin:
# PhysicsNeMo Pull Request

## Description

This PR adds a new `symmetry` namespace within `experimental`, which is intended to house some performant implementations of operations that preserve physical symmetries. What's included in this current PR is support for SO(2) transformations (`SO2Convolution`), and an accompanying rotational equivariance preserving nonlinearity, `GateActivation`. The former is written to take advantage of `einsum`s to efficiently (and correctly) compute linear transforms in a spherical harmonic basis. The new unit tests added include both conventional tests (e.g. expected output shapes), as well as some to check for 2D equivariance.

These layers are nominally general purpose (i.e. any point cloud), but are derived (with heavy modifications to data layout and subsequently performance) from the [`fairchem`](https://github.com/facebookresearch/fairchem/tree/main) repository.

@peterdsharpe this is more or less the same as what you've reviewed already, and @CharlelieLrt please review as much as you want, but including you because this adds new stuff into `physicsnemo.experimental.nn` and figured you might want to monitor for model coding standards.

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [x] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.
- [x] If I am implementing a new model or modifying any existing model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation).

## Dependencies

None

## Review Process

**All PRs are reviewed by the PhysicsNeMo team before merging.**

Depending on which files are changed, GitHub may automatically assign a maintainer for review.

We are also testing AI-based code review tools (e.g., Greptile), which may add automated comments with a confidence score.
This score reflects the AI’s assessment of merge readiness and is **not** a qualitative judgment of your work, nor is
it an indication that the PR will be accepted / rejected.

AI-generated feedback should be reviewed critically for usefulness.
You are not required to respond to every AI comment, but they are intended to help both authors and reviewers.
Please react to Greptile comments with 👍 or 👎 to provide feedback on their accuracy.

PR #1372, opened 2026-02-05
Equivariant layers in three dimensions

laserkelvin:
# PhysicsNeMo Pull Request

## Description

This PR extends #1367 by adding 3D equivariance preserving linear transformations: `SO3LinearGrid` in isolation, and composed together with nonlinearities with `SO3ConvolutionBlock`. Accompanying these changes are also routines to compute 3D rotations, both as part of testing as well as for the actual workflow itself.

The full intended workflow, now with these layers is given some set of edge-wise spherical harmonic embeddings with shape `[num_edges, lmax + 1, mmax + 1, 2, num_channels]`, the Wigner rotation aligns the embeddings along the direction of their edges, reduces an SO3 problem into an SO2, i.e. the hard part is done in SO2[^1]:

```
rotate(x) |> so2_convolution() |> inv_rotate() |> so3_block()
```

The end result should be transformed edge embeddings, and an aggregation operation will get you the updated node embeddings.

Rotations, through the `EdgeRotation` class, can be performed at arbitrary precision as needed: for example, when training at lower precision, we can choose to perform rotations at higher precision to maintain information. Conversely, maybe the rotation could be done at lower precision without too much degradation.

One potential gotcha is that the Wigner matrix is cached: the idea is that, an architecture will comprise multiple SO2/SO3 blocks, and the rotation matrix only needs to be computed once per batch. I've left notes in the docstrings to remind users to *manually clear the cache* after backprop/optimization steps are done as to not have dangling graphs. I've thought about writing a hash-based caching system (i.e. hash the batch, and reuse the matrix only if the hash matches), but part of my design philosophy was to minimize branching with`torch.compile` in mind.

The PR also adds unit tests that more or less mirror the SO2 ones (also updating them to homogenize), with a little more coverage due to the complexity of 3D operations.

[^1]: This is to avoid the expensive tensor product

## Checklist

## Dependencies

None

## Review Process

### DIFF
--- a/examples/cfd/navier_stokes_rnn/navier_stokes_rnn.py
+++ b/examples/cfd/navier_stokes_rnn/navier_stokes_rnn.py
@@ -53,13 +53,12 @@ def prepare_data(
             arrays[k] = np.array(v)
 
         invar = arrays["u"][
-            input_nr_tsteps : input_nr_tsteps + predict_nr_tsteps,
+            0:input_nr_tsteps,
             ...,
             start_idx : start_idx + num_samples,
         ]
         outvar = arrays["u"][
-            input_nr_tsteps + predict_nr_tsteps : input_nr_tsteps
-            + 2 * predict_nr_tsteps,
+            input_nr_tsteps : input_nr_tsteps + predict_nr_tsteps,
             ...,
             start_idx : start_idx + num_samples,
         ]


physicsnemo/experimental/nn/symmetry/so3_block.py
```diff
@@ -0,0 +1,246 @@
...
+        self.lmax = lmax
+        self.mmax = mmax
+
+        # Number of gates needed: one per l>0 degree, times hidden_channels
+        self.num_gates = lmax * hidden_channels
+
+        # Scalar MLP to compute gates from l=0 scalar features
+        # Input: in_channels (from x[:, 0, 0, 0, :])
+        # Output: lmax * hidden_channels (gates for degrees l=1..lmax)
+        self.scalar_mlp = nn.Sequential(
```
peterdsharpe: Might be nice if depth here was configurable?

laserkelvin:
@peterdsharpe @CharlelieLrt ready for review (finally)!

physicsnemo/experimental/nn/symmetry/so3_block.py
peterdsharpe: And can we use ``nn.FullyConnected`` instead for code-reuse? If not, that's ok.

physicsnemo/experimental/nn/symmetry/so3_linear.py
```diff
@@ -0,0 +1,225 @@
...
+            self.bias = nn.Parameter(torch.zeros(out_channels))
+        else:
+            self.register_parameter("bias", None)
+
+        # Initialize weights
+        self.reset_parameters()
+
+        # Create and register the validity mask for (l, m) positions
+        # mask[l, m] = 1.0 if m <= l (valid position), 0.0 otherwise
+        mask = make_grid_mask(lmax, mmax).float()
```
peterdsharpe, line 183: This forces fp32 precision, even under AMP - is this intended?

physicsnemo/experimental/nn/symmetry/wigner.py
```diff
@@ -0,0 +1,1484 @@
...
+        # Check cache is populated
+        if self._cached_wigner is None:
+            raise RuntimeError(
+                "No cached Wigner D-matrices. Call get_wigner_matrices(edge_vecs) first."
+            )
+
+        wigner = self._cached_wigner
+
+        # Get shapes
+        *batch_dims, dim_in, channels = x.shape
```
peterdsharpe, line 1379: Huh, neat - I had no idea you could do splats on partial-tuple unpacking like that

physicsnemo/experimental/nn/symmetry/so3_block.py
laserkelvin: Thanks for that - I actually had looked but I didn't find it in `physicsnemo.nn` so I gave up 😅 

Addressed in 5b837dbd

physicsnemo/experimental/nn/symmetry/so3_linear.py
laserkelvin: That's a good catch, I'm kind of surprised I never ran into type mismatch issues...

Type casting to match input types in 45066148

laserkelvin:
> e.g., can SO3 linear inherit from Linear, so that we don't have to write our own initialization? Can we generalize the SO3 and SO2 layers to be dimensionally-generic?

Yeah let's definitely touch base when it moves out of experimental: I need to play around with composing the layers and running it on real data first, and once I've dogfood'd it a little bit might have a more informed decision to make about these things. If you get the chance, too, let me know if there are pain points

PR #1363, opened 2026-01-31
Refactored diffusion sampler

CharlelieLrt:
# PhysicsNeMo Pull Request

### Overview

This PR introduces a new modular framework for diffusion model sampling, designed to decouple the core components of the sampling process for maximum flexibility and reusability.

#### Decoupled Architecture

- **Model-agnostic**: The sampling function now works with any model implementing the `DiffusionDenoiser` protocol—shape-agnostic, dimension-agnostic, and supports arbitrary conditioning/guidance
- **Solver ↔ Noise Schedule decoupling**: Numerical solvers and noise schedules are now independent, allowing any combination
- **Flexible time-steps**: Time-step discretization can optionally be overridden independently of both solver and noise schedule

##### `Solver` Protocol (`physicsnemo.diffusion.samplers.solvers`)
Defines a minimal interface for numerical integration methods; 4 concrete solver implementations: Euler, Heun, and their respective EDM-stochastic versions.

##### `NoiseScheduler` Protocol (`physicsnemo.diffusion.noise_schedulers`)
Defines minimal interface for training (`add_noise`, `sample_time`) and sampling (`timesteps`, `init_latents`, `get_denoiser`). A base class (`LinearGaussianNoiseScheduler`) implements the protocol for the family of linear-Gaussian schedules; 6 concrete implementation of such noise schedulers: `LinearGaussianNoiseScheduler`, `EDMNoiseScheduler`, `VENoiseScheduler`, `VPNoiseScheduler`, `IDDPMNoiseScheduler`, `StudentTEDMNoiseScheduler`.

##### `DiffusionDenoiser` Protocol (`physicsnemo.diffusion.base`)
Minimal interface for denoisers.

##### New `sample` Function (`physicsnemo.diffusion.samplers.sample`)
Unified entry point for sampling with:
- Configurable solver (string key or custom instance)
- Configurable noise scheduler
- Optional custom time-steps
- Support for intermediate sample collection

## Checklist

- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.
[...]

PR #1381, opened 2026-02-07
DPS and SDA guidance for diffusion

CharlelieLrt:
# PhysicsNeMo Pull Request

### Overview

This PR introduces a modular framework for Diffusion Posterior Sampling (DPS) guidance, enabling constrained generation and data assimilation with diffusion models.

##### `DPSGuidance` Protocol (`physicsnemo.diffusion.guidance`)
Minimal interface for guidance implementations. Any callable `(x, t, x_0) -> Tensor` satisfies this protocol.

##### `DPSDenoiser` (`physicsnemo.diffusion.guidance`)
Transforms an x0-predictor into a score predictor by combining:
- Input denoiser (x0-predictor from trained diffusion model)
- One or more `DPSGuidance` instances
- Conversion callback (`x0_to_score_fn` from noise scheduler)

##### Concrete Guidance Implementations

| Class | Use Case |
|-------|----------|
| `ModelConsistencyDPSGuidance` | Generic (nonlinear) observation operators |
| `DataConsistencyDPSGuidance` | Masked observations (inpainting, sparse probes) |

Both support configurable norm order (L1, L2, etc.) and optional SDA scaling via `gamma` parameter.

### Files Changed

- `physicsnemo/diffusion/guidance/dps_guidance.py` — New guidance implementations
- `physicsnemo/diffusion/guidance/__init__.py` — Updated exports

### Future Work

- Multi-diffusion guidance (planned for future PR)

## Checklist

## Review Process

### DIFF
--- a/physicsnemo/diffusion/noise_schedulers/noise_schedulers.py
+++ b/physicsnemo/diffusion/noise_schedulers/noise_schedulers.py
@@ -587,11 +587,11 @@ class LinearGaussianNoiseScheduler(ABC, NoiseScheduler):
         sigma_dot_t_bc = self.sigma_dot(t_bc)
         alpha_t_bc = self.alpha(t_bc)
         alpha_dot_t_bc = self.alpha_dot(t_bc)
-        g_sq = (
+        g_sq_bc = (
             2 * sigma_dot_t_bc * sigma_t_bc
             - 2 * (alpha_dot_t_bc / alpha_t_bc) * sigma_t_bc**2
         )
-        return g_sq
+        return g_sq_bc
 
     def x0_to_score(
         self,
@@ -765,8 +765,8 @@ class LinearGaussianNoiseScheduler(ABC, NoiseScheduler):
             ) -> Float[Tensor, " B *dims"]:
                 score = score_fn(x, t)
                 f = drift(x, t)
-                g_sq = diffusion(x, t)
-                dx_dt = f - 0.5 * g_sq * score
+                g_sq_bc = diffusion(x, t)
+                dx_dt = f - 0.5 * g_sq_bc * score
                 return dx_dt
 
             return ode_denoiser
@@ -779,10 +779,10 @@ class LinearGaussianNoiseScheduler(ABC, NoiseScheduler):
             ) -> Float[Tensor, " B *dims"]:
                 score = score_fn(x, t)
                 f = drift(x, t)
-                g_sq = diffusion(x, t)
+                g_sq_bc = diffusion(x, t)
                 # Deterministic part of the SDE drift
                 # Note: stochastic term g(t)*dW is handled by the solver
-                dx_dt = f - g_sq * score
+                dx_dt = f - g_sq_bc * score
                 return dx_dt
 
             return sde_denoiser


physicsnemo/diffusion/guidance/dps_guidance.py
```diff
@@ -394,9 +404,11 @@ def A(x_0: Tensor) -> Tensor:
         of ``A``.
     std_y : float
         Standard deviation of the measurement noise :math:`\sigma_y`.
-    norm_order : int, default=2
-        Order of the norm used to compute the residual. Use ``2`` for
-        standard Gaussian likelihood (L2 norm), ``1`` for L1 norm, etc.
+    norm : int | Callable[[Tensor, Tensor], Tensor] | None, default=None
+        Loss function used to compute the residual. ``None`` (default) uses
+        the L2 norm. An ``int`` value uses the corresponding Lp norm. A
```
jleinonen: Nit: can the default just be `2` instead of `None`, if `None` is interpreted as `2` anyway?

physicsnemo/diffusion/guidance/dps_guidance.py
```diff
@@ -0,0 +1,945 @@
...
+        else:
+            variance = self.std_y**2
+
+        # Likelihood score
+        guidance = -grad_x / (2 * variance)
+
+        return guidance
+
+
+class DataConsistencyDPSGuidance(DPSGuidance):
```
jleinonen, line 721: This class duplicates quite a lot of code from `ModelConsistencyDPSGuidance`, would it make sense to reuse that code either by inheritance
(`class DataConsistencyDPSGuidance(ModelConsistencyDPSGuidance):`)
or by composition?

physicsnemo/diffusion/samplers/solvers.py
```diff
@@ -0,0 +1,737 @@
...
+
+        # Euler step from t_hat to t_next
+        t_hat = t_hat_bc.reshape(x.shape[0])
+        d_cur = self.denoiser(x_hat, t_hat)
+        x_next = x_hat + (t_next_bc - t_hat_bc) * d_cur
+
+        return x_next
+
+
+class EDMStochasticHeunSolver(Solver):
```
jleinonen, line 512: This duplicates most of the code from `EDMStochasticEulerSolver`:

* `__init__` is identical except for `alpha`
* `__step__` is identical except for the Heun specific stuff

Could be refactored to share code between the implementations?

CharlelieLrt: @jleinonen do not pay attention to the "files changed" here, because it also includes changes from https://github.com/NVIDIA/physicsnemo/pull/1363 which is not yet merged into main.

Regarding your comment, yes there is some duplicate code, but that's by design. A flatter organization improves code readability. That's a lot of abstractions in this diffusion framework, and adding more level of inheritance will just make things more convoluted.

physicsnemo/diffusion/guidance/dps_guidance.py
CharlelieLrt: Yes, there is some code duplication here, but that's by design to improve code readability.

physicsnemo/diffusion/samplers/solvers.py
jleinonen: While I personally disagree about the increased readability I'll concede that it's a matter of preference. But more than readability I'm concerned about maintainability. Duplicating so much code feels like asking for trouble in the future when we fix a bug or add a feature in the code in one solver and forget to do the same thing in the other identical piece of code. This gets even more problematic if we add more solvers using the same interface going forward.

I do get the point about inheritance being difficult to follow with many abstractions, would it be a better solution to put some of the common code into 1-2 utility functions?

CharlelieLrt: Which part of you thinking of putting in some utility functions exactly?
- The stochastic noise injection? (Used in 2 of the solvers)
- The 1st order Euler step (used in all 4 of the solvers)
- The second order correction (used in 2 of the solvers)

**Note:** by adding two lines of code and an argument `second_order_correction: bool` to the `EDMStochatsicHeunSolver`, it becomes a superset of all 4 solvers and contains all code necessary to run all 4 solvers. Is it what you are suggesting to avoid code duplication?

jleinonen: I like the latter idea, actually. Or perhaps instead of `second_order_correction` just call the class `EDMStochasticSolver` and have an option `integrator: Literal["euler", "heun"]`? It would make the interface more similar to the legacy solvers, but that might not be a bad thing.

Or if you prefer that the Euler and Heun solvers exist as separate classes, you can do as you propose and specialize one as a minimal subclass of the other.

There are many options for reducing redundant code here. When deciding which one to use, one question I would ask is "let's say we want to add a higher-order solver in the future, which code organization makes that most painless?".

CharlelieLrt: I'll think about this, but I am extremely strongly against going back to a single monolithic solver like it was before. Past experience has proved that it is unmaintainable and difficult to develop on it. 

I understand the need to avoid code duplication, but there are also other things to consider:

- How easy is it to relate the code to the math. I think this is critical for diffusion, which is more on the math-heavy side. And that's what advanced users are also interested in: what equations is this abstraction actually solving and with which method. Having a centralized abstraction makes code comprehension much more complicated.
- Centralized code base makes further development more challenging. Unless a new feature is relevant to all solvers, any implementation needs to be extremely careful to avoid undesired side effects on other solvers. E.g. very large part of the new features we review actually break other parts of the codebase they were not intended to even modify, and that has been a very significant time sink in the last year or so.

physicsnemo/diffusion/guidance/dps_guidance.py
```diff
@@ -0,0 +1,945 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES.
```
laserkelvin: License header year

physicsnemo/diffusion/guidance/dps_guidance.py
```diff
@@ -0,0 +1,945 @@
...
+
+@runtime_checkable
+class DPSGuidance(Protocol):
+    r"""
+    Protocol defining the interface for Diffusion Posterior Sampling (DPS)
+    guidance.
+
+    A DPS guidance is a callable that computes a guidance term to steer the
+    diffusion sampling process toward satisfying some observation constraint.
+    A DPSGuidance is expected to be a score-predictor, as it returns a quantity
```
laserkelvin: Wrap ``DPSGuidance`` in backticks

physicsnemo/diffusion/guidance/dps_guidance.py
```diff
@@ -0,0 +1,945 @@
...
+from typing import Callable, Protocol, Sequence, runtime_checkable
+
+import torch
+from jaxtyping import Float
+from torch import Tensor
+
+from physicsnemo.diffusion.base import DiffusionDenoiser
+
+
+@runtime_checkable
```
laserkelvin, line 28: FYI I'm not 100% sure this behaves the way you intended it to. Even with `@runtime_checkable`, I think the `isinstance` part will only check for method and attribute availability, but not function signature.

```python
from typing import Protocol, runtime_checkable


@runtime_checkable
class Expected(Protocol):
    def __call__(self, a: int, b: float): ...


def concrete_func(a: int, b: float, c: str):
    return None

# should be False if the intention is to match __call__ exactly
# but it will return True
print(isinstance(concrete_func, Expected))
```

The question then is whether you do want `@runtime_checkable`, if it doesn't necessarily serve its purpose (it would if you had more methods/attributes)

physicsnemo/diffusion/guidance/dps_guidance.py
```diff
@@ -0,0 +1,945 @@
...
+    >>> t = torch.tensor([1.0])
+    >>> x_0 = x * 0.9
+    >>> output = guidance(x, t, x_0)
+    >>> output.shape
+    torch.Size([1, 3, 8, 8])
+    """
+
+    def __init__(
+        self,
+        A: Callable[[Float[Tensor, " B *dims"]], Float[Tensor, " B *obs_dims"]],
```
laserkelvin: Stylistically, not sure if `operator` might be more descriptive than `A`?

physicsnemo/diffusion/guidance/dps_guidance.py
```diff
@@ -0,0 +1,945 @@
...
+            # Compute gradient of loss w.r.t. x (backprop through x_0)
+            grad_x = torch.autograd.grad(
+                outputs=loss.sum(),
+                inputs=x,
+                create_graph=False,
+            )[0]
+
+        # Compute scaling factor
+        t_bc = t.reshape(-1, *([1] * (x.ndim - 1)))
+        if self.gamma > 0 and self.sigma_fn is not None:
```
laserkelvin: If `torch.compile` is desirable, might want to do some of this validation in the `__init__` instead of runtime to prevent graph breaks

physicsnemo/diffusion/guidance/dps_guidance.py
```diff
@@ -0,0 +1,945 @@
...
+    >>> t = torch.tensor([1.0])
+    >>> x_0 = x * 0.9
+    >>> output = guidance(x, t, x_0)
+    >>> output.shape
+    torch.Size([1, 3, 8, 8])
+    """
+
+    def __init__(
+        self,
+        mask: Float[Tensor, " *mask_shape"],
```
laserkelvin: The hinted shape doesn't match the docstring, which says `(B, *)`, and is also intended to be a bitmask. Would it be better to expect `Bool[Tensor, " B *"]` and then do the typecasting within `__init__`?

physicsnemo/diffusion/guidance/dps_guidance.py
```diff
@@ -0,0 +1,945 @@
...
+            from ``x`` via an x0-predictor to allow gradient backpropagation.
+
+        Returns
+        -------
+        Tensor
+            Likelihood score guidance term of same shape as ``x``.
+        """
+        with torch.enable_grad():
+            # Compute masked predicted and observed values
+            y_pred = self.mask * x_0
```
laserkelvin: Might want to consider `self.mask.to(x_0.dtype, x_0.device) * x_0` to guard against mismatch data types

physicsnemo/diffusion/guidance/dps_guidance.py
NickGeneva: Despite the known short comings of `runtime_checkable`
I dont know of an instance where is actively bad to not use it regardless 🤔

physicsnemo/diffusion/guidance/dps_guidance.py
```diff
@@ -0,0 +1,945 @@
...
+        t : Tensor
+            Batched diffusion time of shape :math:`(B,)`.
+
+        Returns
+        -------
+        Tensor
+            Guided score of same shape :math:`(B, *)` as ``x``. Computed as the
+            sum of the unconditional score and all guidance terms.
+        """
+        x = x.detach().requires_grad_(True)
```
NickGeneva, line 393: Is this detach 100% okay?

physicsnemo/diffusion/guidance/dps_guidance.py
```diff
@@ -0,0 +1,945 @@
...
+        | None = None,
+        gamma: float = 0.0,
+        sigma_fn: Callable[[Float[Tensor, " *shape"]], Float[Tensor, " *shape"]]
+        | None = None,
+        alpha_fn: Callable[[Float[Tensor, " *shape"]], Float[Tensor, " *shape"]]
+        | None = None,
+    ) -> None:
+        if gamma > 0 and sigma_fn is None:
+            raise ValueError("sigma_fn must be provided when gamma > 0")
+        self.mask = mask
```
NickGeneva: Question, should these be modules? I see mask and y in the init, and makes me worried if it will be on the same device as the sample provided to the call?

physicsnemo/diffusion/noise_schedulers/noise_schedulers.py
```diff
@@ -14,14 +14,1730 @@
...
+            Data type of the tensor.
+
+        Returns
+        -------
+        Tensor
+            Initial noisy latent state of shape :math:`(B, *spatial\_shape)`.
+        """
+        ...
+
+    def get_denoiser(
```
NickGeneva, line 204: If I understand correctly, this function `get_denoiser` basically applies a scheduler to a one step denoiser function?

Bit confused with the naming, because we get_denoiser... but we already have one?

physicsnemo/diffusion/noise_schedulers/noise_schedulers.py
```diff
@@ -14,14 +14,1730 @@
...
+    - :meth:`sigma_dot`: Time derivative :math:`\dot{\sigma}(t)`
+    - :meth:`alpha`: Compute the signal coefficient :math:`\alpha(t)`
+    - :meth:`alpha_dot`: Time derivative :math:`\dot{\alpha}(t)`
+    - :meth:`timesteps`: Generate discrete time-steps for sampling
+    - :meth:`sample_time`: Sample diffusion times for training
+
+    **Concrete methods (have default implementations, but can be overridden for
+    custom behavior):**
+
+    - :meth:`drift`: Drift term :math:`f(\mathbf{x}, t)` for ODE/SDE
```
NickGeneva, line 289: Should some of these be private methods?

physicsnemo/diffusion/samplers/samplers.py
```diff
@@ -14,13 +14,337 @@
...
+    >>> tN = torch.tensor([1.0, 1.0])
+    >>> xN = scheduler.init_latents((3, 8, 8), tN)
+    >>>
+    >>> # For this minimal scheduler, denoiser is used directly as ODE RHS
+    >>> denoiser = lambda x, t: -x  # Simple mean-reverting ODE: dx/dt = -x
+    >>> x0 = sample(denoiser, xN, scheduler, num_steps=10, solver="euler")
+    >>> x0.shape
+    torch.Size([2, 3, 8, 8])
+    """
+    solver_options = solver_options or {}
```
NickGeneva: why force none to dict? Could the default just be a dict?

physicsnemo/diffusion/samplers/samplers.py
```diff
@@ -14,13 +14,337 @@
...
+    if isinstance(solver, str):
+        if solver not in SOLVERS:
+            available = ", ".join(f'"{k}"' for k in SOLVERS.keys())
+            raise ValueError(
+                f"Unknown solver '{solver}'. Available solvers: {available}."
+            )
+        solver_cls = SOLVERS[solver]
+        solver_ = solver_cls(denoiser, **solver_options)
+    else:
+        # Assume solver is a Solver-like object with a step method
```
NickGeneva, line 332: Hmmmmmm, Whats the use case to support this? Could you just make solver a required paramter?

physicsnemo/diffusion/samplers/samplers.py
```diff
@@ -14,13 +14,337 @@
...
+        t_steps = time_steps.to(device=xN.device, dtype=xN.dtype)
+    else:
+        t_steps = noise_scheduler.timesteps(num_steps, device=xN.device, dtype=xN.dtype)
+
+    # Main sampling loop
+    samples: List[Tensor] = []
+    x = xN
+    n_steps = len(t_steps) - 1  # Last element is 0 (final time)
+
+    for i in range(n_steps):
```
NickGeneva, line 350: nice work, I really like this is literally just a function with a for loop.
Someone could easily customize this as needed. 👍

physicsnemo/diffusion/samplers/solvers.py
```diff
@@ -0,0 +1,737 @@
...
+            ``t_next``, same shape as ``x``.
+        """
+        # Reshape t for broadcasting: (B,) -> (B, 1, ..., 1)
+        t_cur_bc = t_cur.reshape(-1, *([1] * (x.ndim - 1)))
+        t_next_bc = t_next.reshape(-1, *([1] * (x.ndim - 1)))
+
+        gamma_base = min(self.S_churn / self.num_steps, math.sqrt(2) - 1)
+
+        # Compute perturbed time t_hat with increased noise
+        if self.sigma_fn is not None and self.sigma_inv_fn is not None:
```
NickGeneva: instead of having a conditional branch here... could these function get converted to some pass through in the init if they are none? might clean this up a bit

NickGeneva: Idk, maybe its better like this explicit

physicsnemo/diffusion/samplers/solvers.py
```diff
@@ -0,0 +1,737 @@
...
+        # Validate sigma_fn and sigma_inv_fn
+        if (sigma_fn is None) != (sigma_inv_fn is None):
+            raise ValueError(
+                "sigma_fn and sigma_inv_fn must both be provided or both None."
+            )
+        self.sigma_fn = sigma_fn
+        self.sigma_inv_fn = sigma_inv_fn
+        self.scale_fn = scale_fn
+
+    def step(
```
NickGeneva, line 687: Do you have some analytical check the math is right here? I didnt bother checking it again the paper myself... but one of us can if needed.

Want to make sure these are hardened and trusted

physicsnemo/diffusion/guidance/dps_guidance.py
NickGeneva: iirc this is a literature thing to be consistent with the papers, if its not +1 to considering changing it

physicsnemo/diffusion/guidance/dps_guidance.py
CharlelieLrt: I actually don't use the `@runtime_checkable` here and I am not expecting anyone to use it later. If as Nick said there is no downside in having it, then I'll keep it

physicsnemo/diffusion/guidance/dps_guidance.py
CharlelieLrt: Yeah, you are both right. It's a notation from the original paper, but it hasn't been widely adopted, so I renamed to `observation_operator` instead of `A`.

physicsnemo/diffusion/guidance/dps_guidance.py
laserkelvin: I guess it does introduce a performance overhead but in the grand scheme of things, may not that big a deal 🤷🏼

physicsnemo/diffusion/guidance/dps_guidance.py
CharlelieLrt: Done

physicsnemo/diffusion/guidance/dps_guidance.py
CharlelieLrt: Fixed

physicsnemo/diffusion/guidance/dps_guidance.py
CharlelieLrt: Done

physicsnemo/diffusion/guidance/dps_guidance.py
CharlelieLrt: Good catch. Moved the if statement to the `__init__`

physicsnemo/diffusion/guidance/dps_guidance.py
CharlelieLrt: Good point! I fixed the hint and switched the mask to bool, then convert it to float in the `__init__`

physicsnemo/diffusion/guidance/dps_guidance.py
CharlelieLrt: Right. Done

physicsnemo/diffusion/guidance/dps_guidance.py
CharlelieLrt: Should be. We just want to back-propagate to the input of the diffusion model, not further than that.

physicsnemo/diffusion/guidance/dps_guidance.py
CharlelieLrt: IMO it doesn't make sense to make them Module (ok, there is no real downside in making them Modules, but still...)

You still raised a good point about the device. I modified the `__call__` such that the input `x` becomes the only source of truth for both `device` and `dtype`. That is, if the parameters `self.mask` and `self.y` do not have the same device and dtype, they are automatically converted to that of `x` in the `__call__`

That should be sufficient to mitigate most risks.

physicsnemo/diffusion/noise_schedulers/noise_schedulers.py
CharlelieLrt: All comments below are about the parent PR #1363. It's not yet merged to main so it polluted the history here. I'll address those directly on this PR instead of here.

physicsnemo/diffusion/noise_schedulers/noise_schedulers.py
CharlelieLrt: Note: comment should be on #1363 .

That's a good question. Some are indeed only for internal use within a noise scheduler. They are not expected to be called outside of it. In that case I am inclined on making them private. However, I do see some relatively common use cases where one might want to inherit from an existing noise scheduler and override one of these *possibly-private* method. That's really the only motivation to have them as public methods.

physicsnemo/diffusion/noise_schedulers/noise_schedulers.py
CharlelieLrt: > Bit confused with the naming, because we get_denoiser... but we already have one?


Ok I think that's a very good point. This made me realize that the `DiffusionDenoiser` protocol is actually not well scoped, which makes it very confusing under its current form.

A typical workflow would be:

`Predictor (X0Predictor | ScorePredictor | NoisePredictor | VelocityPredictor) --> convert to ScorePredictor if necessary --> optionally combines with some DPSGuidance (which are themselves instances of ScorePredictor) --> get_denoiser --> DiffusionDenoiser which can be used in a sampler`

Now, all these predictors and denoiser share the same signature `(x: Tensor, t: Tensor) --> Tensor`. What differs between them is the type of tensor being returned (an x0, a score, the noise for a predictor, or the RHS of an ODE/SDE for a denoiser). So, I believe I made an abusive use of the current `DiffusionDenoiser`. Instead there should be a different protocol, maybe just `Predictor`. There could also abemore granularity with `ScorePredictor`, `X0Predictor`, `NoisePredictor`, ... but the danger is that new types of parameterizations will be used and we need to always add new ones. Maybe I am overthinking this.

@NickGeneva thoughts?

physicsnemo/diffusion/guidance/dps_guidance.py
NickGeneva: Sounds good :+1:

physicsnemo/diffusion/samplers/solvers.py
CharlelieLrt: Note: comment should be on https://github.com/NVIDIA/physicsnemo/pull/1363 .

I think that's a valid comment, because these checks might introduce graph breaks if we compile the sampling function.

> could these function get converted to some pass through in the init if they are none


Yes, I'll do that

physicsnemo/diffusion/samplers/solvers.py
CharlelieLrt: Note: comment should be on https://github.com/NVIDIA/physicsnemo/pull/1363 .

Valid comment, not only for the solvers, but also the noise schedulers. I've been looking at all of them, but it's honestly difficult to make sure that everything is right. Automated math verification is also complicated because for some noise schedulers the noise schedule is quite convoluted (and random).
Any manual double-check of the math only is more than welcome!

physicsnemo/diffusion/noise_schedulers/noise_schedulers.py
NickGeneva: imo, I think maybe just renaming the function to something like `apply_schedule`, `apply`, or something would clear things up. Heck could event make this __init__ maybe

Because basically this is just doing a modification of a single step denoiser into one thats not step dependent right?

I dont think a separate protocol is needed, because its still a denoiser at the end of the day right. I just has not constant.

I'm mentally drawing comparisons to torch optim

```
optimizer = optim.SGD(model.parameters(), lr=0.01, momentum=0.9)
scheduler = ExponentialLR(optimizer, gamma=0.9)
```

But for diffusion we have something kinda like

```
denoiser = DPSDenoiser(....)
denoiser = NoiseScheduler(denoiser, ....)
```

physicsnemo/diffusion/noise_schedulers/noise_schedulers.py
NickGeneva: No worries about changing it then :+1:

physicsnemo/diffusion/noise_schedulers/noise_schedulers.py
CharlelieLrt: > Because basically this is just doing a modification of a single step denoiser into one thats not step dependent right?

Yeah kind of, but there are nuances to that:

- The original denoiser can be something else than a "clean data predictor". It could also predict a score, (or a combinations of multiple scores if it is used with DPS guidance), or even in some cases the noise itself. Calling it "denoiser" sounds a little mis-named in this case.
- If it is indeed a "clean data predictor", it is indeed a one step denoiser. However, no-one would ever use that in practice to denoise. It just predicts mean, which is not very useful. It's also where your pytorch mental model is not totally comparable: in pytorch you can definitely use an optimizer without a scheduler, but here it would make no sense to use a 1-step denoiser without a noise scheduler. The noise scheduler is _nearly_ mandatory I would say. So concretely, in your example above, in 99% of cases the `denoiser = DPSDenoiser(....)` cannot be used directly, it has to be converted with `denoiser = NoiseScheduler(denoiser, ....)` to be usable for sampling.

NickGeneva: Makes me wonder if denoise.step and scheudler.step could be a thing :thinking:

physicsnemo/diffusion/samplers/solvers.py
NickGeneva: Sorry apparently I'm reviewing two PRs at once lol

CharlelieLrt: Yes, but that's fine 😆 I know which comment is for which

physicsnemo/diffusion/noise_schedulers/noise_schedulers.py
CharlelieLrt: I mean yes, you could move the solver part to scheduler (that's what is done in HuggingFace Diffuser) and have a `scheudler.step `. The thing is that is does reduce flexibility: it is very common to mix _any_ solver with _any_ scheduler.

The most accurate mental model is just solving on ODE of this form:

$$
\frac{dx}{dt} = F(G(x, t), t)
$$

Here:
- $G$: original trained model + optional guidance (those are all score predictors)
- $F$: the denoiser returned by `get_denoiser`
- The solver is the numerical scheme used to integrate the ODE

NickGeneva: > I mean yes, you could move the solver part to scheduler (that's what is done in HuggingFace Diffuser) and have a scheudler.step . 

Yeah, you would need to have scheduler.step() essentially just modify some attribute in the denoiser... its kinda round about... but perhaps this would then be something like multi-diffusion where you can then do multiple diffusion steps at the same time index (analogous to a batch) and then index your sample step

> If it is indeed a "clean data predictor", it is indeed a one step denoiser. However, no-one would ever use that in practice to denoise.

What about distilled models that do operations with like 1 step sampling?

NickGeneva: Oh wait im seeing what your saying, the input to these functions (denoiser_in) is really just the loss guidance.. not an actual denoiser.

Okay I see

CharlelieLrt: > What about distilled models that do operations with like 1 step sampling?


Well, they don't _always_ do 1 step. Sometimes it's 2-3. But anyways, they fall under the same framework of ODE solve above. The only difference is that the RHS (obtained from `get_denoiser`) is not the same anymore (so basically new noise scheduler), and the original model is not an x0-predictor, but a velocity predictor.

After this discussion I do believe there is a clear conceptual distinction between a predictor (essentially a trained diffusion model + optionally some guidance), which cannot be used as the RHS of an ODE, and a denoiser, which is the actual RHS of an ODE.

Same applies for the guidance: for the guidances, the input should be a (score) predictor, and it is transformed into another (score) predictor. But there is the additional step of converting those to a denoiser to be able to sample with them. Also you can't pass a denoiser to a guidance ( I mean yes sure you could, but it does not make too much sense to me).

So I think there are two options:

- Introducing a new `Predictor` protocol, distinct from `DiffusionDenoiser` (but same signature). This has its danger though, because maybe too much opinionated. So that would allow less flexibility (at the end of the day, we don't care, all these things have the same signature, and maybe there will be some new ways to combine them which we cannot foresee now).
- Just rename the `DiffusionDenoiser` into something maybe a bit more generic that would encompass both predictors and denoisers. This kleaves the dorr open to combine them in any way, as long as they all have the same signature.

NickGeneva: Closing the loop. Clarified things offline.

Tldr, some renaming of protocols for the type hints and also some usability improvements to the denoiser to support either providing a score function or a x0 predictor (model), via force keyword args.

physicsnemo/diffusion/samplers/solvers.py
NickGeneva: Sounds good, no regressions since past release is sufficient.

physicsnemo/diffusion/samplers/solvers.py
```diff
@@ -0,0 +1,737 @@
...
+        x_prime = x + self.alpha * h * d_cur
+
+        # Mask for elements where t_next != 0 (need 2nd order correction)
+        # Shape: (B, 1, ..., 1) for broadcasting
+        mask_bc = (t_next_bc != 0).float()
+
+        # Second RHS evaluation (compute everywhere, masked later)
+        # Avoid division by zero in denoiser by using t_cur where t_prime is 0
+        t_prime = t_prime_bc.reshape(x.shape[0])
+        t_prime_safe = torch.where(t_prime == 0, t_cur, t_prime)
```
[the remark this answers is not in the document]
CharlelieLrt: We are doing `torch.where(t_prime == 0, t_cur, t_prime)`. Here `t_prime` has shape `(B,)` and `t_cur` has the same shape `(B,)`, so there is no problem.

physicsnemo/diffusion/samplers/samplers.py
CharlelieLrt: The current `solver` and `solver_options` parameters are mostly for similarity with [scipy.solve_ivp](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html), which is very well known by most people working on scientific applications
  * For ~80% of use cases, one would just pass a string `solver = "solver_name"`, without specifying any option. That's for non-expert users.
  * Then for 15% of other uses cases, one would pass a string `solver = "solver_name"`, plus a `solver_options = {...}` to override some default settings of the solver. That's for moderately advanced users.
  * Then for the last 5%, one would implement a custom solver from scratch and pass the instance with `solver=my_custom_solver`.
 
That's at least the philosophy in `scipy.solve_ivp`, and that's the one we are trying to reproduce here.

physicsnemo/diffusion/samplers/solvers.py
CharlelieLrt: Done.

physicsnemo/diffusion/noise_schedulers/noise_schedulers.py
```diff
@@ -14,14 +14,1730 @@
...
+        sigma_t_bc = self.sigma(t_bc)
+        sigma_dot_t_bc = self.sigma_dot(t_bc)
+        alpha_t_bc = self.alpha(t_bc)
+        alpha_dot_t_bc = self.alpha_dot(t_bc)
+        g_sq = (
+            2 * sigma_dot_t_bc * sigma_t_bc
+            - 2 * (alpha_dot_t_bc / alpha_t_bc) * sigma_t_bc**2
+        )
+        return g_sq.expand_as(x)
+
```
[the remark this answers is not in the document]
CharlelieLrt: The `expand_as` as been removed. Now `g_sq_bc` is broadcastable to shape of ``x``, but with trailing 1s in its shape.

physicsnemo/diffusion/samplers/samplers.py
```diff
@@ -14,13 +14,337 @@
...
+from physicsnemo.diffusion.base import DiffusionDenoiser
+from physicsnemo.diffusion.noise_schedulers import NoiseScheduler
+
+from .solvers import (
+    EDMStochasticEulerSolver,
+    EDMStochasticHeunSolver,
+    EulerSolver,
+    HeunSolver,
+    Solver,
 )
```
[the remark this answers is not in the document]
CharlelieLrt: This has been fixed with a more informative jaxtyping annotation "N_plus_1" for the shape

PR #1388, opened 2026-02-10
Enhance version checking with less boiler plate and better messages.

coreyjadams:
-------------------------
- main
  - https://github.com/NVIDIA/physicsnemo/pull/1388 :point_left:
    - https://github.com/NVIDIA/physicsnemo/pull/1389
      - https://github.com/NVIDIA/physicsnemo/pull/1390
        - https://github.com/NVIDIA/physicsnemo/pull/1391

<sup>[Stack](https://www.git-town.com/how-to/proposal-breadcrumb.html) generated by [Git Town](https://github.com/git-town/git-town)</sup>

### DIFF
--- a/v2.0-MIGRATION-GUIDE.md
+++ b/v2.0-MIGRATION-GUIDE.md
@@ -35,7 +35,9 @@ Several new packages have been introduced for PhysicsNeMo v2.0.  At a high level
    these layers, where appropriate.
 2. `physicsnemo.domain_parallel` contains the `ShardTensor` object and utilities.
 3. [TBD] `physicsnemo.diffusion`
-4. [TBD] `physicsnemo.mesh`
+4. `physicsnemo.mesh` contains a GPU-accelerated mesh processing library for
+   simplicial meshes of any dimension, including discrete calculus, curvature
+   analysis, subdivision, repair, and visualization.
 5. `physicsnemo.datapipes` brings reusable and generic utilities to standardize
 GPU-centric data pipelines for SciML.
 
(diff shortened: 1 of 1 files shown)


coreyjadams:
This is an attempt to organize this big PR a little better with stacking.  This is the bottom PR in the stack (and the first to be mergeable)

physicsnemo/core/version_check.py
```diff
@@ -16,45 +16,437 @@
...
+
+    return "\n".join(lines)
+
+
+# Registry for package install hints (maps package name to install instructions)
+# Each entry should include:
+#   - The optional dependency group (if applicable)
+#   - Install commands for both uv and pip
+#   - Link to documentation if needed
+_PACKAGE_HINTS: Dict[str, str] = {
```
peterdsharpe, line 164: Two questions, mostly related to the maintenance burden of keeping this in sync:

a) Is there a fallback to this? (E.g., a graceful default message in case we forget to add an install hint)
b) Could we parse the `pyproject.toml` instead? `tomllib` is built in, and this is effectively a reverse dictionary mapping. I guess there are "special" packages like `apex` which have their own instructions, but maybe we could provide overrides only for those packages?

physicsnemo/core/version_check.py
```diff
@@ -102,26 +506,189 @@ def require_version_spec(package_name: str, spec: str = "0.0.0"):
...
             return func(*args, **kwargs)
 
         return wrapper
 
     return decorator
+
+
+# Private registry for OptionalImport instances - ensures same instance
+# is returned for the same module name across the codebase
+_optional_import_registry: Dict[str, "OptionalImport"] = {}
```
peterdsharpe:
```suggestion
_optional_import_registry: dict[str, "OptionalImport"] = {}
```

physicsnemo/core/version_check.py
```diff
@@ -102,26 +487,173 @@ def require_version_spec(package_name: str, spec: str = "0.0.0"):
...
         def wrapper(*args, **kwargs):
-            check_version_spec(package_name, spec, hard_fail=True)
+            # Use OptionalImport to get the nice error message with install hints
+            opt = OptionalImport(package_name)
+            if not opt.available:
+                opt._get_module()  # This will raise RuntimeError with hint
+            # If version specified, also check version
+            if spec != "0.0.0":
+                check_version_spec(package_name, spec, hard_fail=True)
             return func(*args, **kwargs)
```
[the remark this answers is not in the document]
coreyjadams: Resolved.

test/core/test_version_check.py
```diff
@@ -122,3 +128,441 @@ def fn():
...
+                "physicsnemo.core.version_check.metadata.distributions",
+                return_value=[mock_dist],
+            ):
+                # numpy should NOT match numpy-stl (they're different packages)
+                # This test documents current behavior - prefix match WILL find it
+                # because numpy-stl starts with "numpy-"
+                result = get_installed_version("numpy")
+                # Current implementation would find this due to hyphen prefix match
+                assert result == "3.0.0"
+
```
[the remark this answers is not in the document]
coreyjadams: Resolved, this was a good comment.

physicsnemo/core/version_check.py
coreyjadams: We *could* parse pyproject.toml (and if fact the import linter is doing exactly that elsewhere).  IMO its not worth building that infrastructure because this is truly a bandaid until we get lazy imports.

Actually ... If we want to keep things in sync, to me the right way is to parse pyproject.toml and this list in a test that can fail if we're missing a hint. I will see if claude can spin up a test for that.

coreyjadams:
Yes, @peterdsharpe I 100% agree that this is a weaker implementation of `lazy import` and it's good, actually, that the python standard is going to build this into the core repo.  Once we get above 3.15 we can deprecate this :).

physicsnemo/core/version_check.py
coreyjadams: I'm going to defer this - we need to get some of these in to unblock mesh, and this will trigger pyproject.toml and down-pr updates.  So I'll put this into a later PR.

PR #1389, opened 2026-02-10
Add importlib_metadata to pyproject.toml to handle dependency checking.

coreyjadams:
-------------------------
- main
  - https://github.com/NVIDIA/physicsnemo/pull/1389 :point_left:
    - https://github.com/NVIDIA/physicsnemo/pull/1390
      - https://github.com/NVIDIA/physicsnemo/pull/1391

<sup>[Stack](https://www.git-town.com/how-to/proposal-breadcrumb.html) generated by [Git Town](https://github.com/git-town/git-town)</sup>

### DIFF
--- a/physicsnemo/models/srrn/super_res_net.py
+++ b/physicsnemo/models/srrn/super_res_net.py
@@ -4,7 +4,7 @@
 """"""
 
 """
-SRResNet model. This code was modified from, 
+SRResNet model. This code was modified from,
 https://github.com/sgrvinod/a-PyTorch-Tutorial-to-Super-Resolution
 
 The following license is provided from their source,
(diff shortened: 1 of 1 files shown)


peterdsharpe (approved):
Straightforward improvements here, approved.

ktangsali (approved):
LGTM! Thanks for the detailed explanation on the internal chat. Adding it here for context and documentation. 

> Here's the chain of processes:
EntryPoints is the key class we're looking at.  It's in both places. The registry will create EntryPoint objects upon registering all the models.When we do that, most of the time it's fine.There is one doctest that fails because the conditional in module loads from an EntryPoint and it looks like if isinstance(EntryPoint).  It only fails if you import logging first. Digging deeper, either wandb or mlflow (or both) is somehow using importlib_metadata.EntryPoint and all our entry points become that type instead upon loading, if wandb or mlflow is imported first. Even though they are the same class and compatible, the isinstance fails in the doctests because of this. So, I just let importlib_metadata come back and do a check against either in the registry and module loading.

PR #1390, opened 2026-02-10
Ensure datapipes is checked for optional imports too.

coreyjadams:
-------------------------
- main
  - https://github.com/NVIDIA/physicsnemo/pull/1390 :point_left:
    - https://github.com/NVIDIA/physicsnemo/pull/1391

<sup>[Stack](https://www.git-town.com/how-to/proposal-breadcrumb.html) generated by [Git Town](https://github.com/git-town/git-town)</sup>

PR #1397, opened 2026-02-11
HealDA Sensor Embedder

aayushg55:
# PhysicsNeMo Pull Request

## Description

Adds obs sensor embedding modules used in HealDA.

## Checklist

## Review Process


aayushg55:
@NickGeneva @pzharrington Here are the sensor embedding modules.

physicsnemo/experimental/models/healda/point_embed.py
```diff
@@ -0,0 +1,652 @@
...
+
+class MultiSensorObsEmbedding(Module):
+    r"""Multi-sensor observation embedding onto a HEALPix grid.
+
+    Embeds observations from multiple sensor types into a unified representation
+    by applying per-sensor embedders and fusing the results.
+
+    Parameters
+    ----------
+    sensor_configs : list[dict[str, Any]]
```
aayushg55: The list of dicts `sensor_configs` feels a bit messy. Previously, we encapsulated this in a dataclass. Open to any suggestions.

physicsnemo/experimental/models/healda/point_embed.py
```diff
@@ -0,0 +1,652 @@
...
+
+
+def _prod(shape):
+    out = 1
+    for s in shape:
+        out *= s
+    return out
+
+
+def _offsets_to_batch_idx(offsets: torch.Tensor) -> torch.Tensor:
```
NickGeneva: General comment for all Tensor inputs, can these all get updated to JAX typing. This will allow you to spec the dimensions

Here's an example:
https://github.com/NVIDIA/physicsnemo/blob/main/physicsnemo/models/srrn/super_res_net.py#L297

physicsnemo/experimental/models/healda/point_embed.py
```diff
@@ -0,0 +1,652 @@
...
+@torch.compiler.disable
+def _split_by_sensor(
+    obs: torch.Tensor,
+    float_metadata: torch.Tensor,
+    pix: torch.Tensor,
+    local_channel: torch.Tensor,
+    local_platform: torch.Tensor,
+    obs_type: torch.Tensor,
+    offsets: torch.Tensor,
+    expected_num_sensors: int,
```
NickGeneva: why is this needed here? can we just assume offsets.shape[0] is the right dim? Bit confused about the check below with  `expected_num_sensors`

physicsnemo/experimental/models/healda/point_embed.py
```diff
@@ -0,0 +1,652 @@
...
+        raise ValueError(
+            f"offsets first dim ({nsensors}) must match expected_num_sensors ({expected_num_sensors})"
+        )
+
+    out: list[tuple[torch.Tensor, ...]] = []
+    total_obs = obs.shape[0]
+
+    for sensor_idx in range(nsensors):
+        end = offsets[sensor_idx, -1, -1].item()
+        start = 0 if sensor_idx == 0 else offsets[sensor_idx - 1, -1, -1].item()
```
NickGeneva: Confused about these two lines, we have:

```python
offsets[sensor_idx, -1, -1].item()
offsets[sensor_idx - 1, -1, -1].item()
```
so different dimensions... also why if conditional statement `start = 0 if`

physicsnemo/experimental/models/healda/point_embed.py
```diff
@@ -0,0 +1,652 @@
...
+    Outputs
+    -------
+    torch.Tensor
+        Sensor embedding grid with shape :math:`(B, T, N_{pix}, D_{out})`
+        in NEST order.
+    """
+
+    def __init__(
+        self,
+        *,
```
NickGeneva, line 334: why is the general positional args needed here?

physicsnemo/experimental/models/healda/point_embed.py
```diff
@@ -0,0 +1,652 @@
...
+        pix: torch.Tensor,
+        local_channel: torch.Tensor,
+        local_platform: torch.Tensor,
+        hpx_level: int,
+        batch_idx: torch.Tensor,
+        nbatch: int,
+    ) -> torch.Tensor:
+        """Aggregate observations to spatial grid and project to output dimension."""
+        # Convert observation pixels to aggregator grid resolution
+        aggregation_pix = pix // int(4.0 ** (hpx_level - self.hpx_level))
```
NickGeneva: Why is it `hpx_level - self.hpx_level` not following exactly what this line is doing

NickGeneva: Maybe these can have better names, self.hpx_level is the target emb level right... what is hpx_level then? The representitive hpx resolution of the sensor?

physicsnemo/experimental/models/healda/point_embed.py
```diff
@@ -0,0 +1,652 @@
...
+        raise ImportError(
+            (
+                "earth2grid is not installed, cannot use it as a backend for HEALPix padding.\n"
+                "Install earth2grid from https://github.com/NVlabs/earth2grid.git to enable the accelerated path.\n"
+                "pip install --no-build-isolation https://github.com/NVlabs/earth2grid/archive/main.tar.gz"
+            )
+        )
+
+
+def _prod(shape):
```
NickGeneva: does math.prod not work? You import the math module already

physicsnemo/experimental/models/healda/point_embed.py
```diff
@@ -0,0 +1,652 @@
...
+        obs: torch.Tensor,
+        float_metadata: torch.Tensor,
+        pix: torch.Tensor,
+        local_channel: torch.Tensor,
+        local_platform: torch.Tensor,
+        obs_type: torch.Tensor,
+        offsets: torch.Tensor,
+        hpx_level: int,
+    ) -> torch.Tensor:
+        if self.use_checkpoint:
```
NickGeneva: Whats this condition for?

NickGeneva: What is use_checkpoint

physicsnemo/experimental/models/healda/point_embed.py
```diff
@@ -0,0 +1,652 @@
...
+    -------
+    torch.Tensor
+        Embedded observations of shape :math:`(B, D, T, N_{pix})` in HEALPIX_PAD_XY
+        pixel order, where :math:`B` is batch size, :math:`D` is fusion dimension,
+        :math:`T` is time steps, and :math:`N_{pix}` is number of HEALPix pixels.
+    """
+
+    def __init__(
+        self,
+        sensor_configs: list[dict[str, Any]],
```
NickGeneva: If the sensor config is like

- ``name``: sensor name (bookkeeping, unused).
- ``nchannel``: number of sensor channels.
- ``nplatform``: number of sensor platforms.

Could this just be three separate parameters of list[str], list[int], and list[int]  which get validated that the lengths are the same?

physicsnemo/experimental/models/healda/point_embed.py
```diff
@@ -0,0 +1,652 @@
...
+        Flattened float metadata with shape :math:`(N_{obs}, M_{float})`.
+    pix : torch.Tensor
+        Flattened pixel indices of each observation with shape :math:`(N_{obs},)`.
+    local_channel : torch.Tensor
+        Flattened local channel ids of each observation with shape :math:`(N_{obs},)`.
+    local_platform : torch.Tensor
+        Flattened local platform ids of each observation with shape :math:`(N_{obs},)`.
+    obs_type : torch.Tensor
+        Flattened observation type ids with shape :math:`(N_{obs},)`.
+    offsets : torch.Tensor
```
NickGeneva, line 552: Not following how this tensor (S, B, T) is mapped to (N_{obs},).

Why would this just not be of size (S,)

NickGeneva: Is int input of the forward actually (B, T, N_obs)?

physicsnemo/experimental/models/healda/point_embed.py
aayushg55: Sure, we can remove this check

physicsnemo/experimental/models/healda/point_embed.py
aayushg55: `self.hpx_level` is the grid level of the sensor/model, whereas `hpx_level` is the level corresponding to the incoming pix tensor. This accounts for if the loader calculated pix on a higher resolution. Since we use the hpx nest format, converting from higher to lower resolution can be done by dividing by 4^(difference in level). But we don't need this added complexity and can assume pix is of the model level. Can simplify to not pass in hpx level/remove this

physicsnemo/experimental/models/healda/point_embed.py
pzharrington: Not sure if @NickGeneva agrees, but my 2c is that the dataclass is not necessarily the worst thing to have if it is a constructor argument; I assume you refactored bc this module subclasses `physicsnemo.Module` and thus the constructor args need to be JSON-serializable for the `.from_checkpoint()` functionality. If this is more of a helper module, specific to HealDA, imo it is ok to have it just subclass `torch.nn.Module` and then it could accept a dataclass if that's preferred -- not sure if that makes sense though, e.g. if the dataclass is intended to be a user-configurable thing that is passed to the top-level HealDA model (which should be a `physicsnemo.Module`)...

physicsnemo/experimental/models/healda/point_embed.py
aayushg55: `offsets[sensor_idx, -1, -1]` refers to the end of the current sensor.
`offsets[sensor_idx - 1, -1, -1] `refers to the end of the previous sensor at `sensor_idx - 1`. The alignment of the two lines makes it seem like we are indexing into an extra dimension. Can clean up the if statement and add some comments.

physicsnemo/experimental/models/healda/point_embed.py
```diff
@@ -0,0 +1,652 @@
...
+
+    def hpx_grid(*args, **kwargs):
+        """Dummy symbol for missing earth2grid backend."""
+        raise ImportError(
+            (
+                "earth2grid is not installed, cannot use it as a backend for HEALPix padding.\n"
+                "Install earth2grid from https://github.com/NVlabs/earth2grid.git to enable the accelerated path.\n"
+                "pip install --no-build-isolation https://github.com/NVlabs/earth2grid/archive/main.tar.gz"
+            )
+        )
```
coreyjadams: Some recent updates made an effort to reduce this boilerplate.  See, for example this PR: https://github.com/NVIDIA/physicsnemo/pull/1390

The functionality is already merged, that one is just applying it.  It may simplify your life here.

physicsnemo/experimental/models/healda/point_embed.py
aayushg55: No strong preference for using it. I included the keyword only * since `nplatform` and `nplatform` are both `int`, so with keyword only, it would force the user to be clear about what they are setting each to.

physicsnemo/experimental/models/healda/point_embed.py
aayushg55: Simplified logic and added comments to make this clearer

physicsnemo/experimental/models/healda/point_embed.py
aayushg55: Removed

physicsnemo/experimental/models/healda/point_embed.py
aayushg55: Using `.numel()` instead

physicsnemo/experimental/models/healda/point_embed.py
aayushg55: This is for `gradient_checkpointing`. It could be useful for memory saving, although I haven't tested what the saving actually is. Open to removing.

physicsnemo/experimental/models/healda/point_embed.py
aayushg55: Yea I think the separate lists make the most sense. The sensor names are not actually used by the model- I had initially thought to keep them around as metadata so it is easier to map each embedder network to what it actually is, but not sure if it is worth keeping.

physicsnemo/experimental/models/healda/point_embed.py
aayushg55: Thanks, updated to use the `OptionalImport` functionality.

physicsnemo/experimental/models/healda/point_embed.py
aayushg55: Thanks, added jaxtyping annotations everywhere

physicsnemo/experimental/models/healda/point_embed.py
aayushg55: Yea I removed the dataclass due to json-serializaling issues, and it seemed custom dataclasses were not preferred. For now, I switched to having separate parameters, each being a list, and then validated for matching length, as suggested by @NickGeneva.

physicsnemo/experimental/models/healda/point_embed.py
aayushg55: All of our inputs are this packed/flattened `nobs` tensor, where it is packed across time, then batch, then sensor. The offsets then indicate where each "window" or sample in the full flat tensor ends. Basically, it is describing a `(N_sensor,B,T,N_sensor_obs)` tensor, but all flattened so that we don't need to use padding. The batch and time support is to make the module more general, although we only really use B=1 and T=1.

aayushg55:
@greptileai

physicsnemo/experimental/models/healda/point_embed.py
NickGeneva: Oh okay, no fine to leave if implemented I think

physicsnemo/experimental/models/healda/point_embed.py
NickGeneva: Ohhh, i read `offsets[sensor_idx - 1, -1, -1]` as `offsets[sensor_idx, - 1, -1, -1]`

PR #1400, opened 2026-02-11
Bug fixes for ShardTensor+SongUNet

jleinonen:
# PhysicsNeMo Pull Request

## Description

This PR has concise fixes and workarounds that were needed to get `SongUNet` to work with `ShardTensor` in a domain parallel setting. All fixes should be backward compatible with previous functionality. They were tested with both 1-GPU and 2-GPU configurations with domain parallel sizes of 1 and 2 respectively, using `ShardTensor` in both cases.

### 1. `.flip()` error

Using `.flip()` in `SongUNet` was giving the following error: 
```python
NotImplementedError: Operator aten.flip.default does not have a sharding strategy registered.
```
**Solution**: Since the flipped dimension was of size 2, avoid using `.flip()` by replacing it with manual slicing and `torch.concat`.

### 2. Halo with domain parallel size of 1

I was getting this error:
 ```python
File "/usr/local/lib/python3.12/dist-packages/physicsnemo/domain_parallel/shard_utils/halo.py", line 717, in apply_halo_tensors
    shape = list(halo_from_right.shape)
                 ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'shape'
```
This was caused by both `halo_from_left` and `halo_from_right` being `None` when domain parallel size is 1. 

**Solution**: Since `local_tensor` has the same shape as `halo_from_right` except for the dimension `halo_config.tensor_dim`, we can replace `halo_from_right.shape` with `local_tensor.shape` without changing behavior. This allows the code to run even if `halo_from_left` and `halo_from_right` are both `None`.

### 3. Non-contiguous tensor error in attention

In `attention_patches.py`, I was getting errors of non-contiguous tensors in `aten._scaled_dot_product_efficient_attention`.

**Solution**: Add `.contiguous()` calls to creation of `lq`, `lk`, `lv` and `latn_mask` in `ring_sdpa`.

### 4. Error caused by use of `.unbind()` with domain parallel size 2

This only occurred using domain parallel size 2. The call `.unbind(-2)` in `UNetAttention.forward` was causing the error:
```python
  File "/usr/local/lib/python3.12/dist-packages/physicsnemo/domain_parallel/_shard_tensor_spec.py", line 552, in _infer_shard_tensor_spec_from_local_chunks
    raise ValueError(
ValueError: Rank 0 expected local shape torch.Size([1, 0, 8192, 64]) does not match tensor's local shape torch.Size([1, 1, 4096, 64])
```

**Solution**: Replace `.unbind(-2)` with manual slicing.

**Question**: As far as I can tell, if we change the order of permutations above this line such that the qkv dimension comes first, this should lead to contiguous (q, k, v) tensors that can be more efficiently processed by `scaled_dot_product_attention`. It _should_ be backward compatible, is it worth implementing here?

### TBD

Add tests that catch these bugs in case someone reintroduces them.

## Checklist

## Review Process

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Shape mistmatch bug in the Lennard Jones example
+
 ### Security
 
 ### Dependencies
--- a/examples/cfd/swe_distributed_gnn/README.md
+++ b/examples/cfd/swe_distributed_gnn/README.md
@@ -1,5 +1,12 @@
 # Shallow Water Equations - Distributed GraphCast
 
+> **Note**: This example is currently not functional. Distributed mode
+(partition_size > 1) is not supported with the PyG backend.
+Support for distributed message passing using PyG
+backend will likely be included in a future release.
+Until then, this example will raise a `NotImplementedError` when
+attempting to use distributed mode.
+
 This example demonstrates how to leverage a distributed version of GraphCast to scale
 to larger Graph Neural Network (GNN) workloads.
 
@@ -243,12 +250,24 @@ uses the following.
     )
 ```
 
+## Requirements
+
+This example requires the `gnns` optional dependency group to be installed. This includes
+`torch_geometric`, `torch_scatter`, `torch_sparse`, and `torch_cluster` which are needed
+for the distributed GraphCast model.
+
+To install the required dependencies, run:
+
+```bash
+pip install nvidia-physicsnemo[gnns]
+```
+
 ## Getting Started
 
 To get started with this example for a run e.g. using 4 GPUs on a single node, simply run,
 
 ```bash
-torchrun --nodes=1 --nproc-per-node=4 train.py
+torchrun --nnodes=1 --nproc-per-node=4 train.py
 ```
 
 In case you are running on a SLURM cluster and have started a batch job e.g. using 4 GPUs,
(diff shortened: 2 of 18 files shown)


coreyjadams:
Hi @jleinonen - I am only looking at the ShardTensor work here, fyi.  There is nothing wrong with the changes to halo.py, so I plan to approve.  That said, for domain_size==1 it's definitely not recommended to run ShardTensor, it can add some extra performance overhead.

Calling contiguous() is fine, too.

Do you need a sharding path for flip?  Seems like no but it's something relatively easy we can add in the future if it helps.

jleinonen:
> Hi @jleinonen - I am only looking at the ShardTensor work here, fyi. There is nothing wrong with the changes to halo.py, so I plan to approve. That said, for domain_size==1 it's definitely not recommended to run ShardTensor, it can add some extra performance overhead.
> 
> Calling contiguous() is fine, too.
> 
> Do you need a sharding path for flip? Seems like no but it's something relatively easy we can add in the future if it helps.

Thanks @coreyjadams, having a sharding path for `flip` is not necessary for this PR but I only checked the functioning of the SongUNet class with `ShardTensor`. There might be other uses of `flip` left in PNM that will break when someone tries to use them with `ShardTensor`. So maybe it'll be useful to have it in the future.

And yes, I only use sharding with domain parallel size of 1 for testing that the code works with sharded models and tensors (usually I try to get things to work with 1 GPU, then start a 2 GPU job to test actual sharding). Normally I set `self.shard = self.domain_parallel_size > 1` in the training code I'm working on.

PR #1412, opened 2026-02-13
CUDA 12/13 Cross-Compatibility

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Modifies the `pyproject.toml` to allow for installation on both CUDA 12 and CUDA 13 backends. Note that this PR is not a pure upgrade, and comes with some pros and cons that we should discuss.

Current system:
- `uv sync` installs core packages with a CUDA 13 backend.
- `uv sync --extra nn-extras` installs core packages + `nn-extras` with a CUDA 13 backend.
- There is no pre-packaged way to install with a CUDA 12 backend. For this, you'd have to first do something like `uv sync`, and then do `uv pip install -U <insert CUDA 12 package variant>` for CUDA-dependent wheel-based packages, like `torch`, `torchvision`, `cuml`, `pylibraft`, `cupy` etc.

Under the new system from this PR:
- `uv sync` installs core packages with the default PyTorch variant from PyPI. Currently that's CUDA 12.
- `uv sync --extra cu13` installs core packages with CUDA 13 PyTorch.
- `uv sync --extra cu13 --extra nn-extras` installs core packages with CUDA 13 PyTorch + `nn-extras`.
- `uv sync --extra cu12` installs core packages, and forces CUDA 12 backends.

So this is a system of "orthogonal" dependencies:
- For CUDA version, you have `<none>`, `--extra cu13`, or `--extra cu12`
- For level-of-dependency, you have `<none>`, `--extra nn-extras`, `--extra mesh-extras`, etc.

Thoughts?

## Checklist

## Review Process


pyproject.toml
```diff
@@ -152,22 +173,37 @@ dev = [
...
+]
 utils-extras = [
     "wandb",
     "mlflow",
     "line_profiler",
     "vtk",
     "stl",
-    "pylibraft-cu13", # Provides rmm
-    "cupy-cuda13x",
 ]
```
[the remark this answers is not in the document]
peterdsharpe: Fixed in new update.

pyproject.toml
```diff
@@ -62,6 +62,12 @@ Changelog = "https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md"
 
 [tool.uv]
 managed = true
+conflicts = [
+    [
+        { extra = "cu12" },
+        { extra = "cu13" },
+    ],
+]
```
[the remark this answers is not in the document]
peterdsharpe: GitHub doesn't use NVCR containers here; we should check on Blossom though.

coreyjadams:
Well, this is kind of a mess!  Not your fault of course.  I have a couple comments, and a couple questions:

- heads up that dali, transformer_engine need to be accounted for here too.  I have a PR pending that puts them into a "perf" group but we can revisit that too.  We'll have to merge conflicts if both of these go through.
- the gnn packages might also have strange cuda behaviors...

Some questions:
- Seems like this will break our uv-based CI, right?  We have to update install instructions.
- What is the behavior if cu12/cu13 is not selected?  No cuda support?
- Does warp require anything special for cuda?
- When we build wheels for pypi, does this mean building multiple wheels?  Or we pick on as the default?

pyproject.toml
```diff
@@ -62,6 +62,12 @@ Changelog = "https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md"
 
 [tool.uv]
 managed = true
+conflicts = [
```
ktangsali, line 63: TIL! Neat

peterdsharpe:
All great feedback - thanks both! To address questions:

>     * Seems like this will break our uv-based CI, right?  We have to update install instructions.

So I was worried that it would, but surprisingly it seems fine. This is because (from my understanding):

a) The GitHub-side CI doesn't test with GPU functionality, so no issues there.
b) The Blossom-side CI starts from a container base image with dependencies pre-installed for CUDA 13, so it automatically uses a consistent suite.

Because of this, both CI paths are currently passing on this PR. However, if we do switch to a `uv`-based CI install on Blossom (as contrasted against a container-based setup), then we'd need to include an appropriate flag.

>     * What is the behavior if cu12/cu13 is not selected?  No cuda support?

A good question. If neither `--extra cu12` or `--extra cu13` is specified, then:

- `uv sync` will install core dependencies with PyTorch pulled from PyPI, which is currently CUDA 12.8. (In other words, a reasonably safe fallback.) That said, something like `uv sync --extra cu13` is still recommended, since this is explicit about resolution.
- `uv sync --extra util-extras` will unfortunately NOT install RAPIDS packages (e.g., `cuml`), so this is an unsafe fallback. Unfortunately, the fact that RAPIDS packages have different names based on CUDA versions combined with the "no-override" part of pip specs means that this is unavoidable. Users would have to install with `uv sync --extra cu13 --extra util-extras` to ensure they get the full spec.

I've updated the `README.md` with language that goes into what the install process would look like. Basically, you'd always want to specify `--extra cu12` or `--extra cu13` to be safe, though we make best-effort to provide safe fallbacks where possible, within the limits of what can be done per pip-specs.

>     * Does warp require anything special for cuda?

[Warp is (yet another) interesting case](https://github.com/NVIDIA/warp#installing): The PyPI wheels that we install from are built with CUDA 12 runtime. To get CUDA 13, you'd need to install directly from their GitHub.

It turns out this isn't an issue for us at the moment, because unlike PyTorch vs. RAPIDS, they don't have mutually-conflicting child dependencies on `nvidia-cuda-toolkit` versions. So, you can use Warp with CUDA 12 backend on a CUDA 13 system, and it works fine without any dependency conflicts with other packages.

>     * When we build wheels for pypi, does this mean building multiple wheels?  Or we pick on as the default?

We'd still have one wheel, thankfully. Users would install with:

```bash
pip install "nvidia-physicsnemo[cu13]"
```
for base, and
```bash
pip install "nvidia-physicsnemo[cu13,mesh-extras]"
```
for extras. More examples in the `README.md` of this PR.

coreyjadams:
Ok awesome.  a few responses:

- For CI, it's true that the pull request CI is not using cuda, however that's just because I haven't finished things yet.  We can absolutely use CUDA on the nightly CI.  If you want to move this branch from your fork to the core repo, we ought to be able to trigger it manually and see what happens.
- Short term, my hope is to rebuild the CI container after code freeze with a more up to date stack.  We could even do one in 12 and one in 13 if we like pain.  So blossom can use that.
- Long term, I think we will migrate to github-only once the CI pipeline is really robust here, nightly and in PRs.  It's overall a better system (not perfect) but there are things to iron out still.

I appreciate that the wheel is reasonable still.  And I'm glad the default install still comes with torch + cuda :).

I think we should give some time for @NickGeneva and perhaps @dallasfoster / @laserkelvin to weigh in too, since this can be impactful downstream in e2s and alchemi.

laserkelvin:
FYI, I checked out the branch locally on my Cloud VM (x86 Ubuntu 24.04) with an L4, CUDA 12.8.

Running `uv sync` weirdly gave me `cu12` components, because I expected it to default to `cu13` instead...

EDIT: Sorry I have reading comprehension issues - I just scanned the first sentence in the description and my brain interpolated the rest

From what I could think of, it would mean that we likely have to propagate the `cu12`/`cu13` pattern downstream - is this just an exercise we'll have to repeat between CUDA releases (or at least until something like wheel-next becomes available to test)? I'm curious to hear what @NickGeneva's perspective would be for E2S

I do like the ability to explicitly switch versions; in my mind this is as good of a out-of-the-box experience a user could have (if the default version doesn't work for you, just run `uv sync --extra` without needing to do anything else!)

pyproject.toml
coreyjadams, line 63: I am dropping a junk comment here to deliberately hold up a merge.  There are some folks I want to look at this before it goes through.

pyproject.toml
```diff
@@ -62,6 +62,12 @@ Changelog = "https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md"
...
+    [
+        { extra = "cu12" },
```
NickGeneva, line 65: This can quickly blow up. Good idea to have it like this just two groups 👍

PR #1437, opened 2026-02-23
Note for --no-build-isolation for CUDA version mismatch

mnabian:
# PhysicsNeMo Pull Request

## Checklist

## Review Process


README.md
```diff
@@ -409,6 +409,11 @@ For a CUDA 12 environment, replace `cu13` with `cu12`:
 uv sync --extra cu12 --extra nn-extras
 ```
 
+> **Note:** If you encounter a CUDA version mismatch (e.g., "The detected
```
ktangsali: I think we should refine the language here - `--no-build-isolation` should be _the_ way to install these packages correctly. Without that, it can cause a version mismatch.

ktangsali: cc. @peterdsharpe @coreyjadams for viz.

mnabian: How about this?
> **Note:** Extras such as ``gnns``, ``nn-extras``, and ``transformer-engine`` pull in packages that either ship only source distributions or require wheels built for a specific PyTorch and CUDA version. These should be installed with ``--no-build-isolation``, e.g. ``uv sync --extra cu13 --extra gnns --no-build-isolation``.

peterdsharpe:
Oh cool, awesome find! 

I wonder - can we specify this in `pyproject.toml` so it happens automatically for users?

`uv` has support for new syntax using [`no-build-isolation-package`](https://docs.astral.sh/uv/reference/settings/#no-build-isolation-package), which instructs specific package(s) to not isolate when building.

@mnabian if you have a moment, would you be able to try adding this to `pyproject.toml` to see if it works?

```
[tool.uv]
no-build-isolation-package = ["torch_sparse", "torch_geometric"]
```
(Not sure which other GNN packages require no-build-isolation, but we should add those too)

If this works, then we might be able to make this "just work" with `uv sync --extra cu13 --extra gnns`. If it doesn't, then I agree that this docs-based approach is a perfectly fine alternative too!

README.md
ktangsali: I like this better, thanks!

mnabian:
> Oh cool, awesome find!
> 
> I wonder - can we specify this in `pyproject.toml` so it happens automatically for users?
> 
> `uv` has support for new syntax using [`no-build-isolation-package`](https://docs.astral.sh/uv/reference/settings/#no-build-isolation-package), which instructs specific package(s) to not isolate when building.
> 
> @mnabian if you have a moment, would you be able to try adding this to `pyproject.toml` to see if it works?
> 
> ```
> [tool.uv]
> no-build-isolation-package = ["torch_sparse", "torch_geometric"]
> ```
> 
> (Not sure which other GNN packages require no-build-isolation, but we should add those too)
> 
> If this works, then we might be able to make this "just work" with `uv sync --extra cu13 --extra gnns`. If it doesn't, then I agree that this docs-based approach is a perfectly fine alternative too!

@peterdsharpe I tested this out and ran into some other issues. We might be able to get it working, but right before the release, I’d prefer not to make big changes to the build unless you want to give this a try.

peterdsharpe (approved):
Seems like a reasonable change to me!

PR #1379, opened 2026-02-06
StormCast: Training improvements and code refactoring

albertocarpentieri:
## Description

This MR introduces several improvements to the StormCast training pipeline, focusing on configurability, reproducibility, and code maintainability.

### New features

* Added build_optimizer function supporting multiple optimizer types: adam, adamw, stableadamw
* Optimizer parameters (betas, weight_decay, eps, fused) are now configurable via YAML
* Added support for torch schedulers
* Helper functions init_scheduler and step_scheduler for scheduler management

### Performance

Added perf configuration section (similar to CorrDiff example) with:
  * fp_optimizations: Support for fp32, amp-fp16, amp-bf16
  * use_apex_gn: Apex GroupNorm with channels_last memory format

### Training loop

* Refactored monolithic training_loop function into a Trainer class for better readability and maintainability
* Reproducible validation with fixed seeding
* Configurable number of validation steps
* Validation loss saved in checkpoint metadata
* Gradient clipping support
* TensorBoard logging support

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [ ] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.
- [ ] If I am implementing a new model or modifying any existing model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation).

## Dependencies
torch-optimi

PR #1438, opened 2026-02-23
Stormcast refactor with domain parallel + DiT support, testing, Pydantic configs

jleinonen:
# PhysicsNeMo Pull Request

Refactors the StormCast training code as a follow-up to https://github.com/NVIDIA/physicsnemo/pull/1379.

Major additions:

1. Support for domain parallel training, controlled by `training.domain_parallel_size`. The domain parallelism is handled by `ParallelManager` in `utils/parallel.py`. FSDP is now used for parallelism regardless of domain parallel size.
2. Support for `DiT` for diffusion models, with standard configuration with NATTEN loaded by the setup function on `nn.py`.
3. Configuration parsing using Pydantic. This eliminates most of the manual validation and default value checking as those are now handled by the Pydantic validation. Hydra is still used for loading configs but afterwards they are passed into Pydantic.
4. A test suite using pytest. Includes tests for both single- and multi-GPU setups. A mock dataset that produces random data is included for testing. Multi-GPU tests can be run as `torchrun --standalone --nproc-per-node=<number_of_gpus> --no-python pytest test_training.py`.

More minor changes:

1. Logging is now passed through `utils/logging.py` which handles both Tensorboard and WandB logging depending on which are selected.
2. `utils/apex.py` enables using Apex LayerNorm with ShardTensor.
3. Split training-time plotting into multiple plots as this allows one to view them as an animation, which makes it easier to see if the model is learning temporal dynamics properly.

TODO / pending:

1. To support domain parallelism, `utils/loss.py` and `utils/sampler.py` contain versions of `EDMLoss` and `deterministic_sampler` modified to work with ShardTensor. TBD if we want to upstream the support to PhysicsNeMo library.
2. ~~`README.md` still to be updated.~~ Initial updated README.md is added.
3. `inference.py` still to be properly checked for domain parallel support.
 
## Checklist

## Review Process


examples/weather/stormcast/datasets/dataset.py
```diff
@@ -57,6 +57,10 @@ def state_channels(self) -> list[str]:
         """Metadata for the state channels. A list of channel names, one for each channel"""
         pass
 
+    def scalar_condition_channels(self) -> list[str]:
+        """Metadata for the scala condition channels. A list of channel names, one for each channel"""
```
pzharrington: *scalar

examples/weather/stormcast/utils/nn.py
```diff
@@ -87,20 +99,75 @@ def get_preconditioned_architecture(
...
+        condition_dim=condition_dim,
+        conditioning_embedder=conditioning_embedder,
+        **model_kwargs
+    )
+    return EDMPreconditioner(model=ConcatConditionWrapper(dit))
+
+
 def build_network_condition_and_target(
     background: torch.Tensor,
-    state: tuple[torch.Tensor | None, torch.Tensor],
```
pzharrington, line 92: This is reverting a change made in Alberto's PR https://github.com/NVIDIA/physicsnemo/pull/1379 -- intentional or no?

examples/weather/stormcast/utils/sampler.py
```diff
@@ -0,0 +1,570 @@
...
+from torch.distributed.tensor import DTensor
+from torch.distributed.tensor.placement_types import Replicate, Shard
+
+warnings.warn(
+    "The 'deterministic_sampler' function from 'physicsnemo.diffusion.samplers' "
+    "is a legacy implementation that will be deprecated in a future release. "
+    "Updated implementations will be provided in an upcoming version.",
+    LegacyFeatureWarning,
+    stacklevel=2,
+)
```
pzharrington, line 42: Did you copy this in just to remind us to pivot away from the legacy samplers? 😆

examples/weather/stormcast/utils/parallel.py
```diff
@@ -0,0 +1,484 @@
...
+from torch.distributed.tensor.placement_types import Replicate, Shard
+
+from physicsnemo.distributed import DistributedManager
+from physicsnemo.domain_parallel.shard_tensor import ShardTensor, scatter_tensor
+
+from datasets.dataset import worker_init
+from utils.nn import nested_to
+
+
+class ParallelManager:
```
pzharrington: Nit: this name sort of confuses with the `DistributedManager` -- I thought you had written your own version of it for a second. Suggest something like `ParallelHelper` or similar instead

examples/weather/stormcast/utils/parallel.py
```diff
@@ -0,0 +1,484 @@
...
+            Global batch size across data-parallel ranks.
+
+        Returns
+        -------
+        int
+            Per-rank batch size.
+        """
+        return global_batch_size // self.data_parallel_size
+
+    def sharded_dataloader(
```
pzharrington, line 99: Can you explain the rationale for this as opposed to just using e.g. `DistributedSampler` from PyTorch directly as in standard data parallelism?

examples/weather/stormcast/utils/trainer.py
```diff
@@ -90,175 +88,90 @@ class Trainer:
...
+        self.regression_net = self._load_regression_net()
+
+        # Sharding
+        if self.shard:
+            self.logger.info("Sharding model for domain parallelism")
+            self.net = self.parallel_manager.distribute_model(self.net)
+            if self.regression_net is not None:
+                self.regression_net = self.parallel_manager.distribute_model(self.regression_net)
+            if self.invariant_tensor is not None:
+                self.invariant_tensor = self.parallel_manager.distribute_tensor(self.invariant_tensor)
```
pzharrington: This looks like if `self.shard` is False, the model will never get wrapped in FSDP?

examples/weather/stormcast/utils/parallel.py
```diff
@@ -0,0 +1,484 @@
...
+        -------
+        Iterator[torch.Tensor | dict | list]
+            Iterator over (sharded if the shard attribute if True) batches.
+        """
+        data_iter = iter(dataloader)
+
+        i = 0
+        batch = None
+        while True:
+            source_rank_in_mesh = i % self.domain_parallel_size
```
pzharrington, line 192: I find this logic to be quite confusing. While technically the best way to support ShardTensor in the dataloader is to have the datasets provide already sharded data, I think we can leave that out of scope for now and come back to later since we want to be friendly towards custom data use-cases. However I don't see the point of this rotating source rank across iterations. I suggest modifying to:

- Use a standard `DistributedSampler` based on the data parallel rank to handle partitioning dataset samples across the `ddp` mesh
- Within each data parallel rank, always fetch a full batch (i.e., the full domain) each step. (provided by dataloader)
- If `self.shard` is False, yield batch directly.
- If `self.shard` is True, scatter from a deterministic source domain parallel rank (no modulo counter / no i == 0 special case).

examples/weather/stormcast/utils/parallel.py
pzharrington, line 1: I feel like all the different state dict handlers might be able to be cleaned up a bit here and in the training loop.  I do like that you keep the full (non-sharded) state dict the only on-disk format for checkpoints -- maybe there are other constraints with checkpoint save/load I'm not thinking of, but wouldn't it be better to:

- Introduce one shared layout registry (param-name → placement/shard dim), used by:
  - `partition_fn` for runtime param distribution,
  - model-state load adapter,
  - optimizer-state load adapter.
- Replace separate full/local long-lived copies of the model with a simple conversion adapter layer (this might also permit consolidation of the separate model/optimizer state_dict methods you have in the current parallel manager):
   - full checkpoint → runtime mode state,
  - runtime mode state → full checkpoint.
- Optionally keep full model only transiently on rank 0 during save/load to reduce memory.

examples/weather/stormcast/utils/parallel.py
jleinonen: Renamed `ParallelHelper` in https://github.com/NVIDIA/physicsnemo/pull/1438/commits/90bb452840065725f997cc7a7ac0c942038e5ccc.

examples/weather/stormcast/utils/trainer.py
jleinonen: Yeah my usage of `self.shard` was a bit confused as a relic of earlier versions of `ParallelHelper`. I renamed `self.shard` to `self.use_shard_tensor` and now it only indicates whether we use `ShardTensor` in models and inputs. FSDP is always used regardless of ShardTensor usage. (https://github.com/NVIDIA/physicsnemo/pull/1438/commits/46a6787d55a4be3d224c015c6587b85b795b3669)

examples/weather/stormcast/utils/nn.py
jleinonen: I discussed with Alberto and it cannot be `None`. While `None` would be the most natural way to indicate that you don't have a previous state conditioning (that is, you're training a pure downscaling model), unfortunately the PyTorch dataloaders cannot handle `None` values in the data structure returned by `__getitem__`. I would like to have a bit more elegant solution for this than returning dummy tensors, though.

examples/weather/stormcast/datasets/dataset.py
jleinonen: Fixed (https://github.com/NVIDIA/physicsnemo/pull/1438/commits/2f39967c8b1bdb5528bfd8fe2ad75f1066c0f5cb).

examples/weather/stormcast/utils/sampler.py
jleinonen: I didn't put a lot of effort into this file as I wasn't sure if we'd want to port the changes to PNM. Probably we should just refactor the code to use the new samplers?

pzharrington: Yeah. For now I think fine to include these, and we'll have a known working reference commit against which to test the integration of updated diffusion APIs

examples/weather/stormcast/utils/parallel.py
jleinonen: > Introduce one shared layout registry

Definitely agree, actually I planned to implement this before PR but forgot

> Replace separate full/local long-lived copies of the model with a simple conversion adapter layer

Not sure I understand this. Does the answer below address this as well? 

> Optionally keep full model only transiently on rank 0 during save/load to reduce memory.

I'd think keeping the full model in memory is not a big deal because the limiting factor is probably maximum memory utilization and that will be equal regardless of whether the model is kept all the time or not. And re-creating the full model for every `save_checkpoint` adds some time spent creating the model.

pzharrington: Yeah ok fair point about max memory, I was thinking more along the lines of just code cleanliness. I felt like it would be cleaner to have the one (FSDP-ed/sharded) model + optimizer states in the trainer, and the parallel helper would hide away the intermediate full model copies. But probably what you have here is sufficient for a first pass and really we should think about better support in PNM checkpoint save/load to hide away some of the boilerplate when using FSDP and/or ShardTensor. For now let's just go with the current setup, but ideally with one centralized registry specifying which parameters in the state_dicts need scatter/gathering

jleinonen: Centralized function for determining sharded parameters is added in https://github.com/NVIDIA/physicsnemo/pull/1438/commits/3a19949e1dcaed6d658459878f38f8d627847450.

examples/weather/stormcast/README.md
```diff
@@ -20,25 +23,52 @@ regional weather forecasting and climate risk analysis.
 <img src="../../../docs/img/stormcast_rollout.gif"/>
 </p>
 
-The design of StormCast relies on two neural networks:
- 1. A regression model, which provides a deterministic estimate of the next HRRR timestep given the previous timestep's HRRR and background ERA5 states
- 2. A diffusion model, which is given the previous HRRR timestep as well as the estimate from the regression model, and provides a correction to the regression model estimate to produce a final high-quality prediction of the next high-resolution atmospheric state.
+2. [Stormscope](https://arxiv.org/abs/2601.17268),
+a nowcasting/nearcasting model for clouds and precipitation, using GOES satellite radiances and NEXRAD
```
pzharrington: NEXRAD -> MRMS

examples/weather/stormcast/README.md
```diff
@@ -20,25 +23,52 @@ regional weather forecasting and climate risk analysis.
...
+```
+
+The multi-GPU tests might produce some clutter as all processes write to the terminal, but the output should be readable.
+
+The tests take a while to run. Potentially useful `pytest` options include `-k "test_training"` (run only the tests from the `test_training` function) and `-x` (stop at first failure).
 
 ### Configuration basics
 
-StormCast training is handled by `train.py`, configured using [hydra](https://hydra.cc/docs/intro/) based on the contents of the `config` directory. Hydra allows for YAML-based modular and hierarchical configuration management and supports command-line overrides for quick testing and experimentation. The `config` directory includes the following subdirectories:
+Training is handled by `train.py`, configured using [hydra](https://hydra.cc/docs/intro/) based on the contents of the `config` directory. Hydra allows for YAML-based modular and hierarchical configuration management and supports command-line overrides for quick testing and experimentation. The `config` directory includes the following subdirectories:
```
pzharrington: Add some description of `pydantic` updates

examples/weather/stormcast/README.md
```diff
@@ -92,11 +122,25 @@ Both regression and diffusion training can be distributed easily with data paral
 torchrun --standalone --nnodes=1 --nproc_per_node=8 train.py --config-name <your_distributed_training_config>
 ```
 
+#### Domain parallelism
+
+As an advanced form of distributed training, the training code also supports domain parallelism, adapted from Stormcast. It distributes individual training samples across multiple GPUs, and therefore can be used to train large models and/or large domains when even a single sample will not fit in the memory of one GPU. Domain parallelism is controlled by the `training.domain_parallel_size` config setting that specifies how many GPUs each sample should be distributed to. For example, if you have at least two GPUs
```
pzharrington: *StormScope

examples/weather/stormcast/README.md
```diff
@@ -173,13 +225,44 @@ Once you have implemented the custom dataset, create a configuration file in `co
...
+Possible solutions include decreasing batch size, training with more GPUs and using domain parallelism. See the section [Memory management](#memory-management).
+
+### How do I train a model without low-resolution conditioning (e.g. a nowcasting model)?
+
+Ensure that the `model.diffusion_conditions` and `model.regression_conditions` settings for your model don't include `"background"`. Also set `background` in the `__getitem__` function of your dataset to a dummy value such as `np.array([0])`.
+
+### How do I train pure downscaling model without state update?
+
+Ensure that the `model.diffusion_conditions` and `model.regression_conditions` settings for your model don't include `"state"`. Also set `state[0]` in the `__getitem__` function of your dataset to a dummy value such as `np.array([0])`.
+
```
pzharrington, line 245: Tweak descriptions: dummy value only needed if you don't have any background data

PR #1450, opened 2026-02-25
Patching refactor

CharlelieLrt:
# PhysicsNeMo Pull Request

## Checklist

## Review Process

### DIFF
--- a/physicsnemo/diffusion/multi_diffusion/patching.py
+++ b/physicsnemo/diffusion/multi_diffusion/patching.py
@@ -249,6 +249,12 @@ class RandomPatching2D(BasePatching2D):
 
         max_y = self.img_shape[0] - self.patch_shape[0]
         max_x = self.img_shape[1] - self.patch_shape[1]
+        # TODO: use torch.randint instead of random.randint to create
+        # patch indices directly on right device. Note: this will break
+        # non-regression tests because torch.randint and random.randint do not
+        # use the same random number generation process. But for an object taht
+        # is deliberately designed to be random, breaking these non-regression
+        # tests might not be a problem.
         new_indices = torch.tensor(
             [
                 (random.randint(0, max_y), random.randint(0, max_x))


physicsnemo/diffusion/multi_diffusion/patching.py
```diff
@@ -72,40 +63,54 @@ def __init__(
...
         Returns
         -------
         Tensor
-            Patched tensor, shape depends on specific implementation.
+            Patched tensor of shape :math:`(P \times B, C, H_p, W_p)`.
         """
-        pass
+        if isinstance(input, Tensor):
+            return self(input, *args, **kwargs)
+        return super().apply(input)
```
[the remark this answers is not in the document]
CharlelieLrt: Look at line 95... Line 96 will never be called if input is a Tensor!

physicsnemo/diffusion/multi_diffusion/patching.py
```diff
@@ -115,245 +120,277 @@ def fuse(self, input: Tensor, **kwargs) -> Tensor:
...
+        device = self.patch_indices.device if has_buffer else None
+        dtype = self.patch_indices.dtype if has_buffer else torch.long
+
+        new_indices = torch.tensor(
+            [
+                (
+                    random.randint(0, self.img_shape[0] - self.patch_shape[0]),
+                    random.randint(0, self.img_shape[1] - self.patch_shape[1]),
+                )
+                for _ in range(self.patch_num)
```
coreyjadams: Is there ever an exepectation that patch_num is relatively large?  Would be good to avoid this for loop here.

You could probably build this index tensor directly with torch.randint on the device?

test/diffusion/test_patching.py
```diff
@@ -0,0 +1,786 @@
...
+        )
+        x = make_input(INPUT_SHAPE, seed=GLOBAL_SEED, device=device)
+
+        def apply_fn(tensor):
+            return patcher.apply(tensor)
+
+        compiled_fn = torch.compile(apply_fn, fullgraph=False)
+
+        out_eager = patcher.apply(x)
+        out_compiled = compiled_fn(x)
```
coreyjadams: FYI, I think you could reduce some boiler plate and catch future updates using a helper function here:
https://github.com/NVIDIA/physicsnemo/blob/main/test/common/optimization.py#L207

That said, there is nothing wrong with what you are doing.  I tested and fullgraph=True passes and I recommend you use it in all three instances.

physicsnemo/diffusion/multi_diffusion/patching.py
CharlelieLrt: I think that's a good point; `patch_num` could be a few 100s max I would say. I like the idea of creating directly on the device with `torch.randint`, however this will break the non-regression tests that I added in the previous PR, because AFAIK `random.randint` and `torch.randint` use totally different random number generation processes. Even setting them with the same seed should not be sufficient to ensure that these indices are the same.


That being said, we are explicitly working on an object that is called _**Random**_ Patching, so expecting determinism in non-regression tests for this is probably an overkill. If we accept to break the non-regression tests (which are not totally representative of the expected behavior in practical use-cases), then I can switch to `torch.randint`.

What do you think?

test/diffusion/test_patching.py
CharlelieLrt: The `fullgraph=False` is definitely an omission from my side. I changed it to `True`.


Regarding the suggestion of using the helper function, I am not sure I can use it. It seems specialized for compiling `Module` instances, but here I am compiling a function.

CharlelieLrt:
@coreyjadams I launched CI with all reference data from the test PR. If they pass that means we are not breaking anything with this refactor.

Then we can decide what to do with your [comment above](https://github.com/NVIDIA/physicsnemo/pull/1450#discussion_r2885653897). But at least if it breaks after switching to `torch.randint`, we know that it is deliberate.

test/diffusion/test_patching.py
coreyjadams: It's duck typed: I think it will work just fine in your case.  That said, it's not broken and I'm not gonna force a fix :)

CharlelieLrt: Even the `nop_backend`? Sorry I don't know much about setting custom backends, but it seems to expect a `forward` method, right?

coreyjadams:
@CharlelieLrt I think it would be sufficient to put a nice to do in the code summarizing there might be an optimization opportunity to avoid the for loop and use torch.randint, but it might break tests, blha blah blah.  burn that bridge when we get there, so to speak?

To me I think there are bigger tasks of higher priority to ignore first :)

test/diffusion/test_patching.py
coreyjadams: I think that will just prevent the actual compilation after tracing.  Otherwise, tests would take forever to run when we do compilation.  So the backend is named "noop" here I guess"

PR #1457, opened 2026-02-26
Cleanup function spec, plus adding a few functionals

loliverhennigh:
# PhysicsNeMo Pull Request

## Description

This PR is a cleanup pass for `physicsnemo.nn.functional` plus interpolation/test/benchmark updates. This PR will be followed by one more concentrated on CI integration with Github actions.

What’s included:

- Reorganized functionals into clearer category folders (`fourier_spectral`, `geometry`, `neighbors`, `interpolation`, `regularization_parameterization`). This mimics the Docs while not changing API or import paths. Imports still come from `nn.functional.xyz`, this is cleans stuff up internally.
- Migrated interpolation to the newer `grid_to_point_interpolation` structure with proper warp support and backward pass. Also added a `point_to_grid_interpolation` for fun and completness.
- Added two other functionals from ongoing work, `mesh_to_voxel_fraction` and `mesh_poisson_disk_sample`. If its too much to review these I can remove them.
- Updated unit tests for functionals so they all follow the same structure. This required some small but important changes with base FunctionSpec implementation.
- Refreshed benchmark scripts/registry and docs paths so plots/docs line up with the new organization.

Overall goal: make functionals easier to navigate, easier to test, and easier to benchmark.


benchmarks/physicsnemo/nn/functional/benchmark_functionals.py
```diff
@@ -26,6 +26,8 @@
 
 from benchmarks.physicsnemo.nn.functional.registry import FUNCTIONAL_SPECS
 
+_PHASE_ORDER = ("forward", "backward")
```
loliverhennigh: We now run benchmarks on forward and backward (if backward exists). I generate two bar plots now for these instead of just the forward pass before.

loliverhennigh: Again, these files are kinda throw away until I have the ASV setup with the CI/CD system.

benchmarks/physicsnemo/nn/functional/_spec_utils.py
```diff
@@ -0,0 +1,102 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
```
loliverhennigh, line 1: Code in the benchmarks is still a bit junky.

peterdsharpe:
Hi @loliverhennigh - would it be possible to break this PR up into smaller ones for review? It looks like this touches a ton of files, so it's possible to break this down we can divide-and-conquer to review more quickly. If this is genuinely indivisible work, we can also coordinate here to ensure review coverage.

(If it's helpful, the workflow I'm currently experimenting with for breaking up "big work" is to maintain my own branch, and then use `git merge --no-commit` to selectively create sub-PRs from this - see https://github.com/NVIDIA/physicsnemo/pull/1481 for an example.)

docs/_media/functionals/geometry/mesh_poisson_disk_sample/generate.py
```diff
@@ -0,0 +1,227 @@
+#!/usr/bin/env python3
```
loliverhennigh, line 1: This code is completely throwaway. Its just used to make the images in the docs for like the rotating bunny and stuff. If there is any opposition to adding these I will just remove them to keep everything cleaner

PR #1472, opened 2026-03-05
Checkpoint save/load FSDP and ShardTensor support

pzharrington:
# PhysicsNeMo Pull Request

### Summary
- FSDP/ShardTensor-aware checkpoint save and load: `save_checkpoint` and `load_checkpoint` now automatically detect FSDP-wrapped and DTensor/ShardTensor-distributed models and use PyTorch's Distributed Checkpoint (DCP) state-dict APIs to gather/scatter model and optimizer state. In distributed mode all ranks call the functions collectively, while only rank 0 performs file I/O. This eliminates the need for manual parameter gathering/scattering that recipe code (e.g. StormCast) previously had to implement.
- New `load_model_weights` utility: A convenience function for loading a single `.mdlus` or `.pt` file directly into a (potentially distributed) model, handling FSDP + DTensor redistribution automatically.
- StormCast recipe simplification: Removed ~200 lines of manual checkpoint gather/scatter logic from `parallel.py` (`gather_training_state`, `scatter_optimizer_state`, `shard_state_dict`, `scatter_object`, `get_state_dict_shard)` and ~50 lines of rank-0 CPU model/optimizer bookkeeping from trainer.py. All ranks now participate symmetrically in `_resume_or_init`, calling `load_checkpoint` / `save_checkpoint` directly.
- `physicsnemo.core.Module.save`: Added an optional `state_dict` parameter so `save_checkpoint` can pass a pre-gathered full state dictionary for FSDP/DTensor models without calling `self.state_dict()` on the distributed module.
- Minimum torch version bump 2.4 → 2.5: Required because `StateDictOptions.broadcast_from_rank0` (used in the pure-FSDP load path) was introduced in PyTorch 2.5. This option enables rank 0 to broadcast the full state dict to all other ranks without manual scatter, which is the standard non-DTensor distributed load mechanism.

## Checklist

## Review Process

### DIFF
--- a/physicsnemo/core/module.py
+++ b/physicsnemo/core/module.py
@@ -439,7 +439,7 @@ class Module(torch.nn.Module):
         file_name: Path | str | None = None,
         verbose: bool = False,
         legacy_format: bool = False,
-        state_dict: dict | None = None,
+        _state_dict: dict | None = None,
     ) -> None:
         """
         Utility method for saving a ``Module`` instance to a '.mdlus' checkpoint file.
@@ -454,8 +454,8 @@ class Module(torch.nn.Module):
         legacy_format : bool, optional, default=False
             Whether to save the model in legacy tar format. If True, saves as tar archive.
             If False (default), saves as zip archive.
-        state_dict : dict | None, optional, default=None
-            Pre-computed state dictionary to save.  When provided the model's
+        _state_dict : dict | None, optional, default=None
+            Internal pre-computed state dictionary to save.  When provided the model's
             own ``state_dict()`` is **not** called and ``state_dict`` is
             serialized directly.  This is used by
             :func:`~physicsnemo.utils.checkpoint.save_checkpoint` to pass a
@@ -588,7 +588,7 @@ class Module(torch.nn.Module):
                 with zipfile.ZipFile(tmp_path, "w", zipfile.ZIP_STORED) as archive:
                     # Save model state dict
                     state_dict_buffer = io.BytesIO()
-                    _sd = state_dict if state_dict is not None else self.state_dict()
+                    _sd = _state_dict if _state_dict is not None else self.state_dict()
                     torch.save(_sd, state_dict_buffer)
                     archive.writestr("model.pt", state_dict_buffer.getvalue())
 
@@ -612,7 +612,7 @@ class Module(torch.nn.Module):
                 local_path = Path(temp_dir)
 
                 # Save model state dict
-                _sd = state_dict if state_dict is not None else self.state_dict()
+                _sd = _state_dict if _state_dict is not None else self.state_dict()
                 torch.save(_sd, local_path / "model.pt")
 
                 # Save args
--- a/physicsnemo/utils/checkpoint.py
+++ b/physicsnemo/utils/checkpoint.py
@@ -54,6 +54,7 @@ from torch.optim.lr_scheduler import LRScheduler
 
 import physicsnemo
 from physicsnemo.core.filesystem import LOCAL_CACHE, _download_cached
+from physicsnemo.core.module import Module
 from physicsnemo.distributed import DistributedManager
 from physicsnemo.utils.capture import _StaticCapture
 from physicsnemo.utils.logging import PythonLogger
@@ -92,14 +93,7 @@ def _unwrap_ddp_compile(
     return model
 
 
-def _get_model_class_name(model: torch.nn.Module) -> str:
-    """Return the class name of the innermost module (looking through FSDP)."""
-    if isinstance(model, FSDP):
-        return type(model.module).__name__
-    return type(model).__name__
-
-
-def _get_inner_module(model: torch.nn.Module) -> torch.nn.Module:
+def _unwrap_fsdp(model: torch.nn.Module) -> torch.nn.Module:
     """Unwrap one FSDP layer (if present) to reach the user module."""
     if isinstance(model, FSDP):
         return model.module
@@ -246,20 +240,20 @@ def _extract_mdlus_state_dict(
 ) -> dict[str, Any]:
     """Read only the ``state_dict`` from a ``.mdlus`` archive."""
     cached = _cache_if_needed(file_name)
-    if tarfile.is_tarfile(cached):
+    fmt = Module._detect_checkpoint_format(cached)
+
+    if fmt == "tar":
         with tarfile.open(cached, "r") as tar:
             f = tar.extractfile("model.pt")
             return torch.load(
                 io.BytesIO(f.read()), map_location=device, weights_only=False
             )
-    elif zipfile.is_zipfile(cached):
+    else:
         with zipfile.ZipFile(cached, "r") as archive:
             model_bytes = archive.read("model.pt")
         return torch.load(
             io.BytesIO(model_bytes), map_location=device, weights_only=False
         )
-    else:
-        raise IOError(f"Cannot determine checkpoint format for {file_name}")
 
 
 def _get_checkpoint_filename(
@@ -407,7 +401,7 @@ def _unique_model_names(
     model_dict: dict[str, list[torch.nn.Module]] = {}
     for model0 in models:
         model0 = _unwrap_ddp_compile(model0, loading=loading)
-        base_name = _get_model_class_name(model0)
+        base_name = type(_unwrap_fsdp(model0)).__name__
         if base_name in model_dict:
             model_dict[base_name].append(model0)
         else:
@@ -481,10 +475,12 @@ def save_checkpoint(
         Arbitrary key-value pairs persisted alongside the training state
         (e.g. best validation loss, MLflow run ID).
     optimizer_model : torch.nn.Module | None, optional
-        The model whose parameters the ``optimizer`` is tracking.
-        Required by the DCP ``get_optimizer_state_dict`` helper when
-        distributed mode is active.  When ``None``, the first model in
-        ``models`` is used.  Ignored when *not* in distributed mode.
+        The model whose parameters the ``optimizer`` is tracking so that
+        parameter unsharding of optimizer state can be performed correctly.
+        Only required when multiple models are provided, and at least one of
+        them is a distributed model (FSDP/ShardTensor). When ``None``, the
+        first model in ``models`` is used.  Ignored when *not* in distributed
+        mode.
 
     Examples
     --------
@@ -547,7 +543,7 @@ def save_checkpoint(
 
     # == Saving model checkpoint ==
     for name, model in named_models.items():
-        inner = _get_inner_module(model)
+        inner = _unwrap_fsdp(model)
         model_type = "mdlus" if isinstance(inner, physicsnemo.core.Module) else "pt"
         file_name = _get_checkpoint_filename(
             path,
@@ -566,7 +562,7 @@ def save_checkpoint(
             if should_write:
                 state_dict = _cpu_offload_state_dict(state_dict)
                 if isinstance(inner, physicsnemo.core.Module):
-                    inner.save(file_name, state_dict=state_dict)
+                    inner.save(file_name, _state_dict=state_dict)
                 else:
                     with fs.open(file_name, "wb") as fp:
                         torch.save(state_dict, fp)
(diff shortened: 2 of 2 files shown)


pzharrington:
@greptileai

coreyjadams:
@pzharrington Does this need review from shard tensor side too?

examples/minimal/ShardTensorExamples/1_vector_addition/vector_add_sharded.py
```diff
@@ -17,7 +17,8 @@
 import torch
 import time
 
-from physicsnemo.distributed import DistributedManager, scatter_tensor
+from physicsnemo.distributed import DistributedManager
+from physicsnemo.domain_parallel import scatter_tensor
```
coreyjadams, line 21: Should have caught this sooner and I appreciate you got it.  Thanks!

coreyjadams:
I took a look through this PR - over all I think this is much needed (and long overdue) functionality, thank you for finally taking action when no one else would!

Overall I have one concern to discuss: over the next release I think it's important to decouple the ShardTensor(DTensor) inheritance structure (I know that's not a surprise just raising it in _this_ context... ).  Will that break anything you've implemented here?  We probably will need to use distributed tooling but there could be weird behavior introduced if I do that.

What do you think?

pzharrington:
Copying response from Slack here for posterity

> yes pivoting away from ShardTensor(DTensor) inheritance will break some stuff, but the level of refactoring needed I guess would depend on how far ShardTensor ends up drifting from DTensor. I wouldn't let the checkpoint functionality determine much in the "to DTensor or not to DTensor" decision as I assume whatever we end up with, there will be a non-horrible pathway to keeping the same user-facing functionality for checkpoints by shuffling things around under the hood. Bar none, eventually it will need some refactoring to support fsdp2 instead of FSDP (which is deprecating NO_SHARD)

examples/weather/stormcast/utils/nn.py
```diff
@@ -111,7 +111,7 @@ def get_preconditioned_natten_dit(
     patch_size: int = 4,
     attn_kernel_size: int = 31,
     lead_time_steps: int = 0,
-    layernorm_backend: Literal["torch", "apex"] = "apex",
+    layernorm_backend: Literal["torch", "apex"] = "torch",
```
jleinonen, line 114: I assume this change is intentional? It's not mentioned in the PR description. But fine for me since it can be overridden by config file.

pzharrington: Yes, installing apex via `uv`/`pip` is a bit of a pain so fixed the default to torch for a better install experience

physicsnemo/core/module.py
```diff
@@ -453,6 +454,12 @@ def save(
         legacy_format : bool, optional, default=False
             Whether to save the model in legacy tar format. If True, saves as tar archive.
             If False (default), saves as zip archive.
+        state_dict : dict | None, optional, default=None
+            Pre-computed state dictionary to save.  When provided the model's
+            own ``state_dict()`` is **not** called and ``state_dict`` is
+            serialized directly.  This is used by
+            :func:`~physicsnemo.utils.checkpoint.save_checkpoint` to pass a
+            pre-gathered full state dictionary for FSDP / DTensor models.
```
CharlelieLrt: I am not sure I understand how this new `state_dict` argument is used and what it is used for. I find the pattern of doing `some_model.save(..., state_dict=some_other_state_dict)` a little counterintuitive. Also, ideally we would like to keep the `save` method with as few arguments as possible to keep it simple and easy to use. These explanations seem to suggest that the `state_dict` is only used by `physicsnemo.utils.checkpoint.save_checkpoint`, so mostly for _internal_ use only.

I think there are three options, from most to least preferable:

1. `state_dict` is indeed only used by `physicsnemo.utils.checkpoint.save_checkpoint` and not meant to be exposed: in this case hide it from the `save` method here

2. Same case as 1 above, but not really possible to _totally hide it_ from the `save` method: then mark it explicitly for internal use only, for example rename it to `_state_dict`, and clearly explain in the docstring that it is for internal use only.

3. `state_dict` is not only for internal use and is actually meant to be exposed: then I would at least add an example in the docstring to show the typical use-case for this new argument.

physicsnemo/utils/checkpoint.py
```diff
@@ -532,14 +847,268 @@ def load_checkpoint(
...
+        loaded_epoch = checkpoint_dict["epoch"]
 
     if metadata_dict is not None:
         metadata_dict.update(checkpoint_dict.get("metadata", {}))
 
-    return epoch
+    return loaded_epoch
+
+
+def load_model_weights(
```
CharlelieLrt, line 856: Maybe this new method should be added to the corresponding `.rst` file in the API docs?

physicsnemo/utils/checkpoint.py
```diff
@@ -41,12 +61,214 @@
...
+    if isinstance(model, FSDP):
+        return type(model.module).__name__
+    return type(model).__name__
+
+
+def _get_inner_module(model: torch.nn.Module) -> torch.nn.Module:
+    """Unwrap one FSDP layer (if present) to reach the user module."""
+    if isinstance(model, FSDP):
+        return model.module
+    return model
```
CharlelieLrt: I feel like these two methods are redundant. You could replace a call to `_get_model_class_name(model)` by `type(_get_inner_module(model)).__name__`. Also to be extra-safe it would be better to make explicit in the method name that it is only for FSDP. IIRC we have many other helper functions in this module, which get the inner module, but for `compile` or `DDP`. So it would be better to avoid any possible confusion here.

physicsnemo/utils/checkpoint.py
```diff
@@ -41,12 +61,214 @@
...
+def _is_mdlus_archive(path: str) -> bool:
+    """Return ``True`` if *path* is a ``.mdlus`` archive (tar or zip containing ``model.pt``)."""
+    cached = _cache_if_needed(path)
+    if tarfile.is_tarfile(cached):
+        with tarfile.open(cached, "r") as tar:
+            return "model.pt" in tar.getnames()
+    if zipfile.is_zipfile(cached):
+        with zipfile.ZipFile(cached, "r") as archive:
+            return "model.pt" in archive.namelist()
+    return False
```
CharlelieLrt, line 235: Not necessary I think: you can just call the static method `Module._detect_checkpoint_format()`

physicsnemo/utils/checkpoint.py
```diff
@@ -41,12 +61,214 @@
...
+                io.BytesIO(f.read()), map_location=device, weights_only=False
+            )
+    elif zipfile.is_zipfile(cached):
+        with zipfile.ZipFile(cached, "r") as archive:
+            model_bytes = archive.read("model.pt")
+        return torch.load(
+            io.BytesIO(model_bytes), map_location=device, weights_only=False
+        )
+    else:
+        raise IOError(f"Cannot determine checkpoint format for {file_name}")
```
CharlelieLrt: That's a little odd here, because the behavior differs depending on the type of checkpoint (tar vs zip), but the method is not calling `_is_mdlus_archive` (or `Module._detect_checkpoint_format()`).

One would expect to:
1. Call the helper function to determine the type of checkpoint
2. Load it depending on its type

In addition, the overall load logix appears quite different from the one used in [Module.load()](https://github.com/NVIDIA/physicsnemo/blob/125cf76e60f425bdbe3396991ffc0af7b20544e9/physicsnemo/core/module.py#L736-L780). I am in particular concerned because the one here does not use `Module._safe_members()`.

physicsnemo/utils/checkpoint.py
```diff
@@ -260,6 +480,11 @@ def save_checkpoint(
     metadata : dict[str, Any] | None, optional
         Arbitrary key-value pairs persisted alongside the training state
         (e.g. best validation loss, MLflow run ID).
+    optimizer_model : torch.nn.Module | None, optional
+        The model whose parameters the ``optimizer`` is tracking.
```
CharlelieLrt: > The model whose parameters the ``optimizer`` is tracking.

Is there any use-case where it is different from the model contained in `models`?
I'm not totally sure I can think of a case where they differ, but if you have any in mind, it would be good to add it as a docstring example

physicsnemo/utils/checkpoint.py
```diff
@@ -532,14 +847,268 @@ def load_checkpoint(
...
+            inner.load(weights_path)
+        else:
+            cached = _cache_if_needed(weights_path)
+            if is_mdlus:
+                sd = _extract_mdlus_state_dict(weights_path, device)
+            else:
+                sd = torch.load(cached, map_location=device, weights_only=False)
+            inner.load_state_dict(sd)
+        checkpoint_logging.success(f"Loaded model weights from {weights_path}")
+        return
```
CharlelieLrt, line 899: This seems to be doing nearly the same thing with `Module.save()`, right? Would it be possible to just call it here instead?

physicsnemo/core/module.py
pzharrington: Yeah, that's fair. To address this I'm going with (2). Call sites of `Module.save` in the checkpoint saving for distributed models have to first gather the parameters of FSDP/Sharded models appropriately -- however once that is done we still want to make use of the recursive serialization and stuff done by `Module.save` to get the standard `mdlus` format. Putting the gathering into `Module.save` would pollute the core module code too much imo, so I think the better pattern is to just pass an overriding `state_dict` for these cases

pzharrington: I was thinking maaybe there could be some other exotic use-case where passing an overriding state_dict would be useful but can't think of explicit examples so let's just go with (2) instead of (3).

physicsnemo/utils/checkpoint.py
pzharrington: (assuming you mean `Module.load()` instead of `Module.save()`)...The idea here was `load_model_weights` supports loading from `.mdlus` or `.pt` into either `nn.Module` or `physicsnemo.Modle` targets -- `Module.load` only works loading into `physicsnemo.Module`. So we use `Module.load` when that case applies, and otherwise handle the other cases

physicsnemo/utils/checkpoint.py
pzharrington: One could have the case of having several models/Modules in `models` that they are saving into the checkpoint, but only one of them is actually getting optimized and has a corresponding optimizer to save. Can add an example

pzharrington: Actually, I don't think the `>>>` example format will work since this option only really applies to FSDP-wrapped models and that requires a distributed environment. So I will just try to clarify the docstring

physicsnemo/utils/checkpoint.py
pzharrington: Good call, I'll drop `_get_model_class_name` and rename `_get_inner_module` to `_unwrap_fsdp` for clarity.

physicsnemo/utils/checkpoint.py
pzharrington: Well, `Module._detect_checkpoint_format` returns the format type whereas this is just a booelan check on whether or not the file is a `.mdlus` format with model saved into it. I don't think it really cleans it up to add a try-except block using `_detect_checkpoint_format`, would be about the same LOC

physicsnemo/utils/checkpoint.py
pzharrington: I updated this to use `_detect_checkpoint_format` for more consistency. However the difference with `Module.load` is we just want to pull out the state dict from a path we already know to be a `.mdlus` file (call sites use the `_is_mdlus_archive` check beforehand). And correct me if I'm wrong but I don't think we need `Module._safe_members()` since we're reading the tar file directly into memory (`tar.extractfile`), no path traversal issues

physicsnemo/utils/checkpoint.py
CharlelieLrt: Yeah, right

physicsnemo/utils/checkpoint.py
CharlelieLrt: > And correct me if I'm wrong but I don't think we need Module._safe_members() since we're reading the tar file directly into memory (tar.extractfile)

I am not 100% expert on the question, but I think that makes sense

physicsnemo/utils/checkpoint.py
CharlelieLrt: Yeah that's right, I missed this part

PR #1482, opened 2026-03-10
Improves License Check Performance

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Minor Quality-of-Life fix.

Currently, `make license` is kind of slow. On my machine, it runs in **10.68 seconds**:

```
$ time make license
make license  4.44s user 4.12s system 80% cpu 10.681 total
```

With some changes, we can make it way faster (even while running here on all files). After this PR, we're at **0.18 seconds** (**60x speedup**):

```
$ time make license
make license  0.15s user 0.03s system 104% cpu 0.177 total
```

This is nice, since `make license` ordinarily runs in `pre-commit`. So, previously, you would have to either a) incur a ~11 second delay on each commit, or b) reduce scope to not run `--all-files`, which is usually fine, except that some git workflows allow some missed changes to slip through.

There are also some other minor issues with the current implementation:
- We stop file reading only upon `from` or `import` literals, which is fine for `*.py` files but not for other files (e.g., `Dockerfile`). And in cases of Python files without imports, we read the whole thing, which is slow.

## Checklist

## Review Process


coreyjadams:
Good call switching to gitpython.

How is this check passing files to the check?  pre-commit will pass a list of files by default which this check seems to be overriding previously, and now using gitpython might still be overriding?

Anyways, the _easiest_ answer might be, if the bottleneck was walking the file system, let pre-commit send the file list to check and only check those?  That's what I did with import linting and it works nicely.

peterdsharpe:
Great feedback! Let me take a look, I think you're right that we can eke out a bit more perf here.

peterdsharpe:
@coreyjadams @ktangsali 

Latest update (summarized by Opus):

Here's what we landed on:

**subprocess / GitPython is gone entirely.** Rather than switching to GitPython (which would've required `additional_dependencies: ['GitPython']` in the pre-commit config), we took your suggestion further: the script no longer does any file discovery at all. Pre-commit already handles file selection via the `files:` and `exclude:` patterns in `.pre-commit-config.yaml`, and passes matching files as positional arguments to the hook. The `--all-files` flag and `get_all_files()` function were redundant machinery duplicating what pre-commit already does.

**Changes (net -53 lines, 3 files):**

- `header_check.py` is now a pure "check these files" script. Removed `get_all_files`, `get_top_comments`, `--all-files` arg, `import json`, and all `config.json` loading. Added lazy header reading (stops at first non-comment line instead of reading entire files), a `HeaderCheckResult` NamedTuple, type hints, and a module-level compiled regex.
- `config.json` deleted. Its three responsibilities moved elsewhere: `copyright_file` is derived from `__file__`, `include-ext` is the pre-commit `files:` pattern, `exclude-dir` is the pre-commit `exclude:` pattern.
- `Makefile` `license` target changed to `pre-commit run license -a`, matching `black`, `interrogate`, and `lint` targets.

The original bottleneck was `get_all_files()` using `rglob` to walk 18,740 files (including `.venv`) instead of the 1,588 git-tracked files. By letting pre-commit handle discovery, that problem goes away without introducing any new dependencies.

PR #1493, opened 2026-03-12
Update nightly uv pipeline

coreyjadams:
# PhysicsNeMo Pull Request

This is the first PR of several to finally bring GitHub CI in line with Blossom CI.  This is running our pipeline though missing a few critical components still.  I'll highlight what it's doing and why, and at the end go through what is still missing.  This PR is focused only on the **nightly UV testing workflow** but that is intended to be cornerstone of the PR CI eventually.

## Nightly UV workflow.

The [nightly UV workflow](https://github.com/NVIDIA/physicsnemo/blob/update-nightly-uv-pipeline/.github/workflows/github-nightly-uv.yml) runs 4 stages: 

1. Build and cache the environment.  I tried a number of tricks here, they all stunk for various reasons.  We can not cache a container on the CI node on github so we don't have an easy solution.  In the end, I settled on this as the "best" of the mediocre options:  I use the ubuntu24.04 cuda12.8.1-cudnn-devel image.  I bootstrap a number of extra packages into it ([via a custom github action](https://github.com/NVIDIA/physicsnemo/blob/update-nightly-uv-pipeline/.github/actions/bootstrap-cudnn-ci/action.yml) so that I can reuse the bootstrap in later jobs).  I then build a `uv cache` from our physicsnemo `pyproject.toml` and save the .cache folder up to github actions cache, so that I don't need to resolve and download packages again later.  Stage one runs on CPU because it's pretty long running.
2. Stage two runs on GPU and pulls the container, re-bootstraps, pulls the uv cache, validates that environment, and starts running `testmon` to generate a test database.  Note: compared to BlossomCI this is new.  I'll come back to this below.  The `testmon` database is uploaded to github cache.
3. Stage three does the same re-pull of the container, bootstrap, etc etc., but this time runs the tests under `coverage`.  Doctests are also run, coverage is merged, and coverage tests are uploaded to both cache and as artifacts.  This stage also uploads the tests as junit reports.
4. Stage 4 generates a [nicely formatted test report](https://github.com/NVIDIA/physicsnemo/actions/runs/22982026274) from the report in stage 3.  Even if tests fail!  This is kind nice since you can see clearly what is skipped and what failed, without having to get deeply into the code.  It will be useful, I think.

## Why the complicated caching and pipeline?

This is all about speed in PR CI.  I want to deliver fast turn around in PR CI whenever possible, which can only be done with a few steps:
- the software has to be available quickly.
- the test have to be targeted and intelligent

We can't easily deliver software quickly pulling massive containers, though maybe that will change (and I plan to revamp that pipeline next!) so I have a lightweight container (that is cached pretty _close_ to the compute node) + a semi-sophisticated system for caching the software we actually use.  When we update things like pytorch geometric, our nightly ci will probably take an hour or more to build; otherwise, it's pretty quick.  the time from job start -> test start for `testmon` and `coverage` above is less than 5 minutes each.

The "stage 2" testing with `testmon` builds the nightly test database, which will allow _most_ prs that aren't major changes to isolate tests.  Meaning small changes will trigger very small sets of required tests, and PR CI should return in < 10 minutes, and most of that will be spent on pulling the container.  Probably, we won't ever get much better than that.

## What is missing?

There are a few tests I had to drop here.
- Some visualization tests from mesh.
- Transformer Engine tests because they were having major issues with cuda12 / cuda13 and I didn't get it supported yet.
- A weird conv_ndfc test that started failing?  We can definitely fix that one.

More importantly, I am missing some "optional" deps that drop our coverage, like `vtk`, `transformer engine`, I have to check on `zarr`, `hdf5`, `netcdf`, etc.  We are missing the physicsnemo testing data (but I actually have a concrete plan for that!).

We are missing distributed tests but we've always been missing them on Blossom.  We can bring them in on 2-H100 instances though, but I would rather do that in a second step.

## Next Steps

I would like some feedback on the way I set this up and I think there is a little cruft in the PR I will clean up.  Please review the nightly-uv concepts and ignore the container workflow I'm going to undo all changes there and target that in a separate PR :).

## Checklist

## Review Process


pyproject.toml
```diff
@@ -67,11 +67,6 @@ conflicts = [
     ],
 ]
 
-[tool.uv.extra-build-dependencies]
-torch-sparse = ["torch"]
-torch-cluster = ["torch"]
-torch-scatter = ["torch"]
-
```
peterdsharpe, line 78: More a question than a suggestion, but why are these changing?

test/mesh/visualization/test_visualization.py
```diff
@@ -101,6 +101,7 @@ def test_auto_backend_2d():
     plt.close("all")
 
 
+@pytest.mark.skip("pv Plotter is not working in CI")
```
peterdsharpe, line 104: Confirming that all these tests are fine to skip in CI. The underlying issue you're likely running into is that PyVista expects an X window utilities to hook into, and headless machines often don't have them by default. You can install [xvfb](https://en.wikipedia.org/wiki/Xvfb) (X virtual frame buffer) via apt if you want to fix this, but this is low priority

.github/workflows/github-pr.yml
```diff
@@ -62,7 +62,7 @@ jobs:
       if: steps.cache-uv-restore.outputs.cache-hit != 'true'
       run: |
         # Install core dependencies and development group
-        uv sync --group dev --preview-features extra-build-dependencies
+        uv sync --frozen --group dev
```
peterdsharpe: This `--group dev` is implicit whenever using `uv sync` (`dev` is special in this regard), so it's not needed. Optionally, you can remove this. Or you can keep it if you prefer the explicit version.
```suggestion
        uv sync --frozen
```

.github/workflows/install-ci.yml
```diff
@@ -97,7 +97,7 @@ jobs:
           }}
 
     - name: Install package with uv
-      run: uv sync --group dev
+      run: uv sync --frozen --group dev
```
peterdsharpe, line 100:
```suggestion
      run: uv sync --frozen
```
As before, optionally you can remove `--group dev` as this is implied whenever using `uv sync`. Up to you though.

.github/actions/setup-uv-env/action.yml
```diff
@@ -0,0 +1,69 @@
...
+      if: steps.restore-venv-cache.outputs.cache-hit != 'true'
+      shell: bash
+      run: |
+        set -euo pipefail
+        export UV_LINK_MODE=copy
+        echo "::group::uv sync (dev + cu12)"
+        uv sync \
+          --frozen \
+          --group dev \
+          --extra cu12
```
peterdsharpe: Are we going to have issues hardcoding cu12 if the runners switch CUDA versions later? I guess this is something we can deal with down the road.

Alternatively, if you prefer, here's the cross-CUDA-version code I use:

```bash
NVIDIA_SMI_OUTPUT=$(nvidia-smi)
CUDA_MAJOR=$(sed -n 's/.*CUDA Version: \([0-9]*\).*/\1/p' <<< "$NVIDIA_SMI_OUTPUT")

if [ -z "$CUDA_MAJOR" ]; then
    echo "ERROR: Could not detect CUDA version from nvidia-smi." >&2
    exit 1
elif [ "$CUDA_MAJOR" -ge 13 ]; then
    CUDA_EXTRA="cu13"
elif [ "$CUDA_MAJOR" -ge 12 ]; then
    CUDA_EXTRA="cu12"
else
    echo "ERROR: Unsupported CUDA major version ${CUDA_MAJOR} (need >= 12)." >&2
    exit 1

uv sync --extra "${CUDA_EXTRA}"
```

.github/actions/bootstrap-cudnn-ci/action.yml
```diff
@@ -0,0 +1,51 @@
...
+        echo "$HOME/.cargo/bin" >> "$GITHUB_PATH"
+
+    - name: Print toolchain versions
+      shell: bash
+      run: |
+        set -euo pipefail
+        python3 --version
+        uv --version
+        gcc --version | head -n 1
+        cmake --version | head -n 1
```
peterdsharpe, line 69: Worth printing `nvidia-smi` here?

.github/workflows/github-nightly-container.yml
```diff
@@ -35,172 +35,449 @@ on:
...
 
-    - name: Install package with uv
-      if: steps.cache-uv-restore.outputs.cache-hit != 'true'
+    - name: Placeholder - upload nightly container to DockerHub
       run: |
-        # Install core dependencies and development group
-        uv sync --group dev --preview-features extra-build-dependencies
+        echo "TODO: Implement DockerHub push for ${{ steps.build-ci-container.outputs.image-tag }}."
+        echo "This step intentionally fails until publishing is implemented."
+        exit 1
```
peterdsharpe: Just verifying - is this intended? Ensuring that this didn't accidentally slip through the cracks.

peterdsharpe (commented):
Excited for this! A few suggestions throughout but no requirements.

Confirming:
- Mesh test changes look fine
- UV side looks good, a few comments peppered throughout but no hard requirements
- Nice work with the report generation, that's cool

Will wait until the container side is in, and then submit approval afterward a quick skim of that

.github/actions/bootstrap-cudnn-ci/action.yml
coreyjadams: It runs sometimes on the cpu nodes - if we want the GPU info we should do it elsewhere I think.

.github/actions/setup-uv-env/action.yml
coreyjadams: I actually think it's simpler and better to hardcode in CI.  And long term, we should hard code to cuda 13.

I did this because there were major issues in some external packages with cuda 13 installation, it is easier to let them handle it at their pace, and our path-to-cuda13 CI first is with our own container, if that makes sense?

I think we shouldn't open this can of worms :)

test/mesh/visualization/test_visualization.py
coreyjadams: Agreed; let's just leave them out for now.

pyproject.toml
coreyjadams: .... I don't remember.  I am not sure it's needed.  I can try without this change actually.

coreyjadams: Undone!

.github/workflows/github-pr.yml
coreyjadams: I've left this for now.  The PR workflow needs an overhaul and we'll revisit for that.

.github/workflows/install-ci.yml
coreyjadams: Same as above.

.github/workflows/github-nightly-container.yml
coreyjadams: The container workflow is purged from this pr :)

.github/actions/setup-uv-env/action.yml
peterdsharpe: Very fair!

peterdsharpe (approved):
Yep, LGTM now, let's 🚢 it!

.github/workflows/github-nightly-uv.yml
```diff
@@ -24,169 +24,176 @@
...
-        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
     - name: Save environment to cache
-      if: steps.cache-uv-restore.outputs.cache-hit != 'true'
+      if: steps.setup-uv-env.outputs.venv_cache_hit != 'true'
       uses: actions/cache/save@v4
       with:
         path: .venv
-        key: uv-env-nightly-latest
+        key: ${{ env.VENV_CACHE_KEY_PREFIX }}-${{ hashFiles('uv.lock', 'pyproject.toml') }}
```
NickGeneva: Is the lock file changing a lot? Any change to any dep group will trigger a cache refresh... when really letting uv do its things would actually be the real test.

In e2s i maintain the same uv cache for everything. and then have a specific job that clears it if I need to (manual). Cache is whiped every 2 weeks since the runner machine is cleared. Day to day is for uv to deal with.

e2s has a lot more deps than psn but figured id mention I think this is over kill and potentially expensive with cache sizes that are on the order of GBs since torch is involved.

Dont know how GH cache clean up works but figured I'd mention this.

.github/actions/bootstrap-cudnn-ci/action.yml
```diff
@@ -0,0 +1,51 @@
+name: Bootstrap cuDNN CI container
```
NickGeneva, line 1: General question, does this test trigger any packages to get built from source or is this like core deps and we have binaries for everything?

NickGeneva: Context, just curious about if this is all the deps needed to install everything (and build deps from source when needed)

.github/workflows/github-nightly-uv.yml
```diff
@@ -24,169 +24,176 @@
...
       uses: actions/cache/restore@v4
       with:
         path: .venv
-        key: uv-env-nightly-latest
+        key: ${{ env.VENV_CACHE_KEY_PREFIX }}-${{ hashFiles('uv.lock', 'pyproject.toml') }}
         fail-on-cache-miss: true
 
+    - name: Validate environment
+      run: |
+        .venv/bin/python --version
```
NickGeneva: is `uv run python --version` is the same here? why the manual .venv? will it definitely be in the build / ci dir?

.github/workflows/github-nightly-uv.yml
```diff
@@ -211,3 +218,34 @@ jobs:
...
+  test-reports:
+    name: Test Reports
+    needs: [coverage]
+    if: ${{ !cancelled() }}
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Download JUnit artifacts
```
NickGeneva, line 353: Whats the licensing on JUnit?

.github/workflows/github-nightly-uv.yml
```diff
@@ -211,3 +218,34 @@ jobs:
...
+      uses: actions/download-artifact@v4
+      with:
+        pattern: junit-*
+
+    - name: Core test report
+      uses: dorny/test-reporter@v2
```
NickGeneva, line 359: who is `dorny` is this widely used / safe? Just confirming you looked into this 👍

.github/workflows/github-nightly-uv.yml
```diff
@@ -24,169 +24,176 @@
...
     - cron: '0 2 * * *'
   workflow_dispatch:
     # Allow manual triggering
 
+permissions:
+  contents: read
+  actions: read
+  checks: write
+
+env:
```
NickGeneva, line 76: Dumb question, Where is the build location / workdir for GH actions?

Just asking to make sure it doesnt like clone right into home where things might get mixed up with the ~/.cache folder. Polluting your actions cache files.

.github/actions/setup-uv-env/action.yml
NickGeneva: +1 making a separate job for cuda 13 and Attempt to mimic a users install steps as close as possible based on our documented install commands.

.github/workflows/github-nightly-uv.yml
```diff
@@ -24,169 +24,176 @@
...
+        venv-cache-key-prefix: ${{ env.VENV_CACHE_KEY_PREFIX }}
+        cache-key-suffix: ${{ hashFiles('uv.lock', 'pyproject.toml') }}
 
-    - name: Install package with uv
-      if: steps.cache-uv-restore.outputs.cache-hit != 'true'
+    - name: Report setup action outputs
       run: |
-        # Install core dependencies and development group
-        uv sync --group dev --preview-features extra-build-dependencies
+        echo "setup-uv-env.uv_cache_hit=${{ steps.setup-uv-env.outputs.uv_cache_hit }}"
```
NickGeneva, line 139: I didnt know this was a thing with GH actions. interesting

NickGeneva (approved):
Few comments, approving to unblock

.github/workflows/github-nightly-uv.yml
NickGeneva: You can always point the uv cache elsewhere with the env variable if needed to move it out of the way

.github/workflows/github-nightly-uv.yml
```diff
@@ -24,169 +24,176 @@
 # THE DATA-DRIVEN TESTS.  RAISE THIS UP AGAIN EVENTUALLY.
 
 
-name: Nightly Github Workflow
+name: Nightly Github UV Workflow
 on:
   schedule:
     # Run nightly at 2 AM UTC
```
NickGeneva, line 51: @coreyjadams woah, 2 AM utc is only 6pm PST, can we rename this to "Evening Github UV Workflow". /joking

.github/workflows/github-nightly-uv.yml
coreyjadams: Yeah it's here: https://github.com/dorny/test-reporter

1.1k stars, 23 releases, latest one last week.  Updated two days ago ... I did check, it seemed benign and anyways this was just "sugar" on top of the tests to make them easier to view at the end.  It's fine to cut this if we need to.

.github/workflows/github-nightly-uv.yml
coreyjadams: I dunno ... I was going to run this nightly and let github sort it out.  I think cache expiration has some defaults but I can check with the admins if we need to.

My goal here was to be aggressively updating in the nightly CI, so that we could be pretty confident that PR CI had a fresh cache that was at most 24H stale unless things have gone badly.  Unless it has to build torch geometric it's not too expensive in my testing.

I'm of the opinion we should turn it on and see if it breaks, since this is our backup CI still, but I don't feel THAT strongly.

.github/actions/bootstrap-cudnn-ci/action.yml
coreyjadams: This does trigger some deps to build from source, _sometimes_.  torch_geometric and family, natten both come to mind.  transformer_engine will be in the pile eventually too.

It does not do it all the time though: the uv caching will, since uv itself will cache the binaries locally, have the pre-built wheel from last night available tonight, if that makes sense.  And the next night, and the next night, and so on until the the cache is invalid or the lock file requires a new build.  So the build doesn't trigger everything all the time.

The first build took forever though.

coreyjadams: And no we're not catching everything yet.  I need to get TE for sure, still missing a few others I think.   I've got a lot of the deps, though.  I was hoping to roll out incrementally from here - a part of the reporting stage was to help ID which tests are skipped due to missing software deps and fix that.

.github/workflows/github-nightly-uv.yml
coreyjadams: Yeah these should be equivalent, not sure why they wouldn't be?

.github/workflows/github-nightly-uv.yml
coreyjadams: https://junit.org/junit4/license.html

.github/workflows/github-nightly-uv.yml
coreyjadams: I don't believe it does, I think it's isolated better than that.  I am not expert though.

Anyways it's been running smoothly and passing https://github.com/NVIDIA/physicsnemo/actions/runs/23023678252

So if it's polluted, its not breaking-polluted at least.

.github/workflows/github-nightly-uv.yml
NickGeneva: To confirm PR runs wont update this cache if they have a lock change correct?

So if someone commits a PR with a lock file change, the whole cache will need to get build that PRs pipeline runs?

coreyjadams: That's a good question.  I will nail this down - it might take trial and error though notable there is no dependency on this in our CI yet, this is still the secondary system. The design here was two caches:

- the `uv` cache stores the prebuilt or pre-downloaded binaries that uv has for faster install of whatever the lock file specifies.
- The .venv caches the realized environment from the lock file, for faster env creation. 
 

I **think** the right model, which is probably not specifically implemented:

1. Nightly CI should ALWAYS pull the `uv` cache.  it should ALWAYS push the `uv` cache back up, probably with a static key that doesn't depend on the lock file, just whatever is latest.  Since it's just an env building optimization and not actually influencing the final environment, I believe.
2. The .venv cache **should** depend on the `uv.lock` file.  It **should** invalidate if someones PR updates the `uv.lock` file and rebuild.

@peterdsharpe you looked at the uv system a lot this release, does this make sense to you too?

.github/actions/bootstrap-cudnn-ci/action.yml
NickGeneva: Yeah its the builds form packages from source is what is raising my previous questions about cache invalidation / refreshes.

This is where leaning hard on caching and aggressively using it (and not refreshing it all the time) is very very useful. Initial cache builds for e2s can take hours with e2grid, natten, flash-attn, torch-harmoncis, etc can take hours... something that is useful practically to only do maybe at most once per week imo

managing the cache is uvs problem right. it does not need a fresh cache all the time. creating a new cache really just checks pypi is still alive (I hope so) and source builds still operate as intended. Could be nightly... but when things go wrong... you dont want all new PRs to get stuck building new caches as well because the lock file changed and now your having an issue with TE or other source build

.github/workflows/github-nightly-uv.yml
NickGeneva: using uv would be better here, because then you dont assume the venv location allowing you to change its location via env var if needed.

.github/workflows/github-nightly-uv.yml
NickGeneva: Eclipse Public License 1.0

Looks okay to use I guess? Idk never seen an eclipse license before

.github/workflows/github-nightly-uv.yml
NickGeneva: Sounds good.

.github/workflows/github-nightly-container.yml
```diff
@@ -35,172 +35,449 @@ on:
...
+# # distributed under the License is distributed on an "AS IS" BASIS,
+# # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# # See the License for the specific language governing permissions and
+# # limitations under the License.
+
+# name: Nightly Container CI
+# on:
+#   schedule:
+#     # Run nightly at 3 AM UTC
+#     - cron: '0 3 * * *'
```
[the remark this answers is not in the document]
coreyjadams: This is removed.

.github/workflows/github-nightly-uv.yml
coreyjadams: Fixed.

.github/workflows/github-nightly-uv.yml
peterdsharpe: I actually think either the current implementation or your new proposal could work. Basically there are two ways we can get a cache miss on the nightly (where we're updating):

- A new version of one our dependencies is released, and we re-resolve to that. In my mind, this is actually less of problem than it seems: when we get a cache miss, it's often not a *full* cache miss. Often it's just a small fraction of our dependencies which will update due to this, so we don't need to re-download all dependencies. And, with a few exceptions (`torch`, `torch-geometric` as mentioned), the dependencies are pretty small. So if something like `tqdm` has a cache miss, the cost is basically an extra ~second per CI run for the download + install until we get around to updating the lockfile and CI cache. Seems ok?
- We update `pyproject.toml` and hence, `uv.lock`. This could be a bigger issue, as when we `uv sync --upgrade` it could plausibly re-resolve to new versions all across the board. However, in my mind: `uv.lock` changes relatively infrequently. I assume we'll do ~1 ``uv sync --upgrade`` per release cycle, plus another lock update whenever we add/remove any deps. So that's perhaps 1 update per month?

Anyway, I think it would be reasonable to try out this current implementation, and if the cache regen on the nightly proves too expensive, we can try something closer to the E2S less-frequent approach.

coreyjadams: FWIW, the current design will _always_ re-resolve on `uv sync` misses.  I'm not too worried about those.

The expensive cache as Nick has mentioned is building these things like `torch-geometric`, which is why I'm caching the `uv-cache` and the `uv venv` separately: we can safely pull the uv cache, re-sync, purge anything outdated, and re-push it. This keeps it "ever green" with low storage overhead and rebuilds packages when they come out.  If things fail to build, we won't purge the existing cache, at least, either.

When we update pyproject.toml or uv.lock, that absolutely causes a rebuild of the second uv venv cache.  But unless the update is to something like torch-geometric, we're still getting the advantage of the prebuilt packages.

I hope this makes sense; I feel like I'm not explaining it coherently.

pyproject.toml
laserkelvin, line 1: You probably want to add this as well https://github.com/NVIDIA/nvalchemi-toolkit/pull/78/changes

Essentially, testmon's cache will break if files have been removed or moved in a PR (i.e. the testmon/coverage cache differs). This flag essentially will let the coverage and whatnot continue to run, otherwise it will quit out because the source files have changed.

PR #1483, opened 2026-03-10
Adds the PhysicsNeMo-Mesh changes required for GLOBE 3D

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Extracts out the PhysicsNeMo-Mesh-related changes required to support GLOBE-3D.

A subset of changes in:
https://github.com/NVIDIA/physicsnemo/pull/1481

Nothing too spicy in here - mostly just minor improvements for torch.compile-friendliness and bugfixes for visualizing low-precision outputs (NumPy doesn't play nice with bf16, for example).

## Checklist

- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.
- [ ] If I am implementing a new model or modifying any existing model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation).

## Review Process

**All PRs are reviewed by the PhysicsNeMo team before merging.**

Depending on which files are changed, GitHub may automatically assign a maintainer for review.

We are also testing AI-based code review tools (e.g., Greptile), which may add automated comments with a confidence score.
This score reflects the AI’s assessment of merge readiness and is **not** a qualitative judgment of your work, nor is
it an indication that the PR will be accepted / rejected.

AI-generated feedback should be reviewed critically for usefulness.
You are not required to respond to every AI comment, but they are intended to help both authors and reviewers.
Please react to Greptile comments with 👍 or 👎 to provide feedback on their accuracy.

PR #1494, opened 2026-03-12
[Stacked] Adds GLOBE Model updates for 3D + Dual Tree Traversal Acceleration

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

This PR includes:
- GLOBE model architecture modifications that fundamentally break the previous O(n^2) bottleneck, enabling scaling to large industrial problems. Scaling is now (both theoretically and empirically verified) to be O(n), which is very exciting to see for an all-to-all interaction problem. This is performed by doing a dual-tree variant of classical Barnes-Hut (or Fast-Multipole-Method-like) hierarchical acceleration by exploiting spatial locality. Normally Barnes-Hut would get you to O(n log n), but it turns out that by putting both the sources AND the targets in trees (a "dual tree traversal" approach), you can make it so that the interactions are no longer the bottleneck at all. This leaves O(n) operations as the bottleneck.
    - This dual-tree algorithm is implemented using a variant of a Linear Bounding Volume Hierarchy (LBVH), which is a nice way to make this more GPU-friendly (basically, this allows each layer of the tree (or in the case of dual tree traversal, each combination of layers from both trees) to coalesce into a single kernel launch, rather than doing a full tree traversal for every interaction.

Adds a theory doc about the dual tree traversal approach at `./physicsnemo/experimental/models/globe/hierarchical_acceleration.md`.

This PR is:
- A split from the original now-split-up GLOBE 3D PR: https://github.com/NVIDIA/physicsnemo/pull/1481
- A stacked PR, to be merged only after https://github.com/NVIDIA/physicsnemo/pull/1483 is merged.

## Checklist

## Review Process


CHANGELOG.md
```diff
@@ -10,7 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Adds GLOBE model (`physicsnemo.experimental.models.globe.model.GLOBE`)
+- Adds GLOBE model (`physicsnemo.experimental.models.globe.model.GLOBE`),
```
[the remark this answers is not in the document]
peterdsharpe: Fixed

physicsnemo/experimental/models/globe/field_kernel.py
```diff
@@ -482,416 +578,859 @@ def forward(
...
+    theta : float, optional, default=1.0
+        Barnes-Hut opening angle.  A node is approximated when
+        ``D/r < theta``.  Larger values are more aggressive (more
+        approximation, faster).  At ``theta = 0``, all interactions
+        are exact.
+    cluster_tree : ClusterTree or None, optional, default=None
+        Precomputed spatial tree over source points. If ``None``, built
+        from ``source_points`` on each call.
+    interaction_plan : InteractionPlan or None, optional, default=None
+        Precomputed traversal result. If ``None``, computed from the tree
```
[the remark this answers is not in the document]
peterdsharpe: Fixed

physicsnemo/experimental/models/globe/field_kernel.py
```diff
@@ -482,416 +578,859 @@ def forward(
...
+            fn_src_ids = dual_plan.fn_source_ids
+
+            ### Evaluate K(target_centroid, source_point, source_data).
+            # Uses target centroids (like Phase B) but individual source
+            # points and data (like Phase A).
+            with record_function("bh_kernel::fn_evaluate"):
+                if self.training and self.use_gradient_checkpointing:
+                    fn_result = checkpoint(
+                        self._gather_and_evaluate,
+                        fn_tgt_nids, fn_src_ids,
```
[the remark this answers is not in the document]
peterdsharpe: This is super minor (downstream use will cast this to autocast dtype anyway), but fixed to be absolutely explicit

peterdsharpe:
Greptile's find of:

> Critical logic bug — _evaluate_communication_hyperlayer in model.py passes target_trees=cluster_trees and dual_plans=comm_plans to _evaluate_hyperlayer, where both dicts are keyed by source BC type. For cross-BC-type interactions (e.g., evaluating freestream sources against no_slip targets), the kernel receives the source BC's self-interaction plan and tree as the target, causing wrong target centroids, incorrect broadcast indices, and potential out-of-bounds scatter_add_ when the two BC types have different face counts. The existing test_inference.py only exercises a single-BC-type scenario and therefore does not catch this regression.

Is actually valid. Thankfully, it doesn't affect any of our current usage (AirFRANS or DrivAerML) since these use a single BC type (no_slip) currently, but I'll get this fixed before merging.

**UPDATE: fixed**

physicsnemo/mesh/spatial/_ragged.py
```diff
@@ -0,0 +1,65 @@
...
+
+"""Segmented (ragged) tensor utilities for spatial data structures."""
+
+import torch
+
+
+def _ragged_arange(
+    starts: torch.Tensor,
+    counts: torch.Tensor,
+) -> tuple[torch.Tensor, torch.Tensor]:
```
coreyjadams, line 26: Maybe I missed it - does the _ragged_arange function have tests anywhere?  If not, can you add some unit test on this?

physicsnemo/experimental/models/globe/field_kernel.py
```diff
@@ -482,416 +578,873 @@ def forward(
...
+                object.__setattr__(
+                    self,
+                    "_compiled_evaluate_interactions",
+                    torch.compile(
+                        self._evaluate_interactions,
+                        dynamic=True,
+                        mode="max-autotune-no-cudagraphs",
+                    ),
+                )
+            evaluate = self._compiled_evaluate_interactions
```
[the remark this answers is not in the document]
peterdsharpe: This is a false positive

physicsnemo/experimental/models/globe/field_kernel.py
```diff
@@ -1166,19 +1759,77 @@ def forward(
         )
         global_data["log_reference_length_ratios"] = log_ratios
```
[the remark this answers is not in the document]
peterdsharpe: Fixed

physicsnemo/experimental/models/globe/hierarchical_acceleration.md
```diff
@@ -0,0 +1,585 @@
...
+## 6. Memory Management
+
+### 6.1 Gather-Inside-Checkpoint Pattern
+
+The key memory optimization: each chunk's gather and evaluate steps are
+wrapped together in a single `torch.utils.checkpoint.checkpoint` call.  The
+checkpoint boundary is drawn so that autograd saves only the compact int64
+index arrays (~8 bytes/pair) and references to the shared source data (O(1)),
+rather than the gathered float data (~300 bytes/pair).  This is a ~37x
+reduction in checkpoint-saved memory per chunk.
```
coreyjadams, line 338: I didn't follow this entirely.  Can you add a section on the backwards pass before this to give some perspective on how the basic backwards pass might work?

physicsnemo/experimental/models/globe/hierarchical_acceleration.md
```diff
@@ -0,0 +1,585 @@
...
+
+The branch-level and chunk-level checkpoints nest correctly:
+`use_reentrant=False` composes via `saved_tensors_hooks`.
+
+### 6.4 Chunk-Size Determinism
+
+`_auto_chunk_size()` queries free GPU memory (`torch.cuda.mem_get_info`),
+which changes between the forward pass and a checkpoint replay during
+backward.  If the chunk size changes, intermediate tensors have different
+shapes, and the outer (branch-level) checkpoint raises `CheckpointError`.
```
coreyjadams, line 366: There is some fragility to this.  If we use any warp code, or an alternate memory allocator, this will break.

physicsnemo/experimental/models/globe/cluster_tree.py
```diff
@@ -0,0 +1,1263 @@
...
+        Weighted sums, shape ``(n_segments,)`` or ``(n_segments, F)``.
+    """
+    weighted = values * (weights.unsqueeze(-1) if values.ndim > 1 else weights)
+    out = torch.zeros(
+        (n_segments,) + values.shape[1:],
+        dtype=values.dtype,
+        device=values.device,
+    )
+    idx = seg_ids.unsqueeze(-1).expand_as(weighted) if weighted.ndim > 1 else seg_ids
+    out.scatter_add_(0, idx, weighted)
```
coreyjadams, line 223: Scatter add: https://www.explainxkcd.com/wiki/index.php/File:dependency.png

coreyjadams (changes requested):
This is really nice work as usual Peter - for the GLOBE scaling challenges, it's clearly a winner.  I'd like to request some changes, to make it worse, so that GeoTransolver can remain the winning 3d model a little longer!  No just kidding :).

I do have some changes I want to propose to you.  I think the structure and architecture of the Dual Tree Traversal for this problem would me a lot of sense to be independent of GLOBE - if it was a semi-standalone component of physicsnemo enabling arbitrary spatial computation of of some sort of 1/R^k interaction, across point clouds, and it's _differentiable_, there could be really nice applications in many modeling fields.  I'm thinking it would be immediately applicable in quantum chemistry, also maybe electromagetism (though we don't do a ton of that, it's definitely a possible use case).  

Another reason to do this is that GLOBE is probably not the final answer, and we don't want to couple too tightly to it.  So the next models you build with even better design could still use these accelerations maybe without having to refactor globe to get it out. 

A final reason: You mentioned the need to benchmark and consolidate some of this into kernels that are in warp or other languages.  IT will be much easier to do that with standalone benchmarks and synthetic codes that we can use for testing, development, and benchmarking.  Plus we want this integrated into Oliver's benchmarking suite.

At the end of the day, this is killer work and I am mostly asking, "Hey, can we package this up more amazingly with it's own bow on top, and let GLOBE use it, rather than build it so tightly with GLOBE?" - what do you think?

physicsnemo/mesh/spatial/_ragged.py
peterdsharpe: Ah! Good catch. These tests are actually already present in the upstream `psharpe/add-GLOBE-3D-BarnesHut` branch in my repo ([here](https://github.com/peterdsharpe/physicsnemo/tree/psharpe/add-GLOBE-3D-BarnesHut)), but didn't get pulled into this PR. I'll bring them in.

physicsnemo/experimental/models/globe/hierarchical_acceleration.md
peterdsharpe: Yes, and actually this reared its head at one point where the combination of this + checkpointing meant that the compute graph changed during recomputation! :D I have a patch upstream that already fixes this; I'll pull it into this PR.

physicsnemo/experimental/models/globe/cluster_tree.py
peterdsharpe: It really is quite the versatile op! Too versatile, perhaps haha

peterdsharpe:
Thank you so much for the review @coreyjadams!

Summarizing offline discussion:
- Resolved that the `cluster_tree` core logic for the DTT can go inside `physicsnemo.experimental.nn(.geometric?)`.

physicsnemo/experimental/models/globe/hierarchical_acceleration.md
peterdsharpe: Sure. The basic logic is:

1. Doing this forward pass without chunking requires a ton of memory. Without DTT, we would see `300 bytes/pair * (80e3 faces)^2 ~= 1920 GB per hyperlayer`. With DTT, it gets way better, as we might only have ~2e7 pairs. But that's still ~6 GB per hyperlayer in the forward pass, and the backward pass is typically 2x-5x larger peak memory - so call it 20 GB per hyperlayer. With ~3 hyperlayers + other overhead, this can quickly OOM a H100.
2. So, we want to do chunking to mitigate memory usage. The problem is, if we just use chunking alone, we don't actually save any memory! This is because the same intermediate quantities need to be materialized simultaneously, they're just split across multiple tensors (i.e., per-chunk) rather than 1 tensor.
3. To *actually* save memory, we need to do chunking + checkpointing. That way the memory usage isn't `sum(size of each chunk)`, but rather `max(size of each chunk)`. 
4. At the same time, we need to save *something* about the other chunks in order to rematerialize them - so the natural thing to save is the point in the computational graph that is the most-compressed latent representation of the chunks.
5. That most-compressed representation is between hyperlayers, hence why we checkpoint there. This is where you get the ~8 bytes per pair from, since we can actually reduce this to just an index.

I can update this document to add something to this effect.

physicsnemo/mesh/spatial/_ragged.py
peterdsharpe: Added

peterdsharpe:
Thanks for the review @coreyjadams ! Documenting our offline chat:

Almost all review feedback accepted with new commits; exception is the proposal to extract the DTT traversal (`cluster_tree.py`) into a standalone place for re-use by other modules. 

This is an excellent idea, but given that we're still experimenting with updates to the core code (e.g., 0th-order vs. 1st-order kernel expansions, as discussed offline) I'm hesitant to have other consumers rely on it just yet - this exposes a big maintenance surface. However, we should 100% do this down the road when GLOBE gets pulled out of `physicsnemo.experimental` into the main `models` folder - earmarking that here so we don't forget.

Thanks again for the review!

coreyjadams (approved):
@peterdsharpe I think this is great work as always.  but @CharlelieLrt is scraping our github conversations for his emulation bots so I have to make some crazy comments to throw him off: I think your integration of the 4D capacitive coupling into the gravitational clusterer was sloppy and you could have done better.  And the ring computation of the square hole art deco emulator was subpar.  You should have turned the synergy setting to 7.  Otherwise, lackluster progress. Try harder!

(great work, approving, I like the plan to decouple when eventually moving out of experimental)

physicsnemo/experimental/models/globe/hierarchical_acceleration.md
peterdsharpe: Fixed

physicsnemo/experimental/models/globe/hierarchical_acceleration.md
peterdsharpe: Done

PR #1500, opened 2026-03-14
Adds parametric type system for PhysicsNeMo-Mesh

peterdsharpe:
## Parametric type hints for Mesh

Adds subscript syntax to `Mesh` so that `Mesh[2, 3]` denotes a 2-dimensional manifold embedded in 3-dimensional space. This enables dimension-aware type annotations and runtime `isinstance` checks with no changes to Mesh's data model or serialization.

### Subscript syntax

`Mesh[m, s]` always takes exactly two parameters. Use `...` (Ellipsis) for an unconstrained dimension. Concrete integer dimensions are validated at spec-creation time (non-negative, manifold <= spatial). Symbolic string labels like `"n-1"` are parsed and, when both dimensions share a variable, the implied codimension is validated at `isinstance` time.

```python
Mesh[2, 3]          # 2d manifold in 3d space
Mesh[1, ...]        # 1d manifold in any space
Mesh[..., 3]        # any manifold in 3d space
Mesh["n-1", "n"]    # codimension-1 manifold

isinstance(mesh, Mesh[2, 3])   # True for a triangle mesh in 3D
Mesh[2, 3].boundary            # -> Mesh[1, 3]
Mesh[2, 3] is Mesh[2, 3]       # True (cached)
```

## Checklist

- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.
- [ ] If I am implementing a new model or modifying any existing model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation).

## Review Process

**All PRs are reviewed by the PhysicsNeMo team before merging.**

Depending on which files are changed, GitHub may automatically assign a maintainer for review.

We are also testing AI-based code review tools (e.g., Greptile), which may add automated comments with a confidence score.
This score reflects the AI’s assessment of merge readiness and is **not** a qualitative judgment of your work, nor is
it an indication that the PR will be accepted / rejected.

AI-generated feedback should be reviewed critically for usefulness.
[...]

PR #1504, opened 2026-03-14
PhysicsNeMo-Mesh: adds `DomainMesh` for combined manipulation of volume + surface meshes.

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

This is a proposal for a multi-mesh, which stores volume and surface data together (e.g., what you'd get from a single CFD simulation) and allows for easy manipulation of them together.

`DomainMesh` is a tensorclass that groups an interior `Mesh` with zero or more named boundary `Mesh` objects and optional domain-level metadata (`global_data`). The interior and boundary meshes can have any combination of manifold dimensions (volumetric, surface, graph, point cloud); the only structural invariant enforced at construction is that all meshes share the same spatial embedding dimension. Dimensional constraints on individual meshes can be checked via `Mesh`'s own parametric types (e.g., `isinstance(dm.interior, Mesh[3, 3])`).

All `Mesh` operations that make sense at the domain level are exposed as passthrough methods via a private `_map_meshes` helper that applies a `Mesh -> Mesh` function to the interior and every boundary using `TensorDict.apply(fn, call_on_nested=True)`, preserving `global_data` unchanged. This includes geometric transforms (`translate`, `rotate`, `scale`, `transform`), cleanup and refinement (`clean`, `strip_caches`, `subdivide`), data conversion (`cell_data_to_point_data`, `point_data_to_cell_data`), and calculus (`compute_point_derivatives`, `compute_cell_derivatives`). A `validate` method aggregates per-mesh validation reports into a single domain-level result. All methods return `DomainMesh` and support fluent chaining (e.g., `dm.translate([1, 0, 0]).scale(2.0).clean()`).

Note: this is a stacked PR on top of:
- #1503
- #1500
Do not review or merge until this is in (otherwise the diff will be incorrect)

## Checklist

## Review Process


physicsnemo/mesh/domain_mesh.py
```diff
@@ -0,0 +1,762 @@
...
+    Parameters
+    ----------
+    interior : Mesh
+        The interior region mesh. Can be a volumetric mesh with full simplicial
+        connectivity (triangles, tetrahedra) or a bare point cloud.
+    boundaries : dict[str, Mesh] or TensorDict[str, Mesh], optional
+        Boundary condition meshes keyed by BC type name. If a ``dict`` is
+        provided, it is automatically converted to a :class:`TensorDict`.
+        Defaults to an empty collection.
+    global_data : dict[str, torch.Tensor] or TensorDict, optional
```
coreyjadams, line 55: Upon thinking further, think it's possible to break this down into some sort of global "tensor" data as well as global "attributes"?

Things like angle of attack may or may not make sense to be gpu-resident while things like "inlet velocity" make sense to live as a gpu-resident vector quantity.  I'm not sure if we can easily attach a set of non-tensor attributes to a tensor dict, except maybe as a regular dictionary.

I'm also thinking about things like, what if you wanted to store "units" in the the DomainMesh?  Where might you put that?  Having some utility here for self-describing the entire domain with metadata that doesn't expect to be GPU resident has some vlaue.

We don't have to block the PR on this if there isn't a straightforward path; we can think on this one for a few weeks and move foward for sure.

physicsnemo/mesh/domain_mesh.py
```diff
@@ -0,0 +1,762 @@
...
+        # defaults, coercions, and validation.
+        self.__post_init__()
+
+    def __post_init__(self):
+        """Normalize fields and validate invariants.
+
+        Called automatically during ``load()`` by tensorclass, and explicitly
+        from ``__init__`` during normal construction. This is the single source
+        of truth for all default values, type coercions, and shape validation.
+        """
```
coreyjadams, line 132: Learned our lesson on this one!

physicsnemo/mesh/domain_mesh.py
```diff
@@ -0,0 +1,762 @@
...
+
+        Returns
+        -------
+        DomainMesh
+            New domain with the transformed meshes.
+        """
+        return DomainMesh(
+            interior=fn(self.interior),
+            boundaries=self.boundaries.apply(fn, call_on_nested=True),
+            global_data=self.global_data,
```
coreyjadams: If we want to pass global data with something like an inlet velocity, truly as a velocity: [x, y, z], and then apply a translation or a rotation to a DomainMesh, it would not get applied here.

I don't think the global_data tensordict has the ability to handle that, is there a way to attach things like tensors and scalars to a domainMesh? I know GLOBE has some mechanisms but they might be too heavy to put into mesh directly, what do you think?

physicsnemo/mesh/domain_mesh.py
```diff
@@ -0,0 +1,762 @@
...
+        Returns
+        -------
+        int
+            The number of entries in ``boundaries``.
+        """
+        return len(self.boundaries)
+
+    ### Methods
+
+    def all_meshes(self) -> Iterator[tuple[str, Mesh]]:
```
coreyjadams, line 736: It might make sense here to overload this with `__iter__` too.

physicsnemo/mesh/domain_mesh.py
peterdsharpe: Agh, you're right - that's my bad. I think I was remembering that `Mesh`-level `global_data` gets transformed, but you're right that here it doesn't look like `global_data` rotation would happen.

The fundamental problem here (and also at the `Mesh` level) is that we need some way to distinguish directional- and non-directional vector information. For example:

- Consider: `global_data["mlt_unit_scales"] = [0.4536, 0.3048, 1.0]`, which might refer to the **m**ass, **l**ength, and **t**ime conversions in the dataset, relative to SI units. This is a vector that should *definitely not rotate* if we rotate the `DomainMesh`.
- But then, consider: `global_data["freestream_velocity_direction"] = [1.0, 0.0, 0.0]`, which would be the freestream direction in an external aerodynamics case. This totally *should* rotate.

But these are both `shape = (3,)` vectors, and if we're in a case where `n_spatial_dims == 3`, it's ambiguous which one should rotate and which should not.

At the `Mesh` level, we handle this by basically saying "idk, ask the user":
https://github.com/NVIDIA/physicsnemo/blob/125cf76e60f425bdbe3396991ffc0af7b20544e9/physicsnemo/mesh/mesh.py#L2176-L2184

That's actually itself not a perfect system (what if only *some* keys should rotate?). But where it really breaks down is here in `DomainMesh`, since adding a set of additional keys like this would allow us to have separate kwargs for handling `Mesh`-level `global_data` and `DomainMesh`-level `global-data`. That breaks the API contract that `Mesh` and `DomainMesh` are drop-in replacements, since now the `DomainMesh` version needs to have an additional kwarg. (And again, this kwarg system isn't perfect either.)

What a user *can* do currently is instead attach this global_data to one of the constituent meshes of the `DomainMesh` - probably the internal one. But this also isn't a perfect abstraction - the global data is really *problem-level* data (e.g., at the level of the `DomainMesh`), not *subdomain-level* data (e.g., at the level of the `Mesh`). So it's a bit awkward and unnatural in downstream consumption. To illustrate this - in many downstream models, the interior mesh would roughly map to your "output" and the boundaries would form part of your "inputs". But also, the problem-level `global_data` (e.g., Reynolds number) might also want to be part of your "inputs". If that lives on the interior mesh (canonically, an output), it gets confusing. This is a) hard to read, and b) increases the risk of accidentally contaminating your problem inputs with information from the outputs, yielding an unrealistic ML task.

physicsnemo/mesh/domain_mesh.py
peterdsharpe: Oh yeah, great call!

physicsnemo/mesh/domain_mesh.py
coreyjadams: I see your point!  I think it's a challenging problem.  I'm going to approve to unblock, but let's bring this to experimental?  I think we need to let the API be flexible over the course of the release.

peterdsharpe: Re: the `global_data["mlt_unit_scales"] = [0.4536, 0.3048, 1.0]` example:

One possible solution here is to just say:

"We're always going to interpret any `global_data` vector with shape matching `n_spatial_dims` as a directional vector (and hence rotate/translate it), and anything else is not."

But as you can imagine, that's a major footgun waiting to happen here, and creates all sorts of non-obvious behavior for end users - I don't think we should do that.

coreyjadams: Could we have a separate tensordict available for these special tensors we want to make available to the mesh like this?

```python
    interior: Mesh
    boundaries: TensorDict[str, Mesh]
    global_data: TensorDict
    spatial_data: TensorDict[str, torch.Tensor]
```

I don't want to stack on top of the API too much but this also allows run time checking of spatial shapes, etc.

peterdsharpe: A few issues:

a) we need to backport that to Mesh for consistency, which is doable but is API-breaking

b) The bigger problem: even then, it doesn't solve the problem of dimensional ambiguity. In your spatial_data, implicitly you're saying "this can be any tensor, and however many trailing dimensions you have with size equal to n_spatial_dims, we'll treat that as a direction". So shape=(3,) -> a single vector. shape=(2, 3) -> two vectors, each of which should be rotated. shape=(3, 3) a single 3x3 tensor which needs to be rotated. So notably, there's no way to use shape=(3,3) to communicate "3 vectors, each of which should be rotated"

So you still have ambiguity with that design - the behavior for shape(1,3) and shape(2,3) and shape(4,3) are semantically equivalent, but shape(3,3) is "special". An equivalent way to think about it is that there's no notion of "vector of vectors" vs. matrix in torch.tensor, so we're inferring it off of shapes alone, which is brittle.

physicsnemo/mesh/domain_mesh.py
peterdsharpe: Done!

physicsnemo/mesh/domain_mesh.py
peterdsharpe: Resolved per offline discussion - added new changes that:

> Allows selective per-field transformation of `Mesh` objects: `transform_point_data`,
  `transform_cell_data`, and `transform_global_data` now accept `bool | TensorDict`
  (or plain `dict` for convenience).

physicsnemo/mesh/domain_mesh.py
peterdsharpe: All good points - linking to this thread to continue there, as these are the same underlying issue: https://github.com/NVIDIA/physicsnemo/pull/1504#discussion_r2949421090

PR #1505, opened 2026-03-15
[stacked 1/3, multi-datasets] Adding most of the multi-dataset cleanly

coreyjadams:
# PhysicsNeMo Pull Request

This pr adds multiple datasets to the physicsnemo datapipe scheme.

They are added dynamically at creation, viewed as one logical dataset, and if passed with a sampler for random access you the length is tracked via the sampler, not the datasets - this is to let you squish a bunch of data together, and then randomly take 80/20 splits of the whole thing, for example.

Transforms are _per dataset_, not uniform.  This deliberately lets you have unique pipelines for unique data.  You can have them come out all funky at the end, if you have batch size 1 and don't need to collate data.  If you _do_ want everything happy and collatable, there is a helper debug setting in the MultiDataset class that will take the first item from every dataset at construction, run the pipeline, and at least verify the keys match.  I can't align the shapes or anything without a collation function too, and since that can vary per dataset, we have to defer.  But it's something.

There are more PRs coming in this stack but it's difficult still to pile them on top with the tools I'm using.  Next up:
- some transform work to clean up the pipelines for an example.
- A new exmaple in the darcy flow space enabling the transolver example with multiple datasets at once.

There is a parallel pr for geotransolver to enable it for 2D data.  With that in, we can do GeoTransolver on Darcy with multiple datasets too.

## Checklist

## Review Process

### DIFF
--- a/physicsnemo/datapipes/multi_dataset.py
+++ b/physicsnemo/datapipes/multi_dataset.py
@@ -91,9 +91,10 @@ class MultiDataset:
 
     Parameters
     ----------
-    datasets : Sequence[Dataset]
-        One or more Dataset instances (Reader + transforms each). Order defines
-        index mapping: first dataset occupies 0..len(ds0)-1, etc.
+    *datasets : Dataset
+        One or more Dataset instances passed as positional arguments
+        (Reader + transforms each). Order defines index mapping: first
+        dataset occupies 0..len(ds0)-1, etc.
     output_strict : bool, default=True
         If True, require all datasets to produce the same TensorDict keys (output
         keys after transforms) so :class:`DefaultCollator` can stack batches. If
@@ -106,7 +107,7 @@ class MultiDataset:
     Raises
     ------
     ValueError
-        If ``len(datasets) < 1`` or if ``output_strict=True`` and output keys differ.
+        If no datasets are provided or if ``output_strict=True`` and output keys differ.
 
     Notes
     -----
@@ -147,7 +148,7 @@ class MultiDataset:
     >>> from physicsnemo.datapipes import Dataset, MultiDataset, HDF5Reader, Normalize
     >>> ds_a = Dataset(HDF5Reader("a.h5", fields=["x", "y"]), transforms=None)  # doctest: +SKIP
     >>> ds_b = Dataset(HDF5Reader("b.h5", fields=["x", "y"]), transforms=None)   # doctest: +SKIP
-    >>> multi = MultiDataset([ds_a, ds_b], output_strict=True)                   # doctest: +SKIP
+    >>> multi = MultiDataset(ds_a, ds_b, output_strict=True)                     # doctest: +SKIP
     >>> len(multi) == len(ds_a) + len(ds_b)                                      # doctest: +SKIP
     True
     >>> data, meta = multi[0]   # from ds_a                                      # doctest: +SKIP
@@ -156,8 +157,7 @@ class MultiDataset:
 
     def __init__(
         self,
-        datasets: Sequence[Dataset],
-        *,
+        *datasets: Dataset,
         output_strict: bool = True,
     ) -> None:
         if len(datasets) < 1:
--- a/test/datapipes/core/test_multi_dataset.py
+++ b/test/datapipes/core/test_multi_dataset.py
@@ -29,7 +29,7 @@ class TestMultiDatasetBasic:
         """MultiDataset with two datasets has combined length."""
         ds_a = dp.Dataset(dp.NumpyReader(numpy_data_dir))
         ds_b = dp.Dataset(dp.NumpyReader(numpy_data_dir))
-        multi = dp.MultiDataset([ds_a, ds_b], output_strict=True)
+        multi = dp.MultiDataset(ds_a, ds_b, output_strict=True)
 
         assert len(multi) == len(ds_a) + len(ds_b)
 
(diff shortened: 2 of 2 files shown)


physicsnemo/datapipes/multi_dataset.py
```diff
@@ -0,0 +1,378 @@
...
+        if index < 0 or index >= n:
+            raise IndexError(
+                f"Index {index} out of range for MultiDataset with {n} samples"
+            )
+        # Find k such that cumul[k] <= index < cumul[k+1]
+        for k in range(len(self._cumul) - 1):
+            if self._cumul[k] <= index < self._cumul[k + 1]:
+                return k, index - self._cumul[k]
+        # Fallback (should not be reached)
+        return len(self._datasets) - 1, index - self._cumul[-2]
```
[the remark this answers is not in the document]
peterdsharpe: Good bot! 🤖

physicsnemo/datapipes/multi_dataset.py
```diff
@@ -0,0 +1,378 @@
...
+    >>> multi = MultiDataset([ds_a, ds_b], output_strict=True)                   # doctest: +SKIP
+    >>> len(multi) == len(ds_a) + len(ds_b)                                      # doctest: +SKIP
+    True
+    >>> data, meta = multi[0]   # from ds_a                                      # doctest: +SKIP
+    >>> meta["dataset_index"]   # 0                                              # doctest: +SKIP
+    """
+
+    def __init__(
+        self,
+        datasets: Sequence[Dataset],
```
laserkelvin: This is purely stylistic, but I would prefer `*datasets: Dataset` so users don't have to form a list, or if they do, can just unpack.

physicsnemo/datapipes/multi_dataset.py
```diff
@@ -0,0 +1,378 @@
...
+
+        # Cumulative lengths: cumul[k] = sum(len(datasets[j]) for j in range(k))
+        # So index i is in dataset k when cumul[k] <= i < cumul[k+1], local = i - cumul[k]
+        cumul = [0]
+        for ds in self._datasets:
+            cumul.append(cumul[-1] + len(ds))
+        self._cumul = cumul
+
+        if output_strict:
+            self._field_names = _validate_strict_outputs(self._datasets)
```
laserkelvin, line 184: While keys are validated, what about shapes and data types?

docs/api/datapipes/physicsnemo.datapipes.rst
```diff
@@ -153,35 +155,60 @@ the ``Dataset`` is responsible for the threaded execution of ``Reader``s and
...
-container chains multiple transforms into a single callable.  See
+container chains multiple transforms into a single callable.  Refer to
 :doc:`physicsnemo.datapipes.transforms` for the base class API, ``Compose``,
 and all built-in transforms.
 
 Collation
 ^^^^^^^^^
 
-Combining a set of tensordict objects into a batch of data can, at times,
+Combining a set of TensorDict objects into a batch of data can, at times,
```
megnvidia: This looked like one of those I was typing some text and then a bunch of notifications hit my screen and I lost track of where my cursor was and I thought i was done with what i had written in the one window....and then someone walked by and started talking to me about something.......

I took a guess at what a logical way to complete the sentence was.

physicsnemo/datapipes/multi_dataset.py
coreyjadams: Good call.

physicsnemo/datapipes/multi_dataset.py
coreyjadams: Shapes and data types are not validated.  data types I could add, there are not really any scenarios I can think of where batches are added together and mixing of dtypes happens.

Shape compatibility depends entirely on what collation function is used, and that is not exposed to the dataset.  Only to the DataLoader.  So, I can't validate it here.  If we really need that validation, I think we encourage users to try using the whole pipeline?

coreyjadams:
@laserkelvin thanks for taking a look!

- yes, there is a line of sight towards DDP.  #1507 includes this PR + a training example.  I didn't build DDP into it, but can (and now will!) but tl;dr it's not hard;  we use the usual machinery of distributed samplers to coordinate over ranks and it works as usual.  Datasets are exposed as "one" unit so the sampler indexes into the unit as a whole.
- Users should expect to balance batch composition via samplers, if they want that behavior.  The idea here was that if you compose datasets into a multi-dataset, they become "one" and your sampling treats them as one.  Perhaps you think we should support some sort of custom sampler?  We run the risk of biasing the training of course in that case...
- the intention is not solely for concatenation.  In fact, for many datasets collation via concatenation is not correct.  So I can't do something more sophisticated than check "hey, does every dataset produce the same keys?"  It is not strictly mandatory, though, you could have a different output from each dataset happily, if you are not collating (aka batch size 1).
- yes

docs/api/datapipes/physicsnemo.datapipes.rst
coreyjadams: Ah, it was just a small typo needing to join up with the next sentence.  I'll fix it.  Thanks!

laserkelvin:
> * Users should expect to balance batch composition via samplers, if they want that behavior.  The idea here was that if you compose datasets into a multi-dataset, they become "one" and your sampling treats them as one.  Perhaps you think we should support some sort of custom sampler?  We run the risk of biasing the training of course in that case...

Yeah nah I don't think we need to necessary provide the custom sampler - that will depend a lot on the dataset compositions, and we can't know that _a priori_ generally, I don't think. **But**, I think it is important to highlight that aspect, maybe in an example or something when it comes to it. I imagine there's going to be a lot of imbalanced datasets that, if naively just shuffled, will end up yielding training signals skewed towards the bigger datasets. As long as there's a path forward for then I'm happy with that.

Another thing came to mind as well: how do you envision dealing with different splits?

PR #1519, opened 2026-03-19
Natten ShardTensor and ND support

pzharrington:
# PhysicsNeMo Pull Request

## Description
Extends support for 1D and 3D natten in addition to 2D, and exposes them in `nn.functional`. The lightweight wrappers in `nn.functional` are used to more seamlessly support `ShardTensor` usage via the `register_function_handler` pattern (no more needing to import `partial_na2d` from the natten patches file).

This PR also pulls out the `natten` dependency from the `nn-extras` group into its own dedicated optional dependency group; uses the wheels they release against pinned torch+CUDA combos which is much faster than building natten each install.

## Checklist

## Review Process


pzharrington:
@greptileai

pzharrington:
@greptileai

peterdsharpe:
Hey @pzharrington - would it be possible to streamline the `natten` install process as part of this PR (or in a sibling PR, if you prefer)? As `natten` becomes used more in `nn` and other core parts of PhysicsNeMo, having a smoother install process would be very helpful - particularly for end-users who are not using `natten` functionality but need to install `natten` as a transitive dependency anyway via our layered dependency stack. I know that install burden (e.g., building packages like `transformer_engine`, `torch_geometric`, etc.) was a major motivation for the refactor, so I want to be sure that we keep things as easy for end-users to `uv sync` or `pip install nvidia-physicsnemo` as possible.

It looks like `natten` [provides pre-built wheels](https://natten.org/install/) at https://whl.natten.org, which could be embedded into `pyproject.toml`. What do you think?

pzharrington:
@greptileai

physicsnemo/domain_parallel/shard_utils/natten_patches.py
```diff
@@ -183,147 +179,89 @@ def partial_na2d(
...
     for halo_config in halo_configs:
         lq = halo_padding(lq, q._spec.mesh, halo_config)
         lk = halo_padding(lk, k._spec.mesh, halo_config)
         lv = halo_padding(lv, v._spec.mesh, halo_config)
 
-    # Apply native na2d operation (dilation explicit; other options via na2d_kwargs)
-    x = base_func(lq, lk, lv, kernel_size=kernel_size, dilation=dilation, **na2d_kwargs)
+    x = base_func(
+        lq, lk, lv, kernel_size=kernel_size, dilation=dilation, **natten_kwargs
+    )
```
coreyjadams, line 203: Did you kill all the comments here deliberately or was that Claude?  For some reason, AI really hates commented code in my experience.  I think it's worried about its own job security ;).

physicsnemo/nn/functional/natten.py
```diff
@@ -0,0 +1,179 @@
...
+        Additional keyword arguments forwarded to :func:`natten.functional.na2d`
+        (e.g. ``is_causal``, ``scale``).
+
+    Returns
+    -------
+    torch.Tensor
+        Output tensor of the same shape as ``q``.
+    """
+    if has_torch_function((q, k, v)):
+        return handle_torch_function(
```
coreyjadams, line 116: This syntax,  while valid, might be cleaner if we're using a decorator syntax.  Meaning, and this is just brainstorming here, we might import the `na2d` function and then decorate it with a `custom_op` wrapper from pytorch like we do here on the knn operator: https://github.com/NVIDIA/physicsnemo/blob/main/physicsnemo/nn/functional/knn/_cuml_impl.py#L33

An advantage of this would be that we don't have to duplicate the natten API: we can pass through the decorator and just be registering a key with pytorch, instead of creating a transient full function.  Might look like this:

```python
na1d = torch.library.custom_op("physicsnemo::na1d_wrapper",mutates_args=())(na1d)
na2d = torch.library.custom_op("physicsnemo::na2d_wrapper",mutates_args=())(na2d)
na3d = torch.library.custom_op("physicsnemo::na3d_wrapper",mutates_args=())(na3d)
```

Then, in the ShardTensor wrapper, instead of the function handler, we can catch it with the _named_ function handler, again like the kNN is doing: 
https://github.com/NVIDIA/physicsnemo/blob/main/physicsnemo/domain_parallel/shard_utils/knn.py#L283

What do you think?  I can't promise it'll work since I haven't tried it, but it would make natten updates much easier to keep up with here.

pzharrington: This got messy as the natten functional interface args include some types that `torch.library.infer_schema` complains about when directly decorating with -- I think the easiest way to support all the kwargs (which are nice for perf tuning, customization, etc) is with the duplicated API, even though it's a bit verbose

physicsnemo/domain_parallel/shard_utils/natten_patches.py
pzharrington: Ah yeah that was Claude, let me reintroduce those

pzharrington:
> We might advertise, in the docs, though not needed in this PR, that you could now use neighborhood attention layers and modules freely as building blocks with domain parallelism.

@coreyjadams yup this PR is just to get the ShardTensor and functional interfaces in place, next step is to move some of the attention layer modules into `nn` out of `models.dit` so they can be reused. In that we can shore up the documentation

loliverhennigh (approved):
The natten is not in the FunctionSpec format but I can clean this up in a clean up PR I have for the functional unit tests.

PR #1520, opened 2026-03-19
Refactor functional module layout into category packages

loliverhennigh:
# PhysicsNeMo Pull Request

## Description
Breaking up this PR, 
https://github.com/NVIDIA/physicsnemo/pull/1457

This is just module layout a bit

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -70,6 +70,7 @@ physicsnemo/nn/ @loliverhennigh
 physicsnemo/nn/functional/ @loliverhennigh
 physicsnemo/nn/functional/knn/ @coreyjadams @peterdsharpe @loliverhennigh
 physicsnemo/nn/functional/radius_search/ @coreyjadams @peterdsharpe @loliverhennigh
+physicsnemo/nn/functional/natten/ @pzharrington
 physicsnemo/nn/functional/sdf.py @RishikeshRanade @loliverhennigh
 
 # Module layers
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (or plain `dict` for convenience).
 - Adds `physicsnemo.mesh.remeshing` subpackage with `partition_cells()` for
   creating Voronoi regions around seed points. BVH-accelerated.
+- Added support for 1D, 2D, and 3D neighborhood attention (natten) via
+  `physicsnemo.nn.functional` interface, with full `ShardTensor` support.
 
 ### Changed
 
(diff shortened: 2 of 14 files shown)


physicsnemo/nn/functional/neighbors/radius_search/radius_search.py
peterdsharpe: Can we add `jaxtyping` type hints here, and in `torch_forward`? Doesn't have to come in this PR specifically (though it could), but if we can merge this somewhere into this stack of PRs (wherever you think makes most sense), it'd be nice.

physicsnemo/nn/functional/neighbors/radius_search/radius_search.py
```diff
@@ -73,18 +73,29 @@ class RadiusSearch(FunctionSpec):
...
+            - ``points`` (optional): Neighbor points when ``return_points=True``
+            - ``distances`` (optional): Neighbor distances when ``return_dists=True``
 
     Raises:
         KeyError: If an explicit implementation name is not registered.
         ImportError: If the selected implementation is unavailable.
 
     """
 
+    _BENCHMARK_CASES = (
```
peterdsharpe, line 90: This might be more readable as a dict[str, dict[str, Any]] that get kwarg-splatted?

`("small-p1024-q512-r0p1-m32", 1024, 512, 0.1, 32)`
becomes
`{"small": {"num_points": 1024, "num_queries": 512}`
and so on.

physicsnemo/nn/functional/fourier_spectral/fft.py
```diff
@@ -361,16 +373,18 @@ def torch_forward(input: Tensor) -> Tensor:
         return input
 
     @classmethod
-    def make_inputs(cls, device: torch.device | str = "cpu"):
+    def make_inputs_forward(cls, device: torch.device | str = "cpu"):
```
loliverhennigh, line 377: The changes for function_spec.py that match this will come in the next PR. For now could you let this run through if that's alright.

peterdsharpe (approved):
Some feedback, but overall looks reasonable to me. This is technically a semver-violating change (in that it breaks external imports of files which were not previously `_`-prefixed, and hence might break end-user code upon a minor version bump). If the engineering motivation for this is strong, though, we can do it. Let's at-minimum add a `CHANGELOG.md` note for this, and if there's anything we can do to signpost this for end-users to reduce the pain, that would be great.

I'll let Corey review the minor changes to domain_parallel and friends.

physicsnemo/nn/functional/neighbors/radius_search/radius_search.py
loliverhennigh: I would like to keep it this way as I use this string for the benchmark plots and would like to keep it flexible in what I write here.

physicsnemo/nn/functional/neighbors/radius_search/radius_search.py
loliverhennigh: Added them to all the functionals.

PR #1528, opened 2026-03-23
adding underfill case

ghasemiAb:
# PhysicsNeMo Pull Request

## Checklist

## Review Process


examples/cfd/underfill_dispensing/conf/config.yaml
```diff
@@ -0,0 +1,36 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+hydra:
+  job:
+    chdir: True
+  run:
+    dir: ./outputs/
+
+experiment_name: "Unified-Training"
```
coreyjadams, line 23: I think these names are outdated?

examples/cfd/underfill/datapipe.py
```diff
@@ -0,0 +1,585 @@
...
+        
+    Returns:
+        numpy.ndarray
+    """
+    if isinstance(value, torch.Tensor):
+        return value.detach().cpu().numpy()
+    elif isinstance(value, np.ndarray):
+        return value
+    else:
+        return np.asarray(value)
```
coreyjadams: What is all of this conversion code doing?  Is it necessary?  Is this to save statistics or something else?

examples/cfd/underfill/vtp_reader.py
```diff
@@ -0,0 +1,346 @@
...
+# Patterns ordered from most specific to most permissive.
+# Each pattern must expose a named group ``idx`` used for sorting.
+_TIME_SERIES_PATTERNS: list[re.Pattern] = [
+    # <field>_step00, <field>_step01, ...
+    re.compile(r"^(?P<field>.+?)_step(?P<idx>\d+)$"),
+    # <field>_t0.000, <field>_t0.005, ...  (float time label)
+    re.compile(r"^(?P<field>.+?)_t(?P<idx>\d+\.\d+)$"),
+    # <field>_00, <field>_01, ...  (bare numeric suffix)
+    re.compile(r"^(?P<field>.+?)_(?P<idx>\d+)$"),
+]
```
coreyjadams: Is there a better way to do this?  In general, unless there is a performance reason to want to use regular expressions I'm not a huge fan.  I don't know how to cast spells so easily and its harder to maintain.  Can we use some sort of glob to pattern match, unless that is too slow?

examples/cfd/underfill/vtp_reader.py
```diff
@@ -0,0 +1,346 @@
...
+            first_idx = entries[0][0]
+            last_idx = entries[-1][0]
+            print(
+                f"      Field '{name}': {len(entries)} steps "
+                f"(idx {first_idx} → {last_idx})"
+            )
+
+        arrays: list[np.ndarray] = []
+        for _idx, key in entries:
+            arr = np.asarray(mesh.point_data[key], dtype=np.float64)
```
coreyjadams: Are you sure float64 is necessary?

examples/cfd/underfill/rollout.py
```diff
@@ -0,0 +1,160 @@
...
+
+class TransolverAutoregressiveRollout(GeoTransolver):
+    """
+    GeoTransolver with autoregressive rollout for Epoxy VOF prediction.
+
+    Always processes ALL nodes. Interface masking is handled in the loss
+    (train.py), not here.
+    """
+
+    def __init__(self, *args, **kwargs):
```
coreyjadams: I would prefer this function and constructor to have a proper syntax with named arguments and docstring, unless there is a compelling reason to do it this way.

examples/cfd/underfill/rollout.py
```diff
@@ -0,0 +1,160 @@
...
+            else:
+                delta = self._forward_step(
+                    fx_t.unsqueeze(0), coords.unsqueeze(0)
+                ).squeeze(0)
+
+            vof_next = torch.sigmoid(delta)
+            outputs.append(vof_next)
+            vof_t = vof_next
+
+        return torch.stack(outputs, dim=0)  # [T, N, 1]
```
coreyjadams: We get a lot of questions about doing rollout with models like this.  Please add some comments here to explain what you're doing and why - users will be reading this quite closely, I expect, and I think it will be really valuable to have this section in particular really well documented.

examples/cfd/underfill/rollout.py
```diff
@@ -0,0 +1,160 @@
...
+
+    band = (coords[:, axis] >= z_min - expansion) & (
+        coords[:, axis] <= z_max + expansion
+    )
+    return band
+
+
+
+
+class TransolverAutoregressiveRollout(GeoTransolver):
```
coreyjadams: GeoTransolver is an experimental model and syntax can (and probably WILL) change before it lands in models eventually.  Are you sure about inheriting here vs another design pattern?  It might make sense to have this object own a GeoTransovler instance instead?

examples/cfd/underfill/rollout.py
```diff
@@ -0,0 +1,160 @@
...
+        )
+        phases = coords.unsqueeze(-1) * (2.0 * torch.pi * freqs)
+        sin_enc = torch.sin(phases)
+        cos_enc = torch.cos(phases)
+        enc = torch.cat([sin_enc, cos_enc], dim=-1)
+        enc = enc.reshape(B, N, 2 * 2 * self.num_fourier_frequencies)
+
+        if squeeze:
+            enc = enc.squeeze(0)
+        return enc
```
coreyjadams: We have a lot of fourier feature building available in physicsnemo.nn - please use one of the if at all possible instead of writing another?

examples/cfd/underfill_dispensing/train.py
```diff
@@ -0,0 +1,708 @@
...
+            else:
+                step_fn(closure)
+
+    def state_dict(self):
+        return {"optimizers": [opt.state_dict() for opt in self.optimizers]}
+
+    def load_state_dict(self, state_dict):
+        for opt, sd in zip(self.optimizers, state_dict["optimizers"]):
+            opt.load_state_dict(sd)
+        self.param_groups = [g for opt in self.optimizers for g in opt.param_groups]
```
coreyjadams, line 108: This is upstreamed in physicsnemo and needs to be removed here.  Look in physicsnemo.optim now, please :)

examples/cfd/underfill_dispensing/train.py
```diff
@@ -0,0 +1,708 @@
...
+# ═══════════════════════════════════════════════════════════════════════════════
+# Per-Timestep Interface Loss
+# ═══════════════════════════════════════════════════════════════════════════════
+
+class PerTimestepInterfaceLoss(torch.nn.Module):
+    """
+    Per-timestep interface-only loss.
+    
+    Uses absolute_expansion for predictable behavior on normalized coords.
+    """
```
coreyjadams, line 179: Please expand this docstring with more explanation about this loss construction and its action in the roll-out example?

examples/cfd/underfill_dispensing/train.py
```diff
@@ -0,0 +1,708 @@
...
+        logger0.info("")
+        logger0.info(f"  Total parameters:     {total_params:>12,}")
+        logger0.info(f"  Trainable parameters: {trainable_params:>12,}")
+        logger0.info(f"  Model size:           {total_params * 4 / 1024**2:>11.2f} MB  (fp32)")
+
+        logger0.info("")
+        logger0.info("=" * 72)
+        logger0.info("  STARTING TRAINING")
+        logger0.info("=" * 72)
+        logger0.info("")
```
coreyjadams, line 687: This is a lot of boilerplate printout. You could consider a cleanup with something like `tabulate` but it's not mandatory.

examples/cfd/underfill/inference.py
```diff
@@ -0,0 +1,603 @@
...
+        return cls(
+            mean=float(arr.mean()),
+            std=float(arr.std()),
+            min_val=float(arr.min()),
+            max_val=float(arr.max()),
+            filled_pct=float((arr > threshold).sum() / len(arr) * 100),
+        )
+
+
+def print_header(title: str, width: int = 80):
```
coreyjadams: There are plenty of libraries to simply work like this, FYI, that can reduce clutter in physicsnemo.

examples/cfd/underfill/inference.py
```diff
@@ -0,0 +1,603 @@
...
+def _to_tensor(value, dtype=torch.float32) -> torch.Tensor:
+    """Safely convert a value to a torch tensor."""
+    if isinstance(value, torch.Tensor):
+        return value.to(dtype=dtype)
+    return torch.as_tensor(value, dtype=dtype)
+
+
+def _stats_to_device(stats: dict, device: torch.device, dtype=torch.float32) -> dict:
+    """Convert stats dict to tensors and move to device."""
+    return {k: _to_tensor(v, dtype=dtype).to(device) for k, v in stats.items()}
```
coreyjadams: Duplicated in train.py, no?

coreyjadams (dismissed):
Hi @ghasemiAb ,

Thanks for this PR!  Overall I think it looks really good.  I have scattered some comments here and there, mostly it is request for more organization / code clean up / better comments / docstrings etc.  I think this example will get a lot of attention so I am pushing a little to encourage you to give it an extra level of polish.

The addition of a rollout example for GeoTransolver on a static mesh is pretty cool.  Users have been asking about this type of workload, so this _will_ get attention.  It's really a nice addition to physicsnemo and I'm happy you've contributed it!  You're choice to drive the loss function by the boundary layer is also good, rather than the static components of the mesh.

I have a couple of mandatory updates for approval, if you don't mind:

- Please add a requirements.txt file.  There are some things that have to be included there (like pyvista, for IO), and torch>(whenever muon was included, maybe that was 2.9?).
- Your readme has some math rendering errors, can you make sure to go through it again and fix?
- Please add some example convergence plots and visualizations to the README, if you can.

Hopefully, these aren't too much!

ghasemiAb:
Hi @coreyjadams @ram-cherukuri I considered the comment you made already. Please check and if okay confirm the PR.

examples/cfd/underfill/img/G20.gif
RishikeshRanade, line 1: Are we allowed to publish this result?

examples/cfd/underfill/datapipe.py
```diff
@@ -0,0 +1,687 @@
...
+    Expected VTP data format:
+        - coords: [N, 3]
+        - epoxy_vof: [T, N, 1] with T timesteps
+
+    Statistics exposed (used by train.py and inference.py):
+        - node_stats: {"pos_mean": [3], "pos_std": [3]}
+        - feature_stats: {"feature_mean": [1], "feature_std": [1]}
+    """
+
+    NUM_FEATURES = 1  # Scalar field (epoxy_vof)
```
RishikeshRanade, line 299: Should this be part of config?

examples/cfd/underfill/conf/datapipe/graph.yaml
RishikeshRanade, line 1: Please change the name of the example to underfill_dispensing

examples/cfd/underfill/img/G20.gif
ram-cherukuri: We will share the PR with Arvind so they can call it out if they are not ok with sharing. May be now is a good time to share the link to the readme to make sure they are ok with everything.

examples/cfd/underfill_dispensing/train.py
ghasemiAb: I let it be in the original format

examples/cfd/underfill_dispensing/conf/datapipe/graph.yaml
ghasemiAb: done!

examples/cfd/underfill/inference.py
ghasemiAb: yes! I mentioned specifically that we have this two function in two places. In the next version I will try to have a utils.py and push some fucntions into that.

examples/cfd/underfill/datapipe.py
```diff
@@ -0,0 +1,585 @@
...
+            else:
+                self._log("\n  WARNING: No saved statistics found, computing from current split")
+                self._log("           Run training first to generate statistics!")
+                self.node_stats = self._compute_node_stats()
+                #self.feature_stats = self._compute_feature_stats()
+                # Hardcode feature stats to make normalization a no-op
+                self.feature_stats = {
+                    "feature_mean": torch.zeros(1, dtype=torch.float32),
+                    "feature_std": torch.ones(1, dtype=torch.float32),
+                }
```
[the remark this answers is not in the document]
ghasemiAb: Addressed via USE_VOF_NORMALIZATION = False flag at module level. All three commented-out #self.feature_stats = self._compute_feature_stats() lines removed. _compute_feature_stats is no longer dead code — it's now the True branch of the flag.

The runtime log explicitly states which mode is active on every dataset construction, and the flag has a docstring explaining why identity is the default (VOF is already in [0, 1] with a near-bimodal distribution, so z-score provides negligible benefit while complicating post-processing).

examples/cfd/underfill/datapipe.py
```diff
@@ -0,0 +1,585 @@
...
+        self._log(f"  Split:       {split}")
+        self._log(f"  Data dir:    {self.data_dir}")
+        self._log(f"  Num samples: {num_samples}")
+        self._log(f"  Num steps:   {self.num_steps} (T)")
+        self._log(f"  Rollout:     {self.num_steps - 1} steps (T-1)")
+        self._log(f"  Feature:     epoxy_vof (scalar)")
+
+        # Create stats directory
+        self._stats_dir = STATS_DIRNAME
+        os.makedirs(self._stats_dir, exist_ok=True)
```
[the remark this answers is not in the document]
ghasemiAb: Fixed. Replaced the relative STATS_DIRNAME assignment with a _resolve_stats_dir() helper that:

Accepts an explicit stats_dir override via the dataset config.
Falls back to <hydra.run.dir>/stats/ via HydraConfig.get().runtime.output_dir — this gives a stable absolute path regardless of whether hydra.job.chdir is True/False, so training and later inference invocations resolve to the same directory.
Falls back to ./stats/ only when not running under Hydra (e.g., unit tests).
The dataset now logs the resolved absolute path on construction, so normalization drift would be visible immediately. Training, validation, and inference all read/write the same stats/ directory now.

docs/img/underfill/G20.gif
mnabian, line 1: Images typically live under `docs/img`.

examples/cfd/underfill_dispensing/README.md
```diff
@@ -0,0 +1,437 @@
...
+│              TransolverAutoregressiveRollout                   │
+│                                                                │
+│   Input:  coords [N, 3]  +  vof_t [N, 1]                       │
+│                                                                │
+│   For each timestep t = 0, 1, ..., T-2:                        │
+│   ┌───────────────────────────────────────────┐                │
+│   │  1. Fourier Features(coords[:, [x,z]])    │ → [N, 12]      │
+│   │  2. Concat [vof_t, coords, fourier]       │ → [N, 16]      │
+│   │  3. GeoTransolver forward pass            │ → [N, 1]       │
+│   │  4. Sigmoid activation                    │ → [N, 1]       │
```
mnabian, line 203: Maybe explain more about why we decided to use Sigmoid activation.

examples/cfd/underfill_dispensing/train.py
```diff
@@ -0,0 +1,766 @@
+# train.py
```
mnabian: All your license headers should be corrected.

examples/cfd/underfill_dispensing/conf/config.yaml
```diff
@@ -0,0 +1,36 @@
...
+experiment_desc: "unified training recipe for crash models"
+run_desc: "unified training recipe for crash models"
```
mnabian: These are copied from crash recipe and need to be updated.

examples/cfd/underfill_dispensing/conf/training/default.yaml
```diff
@@ -0,0 +1,54 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# ┌───────────────────────────────────────────┐
+# │                   Data                    │
+# └───────────────────────────────────────────┘  
+
+raw_data_dir:  "/workspace/data/train"
```
mnabian: Intstead of these hard-coded paths, you can use `???`.

examples/cfd/underfill_dispensing/conf/inference/default.yaml
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+output_dir_pred: "./predicted_vtps"
+output_dir_exact: "./exact_vtps"
+raw_data_dir_test: "/workspace/data/val"
+#raw_data_dir_test: raw_data_dir_test:"/workspace/isv/ansys/data/converted_output_singleVTU-VTP/val"
```
mnabian: Remove the hard-coded paths.

examples/cfd/underfill_dispensing/inference.py
```diff
@@ -0,0 +1,630 @@
...
+    logger0 = RankZeroLoggingWrapper(logger, dist)
+    logger0.file_logging()
+    logging.getLogger().setLevel(logging.INFO)
+
+    parent_dir = to_absolute_path(cfg.inference.raw_data_dir_test)
+    if not os.path.isdir(parent_dir):
+        logger0.error(f"Parent directory not found: {parent_dir}")
+        return
+
+    run_dirs = sorted(d.path for d in os.scandir(parent_dir) if d.is_dir())
```
mnabian, line 608: Inference treats `cfg.inference.raw_data_dir_test` as a parent containing one **subdirectory per run**:

```python
run_dirs = sorted(d.path for d in os.scandir(parent_dir) if d.is_dir())
if len(run_dirs) == 0:
    logger0.error(f"No run directories found under: {parent_dir}")
    return
```

but the README (lines 92–101) documents the dataset as a flat directory of VTP files (`val/G24.vtp`, …), and training reads VTPs directly from `raw_data_dir`/`raw_data_dir_validation`. With the documented layout `raw_data_dir_test` contains no subdirectories, so inference logs an error and exits without processing anything. Either change discovery to fall back to scanning `*.vtp` directly under `parent_dir`, or update the README and example data layout to put each run in its own subdir.

examples/cfd/underfill_dispensing/requirements.txt
```diff
@@ -0,0 +1,5 @@
+pyvista>=0.43.0
+tabulate>=0.9.0
+tensorboard>=2.20.0
+torchinfo>=1.8
+torch>=2.7
```
mnabian: `train.py` calls `torch.optim.Muon(...)` unconditionally (line 337), but `torch.optim.Muon` was only added in PyTorch 2.9.

examples/cfd/underfill_dispensing/conf/training/default.yaml
```diff
@@ -0,0 +1,54 @@
...
+num_validation_samples: 5
+validation_freq: 25
+save_chckpoint_freq: 25
+
+interface_mask:
+  vof_lo: 0.01         # nodes with VOF > this AND < vof_hi are "interface"
+  vof_hi: 0.99
+  band_fraction: 0.05 # dilate by 10% of domain extent along interface-normal axis
+  interface_axis: -1    # -1 = auto-detect per sample
+  absolute_expansion: 0
```
mnabian: `absolute_expansion: 0` is the literal integer `0`, not `null`. In `compute_interface_band` (rollout.py:84), the branch is `if absolute_expansion is not None: expansion = absolute_expansion else: expansion = band_fraction * domain_extent`, so with this default the `band_fraction: 0.05` immediately above is **ignored** and the band uses zero expansion. The README and the inline comment ("0 = no fractional dilation", "0 = core interface only") suggest both knobs do something useful at 0, which is misleading. Also note the band still selects all nodes whose axis-coordinate is within `[z_min, z_max]` of the core (often many more than the core itself), not "core only". Recommend setting the default to `null` so `band_fraction` actually applies, or documenting that `absolute_expansion=0` disables `band_fraction`.

examples/cfd/underfill_dispensing/datapipe.py
```diff
@@ -0,0 +1,588 @@
...
+            debug: Enable verbose output
+            stats_dir: Explicit path for normalization statistics directory.
+                       If None (default), resolves to ``<hydra.run.dir>/stats/``
+                       (i.e., ``./outputs/stats/`` with the default config).
+        """
+        self.name = name
+        self.data_dir = data_dir or "."
+        self.split = split
+        self.num_samples = num_samples
+        self.num_steps = min(num_steps, 20)  # Max 20 steps (step00 to step19)
```
mnabian: `self.num_steps = min(num_steps, 20)` silently caps the requested timesteps at 20 with no warning, while `train.py` derives `self.rollout_steps = cfg.training.num_time_steps - 1` from the **uncapped** config.

mnabian (changes requested):

mnabian (approved):
LGTM now.

PR #1534, opened 2026-03-25
Fix Model Sigma Data

albertocarpentieri:
# PhysicsNeMo Pull Request

## Description
- The EDM preconditioner (EDMPrecond) defaults to sigma_data=0.5, ignoring the value set in training.loss.sigma_data. This causes a mismatch between the loss weighting and the preconditioning coefficients (c_skip, c_out, c_in), leading to systematic bias in diffusion model predictions.
- Forward training.loss.sigma_data into the model hyperparameters when building the diffusion network, so the preconditioner and loss use the same value.

## Checklist

- [x ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).


jleinonen:
As per the Greptile comment, this doesn't address the case of `sigma_data` passed as a list. Should we implement it now or leave it for later?

jleinonen:
Can we add a test that makes sure the loss and the model have the same sigma_data?

pzharrington:
Did this bug only apply to `EDMPrecond` (used with songunet) and not `EDMPreconditioner`? If so these changes may be superseded fairly soon. I started working on full adoption of the new `physcisnemo.diffusion` interfaces, thus eliminating use of `EDMPrecond`, still need to figure out the specific changes to support channel-wise `sigma_data` though.

jleinonen:
> Did this bug only apply to `EDMPrecond` (used with songunet) and not `EDMPreconditioner`? If so these changes may be supersedes fairly soon. I started working on full adoption of the new `physcisnemo.diffusion` interfaces, thus eliminating use of `EDMPrecond`, still need to figure out the specific changes to support channel-wise `sigma_data` though.

It seems to affect both and this change should fix both, since both have a `sigma_data` parameter in their call signature.

pzharrington:
Ah ok I was just going off the greptile summary. Regarding the channel-wise `sigma_data`, how impactful has it been to tune that rather than use a scalar or the default 0.5? I don't think StormScope uses per-channel `sigma_data` IIRC

jleinonen:
> Ah ok I was just going off the greptile summary. Regarding the channel-wise `sigma_data`, how impactful has it been to tune that rather than use a scalar or the default 0.5? I don't think StormScope uses per-channel `sigma_data` IIRC

It's mostly relevant for regression-diffusion models. For those the regression net often has very different errors for the channels and then we should choose the RMSE of the regression for each channel as `sigma_data`. Whereas for pure diffusion models we usually have the data normalized to unit variance, in which case `sigma_data = 1.0` should be the most mathematically justified choice.

PR #1544, opened 2026-04-01
GeoTransolver Guard 🛡️ 

mnabian:
# PhysicsNeMo Pull Request

## Problem

When users apply a pretrained GeoTransolver checkpoint to inputs outside the training distribution (e.g., running a DrivAerML-trained model on motorcycles or aircraft), the model silently produces unreliable predictions. There is no mechanism to detect or warn about out-of-distribution (OOD) inputs at inference time.

## Solution

We add two lightweight OOD guards that integrate seamlessly into the existing training and inference workflow. Both guards are controlled by a single knob (`guard_buffer_size`) and require no additional scripts, calibration steps, or changes to the training loop.

- **During training**: guards passively collect calibration data (zero impact on model predictions or gradients).
- **During inference**: guards check inputs against the training distribution and emit `warnings.warn()` if OOD is detected.
- **At checkpoint save**: the kNN threshold is automatically computed from collected data.

## Guard 1: Global Parameters — Bounding Box

**What it monitors:** The raw `global_embedding` input tensor (e.g., air density, stream velocity for DrivAerML).

**How it works:**
- During training, tracks per-dimension running min/max across all batches.
- At inference, compares each dimension of the input against the stored bounds.
- Emits a warning per dimension that falls outside the training range.

**Why input space:** Global parameters are low-dimensional scalars (2-3 dims). A bounding box is simpler, more interpretable, and more reliable than latent-space methods at this dimensionality.

**Example warning:**
```
OOD Guard: global_embedding dim 0 value 0.7500 below training min 0.9000
```

## Guard 2: Geometry Context — kNN Distance

The implementation is based on this paper: https://arxiv.org/pdf/2204.06507
 
**What it monitors:** The geometry context vector produced by `GlobalContextBuilder.geometry_tokenizer` -- a learned 32-dimensional representation of the input geometry, mean-pooled over attention heads and slices.

**How it works:**
- During training, accumulates pooled geometry embeddings into a fixed-size FIFO rolling buffer.
- At checkpoint save, computes a kNN distance threshold from the buffer (99th percentile of leave-one-out k-th nearest neighbor distances).
- At inference, L2-normalizes the input embedding, computes its k-th nearest neighbor distance to the stored training embeddings, and warns if above threshold.

**Why latent space:** Geometry is a variable-size point cloud -- there is no fixed-dimensional input representation to bound. The post-ContextProjector embedding compresses geometry into a fixed 32-dim vector suitable for distance-based methods.

**Why kNN:**
- Training set is small (~400 samples for DrivAerML). Covariance estimation (Mahalanobis) is unreliable.
- No distributional assumption -- geometries may cluster multimodally.
- No additional trainable components needed.

**Why not monitor multi-scale local features:**
- Local features are derived from geometry via ball queries -- not an independent signal.
- They dominate the context dimension (768/832 = 92%), drowning out other signals in a combined detector.
- 768 dims with ~400 samples is a poor regime for kNN (curse of dimensionality).

**Example warning:**

image

### Enabling guards

Add to model config (or pass to constructor):

```yaml
guard_buffer_size: 500  # set to dataset size; null to disable
guard_knn_k: 10         # k for geometry kNN (optional, default 10)
```

Both guards are enabled when `guard_buffer_size` is set, and disabled when it is `None`.

### Training

No changes to the training script. Guards collect data automatically during `model.train()` forward passes. The kNN threshold is computed automatically when the checkpoint is saved (via `state_dict()` override).

### Inference

No changes to the inference script. Guards run automatically during `model.eval()` forward passes and emit Python warnings for OOD inputs.

## Configuration

| Parameter | Default | Description |
|-----------|---------|-------------|
| `guard_buffer_size` | `None` (disabled) | FIFO buffer size for geometry embeddings. Set to dataset size. |
| `guard_knn_k` | `10` | k for k-th nearest neighbor distance. Range 5-15 recommended. |

### Threshold

The kNN threshold is set at the 99th percentile of training-set leave-one-out kNN distances. This means ~1% false alarm rate on in-distribution data -- near-zero warnings on validation/test sets that are in-distribution.

## Multi-GPU / DDP

- Each rank maintains its own buffer independently -- no cross-rank communication.
- The distributed sampler shuffles data each epoch, so after a few epochs each rank's FIFO buffer covers most of the training set.
- Rank 0 saves the checkpoint; its buffer and threshold are what persist.
- Recommendation: set `guard_buffer_size >= dataset_size` to ensure good coverage per rank.

## Checkpoint Compatibility

- **Pre-guard checkpoint loaded into guard-enabled model:** Guard buffers retain their initial values (inf/-inf). Guards remain inactive until training populates the buffers.
- **Guard checkpoint loaded into non-guard model:** Extra `guard_*` keys in the state dict. Requires `strict=False` when loading.

## Tests

This is currently tested for the Crash recipe on the bumper beam dataset. No OOD warnings for inference on test samples. For new OOD samples which have either OOD global parameters or a scaled geometry by a small factor (1.05), the OOD warning is raised.

### Registered buffers

All guard state is stored as registered buffers (persistent, non-trainable):

| Buffer | Shape | Description |
|--------|-------|-------------|
| `guard_global_min` | `(global_dim,)` | Per-dimension training min |
| `guard_global_max` | `(global_dim,)` | Per-dimension training max |
| `guard_geo_embeddings` | `(buffer_size, dim_head)` | FIFO geometry embedding store |
| `guard_geo_ptr` | `(1,)` | FIFO write pointer |
| `guard_knn_threshold` | scalar | 99th percentile kNN distance |
| `guard_knn_k` | scalar | k for kNN |

### Key methods

| Method | When called | Purpose |
|--------|------------|---------|
| `_guard_collect()` | Every training forward pass | Update global bounds, append geometry embeddings to FIFO |
| `_guard_check()` | Every inference forward pass | Check bounds and kNN distance, emit warnings |
| `compute_guard_threshold()` | At `state_dict()` call (before save) | Compute 99th percentile kNN threshold from buffer |

## Checklist

## Dependencies

None


physicsnemo/experimental/models/geotransolver/context_projector.py
```diff
@@ -811,7 +811,9 @@ def build_context(
 
         # Tokenize geometry features
         if self.geometry_tokenizer is not None and geometry is not None:
-            context_parts.append(self.geometry_tokenizer(geometry))
+            geometry_context = self.geometry_tokenizer(geometry)
+            self._last_geometry_context = geometry_context
```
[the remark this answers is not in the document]
mnabian: Addressed.

examples/structural_mechanics/crash/conf/bumper_geotransolver_oneshot.yaml
```diff
@@ -38,9 +38,9 @@ defaults:
 # └───────────────────────────────────────────┘
 
 training:
-  raw_data_dir: ???              # set in config or via CLI: training.raw_data_dir=/path/to/train
-  raw_data_dir_validation: ???   # set in config or via CLI: training.raw_data_dir_validation=/path/to/validation
-  global_features_filepath: ???  # set in config or via CLI: training.global_features_filepath=/path/to/global_features.json
+  raw_data_dir: /code/datasets/bumper_beam/train/               # set in config or via CLI: training.raw_data_dir=/path/to/train
+  raw_data_dir_validation: /code/datasets/bumper_beam/validation/   # set in config or via CLI: training.raw_data_dir_validation=/path/to/validation
+  global_features_filepath: /code/datasets/bumper_beam/global_features.json  # set in config or via CLI: training.global_features_filepath=/path/to/global_features.json
```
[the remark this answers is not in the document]
mnabian: Addressed.

physicsnemo/experimental/models/geotransolver/geotransolver.py
```diff
@@ -23,6 +23,8 @@
 
 from __future__ import annotations
 
+import logging
+import warnings
```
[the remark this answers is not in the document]
mnabian: Addressed.

physicsnemo/experimental/models/geotransolver/geotransolver.py
```diff
@@ -556,4 +617,111 @@ def forward(
...
+                k = self.guard_knn_k.item()
+                kth_dists = dists.topk(k, largest=False).values[:, -1]  # (B,)
+                threshold = self.guard_knn_threshold.item()
+                for i in range(kth_dists.shape[0]):
+                    dist_val = kth_dists[i].item()
+                    if dist_val > threshold:
+                        logging.warning(
+                            f"{_RED}OOD Guard: geometry sample {i} kNN distance "
+                            f"{dist_val:.4f} above threshold {threshold:.4f}{_RESET}"
+                        )
```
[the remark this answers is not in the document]
mnabian: Addressed.

physicsnemo/experimental/models/geotransolver/geotransolver.py
```diff
@@ -556,4 +617,111 @@ def forward(
...
+        n_valid = min(ptr, buf_size)
+        store = self.guard_geo_embeddings[:n_valid]
+        store_norm = store / (store.norm(dim=-1, keepdim=True) + 1e-8)
+        dists = torch.cdist(store_norm, store_norm)  # (N, N)
+        dists.fill_diagonal_(float("inf"))
+        k = self.guard_knn_k.item()
+        kth_dists = dists.topk(k, largest=False).values[:, -1]  # (N,)
+        # 99th percentile — near-zero false alarms on in-distribution data
+        threshold = torch.quantile(kth_dists, 0.99)
+        self.guard_knn_threshold.copy_(threshold)
```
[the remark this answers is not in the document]
mnabian: Addressed.

mnabian:
Note: Tests will be added after the initial review is done for the overall design and implementation of this guardrail.

physicsnemo/experimental/models/geotransolver/geotransolver.py
```diff
@@ -589,6 +650,8 @@ def forward(
         else:
             x = tuple(x)
 
+        return x
+
         if return_embedding_states:
             return x, embedding_states
-        return x
\ No newline at end of file
+        return x
```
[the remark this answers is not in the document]
mnabian: This is addressed.

physicsnemo/experimental/guardrails/embedded/ood_guard.py
```diff
@@ -0,0 +1,383 @@
...
+        if n_valid == 0:
+            return
+        store = self.geo_embeddings[:n_valid]
+        store_norm = store / (store.norm(dim=-1, keepdim=True) + 1e-8)
+        dists = torch.cdist(store_norm, store_norm)
+        dists.fill_diagonal_(float("inf"))
+        k = min(self.knn_k, n_valid - 1)
+        if k <= 0:
+            return
+        avg_knn_dists = dists.topk(k, largest=False).values.mean(dim=-1)
```
coreyjadams: Can this switch to physicsnemo.nn.functional's knn for a non-brute force method?

physicsnemo/experimental/guardrails/embedded/ood_guard.py
```diff
@@ -0,0 +1,383 @@
...
+        n_valid = self._n_valid()
+        if n_valid == 0:
+            return
+        store = self.geo_embeddings[:n_valid]
+        store_norm = store / (store.norm(dim=-1, keepdim=True) + 1e-8)
+        dists = torch.cdist(z, store_norm)  # (B, n_valid)
+        # Query is not in the store, so no -1 needed; clamp to buffer size.
+        k = min(self.knn_k, n_valid)
+        avg_knn_dists = dists.topk(k, largest=False).values.mean(dim=-1)  # (B,)
+        over = avg_knn_dists > self.knn_threshold
```
coreyjadams: Same here, for a non-brute-force knn

physicsnemo/experimental/models/geotransolver/geotransolver.py
```diff
@@ -562,10 +605,23 @@ def forward(
...
+                local_embedding, local_positions, geometry, global_embedding
+            )
         )
 
+        # --- OOD Guard ---
+        if self.ood_guard is not None:
+            # Pool (B, H, S, D) -> (B, D); guard expects pre-pooled latents.
+            geo_latent = (
+                geo_ctx.mean(dim=(1, 2)) if geo_ctx is not None else None
+            )
```
coreyjadams, line 619: Is there any change needed here if the backend attention is FLARE instead of PhysicsAttention?

coreyjadams (approved):
Approving to unblock, overall it looks good, I do think we should use a non-brute force kNN.

I've got a unified training recipe for geotransolver that makes some small tweaks to geotransolver too, so whoever merges second will have minor conflicts to deal with but it will not be difficult I expect.

physicsnemo/experimental/models/geotransolver/geotransolver.py
mnabian: No change needed. `geo_ctx` comes from the context builder, which runs before the attention blocks. Its shape is the same whether you use GALE or FLARE — the attention type only changes what happens after, not the context builder. So the pooling works for both.

coreyjadams:
/ok to test [9c59ee4](https://github.com/NVIDIA/physicsnemo/pull/1544/commits/9c59ee44703b87d25c5f4cb83ae8c4faadcfa57c)

physicsnemo/experimental/guardrails/embedded/ood_guard.py
mnabian: Switched to physicsnemo.nn.functional's knn

physicsnemo/experimental/guardrails/embedded/ood_guard.py
mnabian: Switched to physicsnemo.nn.functional's knn

examples/structural_mechanics/crash/README.md
```diff
@@ -211,6 +211,47 @@ torchrun --nproc_per_node=<NUM_GPUS> inference.py --config-name=bumper_geotranso
...
+- **Geometry** — k-nearest-neighbour distance on a pooled geometry latent.
+
+Enable it through the model config by setting `guard_config` to a mapping
+(leave `null` to disable):
+
+```yaml
+# conf/my_experiment.yaml
+model:
+  guard_config:
+    buffer_size: 121       # FIFO buffer; typically = num_training_samples
```
RishikeshRanade, line 231: Can you add more details about what these config control? and a recommendation setting for different datasets.

examples/structural_mechanics/crash/conf/bumper_geotransolver_oneshot.yaml
```diff
@@ -75,4 +75,12 @@ datapipe:
 model:
   functional_dim: 3   # coords (3)
   out_dim: 250        # (num_time_steps - 1) * 5 = 50 * 5
-  global_dim: 3       # must match len(datapipe.global_features)
\ No newline at end of file
+  global_dim: 3       # must match len(datapipe.global_features)
+  # OOD guard (disabled by default — set `guard_config` to a mapping to enable)
+  guard_config: null
```
RishikeshRanade, line 80: Can we make sure that these changes are reflected in other geotansolver recipes and configs and they dont fail because of this? For example, external aero

physicsnemo/experimental/models/geotransolver/context_projector.py
```diff
@@ -798,13 +800,17 @@ def build_context(
...
         -------
-        tuple[torch.Tensor | None, list[torch.Tensor] | None]
+        tuple[torch.Tensor | None, list[torch.Tensor] | None, torch.Tensor | None]
             - ``context``: Concatenated context tensor of shape :math:`(B, H, S, D_c)`
               where :math:`D_c` is the total context dimension, or ``None`` if no
               context sources are provided.
             - ``local_features``: List of local feature tensors, one per input type,
               each of shape :math:`(B, N, D_l)`, or ``None`` if local features are
               disabled.
+            - ``geometry_context``: Detached geometry-tokenizer output of shape
```
RishikeshRanade: nit: needs to be geometry_context_detached

test/models/geotransolver/test_geotransolver.py
```diff
@@ -748,3 +748,107 @@ def test_geotransolver_metadata():
...
+        dropout=0.0,
+        n_head=4,
+        act="gelu",
+        mlp_ratio=2,
+        slice_num=8,
+        use_te=False,
+        time_input=False,
+        plus=False,
+        include_local_features=False,
+        guard_config=guard_config,
```
RishikeshRanade, line 776: We are releasing a checkpoint trained on the model with the guard config in pnemo-cfd. After looking at the code, I am of the opinion that it should be okay and the checkpoint will break. Do we need to test it? What do you think?

RishikeshRanade (approved):

examples/structural_mechanics/crash/conf/bumper_geotransolver_oneshot.yaml
mnabian: By default, the guardrail is disabled and the examples that are using geotransolver will run like before with no changes to the config files.

test/models/geotransolver/test_geotransolver.py
mnabian: I'm not sure if I fully understand this comment. But previous trained checkpoints should work fine with this model if we set guard_config to None.

physicsnemo/experimental/models/geotransolver/context_projector.py
mnabian: Fixed

examples/structural_mechanics/crash/README.md
mnabian: Added

PR #1556, opened 2026-04-09
ShardTensor Refactor

coreyjadams:
# PhysicsNeMo Pull Request

This pull request implements a substantial refactor of shard tensor components.  I'll add some motivations as comments on the change log, for clarity, but major highlights:

- ShardTensor becomes its **own** class inheriting straight from torch.Tensor, instead of DTensor.  We no longer are tied to DTensor API.  This frees us to make future decisions about API.  It doesn't change today's programming model.
- One challenge here was ensure we manage the tensor properties about `requires_grad` and `is_leaf` correctly, which requires care since calling setters on them will go through the dispatch mechanism, fall back to DTensor, and eventually no-op unless we guard it.  This took me an embarrassingly long time to debug 😭 .
- The logic of the dispatch mechanism has been streamlined and clarified.  The challenge is that the `__torch_function__` ops are differentiable, while `__torch_dispatch__` ops are not and attach a gradient function to the outputs.  So a lot of care has to be taken, there.  I've tried to make the logic chain clean and clear in the codebase... hopefully it is.
- I've implemented several components of `ShardTensor` that are necesary for `torch.compile` support.  Ops like `__tensor_flatten__` and `__tensor_unflatten__` need to be available and we need to implement them ourself for compilation to work properly.
- The unbind op, which was previously implemented as an extension to DTensor and we use it as a fall back, is now just implemented as a ShardTensor op.  This was causing issues because the PyTorch API kept changing .... 
- I also fixed a couple small bugs.  Thanks Claude.

I also updated the tests slightly: I increased the conv2d image sizes for stability, and added some tests for new ops.

From a user perspective, this should be nearly transparent: no API changes, all tests still pass, etc.  It's basically an under-the-hood refactor.  But if we want to enable `torch.compile` for `ShardTensor` we are going to need this.

## Checklist

## Review Process


physicsnemo/domain_parallel/custom_ops/_tensor_ops.py
```diff
@@ -16,130 +16,215 @@
...
 
-if check_version_spec("torch", "2.10.0a"):
-    from torch.distributed.tensor._ops.registration import (
-        register_prop_rule,
-    )
-else:
-    from torch.distributed.tensor._ops.utils import (
-        register_prop_rule,
-    )
-
```
coreyjadams, line 50: This was fragile and causing issues in CI, and the primary reason for refactoring this into a shardtensor op instead of DTensor.

physicsnemo/domain_parallel/shard_utils/normalization_patches.py
```diff
@@ -282,13 +282,13 @@ def backward(
...
 
-        if weight is not None and weight.requires_grad:
+        if weight is not None and ctx.needs_input_grad[3]:
             # grad_weight_c = sum_{n, spatial} grad_output * y  (per-channel)
             y_c = y.view(N, C, HxW_local)
             grad_out_c = local_grad_output.view(N, C, HxW_local)
             grad_weight = (grad_out_c * y_c).sum(dim=(0, 2))  # (C,)
 
-        if bias is not None and bias.requires_grad:
+        if bias is not None and ctx.needs_input_grad[4]:
```
coreyjadams, line 291: This was a nasty bug: we have to check the context if something needs a grad, not the tensor itself.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -42,67 +45,262 @@
...
 
-def _shard_tensor_to_dtensor(st: "ShardTensor") -> DTensor:
-    r"""Convert a ShardTensor to a plain DTensor for dispatch.
+# ======================================================================
 
-    Creates a DTensor with the same internal state as the ShardTensor,
-    which allows DTensor's dispatch to handle it correctly.
+# ============================================================================
+# Layer 1 -- Semi-private conversions (no autograd, no spec inference)
+# ============================================================================
```
coreyjadams, line 55: I opted for a design which builds up this dispatch fallback mechanism in stages, so it can be careful and deliberate about what is routing where and when.

So I am calling it with "layers" and each layer is getting more complex in the autograd graph and ShardTensorSpec / TensorSpec interpolation.  This one here is the least complex, not doing auto grad, not  figuring out specs.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -42,67 +45,262 @@
...
-        strides=st._spec.tensor_meta.stride,
-        dtype=st.dtype,
-        device=st.device,
-        layout=st.layout,
-        requires_grad=st.requires_grad,
-    )
+    if hasattr(torch.Tensor, "_dtensor__new__"):
+        dtensor = torch.Tensor._dtensor__new__(
+            DTensor, st._local_tensor, st._spec, requires_grad=st.requires_grad
+        )
```
coreyjadams, line 67: This is a newer API.  so the fall back is also available too.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -42,67 +45,262 @@
...
+    Backward: :func:`_shard_tensor_to_dtensor`.
+    """
+
+    @staticmethod
+    def forward(ctx, dtensor: DTensor, spec: ShardTensorSpec) -> "ShardTensor":
+        return _dtensor_to_shard_tensor(dtensor, spec)
+
+    @staticmethod
+    def backward(ctx, grad_output: "ShardTensor"):
+        return _shard_tensor_to_dtensor(grad_output), None
```
coreyjadams, line 122: The differentiable conversions just let autograd attach grad functions correctly to the conversion functions in level 1.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -42,67 +45,262 @@
...
-    elif isinstance(arg, Mapping):
-        return type(arg)({k: _convert_args_to_dtensor(v) for k, v in arg.items()})
-    elif isinstance(arg, Iterable) and not isinstance(arg, (str, bytes)):
-        converted = [_convert_args_to_dtensor(a) for a in arg]
-        return type(arg)(converted)
-    return arg
 
+    @staticmethod
+    def forward(ctx, st: "ShardTensor") -> DTensor:
+        ctx.shard_tensor_spec = st._spec
```
coreyjadams, line 137: We cache the spec, because in the backward pass it's likely that the spec from the forward pass is correct for the input gradients in the backward pass.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -42,67 +45,262 @@
...
+            and dtensor._spec.placements == arg._spec.placements
+        ):
+            return arg._spec
+    return _infer_shard_tensor_spec_from_local_chunks(
+        dtensor._local_tensor,
+        dtensor._spec.mesh,
+        dtensor._spec.placements,
+        sharding_shapes="chunk",
+        global_shape=dtensor.shape,
+    )
```
coreyjadams, line 172: The goal here is to figure out a correct spec for a DTensor -> ShardTensor promotion based on other args.  So, if we're doing a unary op, for example:  `y = 2*x`, we fall back to DTensor, do the op, and get a DTensor back.  We don't want to communicate to figure out the sharding shapes, though, so this looks at all the input args to figure out, "Hey, does any input arg match this output well enough to use it's spec?"

We match on the tensor meta (overall shape) + placements + input arg has to be a shard tensor, which seems to be robust here.

Otherwise, we assume the DTensor has DTensor chunking and promote that.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -42,67 +45,262 @@
...
+    with _conversion_scope():
+        converted_args = tuple(_convert_args_to_dtensor(arg, use_autograd=False) for arg in args)
+        converted_kwargs = {
+            k: _convert_args_to_dtensor(v, use_autograd=False) for k, v in (kwargs or {}).items()
+        }
+        
+    dispatch_res = func(*converted_args, **(converted_kwargs or {}))
+    
+    with _conversion_scope():
+        return _convert_results_to_shard_tensor(dispatch_res, args, use_autograd=False)
```
coreyjadams, line 221: There are two things going on here.  This one is critical, actually.

First, there is a possible infinite recurision depth if we're dumb: because the dispatch fallbatch itself uses torch functions to attach autograd functions, for the differentiable paths, if we do not prevent that we will infinitely dispatch and reach the recursion depth in python.  So that's fun.  So we use this conversion scope and call it in `__torch_function__` below, in shard tensor itself, to skip the dispatch conversion if _we're already within a conversion scope_.

And, we make it thread safe with a thread local counter rather than a boolean.  Got my eye on you, free threaded python ...

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -42,67 +45,262 @@
...
+            # DTensor can be iterable; exit early deliberately
+            return arg
+        case Mapping():
+            return type(arg)({k: _convert_args_to_dtensor(v, use_autograd) for k, v in arg.items()})
+        case tuple():
+            return tuple(_convert_args_to_dtensor(a, use_autograd) for a in arg)
+        case list():
+            return [_convert_args_to_dtensor(a, use_autograd) for a in arg]
+        case _:
+            return arg
```
coreyjadams, line 277: I think I caught most of the cases, here, but if I missed any we can add them.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -42,67 +45,262 @@
...
+
+
+def _convert_results_to_shard_tensor(
+    result: object, input_args: tuple, use_autograd: bool = False
+) -> object:
+    r"""Recursively replace DTensors with ShardTensors in an op result.
+    
+    If use_autograd is True, uses Layer 2 to preserve the graph connection.
+    Handles None returns gracefully for inplace ATen operations.
+    """
```
coreyjadams, line 287: This is the function that is meant to take outputs from DTensor and turn them back to ShardTensor, if they shouold be ShardTensor.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -296,74 +500,7 @@ def backward(
...
-        Returns
-        -------
-        Tuple[DTensor, None]
-            The gradient for the DTensor input, and ``None`` for the spec.
-        """
-        return grad_output, None
-
-
-class ShardTensor(DTensor):
+class ShardTensor(torch.Tensor):
```
coreyjadams, line 518: Now straight from the torch.Tensor tap!

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -488,6 +625,71 @@ def register_named_function_handler(cls, func_name: str, handler: Callable) -> N
...
+    #     # Set requires_grad AFTER _spec/_local_tensor are assigned, using
+    #     # the C-level setter directly (bypassing our Python property
+    #     # override) so the autograd engine sees the correct flag.
+    #     if requires_grad:
+    #         torch.Tensor.requires_grad.__set__(ret, True)
+
+    #     cls._enable_shard_patches = True
+
+    #     return ret
+
```
coreyjadams: We can probably delete this.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -496,254 +698,205 @@ def __new__(
...
+        return self._spec.placements
 
-        Returns
-        -------
-        ShardTensor
-            Equivalent ShardTensor with the same local tensor and inferred spec.
-        """
-        return cls._maybe_promote_dtensor(dtensor, ())
+    def __tensor_flatten__(self):
+        return ["_local_tensor"], (self._spec, self.requires_grad)
```
coreyjadams, line 702: This isn't really doing anything yet.  We're going to need this for compilation, but it's not yet enabled... we could cut this for this PR if you like.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -496,254 +698,205 @@ def __new__(
...
+            tensor_meta=unflatten_meta,
+            _local_shape=local_tensor.shape,
+            _sharding_shapes=spec._sharding_shapes,
+        )
+        return ShardTensor.__new__(
+            ShardTensor,
+            local_tensor=local_tensor.requires_grad_(requires_grad),
+            spec=unflatten_spec,
+            requires_grad=requires_grad,
+        )
```
coreyjadams, line 725: Similar to __flatten__.  We could cut for now.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -496,254 +698,205 @@ def __new__(
...
-                for d in result
-            )
-
-        return result
+        if isinstance(dtensor, ShardTensor):
+            return dtensor
+        spec = _resolve_spec_for_dtensor(dtensor)
+        if dtensor.grad_fn is not None:
+            return _DTensorToShardTensor.apply(dtensor, spec)
+        return _dtensor_to_shard_tensor(dtensor, spec)
```
coreyjadams, line 844: Fall back to workers, above.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -496,254 +698,205 @@ def __new__(
...
+            # When converting shard tensor to dtensor, or dtensor to shard tensor,
+            # we just run the function without ShardTensor dispatch.
+            with torch._C.DisableTorchFunctionSubclass():
+                return func(*args, **kwargs)
+        if func in cls._function_registry and cls._enable_shard_patches:
+            return cls._function_registry[func](func, types, args, kwargs)
+        if str(func) in cls._named_function_registry and cls._enable_shard_patches:
+            return cls._named_function_registry[str(func)](func, types, args, kwargs)
+        res = _torch_function_fallback_via_dtensor(func, args, kwargs)
+        return res
```
coreyjadams, line 860: This is now all streamlined to the above functions.  It's really just a conversion check (are we converting?  Yes?  Then the function being run _is_ the conversion, so run it right away and return) otherwise check the handlers and then go to the fall back handler.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -496,254 +698,205 @@ def __new__(
...
-            # Promote any DTensor results back to ShardTensor.
-            return cls._promote_dtensor_results(dispatch_res, args)
+        # Use a handler, if we have one:
+        handler = cls._dispatch_registry.get(func)
+        if handler is None:
+            handler = cls._dispatch_registry_by_name.get(str(func))
+        if handler is not None:
+            return handler(*args, **kwargs)
+        # Otherwise, try the dtensor route:
+        return _dispatch_fallback_via_dtensor(func, args, kwargs)
```
coreyjadams, line 877: Likewise, streamlined.  Check the registry, and then go for the fallback.

No conversion check here, that's not a dispatch op.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -962,9 +1115,43 @@ def backward(self, *args, **kwargs):
...
+    device_mesh: DeviceMesh,
+    partition_fn: (Callable[[str, nn.Module, DeviceMesh], None] | None) = None,
+) -> nn.Module:
+    """Distribute a module over a domain mesh and adapt outputs for FSDP."""
+    distributed_module = distribute_module(
+        module,
+        device_mesh=device_mesh,
+        partition_fn=partition_fn,
+    )
+    return wrap_for_fsdp(distributed_module)
```
coreyjadams, line 1130: These are new, which I did to let us plug models into FSDP more easily and try to fix some of these bugs.  I am not certain they are necessary or not any more with some of the other fixes.

test/domain_parallel/ops/test_view_ops.py
```diff
@@ -534,20 +534,23 @@ def test_view_trailing_dims_1d_to_3d(
     distributed_mesh,
     backward,
 ):
-    """Test view (6,) -> (2, 3, 1) with Shard(0): trailing dim must stay in group.
+    """Test view (48,) -> (8, 6, 1) with Shard(0): trailing singleton in target.
```
coreyjadams, line 537: Just making it bigger so the view is actually defined properly on more than 2 GPUs.

test/domain_parallel/ops/test_convolution.py
```diff
@@ -293,7 +293,7 @@ def test_conv_transpose_2d_1dmesh(
 
 
 @pytest.mark.multigpu_static
-@pytest.mark.parametrize("H", [32, 256])
+@pytest.mark.parametrize("H", [128, 256])
```
coreyjadams, line 296: This is just for cudnn stability, 32 /2 and 32 / 4 were really numerically unstable.

test/domain_parallel/test_initialization.py
```diff
@@ -121,6 +121,97 @@ def init_from_data_rank_worker(mesh):
...
+        mesh,
+        placements,
+        global_shape=torch.Size(global_shape),
+        dtype=torch.float32,
+        requires_grad=requires_grad,
+    )
+
+    assert st.requires_grad is requires_grad
+    if requires_grad:
+        assert st.is_leaf
```
coreyjadams, line 150: This is a new test to make sure if we use `scatter_tensor`, and say `require_grad`, it actually obeys that.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -138,11 +336,17 @@ def forward(
         ctx.grad_placements = grad_placements
         local_tensor = input._local_tensor
```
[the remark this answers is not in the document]
coreyjadams: Valid, fixed.

jleinonen:
Hi @coreyjadams, could you run the test suite here https://github.com/NVIDIA/physicsnemo/blob/main/examples/weather/stormcast/test_training.py before merging to make sure the PR doesn't break something unexpected? That test suite is part of an example so it won't run automatically as part of CI but contains many end-to-end tests using ShardTensor. See here https://github.com/NVIDIA/physicsnemo/tree/main/examples/weather/stormcast#testing for instructions on how to run the tests (including multi-GPU tests).

Or even better, you could enable torch compiling in the tests and check that it works with domain parallelism!

coreyjadams:
@jleinonen Thanks for bringing it to my attention - I'll run your tests before merge, yes.

physicsnemo/domain_parallel/shard_utils/normalization_patches.py
pzharrington: Only since `.requires_grad` is implicated, want to ask if we'll need to modify/update this fix I'm adding: https://github.com/NVIDIA/physicsnemo/pull/1566

It is more related to frozen params with `nn.Parameter(..., requires_grad=False)` getting silently updated by `distribute_module` in PyT <= 2.10. Relevant bugged PyT code is [here](https://github.com/pytorch/pytorch/blob/449b1768410104d3ed79d3bcfe4ba1d65c7f22c0/torch/distributed/tensor/_api.py#L972), it has since been fixed on main [here](https://github.com/pytorch/pytorch/commit/cc08f9222c8c7b60562ee58f7283a42d2fc1009a)

pzharrington:
Those stormcast end-to-end tests are likely to catch things, but yeah also the domain parallel diffusion and distributed checkpoint save/load multigpu tests I added should provide a second layer of sanity.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -136,13 +351,17 @@ def forward(
         """
         ctx.shard_tensor_spec = input._spec
         ctx.grad_placements = grad_placements
-        local_tensor = input._local_tensor
+        # # JUST LIKE DTENSOR:
+        # # We need to return a fresh Tensor object there as autograd metadata
+        # # will be inplaced into it. So we don't want to pollute the Tensor
+        # # object stored in the _local_tensor of this ShardTensor.
+        # return local_tensor.view_as(local_tensor)
```
pzharrington, line 358: Can drop the commented out bit

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -496,254 +648,233 @@ def __new__(
...
         ret._spec = spec
         ret._local_tensor = local_tensor
 
-        cls._enable_shard_patches = True
+        # Set requires_grad AFTER _spec/_local_tensor are assigned, using
+        # the C-level setter directly (bypassing __torch_function__ which
+        # would convert to DTensor and set on a temporary).
+        if requires_grad:
+            with torch._C.DisableTorchFunctionSubclass():
+                torch.Tensor.requires_grad.__set__(ret, True)
```
pzharrington, line 669: How stable do you think this approach will be across PyT versions?

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -496,254 +648,233 @@ def __new__(
...
+            ShardTensor,
+            local_tensor=local_tensor.requires_grad_(requires_grad),
+            spec=unflatten_spec,
+            requires_grad=requires_grad,
+        )
+
+    # -- Autograd property overrides -------------------------------------------
+    # The C-level requires_grad is authoritative for autograd engine
+    # decisions; we read it first and fall back to _local_tensor for the
+    # case where _make_wrapper_subclass didn't propagate it correctly.
```
pzharrington, line 730: > fall back to _local_tensor for the case where _make_wrapper_subclass didn't propagate it correctly.

Can you explain the scenario that would lead to this happening a bit more?

pzharrington:
@coreyjadams general question on this:

> From a user perspective, this should be nearly transparent: no API changes, all tests still pass, etc. It's basically an under-the-hood refactor. 

Does that mean the recommended procedure for data and domain parallel training is still to `distribute_module` for any params that need sharding, then wrap with FSDP?

physicsnemo/domain_parallel/shard_tensor.py
pzharrington: > I did to let us plug models into FSDP more easily and try to fix some of these bugs.

Can you elaborate on the bugs in question? 😅

physicsnemo/domain_parallel/shard_tensor.py
pzharrington: I can tell you had fun debugging this 😱 I find Layer 3 to be the gnarliest among the conversion layers

physicsnemo/domain_parallel/shard_utils/normalization_patches.py
coreyjadams: Your PR shouldn't be affected.  This code here is fragile, I believe, because this requires grad state is not reliable inside of this torch.autograd.Function context.  Outside of that its OK still.

This isn't a ShardTensor thing, in particular - I think I was doing this subtly wrong previously before.

physicsnemo/domain_parallel/shard_tensor.py
coreyjadams: It actually took me 1+ month, off and on, coming back to this and poking at it to track down what the heck was happening in all of this.

physicsnemo/domain_parallel/shard_tensor.py
coreyjadams: I believe we had some reported user issues with FSDP, no?  Didn't that bug report come from a user to me via you? 😅 

I guess if we can't reproduce any bugs we don't need this API....

physicsnemo/domain_parallel/shard_tensor.py
coreyjadams: As stable as possible - I don't see any issues with this.  We were really getting bit on PyTorch moving the DTensor wrapper APIs around in prereleases, which is what NVIDIA was building containers on.  That's the #1 reason I pulled the `unbind` interface up from DTensor into ShardTensor in this PR too, so that we don't have exposure to that API.

coreyjadams:
>Does that mean the recommended procedure for data and domain parallel training is still to distribute_module for any params that need sharding, then wrap with FSDP?

Yes, for today this is still the case.  I know we've been mulling that API change and whether it's a good design decision or not.  That's not part of this PR at least.  I propose we at least postpone any major API changes to next release if you don't mind, and focus on getting to a stable product here :)

physicsnemo/domain_parallel/shard_tensor.py
coreyjadams: I had Claude make a diagram!

image

tl;dr once scenario is very much an edge case right now but will become a real concern when we enable compile in the future and start dealing with flatten / unflatten.

The second case is if someone manually sets the grad on _local_tensor, somewhere, but shardtensor didn't get set.  So we can catch that too.

coreyjadams:
Updating per conversation with @pzharrington:
- This PR seems to interfere with StormScope training and FSDP issues, which is unfortunate.  The FSDP issue appears as an "unexpected state" error during the backward pass: Using shard tensor appears to leave FSDP parameters in an unexpected state for stormscope.
- The solution is almost certainly to refactor StormScope training recipe onto FSDP 2, or ShardTensor onto DDP-only compatibility at the same time, so that we can avoid FSDP1.
- This PR is likely to defer until *after* 26.05 release, due to time constraints.

So, I'll leave this open, and not merge it yet.

coreyjadams:
@pzharrington @negin513  Now that the FSDP2 PR has merged for storm scope, where does that leave this PR RE:dont-break-stormscope, if you don't mind me asking?

CharlelieLrt (approved):
Approving for all changes in `physicsnemo/diffusion` - did not check the rest

physicsnemo/domain_parallel/shard_tensor.py
coreyjadams: This has been purged because it's no longer needed.

coreyjadams:
/ok to test [fb8691f](https://github.com/NVIDIA/physicsnemo/pull/1556/commits/fb8691f7d6fee778022942956e490377748cd820)

coreyjadams:
/ok to test [b374888](https://github.com/NVIDIA/physicsnemo/pull/1556/commits/b374888b45d53502d5c86871a60f5bbc6e1e9cc7)

coreyjadams:
For posterity, since the multiGPU PR is in flight and we can't test it here.

physicsnemo 4x GB200 tests, --multigpu-static:
```
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============= 551 passed, 15082 skipped, 235 xfailed, 221 warnings in 233.77s (0:03:53) =============
[ble: elapsed 255.221s (CPU 277.6%)] uv run torchrun --nproc-per-node 4 -m pytest --multigpu-static t
```

physicsnemo, 4xGB200, --multigpu-dynamic:
```
==================== 14 passed, 15854 skipped, 24 warnings in 237.77s (0:03:57) =====================
```

Stormcast, 4xGB200:
```
❯ uv run torchrun --nproc-per-node 4 -m pytest test_training.py
....

====================== 5 passed, 105 skipped, 33 warnings in 94.34s (0:01:34) =======================
```


Stormcast, 2xGB200:
```
❯ uv run torchrun --nproc-per-node 2 -m pytest test_training.py
....
===================== 34 passed, 76 skipped, 327 warnings in 627.75s (0:10:27) ======================
```

coreyjadams:
/ok to test [b8a827c](https://github.com/NVIDIA/physicsnemo/pull/1556/commits/b8a827cd2009388df741ce431be07b987d4a876b)

pzharrington (approved):

examples/weather/stormcast/requirements.txt
```diff
@@ -1,4 +1,9 @@
...
 pydantic
 torch>=2.10
-tensorboard
\ No newline at end of file
+tensorboard
+matplotlib
+xarray
+wandb
+dask
+natten
```
pzharrington, line 9: @coreyjadams FYI these added packages are taken care of by the `uv sync` command in the README, but I guess harmless to include them here too

coreyjadams: What's a README?  😆

PR #1565, opened 2026-04-13
🚀[FEA]: Add TensorRT compilation utility and hybrid Warp example

manmeet3591:
This PR adds a new inference optimization capability to PhysicsNeMo and a corresponding example for hybrid Warp + TensorRT execution.

### **New Features**
- **TensorRT Compilation Utility**: Added `physicsnemo/utils/inference.py` containing a `compile_to_trt` function. This utility wraps `torch_tensorrt` to simplify the process of optimizing PhysicsNeMo models for high-performance inference.
- **Hybrid Inference Example**: Added `examples/minimal/inference/torch_trt_warp_inference.py`. This example demonstrates how to integrate NVIDIA Warp (for geometric tasks like neighbor search) with a TensorRT-optimized neural network in a single, zero-copy pipeline.

### **Why this is useful**
Many Physics-AI models require complex geometric preprocessing (best handled by Warp) and high-speed neural network inference (best handled by TensorRT). This PR provides the necessary utilities and patterns to build such hybrid pipelines efficiently.

### DIFF
--- a/physicsnemo/mesh/README.md
+++ b/physicsnemo/mesh/README.md
@@ -422,7 +422,7 @@ neighbors of mesh elements (i.e., based on the mesh connectivity,as opposed to
 Note that these use an efficient sparse (`indices`, `offsets`) encoding of the
 adjacency relationships, which is used internally for all computations. (See the
 dedicated
-[`physicsnemo.mesh.neighbors._adjacency.py`](physicsnemo/mesh/neighbors/_adjacency.py)
+[`physicsnemo.mesh.neighbors._adjacency.py`](./neighbors/_adjacency.py)
 module.) You can convert these to a typical ragged list-of-lists representation
 with `.to_list()`, which is useful for debugging or interoperability, at the
 cost of performance:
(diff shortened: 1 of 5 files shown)


physicsnemo/mesh/README.md
```diff
@@ -422,7 +422,7 @@ neighbors of mesh elements (i.e., based on the mesh connectivity,as opposed to
 Note that these use an efficient sparse (`indices`, `offsets`) encoding of the
 adjacency relationships, which is used internally for all computations. (See the
 dedicated
-[`physicsnemo.mesh.neighbors._adjacency.py`](physicsnemo/mesh/neighbors/_adjacency.py)
+[`physicsnemo.mesh.neighbors._adjacency.py`](./neighbors/_adjacency.py)
```
peterdsharpe, line 425: Please remove these unrelated changes, or bring them in with a separate PR.

physicsnemo/nn/functional/neighbors/radius_search/_warp_impl.py
```diff
@@ -53,7 +53,7 @@ def count_neighbors(
     grid: wp.HashGrid,
     wp_points: wp.array(dtype=wp.vec3),
     wp_queries: wp.array(dtype=wp.vec3),
-    wp_launch_device: wp.context.Device | None,
+    wp_launch_device: wp.Device | None,
```
peterdsharpe, line 56: This is a good idea, and fixes a deprecation warning, but is unrelated to the main PR here - please bring these changes in a separate PR.

physicsnemo/utils/inference.py
```diff
@@ -0,0 +1,113 @@
...
+except ImportError:
+    _TORCH_TRT_AVAILABLE = False
+
+logger = logging.getLogger(__name__)
+
+
+def compile_to_trt(
+    model: torch.nn.Module,
+    input_signature: List[torch.Tensor],
+    enabled_precisions: Optional[Set[torch.dtype]] = None,
```
peterdsharpe, line 34:
```suggestion
    enabled_precisions: Set[torch.dtype] | None = None,
```
Modernizes type-hint syntax

physicsnemo/utils/inference.py
```diff
@@ -0,0 +1,113 @@
...
+    _TORCH_TRT_AVAILABLE = True
+except ImportError:
+    _TORCH_TRT_AVAILABLE = False
+
+logger = logging.getLogger(__name__)
+
+
+def compile_to_trt(
+    model: torch.nn.Module,
+    input_signature: List[torch.Tensor],
```
peterdsharpe, line 33:
```suggestion
    input_signature: list[torch.Tensor],
```

physicsnemo/utils/inference.py
```diff
@@ -0,0 +1,113 @@
...
+
+    try:
+        # Use torch.compile with tensorrt backend if using PyTorch 2.x style
+        # or fall back to torch_tensorrt.compile for explicit conversion.
+        # Here we prefer the explicit torch_tensorrt.compile for better control
+        # over the conversion process in static inference scenarios.
+        trt_model = torch_tensorrt.compile(model, **compile_spec)
+        logger.info("TensorRT compilation successful.")
+        return trt_model
+    except Exception as e:
```
peterdsharpe, line 100: I'd recommend allowing this to fail, rather than re-raising here. If we do choose to re-raise, perhaps we can narrow scope to tighter than a bare `Exception`?

physicsnemo/utils/inference.py
```diff
@@ -0,0 +1,113 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+from typing import Any, List, Optional, Union
+
+import torch
+
+try:
+    import torch_tensorrt
```
peterdsharpe, line 23: violates repo-wide optional import conventions; will trip up `importlinter`

peterdsharpe (changes requested):
Hi @manmeet3591,

Thank you for opening this PR! 

While this contribution is interesting, I'm not sure that it belongs in PhysicsNeMo-Core - it:
a) adds a new optional dependency of TensorRT that is used in this example, and we'd like to keep the dependency list for PhysicsNeMo minimal
b) the main compilation logic in `inference.py` is basically a thin wrapper around `torch_tensorrt.compile()`, which could be instead used directly by end-users in their downstream training utilities.

For example, by analogy: we don't include any PhysicsNeMo-specific wrappers for `torch.compile` support, yet many downstream users `torch.compile()` their PhysicsNeMo models in their own scripts.

Is there an angle of the value proposition here that I'm not fully understanding?

physicsnemo/utils/inference.py
manmeet3591: Yes, agree.

coreyjadams:
Hi @manmeet3591 I want to echo @peterdsharpe:

1. Thanks for opening this, I actually do think there is significant value here.  We've put a lot of work in physicsnemo to support warp kernels without graph breaks, for example.  TensorRT inference is a similar story and important for our models.
2. That said, I also can't understand quite where the "secret sauce" is here: I don't see any special treatment of warp in the tensorRT steps, so I'm confused what steps are necessary for enablement.  Is there something that, if you didn't have it here, would prevent this from working?

It's an interesting direction.  I think it's worth exploring, if we can see the benefit?  I am certainly grateful you've started this conversation.

PR #1566, opened 2026-04-13
Fix for partition function

pzharrington:
# PhysicsNeMo Pull Request

## Description
Fixes a bug related to `distribute_module` PyTorch<=2.10 that silently flips `requires_grad` on frozen parameters. Any non-sharded parameters should fall back to `Replicate` placement, but wrapping them in `nn.Parameter` should preserve `requires_grad`. This should not be needed in PyTorch>2.10 but  might as well include the fix here for robustness and clarity.

## Checklist

## Review Process


jleinonen (approved):

physicsnemo/nn/module/embedding_layers.py
```diff
@@ -127,7 +127,16 @@ def __init__(
         freqs = torch.arange(start=0, end=self.freq_embed_dim // 2, dtype=torch.float32)
         freqs = freqs / (self.freq_embed_dim // 2 - (1 if self.endpoint else 0))
         freqs = (1 / self.max_positions) ** freqs
-        self.register_buffer("freqs", freqs, persistent=False)
+        self.register_buffer("freqs", freqs)
+        self.register_load_state_dict_pre_hook(self._fill_missing_freqs)
+
+    @staticmethod
+    def _fill_missing_freqs(module, state_dict, prefix, *_args, **_kwargs):
+        """Backward compat: old checkpoints saved freqs as non-persistent.
```
CharlelieLrt, line 135: Just a FYI, we have both type of checkpoints I think:
- some with the buffer
- some without the buffer
It's because iirc the `persistent=False` was added only < 1 year ago, so checkpoints from before that had the buffer.

But anyways, your solution works for both I think, so approving

CharlelieLrt (approved):
LGTM!

physicsnemo/nn/module/embedding_layers.py
pzharrington: Actually the change came here: https://github.com/NVIDIA/physicsnemo/pull/1343

Previously the `freqs` were stored as a plain tensor (not in the state dict), so when they were refactored into a buffer we used `persistent=False` to avoid having to do a state dict hook like in this PR. Unfortunately we still need that now to support some checkpoints trained by NVR 😭

PR #1572, opened 2026-04-16
Geometry Functionals

loliverhennigh:
# PhysicsNeMo Pull Request

## Description
Two functionals for poisson disk sampling and mesh to voxel fraction.


docs/img/nn/functional/geometry/mesh_poisson_disk_sample/mesh_poisson_modes.gif
ktangsali, line 1: It is very hard to see the difference between the two techniques here. Rotation is making it even more difficult. Can we have a static plot? And then maybe zoom in a certain region to show the differences?

physicsnemo/nn/functional/geometry/sdf.py
ktangsali, line 42: Underlines are missing?

ktangsali (approved):
Some minor comments.

PR #1593, opened 2026-04-27
cuML KNN Compatibility Fix

peterdsharpe:
## Original Error

The failure was discovered through CUDA remeshing tests, but the failing code
path is the generic KNN cuML backend. Any CUDA caller that selected
`knn(..., implementation="cuml")` would hit the same issue.

The traceback failed here:

```text
physicsnemo/nn/functional/neighbors/knn/_cuml_impl.py:53: in knn_impl
    handle = cuml.Handle(stream=ptr)

AttributeError: module cuml has no attribute Handle
```

The failing tests were:

```text
test/mesh/remeshing/test_partition.py::TestDeviceCompat::test_assignments_and_area_conservation[cuda]
test/mesh/remeshing/test_partition.py::TestDeviceCompat::test_normals_on_surface[cuda]
```

## Observed Versions

This was observed in the current CUDA 13 environment:

```text
torch 2.11.0+cu130
cuml 26.04.000
cupy 14.0.1
```

The relevant API probe:

```text
has cuml.Handle: False
has cuml.neighbors.NearestNeighbors: True
```

Behavior on this environment:

| Implementation | cuML 26.04.000 / CuPy 14.0.1 |
|---|---|
| old `cuml.Handle(...)` + `NearestNeighbors(..., handle=...)` | fails with `AttributeError: module cuml has no attribute Handle` |
| new `NearestNeighbors(n_neighbors=k)` public API | passes |

## Root Cause

`knn(...)` auto-selects the cuML backend for CUDA tensors. The backend was
probing availability only by checking that `cuml>=24.0.0` and `cupy>=13.0.0`
could be imported. That check passed in this environment.

However, the installed RAPIDS/cuML package exposes the modern public estimator
API:

```python
cuml.neighbors.NearestNeighbors(n_neighbors=k)
```

and does **not** expose the older stream-handle API used by the implementation:

```python
cuml.Handle(stream=...)
cuml.neighbors.NearestNeighbors(..., handle=...)
```

So the backend was marked available, but then failed at runtime when it tried to
construct `cuml.Handle`.

## Chosen Fix

Use cuML's public `NearestNeighbors` constructor directly:

```python
knn = cuml.neighbors.NearestNeighbors(n_neighbors=k)
```

This matches the installed cuML API and leaves stream / handle management to
cuML. The tensors are still passed zero-copy through DLPack:

```python
points = cp.from_dlpack(points)
queries = cp.from_dlpack(queries)
```

Then the code continues as before:

```python
knn.fit(points)
distance, indices = knn.kneighbors(queries)
```

## Why This Is the Right Tradeoff

- It uses the supported public cuML estimator API instead of relying on an
  implementation-specific handle object.
- It keeps the backend selection behavior unchanged: CUDA tensors still prefer
  cuML when cuML is available.
- It avoids per-call feature probing or signature inspection in the KNN hot path.
- It fixes the observed RAPIDS/cuML 26.04.000 runtime failure while preserving the
  existing torch and scipy fallback implementations.

## Validation

The failing remeshing tests and the cuML KNN backend tests now pass:

```text
uv run --no-sync pytest \
  test/mesh/remeshing/test_partition.py::TestDeviceCompat \
  test/nn/functional/neighbors/test_knn.py::test_knn_cuml \
  test/nn/functional/neighbors/test_knn.py::test_knn_backend_forward_parity \
  test/nn/functional/neighbors/test_knn.py::test_knn_opcheck \
  -q

10 passed, 2 skipped
```


peterdsharpe:
Hi @loliverhennigh - submitting this for your review. To me, this makes sense, but in the spirit of [Chesterton's Fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence), there may be considerations at play here that I'm not aware of. Please feel free to shoot this one down if this breaks other stuff!

coreyjadams (changes requested):
Hi @peterdsharpe thanks for updating, it's frustrating when things like this shift from under us.

The handle interface from rapids was always a little clunky.  The motivation here as I'm sure you inferred is that we may not have any idea what stream the knn kernel is being submitted under, and it's likely not the default stream: in face the most common usecase IMO is in datapipes where it's deliberately in a non-default stream.

If dlpack really does this stream management under the hood, that's a great benefit, but we should explicitly be testing this.  Can we add a test in this PR to submit kNN kernels in non-default streams on CUDA devices to ensure they enque and sync properly?  If this isn't actually stream safe we will cause major headaches if anyone uses this in a multi-stream setting.

peterdsharpe:
Thanks @coreyjadams! Super fair - I will add a test.

peterdsharpe:
Adding another to-do from offline @coreyjadams discussion: We should also probably pin the versioning in pyproject.toml to whatever rapids introduced this or higher

peterdsharpe:
Update: added stream test + version pins.

coreyjadams:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1593/commits/e0e30f2202e7b94311672bd94e157ff671382fb6

loliverhennigh (approved):
LGTM, I dont think this should effect anything

coreyjadams (approved):
Thanks for adding the test, approving to unblock.

PR #1595, opened 2026-04-27
Synchronizes `GLOBE` model progress for 26.05

peterdsharpe:
# Synchronizes GLOBE model progress for 26.05

## Description

Syncs the in-flight GLOBE model work for the 26.05 release.  All model-class
changes land in `physicsnemo/experimental/`, where back-compatibility is not
guaranteed (workspace rule MOD-002a).

### GLOBE kernel: self-regularized Padé activation

Adds an optional `self_regularization_beta: float | None = None` parameter to
[`Pade.__init__`](physicsnemo/nn/module/pade.py) and plumbs it through
[`Kernel`, `BarnesHutKernel`, and `MultiscaleKernel`](physicsnemo/experimental/models/globe/field_kernel.py)
and [`GLOBE`](physicsnemo/experimental/models/globe/model.py).  When set, adds
a `beta * varphi_n^2` term to the denominator, bounding output magnitude
strictly below `1/beta`.  The term is `C^infinity` smooth and preserves the
`N == D` constant-asymptote property of the Padé form.  Default is `None`
(no behavior change).

### GLOBE latent-feature compression

In [`physicsnemo/experimental/models/globe/model.py`](physicsnemo/experimental/models/globe/model.py),
Phase 2 (communication-layer interactions) now compresses latent features via
arcsinh-based smooth log compression.  This prevents the communication-layer
amplification loop from producing `O(10^4)` intermediate values.  Phase 3
(prediction-point evaluation) is now chunked.

### GLOBE non-regression tests

New [`test/models/globe/test_nonregression.py`](test/models/globe/test_nonregression.py)
plus reference data at `test/models/globe/data/globe_nonregression_output.pth`.
Locks full-forward-pass numerics against any silent drift in indexing,
gather/scatter, or feature engineering.  Reference is regenerated on first
run and compared on subsequent runs.

### `physicsnemo.experimental.utils` package

New utility package, broken out from the GLOBE-specific tree:

- [`compile_logging.py`](physicsnemo/experimental/utils/compile_logging.py):
  `CompileDiagnosticsCollector` (a `logging.Filter` that captures
  `torch.compile` graph-break and recompile events as compact summaries) and
  `disable_autotune_printing`.
- [`prefetch.py`](physicsnemo/experimental/utils/prefetch.py): `prefetch_map`,
  a one-step lookahead iterator wrapper that preserves `__len__` so `tqdm`
  progress bars still work.
- [`cached_dataset.py`](physicsnemo/experimental/utils/cached_dataset.py):
  `CachedPreprocessingDataset` migrated from
  `physicsnemo/experimental/models/globe/utilities/cached_dataset.py` (94%
  similarity rename) so it is no longer GLOBE-specific.

### AirFRANS example sync

Refresh of [`examples/cfd/external_aerodynamics/globe/airfrans/`](examples/cfd/external_aerodynamics/globe/airfrans/):

- `dataset.py`: `AirFRANSSample` switched from `@tensorclass` decorator plus a
  `TYPE_CHECKING` `to`-method stub to `class AirFRANSSample(TensorClass):`.
  Runtime behavior is identical; the change exposes every auto-generated
  tensorclass method (`to`, `clone`, `__getitem__`, `cuda`, `cpu`, `detach`,
  ...) to static type-checkers via `tensordict`'s shipped `tensorclass.pyi`,
  with no per-method stubs needed.  Also adds `compute_surface_force_coefficients`
  and a unified `postprocess` / `visualize_comparison` pair.
- `train.py`: scheduler-patience normalization, first-launch diagnostics
  (auto-disabled after the first batch), explicit `prepare_sample` helper,
  periodic surface-force-coefficient logging, and a `save_ckpt` helper.
- Minor: `inference.py`, `mlflow_launch.sh`, `requirements.txt`, `run.sh`,
  `utilities.py`.

### BVH refactor

In [`physicsnemo/mesh/spatial/bvh.py`](physicsnemo/mesh/spatial/bvh.py),
`find_candidate_cells` is split into `_validate_query_points` and
`_traverse` (+62 / -49, no behavior change).

### CODEOWNERS

Note: lots of CODEOWNERS are getting pulled in since I'm actually not an owner on `physicsnemo/nn/module/pade.py` - this PR also fixes that.

Self-add as a codeowner for [`physicsnemo/nn/module/pade.py`](physicsnemo/nn/module/pade.py).

## Checklist

## Dependencies

No changes to top-level `pyproject.toml`.  The AirFRANS example's local
[`requirements.txt`](examples/cfd/external_aerodynamics/globe/airfrans/requirements.txt)
adds `tenacity~=9.1.4`, used in
[`utilities.py`](examples/cfd/external_aerodynamics/globe/airfrans/utilities.py)
to retry transient MLflow logging failures.


loliverhennigh (approved):
Looked at the pade.py part. Looks good to me

examples/cfd/external_aerodynamics/globe/airfrans/train.py
```diff
@@ -620,26 +688,55 @@ def checkpoint_metadata() -> dict[str, Any]:
...
+                        )
+                        if use_mlflow:
+                            log_metrics(
+                                {
+                                    f"force_coeffs/{split}_{k}_{src}": coeffs[k]
+                                    for src, coeffs in [
+                                        ("pred", pred_coeffs),
+                                        ("true", true_coeffs),
+                                    ]
+                                    for k in pred_coeffs
```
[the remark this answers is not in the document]
peterdsharpe: False positive (greptile is misreading indents)

physicsnemo/experimental/utils/cached_dataset.py
```diff
@@ -51,6 +52,11 @@ def __init__(
         self.cache_dir = Path(cache_dir) if cache_dir is not None else None
         self.use_ram_caching = use_ram_caching
 
+        if not self.sample_paths:
+            raise ValueError(
+                "sample_paths is empty; the dataset must contain at least one sample."
+            )
+
         if self.cache_dir is not None:
             self.cache_dir.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
peterdsharpe: Fixed by using lru_cache with bound

physicsnemo/nn/module/pade.py
```diff
@@ -75,6 +75,14 @@ class Pade(Module):
     use_separate_mlps : bool, optional, default=True
         If ``True``, uses separate MLPs for numerator and denominator.
         If ``False``, uses a single MLP with split outputs.
+    self_regularization_beta : float or None, optional, default=None
+        If not ``None``, adds :math:`\beta \varphi_n^2` to the denominator,
+        bounding the output magnitude strictly below :math:`1/\beta`.
+        This prevents unbounded growth when the numerator MLP produces
+        large outputs while the denominator is small. The term
+        :math:`\varphi_n^2` is :math:`C^\infty` smooth and preserves
+        the :math:`N = D` far-field constant-asymptote property.
```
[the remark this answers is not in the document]
peterdsharpe: Fixed in docstring

physicsnemo/nn/module/pade.py
```diff
@@ -138,6 +158,7 @@ def __init__(
         self.denominator_order = denominator_order
         self.share_denominator_across_channels = share_denominator_across_channels
         self.use_separate_mlps = use_separate_mlps
+        self.self_regularization_beta = self_regularization_beta
```
coreyjadams, line 167: Consider a guard, here, on beta > 0?

If beta < 0, since raw_numerator^2 >= 0 you will be strictly adding a negative term to the denominator.  There are cases this could take the denominator to 0, when in fact beta is meant to do the opposite I think.

physicsnemo/experimental/models/globe/model.py
```diff
@@ -566,6 +574,21 @@ def _evaluate_communication_hyperlayer(
...
+            ### Compress latent features to prevent the communication-layer
+            ### amplification loop from producing O(10^4) intermediate values.
+            ### Uses arcsinh for C-infinity smooth logarithmic compression.
+            if self.latent_compression_scale is not None:
+                C = self.latent_compression_scale
+                eps_sq = self.smoothing_radius**2
+
+                def _compress(t: torch.Tensor) -> torch.Tensor:
+                    if t.ndim == 1:
+                        return C * torch.arcsinh(t / C)
```
coreyjadams, line 611: Considering "0.0" is sometimes a valid standin for "None", but in this case would be a disaster, a guard against that would be good.

physicsnemo/experimental/utils/compile_logging.py
```diff
@@ -0,0 +1,205 @@
...
+        collector = CompileDiagnosticsCollector()
+        collector.install()
+        # ... run one training batch ...
+        torch._logging.set_logs(graph_breaks=False, recompiles=False)
+        collector.uninstall()
+        print(collector.summary())
+        # For full stack traces when investigating:
+        # print(collector.detailed_summary())
+    """
+
```
coreyjadams: Can you elaborate a little on what prompted the development of this class, in GLOBE?

coreyjadams: What I'm getting at here is I'm not convinced that stifling complie warnings is generally a useful thing we should be doing; I worry it will lead to more "be quiet dynamo" and less problem fixing.

physicsnemo/experimental/utils/prefetch.py
```diff
@@ -0,0 +1,81 @@
...
+        return len(self._iterable)  # type: ignore[arg-type]
+
+
+def prefetch_map(iterable: Iterable[_T], fn: Callable[[_T], _U]) -> _PrefetchMap[_U]:
+    """Apply *fn* to each element, overlapping ``fn(next)`` with consumption of current.
+
+    Submits ``fn(element)`` to a single background thread one step ahead
+    of the consumer.  Tensor operations release the GIL during C++/CUDA
+    execution, so CPU-bound preparation of the next sample overlaps with
+    GPU-bound processing of the current sample.
```
coreyjadams: This could be misleading to users: Thread pool executor doesn't actually release the GIL, and Tensor ops may or may not release the GIL, some can actually require a sync and be blocking which would stall your thread pool.  We don't need to fix it in the code here, but it's worth updating the docstring to emphasize that this really only works on truly async code that releases the GIL, and the prefetch map itself is not a GIL-workaround.

physicsnemo/experimental/utils/cached_dataset.py
```diff
@@ -24,32 +23,45 @@
 
 
 class CachedPreprocessingDataset(Dataset, ABC):
```
coreyjadams, line 25: Next release (not here) I think we could actually have a path to upstream this functionality to core: now that mainline datapipes is reproducible, preprocessing datapipes can even work on subsampling and similar pipelines.  So we might be able to do this.

No action needed.

coreyjadams (commented):
I have reviewed all the code in the physicsnemo space.  Moving on to examples but dropping this batch of comments first.

examples/cfd/external_aerodynamics/globe/airfrans/train.py
```diff
@@ -73,27 +88,33 @@ def main(
...
         "default", "max-autotune-no-cudagraphs"
     ] = "max-autotune-no-cudagraphs",
-    points_per_iter: int = 2048,
+    n_prediction_points: int = 2048,
     learning_rate: float = 1e-3,
     weight_decay: float = 1e-4,
     use_muon: bool = True,
-    muon_method: Literal["original", "match_rms_adamw"] = "original",
-    train_face_downsampling_ratio: float = 1.0,
+    muon_method: Literal["original", "match_rms_adamw"] = "match_rms_adamw",
```
coreyjadams, line 95: What prompted this change to the optimizer?

examples/cfd/external_aerodynamics/globe/airfrans/train.py
```diff
@@ -553,22 +637,14 @@ def checkpoint_metadata() -> dict[str, Any]:
 
             scheduler.step(loss["train"])
 
-            if profiler is not None:
-                profiler.step()
+            if dist.world_size > 1:
+                barrier()
```
coreyjadams, line 636: Do you need a barrier here?

examples/cfd/external_aerodynamics/globe/airfrans/train.py
```diff
@@ -620,26 +688,55 @@ def checkpoint_metadata() -> dict[str, Any]:
...
+                                        ("true", true_coeffs),
+                                    ]
+                                    for k in pred_coeffs
+                                },
+                                step=epoch,
+                            )
+
                 last_image_epoch, last_image_loss = epoch, loss["train"]
+                if dist.world_size > 1:
+                    barrier()
```
coreyjadams, line 734: Same - you you need a barrier?

examples/cfd/external_aerodynamics/globe/airfrans/dataset.py
```diff
@@ -155,11 +118,23 @@ def make_dataloader(
...
+                (the default), auto-computed as ``floor(n_cpus / n_gpus)``
+                to fully overlap data loading with GPU training.
 
         Returns:
             Configured DataLoader with distributed sampling.
         """
+        if num_workers is None:
+            n_cpus = os.cpu_count() or 1
+            n_gpus = max(1, torch.cuda.device_count())
+            num_workers = n_cpus // n_gpus
```
coreyjadams: Without dedicated attention to thread affinity, this likely will not do what you think it does.  Additionally, even in a perfect world, its usually ideal to leave CPU 0 per rank open to orchestrate kernel launches, OS needs, thread scheduling, and any other jobs.  

typically to get good CPU side concurrency its necessary to pin a thread to a CPU which requires spawning that as a process, not a thread, in python.  Check out `psutil` (only on linux, really) which lets you do this via affinity + subprocess spawning.  It's not what you want here anyways.

Plus, it's probably good to verify the `os.cpu_count()` return, is it hardware cores or virtual threads?  

tl;dr I recommend you back off and leave a dedicated hardware core, per rank, open for model orchestration and OS tasks.

coreyjadams (approved):

coreyjadams:
Arg I clicked the "Approve" button with no comments.  Here's what I meant to say.

Overall I think this PR looks good.  I left lots of small comments, I think there are a couple places where some input sanitation might save adventurous users from footguns, but otherwise it's in good shape.  Well done!

I don't have any hard changes requested, I asked a couple questions for my own interest here, but overall I think this is about ready to go!

physicsnemo/experimental/utils/compile_logging.py
peterdsharpe: It's a valid concern! The motivation here is that graph breaks were emitting like 50-200 lines of code each, and recompiles were similar. And that's per-rank - so in a training run with, say, 1 graph break and 10 recompiles, and with 16 ranks, you could be looking at thousands or tens of thousands of lines of code in the logfile.

At that volume, it's not very actionable. This change doesn't suppress/stifle the compile warnings, so much as summarize them - each graph break or recompile gets one line that's roughly `file:lineno   reason   <- caller`. 

That said, this is definitely an *optional* bell-and-whistle - more a "nice to have" than a must-merge thing, if you'd rather keep this separate.

coreyjadams: IMO the Right :tm: move here is maybe not to block compile logs.  You could quite easily just route non-error logs from Rank >0 into the trash?

physicsnemo/nn/module/pade.py
peterdsharpe: Done in the next push - added a constructor guard:

```python
if self_regularization_beta is not None and self_regularization_beta <= 0:
    raise ValueError(
        f"self_regularization_beta must be positive (use None to disable), "
        f"got {self_regularization_beta=!r}"
    )
```

Also mirrored the same guard at the GLOBE level (rather than only at `Pade`) so the error points at the user-facing API surface.

physicsnemo/experimental/utils/compile_logging.py
peterdsharpe: Hmmm let me give that a shot! I seem to recall some months back having some issues due to `torchrun` rank handling with that, but perhaps there's a way here

physicsnemo/experimental/utils/cached_dataset.py
peterdsharpe: Sounds good! Yeah, to be honest this whole AirFRANS demo will likely get superseded long-term by the unified external aero datapipe, so I'm not too worried here

examples/cfd/external_aerodynamics/globe/airfrans/train.py
peterdsharpe: Seems like the [research consensus is that this "match_rms_adamw" is better-practice, as it results in optimal learning rates that are similar between Muon and Adam-based optimizers](https://arxiv.org/pdf/2502.16982) - the "original" is really just the default for historical reasons. I tested this new setting vs. the old one on AirFRANS on HSG, and I'm seeing very similar loss curves, so figured I'd baseline to the "better-practice" one.

physicsnemo/experimental/models/globe/model.py
peterdsharpe: Fixed

examples/cfd/external_aerodynamics/globe/airfrans/train.py
peterdsharpe: Strictly speaking no - the `all_reduce` inside run_epoch 1 already synchronizes ranks. Kept as defensive belt-and-suspenders so the rank-0 checkpoint and best-model save that follow happen with all ranks demonstrably aligned. Happy to drop it if you want the leaner version.

examples/cfd/external_aerodynamics/globe/airfrans/train.py
peterdsharpe: This one earns its keep. The rank 0 visualization branch does heavy work (forward at viz resolution, matplotlib render, MLFlow upload) - it can take tens of seconds. Without the barrier, ranks 1..N race ahead into the next epoch's data loading and the next all_reduce then waits on rank-0 mid-epoch, which is harder to reason about than an explicit barrier here. And, in pathological cases where the viz errors, this makes debugging or recovery messier.

physicsnemo/experimental/utils/compile_logging.py
peterdsharpe: Did the rank-aware silencer approach. Replaced `CompileDiagnosticsCollector` with `silence_compile_logs_on_non_zero_ranks(rank)`.

`train.py`'s first-launch block now just sets `torch._logging.set_logs(graph_breaks=True, recompiles=True)` on rank 0 for the first batch and prints raw Dynamo output, which is what you wanted.

Function is in `physicsnemo.experimental.utils` so other recipes can pick it up with a one-line import.

physicsnemo/experimental/utils/prefetch.py
peterdsharpe: fixed

examples/cfd/external_aerodynamics/globe/airfrans/dataset.py
peterdsharpe: Added a much cleaner fix :)

examples/cfd/external_aerodynamics/globe/airfrans/train.py
coreyjadams: It's not a big deal, and it's already merged anyways, but if visualization is that expensive you could consider dumping that necessary data into a CPU subprocess instead of a distributed barrier, so that you don't idle all CPUs while rank-0 writes images?

PR #1602, opened 2026-04-28
Batched ball query

coreyjadams:
# PhysicsNeMo Pull Request

This PR is long overdue, honestly.

This adds batching to the radius search and ball query.  For models like DoMINO and GeoTransolver with local features, this enables us to run with BatchSize > 1.  In external aero use cases this isn't such a big deal, since we have always operated at the limit of GPU memory anyways with subsampling.  For other domains, however, the point clouds are not so large and batching is a need.

### Torch

The pytorch implementation is easy: batching with the brute force method is trivial.

### Warp

Radius search with warp has two distinct code paths, which have unique computational performance profiles and have to be treated separately.  In both cases, we have to perform a loop over the batch at the start to construct a hash map per batch item: this is unavoidable and in my opinion there isn't any low hanging way to do better with a single hash map.

Once each batch item B has a hash map, the code diverges into two branches based on the value of `max_points`.  When `max_points` is None, the algorithm must first count how many points will be in the output space for every query, sum this, and allocate the appropriate output memory.  Without GPU-side allocation, we have to perform a CPU-GPU sync here.  In the batched implementation I deliberately sync all kernels at once, to only perform one block per call.  After, the second kernel launches to fill in the selected points, per batch and per query.  This code path is mostly unused in the repo, only in FigConvNet and I think it still has it's own funky implementation.

The other path, more important and used by Domino and GeoT, sets `max_points=N`.  This will launch a 2D warp kernel now (instead of 1D) where the first dimension is the batch and indexes which hashmap index the kernel will use.  Otherwise, it's largely unchanged.

## Tests

Tests are updated to verify that batching works, especially in models like GeoT with local features enabled.

## Domain Parallelism

Domain parallelism doesn't _really_ target batch size > 1, however, I have made sure to update the wrapper to support this and ensure it still works.

## Checklist

## Review Process


coreyjadams:
@loliverhennigh Is it possible to test this PR with your ASV benchmarks and check it for regressions before merge?

physicsnemo/nn/functional/neighbors/radius_search/_warp_impl.py
```diff
@@ -49,6 +46,24 @@
 BLOCK_DIM = 32
 
 
+def _validate_inputs(points: torch.Tensor, queries: torch.Tensor):
```
loliverhennigh: Not super important but you can put this in the utils.py as its a repeat of whats in _torch_impl.py

loliverhennigh (approved):
Minor comment but went through everything else and looks good to me.

peterdsharpe (approved):
Partial review on `interrogate_baseline.txt` only, where I am codeowner - approving to unblock

RishikeshRanade (approved):
LGTM

physicsnemo/nn/functional/neighbors/radius_search/_warp_impl.py
coreyjadams: Done!

ktangsali (approved):
LGTM!

PR #1608, opened 2026-04-30
mesh: enable ShardTensor support for mesh conversion/geometry paths

loliverhennigh:
# PhysicsNeMo Pull Request


physicsnemo/domain_parallel/__init__.py
```diff
@@ -47,7 +47,7 @@
     # In minumum versions are met, we can import the shard tensor and spec.
 
     from ._shard_tensor_spec import ShardTensorSpec
-    from .shard_tensor import ShardTensor, scatter_tensor
+    from .shard_tensor import ShardTensor, replicated_zeros_like, scatter_tensor
```
loliverhennigh: Hey @coreyjadams, if I am adding the zeros_like correct I might add a few similar ops just consistency even though they are not needed.

loliverhennigh:
Blaa, some stuff got messed up and not quite ready for review. Ill fix tonight...

physicsnemo/mesh/utilities/_scatter_ops.py
```diff
@@ -24,9 +24,28 @@
 import torch
 from jaxtyping import Float, Int
 
+from physicsnemo.domain_parallel import replicated_zeros_like
 from physicsnemo.mesh.utilities._tolerances import safe_eps
 
 
+def _is_sharded_tensor(tensor: torch.Tensor) -> bool:
+    return hasattr(tensor, "_spec") and hasattr(type(tensor), "from_local")
```
peterdsharpe: @coreyjadams is this best-practices for type-narrowing ShardTensor?

Feels like `isinstance(tensor, ShardTensor)` would be better, unless there's something I'm missing?

(And if this does get replaced with `isinstance`, then this can be inlined rather than keeping a separate function.)

test/mesh/mesh/test_geometry_properties.py
```diff
@@ -35,11 +43,110 @@
...
+        pytest.skip("ShardTensor mesh geometry tests currently run on CPU tensors")
+    return ShardTensor.from_local(
+        tensor,
+        mesh_shard_device_mesh,
+        [placement],
+        sharding_shapes="infer",
+    )
+
+
+def _mesh_to_mode(mesh: Mesh, *, mesh_tensor_mode: str, mesh_shard_device_mesh) -> Mesh:
```
peterdsharpe: Docstrings needed (here and in above functions)

test/mesh/mesh/test_geometry_properties.py
```diff
@@ -35,11 +43,110 @@
...
+)
+_MESH_TENSOR_MODES = (
+    ["dense", "shard_replicate", "shard_sharded"]
+    if _ENABLE_SHARD_MESH_TEST_MODES
+    else ["dense"]
+)
+
+
+@pytest.fixture(params=_MESH_TENSOR_MODES)
+def mesh_tensor_mode(request):
```
peterdsharpe: Type hints?

test/mesh/mesh/test_geometry_properties.py
```diff
@@ -35,11 +43,110 @@
...
+        )
+        try:
+            yield
+        finally:
+            if dist.is_initialized():
+                dist.destroy_process_group()
+
+
+@pytest.fixture
+def mesh_shard_device_mesh(mesh_tensor_mode, _single_rank_dist_group):
```
peterdsharpe: Type hints + docstrings needed

test/mesh/mesh/test_data_conversion.py
```diff
@@ -20,11 +20,211 @@
...
+
+
+def _convert_data_for_mode(
+    data: dict[str, torch.Tensor] | None,
+    *,
+    n_points: int,
+    n_cells: int,
+    point_placement: Replicate | Shard,
+    cell_placement: Replicate | Shard,
+):
```
peterdsharpe: docstring needed

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -965,6 +997,71 @@ def backward(self, *args, **kwargs):
...
+def replicated_zeros_like(
+    tensor: torch.Tensor,
+    shape: Sequence[int] | torch.Size,
+    *,
+    dtype: torch.dtype | None = None,
+) -> torch.Tensor:
+    r"""Create zeros matching a tensor's device/mesh semantics.
+
+    For ``ShardTensor`` inputs this returns a replicated ``ShardTensor`` on the
+    same mesh. For regular tensors this falls back to ``torch.zeros`` on the
```
peterdsharpe: Given that this is part of the public API, please add complete NumPy-style docstring

physicsnemo/mesh/mesh.py
```diff
@@ -1463,15 +1463,14 @@ def cell_data_to_point_data(self, overwrite_keys: bool = False) -> "Mesh":
...
-        ).repeat_interleave(n_vertices_per_cell)
+        # Repeat each cell value once per incident vertex.
+        # This avoids advanced indexing with an explicit cell-index tensor.
 
         converted = self.cell_data.apply(
             lambda cell_values: scatter_aggregate(
-                src_data=cell_values[cell_indices],
+                src_data=cell_values.unsqueeze(1)
+                .expand(-1, n_vertices_per_cell, *cell_values.shape[1:])
+                .reshape(-1, *cell_values.shape[1:]),
```
peterdsharpe, line 1495: Why is `Mesh.cell_data_to_point_data()` updated, but `Mesh.point_data_to_cell_data()` is not? The asymmetry seems suspect - seems like they should either both need updates for ShardTensor or neither should, but I might be missing something?

Or is this just an unrelated change (and if so, what motivated this)?

test/mesh/mesh/test_data_conversion.py
```diff
@@ -20,11 +20,211 @@
...
+    return ShardTensor.from_local(
+        tensor,
+        _ACTIVE_MESH_DEVICE_MESH,
+        [placement],
+        sharding_shapes="infer",
+    )
+
+
+def _convert_data_for_mode(
+    data: dict[str, torch.Tensor] | None,
```
peterdsharpe: This function incorrectly:
a) accepts `dict[str, torch.Tensor]` instead of `TensorDict`
b) iterates only over top-level keys.

Combined with downstream uses that pass in `TensorDict`, this causes issues in hierarchical dictionaries. (Specifically, this will cause nested values to remain as `torch.Tensor`.)

Consider using `TensorDict.apply()` instead.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -908,6 +908,38 @@ def to_local(
 
         return _ToTorchTensor.apply(self, grad_placements)
 
+    def new_replicated_zeros(
```
coreyjadams: To add this function as an API call on shard tensor, vs supporting the underlying dispatch call, has to have a really good motivation.

What is the value of this vs. supporting the backend of `torch.zeros_like(a)` when a is a shard tensor?  and, in fact, I think that should already work?

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -908,6 +908,38 @@ def to_local(
...
+        ShardTensor
+            A replicated ShardTensor of zeros on the same mesh.
+        """
+        out_dtype = self.dtype if dtype is None else dtype
+        local = torch.zeros(tuple(shape), dtype=out_dtype, device=self.device)
+        return ShardTensor.from_local(
+            local,
+            self._spec.mesh,
+            [Replicate() for _ in range(self._spec.mesh.ndim)],
+            sharding_shapes="infer",
```
coreyjadams: This is unusual here: typically I'd object loudly to passing "infer" as sharding shapes since that can trigger a blocking allreduce and is a major perf headache.  but you're replicating on all ranks.  I don't think I understand this function's role, really.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -965,6 +997,71 @@ def backward(self, *args, **kwargs):
         return self.to_local().backward(*args, **kwargs)
 
 
+def replicated_zeros_like(
+    tensor: torch.Tensor,
+    shape: Sequence[int] | torch.Size,
+    *,
+    dtype: torch.dtype | None = None,
+) -> torch.Tensor:
```
coreyjadams: I still think it makes more sense to support `torch.zeros_like(a)` for a as a replicated tensor.

tl;dr the ShardTensor design philosophy is to make zero code changes on user side, whenever possible, so we support torch calls on shard tensors first and foremost rather then introduce new API.  Is it possible to implement your work without this?

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -965,6 +997,71 @@ def backward(self, *args, **kwargs):
...
+    out_dtype = tensor.dtype if dtype is None else dtype
+    return torch.zeros(tuple(shape), dtype=out_dtype, device=tensor.device)
+
+
+def _cross_wrapper(func, types, args, kwargs):
+    if kwargs is None:
+        kwargs = {}
+
+    if kwargs.get("out", None) is not None:
+        raise RuntimeError("torch.linalg.cross(out=...) is not supported for ShardTensor.")
```
coreyjadams: If this is a function overload, it's in the wrong place.  `shard_tensor.py` is for the core tensor object only.  There is a sub folder for ops.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -965,6 +997,71 @@ def backward(self, *args, **kwargs):
...
+    dim = kwargs.get("dim", -1)
+    if len(args) > 2:
+        dim = args[2]
+
+    if not isinstance(input_tensor, ShardTensor) or not isinstance(
+        other_tensor, ShardTensor
+    ):
+        raise RuntimeError(
+            "torch.linalg.cross with ShardTensor inputs requires both arguments to be ShardTensor."
+        )
```
coreyjadams: It looks like what we want to be doing is implementing a wrapper for `torch.linalg.cross` on shard tensor objects.  There is not a need to check that all objects are ShardTensor at this time.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -965,6 +997,71 @@ def backward(self, *args, **kwargs):
...
+    if input_tensor._spec.placements != other_tensor._spec.placements:
+        raise RuntimeError(
+            "torch.linalg.cross requires both ShardTensor inputs to have identical placements."
+        )
+
+    local_result = torch.linalg.cross(
+        input_tensor.to_local(),
+        other_tensor.to_local(),
+        dim=dim,
+    )
```
coreyjadams: This appears to assume there aren't funky placements.  But does this work if input and other are both sharded?  Both replicated?  What about the tensor_dim of sharding?  Any constraints on sharding shapes?  Is cross purely an op on the `dim` channel, and changes the `dim` channel?

physicsnemo/mesh/utilities/_scatter_ops.py
```diff
@@ -93,7 +112,9 @@ def scatter_aggregate(
 
     ### Fast path: unweighted sum is a single scatter_add_ with no extra work
     if weights is None and aggregation == "sum":
-        aggregated_data = torch.zeros((n_dst, *data_shape), dtype=dtype, device=device)
+        aggregated_data = _replicated_zeros_like(
+            src_data, (n_dst, *data_shape), dtype
+        )
```
coreyjadams: The distributed-friendly update to make here is, if mesh can support it, to port `torch.zeros` away from this shape and onto `torch.zeros_like`.  So whatever is building data_shape as input, we build `zeros_like(that_object)` and then mesh can work on single device and sharded inputs too.

physicsnemo/mesh/utilities/_scatter_ops.py
```diff
@@ -102,7 +123,10 @@ def scatter_aggregate(
 
     ### Initialize weights if not provided
     if weights is None:
-        weights = torch.ones(len(src_to_dst_mapping), dtype=dtype, device=device)
+        if _is_sharded_tensor(src_data):
+            weights = torch.ones_like(src_to_dst_mapping, dtype=dtype)
+        else:
+            weights = torch.ones(len(src_to_dst_mapping), dtype=dtype, device=device)
```
coreyjadams: This is much closer to the "right" way for domain parallelism, but in fact we can probably consolidate to just "ones_like" for all paths.

physicsnemo/mesh/utilities/_scatter_ops.py
```diff
@@ -114,11 +138,7 @@ def scatter_aggregate(
     weighted_data = src_data * weights.view(weight_shape)
 
     ### Scatter-add weighted data to destinations
-    aggregated_data = torch.zeros(
-        (n_dst, *data_shape),
-        dtype=dtype,
-        device=device,
-    )
+    aggregated_data = _replicated_zeros_like(src_data, (n_dst, *data_shape), dtype)
```
coreyjadams: Same here re:zeros_like

physicsnemo/mesh/utilities/_scatter_ops.py
```diff
@@ -134,7 +154,7 @@ def scatter_aggregate(
     ### Normalize weighted sum to weighted mean
     if aggregation == "mean":
         ### Compute sum of weights at each destination
-        weight_sums = torch.zeros(n_dst, dtype=dtype, device=device)
+        weight_sums = _replicated_zeros_like(src_data, (n_dst,), dtype)
```
coreyjadams: Same here re:zeros_like

physicsnemo/mesh/mesh.py
```diff
@@ -1463,15 +1463,14 @@ def cell_data_to_point_data(self, overwrite_keys: bool = False) -> "Mesh":
...
-        # Shape: (n_cells * n_vertices_per_cell,)
-        cell_indices = torch.arange(
-            self.n_cells, device=self.points.device
-        ).repeat_interleave(n_vertices_per_cell)
+        # Repeat each cell value once per incident vertex.
+        # This avoids advanced indexing with an explicit cell-index tensor.
 
         converted = self.cell_data.apply(
             lambda cell_values: scatter_aggregate(
-                src_data=cell_values[cell_indices],
```
coreyjadams, line 1499: Indexing a ShardTensor with a ShardTensor index should work fine?

coreyjadams (changes requested):
Hi @loliverhennigh - I have left a number of comments on the PR.  I think there are some updates to make, though I'm happy to see the changes overall are not too extensive.  I think there are some design choices that cut against the philosophy of ShardTensor we should tweak, but it doesn't look likes going to be too much.

Happy to discuss this with as much detail as you'd like offline!

physicsnemo/mesh/mesh.py
loliverhennigh: `cell_indices` is not a ShardTensor the way it was before. I was coming from just the `torch.arange` function

physicsnemo/mesh/mesh.py
loliverhennigh: Its because we needed a shardTensor to index hence the change. In the point_data_to_cell_data we use cells to index but that is already a ShardTensor.

loliverhennigh:
\blossom-ci

loliverhennigh:
\blossom-ci

loliverhennigh:
Hi @coreyjadams - quick follow-up on this PR. The latest GitHub Actions checks are green, and I re-triggered Blossom with `\blossom-ci`, but I still do not see a Blossom status/check showing up on the PR. Could you take a look when you have a chance, or let me know if there is another validation step I should trigger?

physicsnemo/mesh/utilities/_scatter_ops.py
```diff
@@ -93,7 +92,7 @@ def scatter_aggregate(
 
     ### Fast path: unweighted sum is a single scatter_add_ with no extra work
     if weights is None and aggregation == "sum":
-        aggregated_data = torch.zeros((n_dst, *data_shape), dtype=dtype, device=device)
+        aggregated_data = src_data.new_zeros((n_dst, *data_shape), dtype=dtype)
```
peterdsharpe, line 95: @coreyjadams out of curiosity, are these changes to `_scatter_ops.py` required for ShardTensor compat?

From a Mesh side, these all look find to me, and I can imagine where they come from (allowing duck typing where a `ShardTensor` creates another `ShardTensor`) - but my understanding was that `ShardTensors` would safely "type-promote" here. In other words, no changes should be needed to downstream code.

But like I said, if this *is* needed for ShardTensor, then these changes are fine with me too.

test/mesh/mesh/test_geometry_properties.py
```diff
@@ -26,20 +26,91 @@
...
 
 ATOL = 1e-4
 RTOL = 1e-4
-
+_SHARD_MESH_TENSOR_MODES = ("shard_replicate", "shard_sharded")
 
 ### Helper Functions ###
 
 
+def _to_dense_tensor(tensor: torch.Tensor) -> torch.Tensor:
```
peterdsharpe, line 46: Type hint is inaccurate here? You're expecting a ShardTensor input optionally, too.

By the way, is this function needed? Can't we `torch.Tensor(tensor)` cast at point-of-use instead? This would be more readable and equally verbose

test/mesh/mesh/test_geometry_properties.py
```diff
@@ -26,20 +26,91 @@
...
+
+def _assert_allclose(a: torch.Tensor, b: torch.Tensor, **kwargs) -> None:
+    """Assert equality after materializing any distributed tensor inputs."""
+    assert torch.allclose(_to_dense_tensor(a), _to_dense_tensor(b), **kwargs)
+
+
+def _assert_shard_tensor(tensor: torch.Tensor) -> None:
+    """Assert that a tensor is backed by ShardTensor."""
+    assert ShardTensor is not None
+    assert isinstance(tensor, ShardTensor)
```
peterdsharpe: Is this needed? Seems like this should be inlined?

physicsnemo/mesh/utilities/_scatter_ops.py
loliverhennigh: Done.

test/mesh/mesh/test_geometry_properties.py
loliverhennigh: Done.

test/mesh/mesh/test_geometry_properties.py
loliverhennigh: Done.

loliverhennigh:
\blossom-ci

physicsnemo/mesh/utilities/_scatter_ops.py
loliverhennigh: I think this is needed as-is. The old `torch.zeros(...)` allocation only had shape/dtype/device inputs, so there was no ShardTensor operand for dispatch/type-promotion to preserve placement from. Using `src_data.new_zeros(...)` makes the allocation tensor-derived, so it preserves the ShardTensor behavior while keeping the dense path equivalent.

coreyjadams:
I took a look at trying to update the tests to align them with the domain parallelism tests; the test coverage here was only ~60 tests, all on CPU, for domain parallelism.  It isn't aligned with the torchrun syntax that domain parallel tests expect (look at /test/plugins/distributed_fixtures.py for more info).

The cross operation looks like it's pretty much implemented OK but the tests aren't quite aligned with the way the op testing is done in the domain parallelism suite, so we're falling short there too.

I just don't think there is sufficient test coverage to say if we've achieved domain parallelism on PhysicsNeMo mesh.  I don't think this is ready to go for the RC.  There are ~2000 tests for mesh, and while we don't need all of them, being able to test against most of the core functionality + things we would want in datapipes, models, preprocessing etc for domain parallelism is, I think, a prerequisite for merge here.  60 tests (3% coverage max, if we're generous) all on CPU I think is probably short?

@peterdsharpe Any suggestions of what core mesh functionality should be prioritized for testing in a distributed mode?  I'm thinking we need generic mesh operations (slicing, scatter_ops, vertex_to_cell and the opposite) as well as manifold projection operations probably.  What else?

I think we'd also want to be able to have good distributed-mesh-init testing, like how to initialize a mesh with sharded tensors and make sure it works, gradients flow, we can resize / reshape / redistribute it, etc.  Core tensor ops are covered with shard tensor already, so just mesh-specific things are fine.

peterdsharpe:
Yeah, I think these are all great calls @coreyjadams . Re Mesh testing, the high-risk areas that might be good to add testing coverage mesh for a sharded Mesh are as follows:

- `test/mesh/mesh` and `test/mesh/geomery` - these are kind of the "most core" Mesh data structure ops (cheap, unlikely to be broken by sharding, but extremely widespread usage)
- `test/mesh/test_transformations.py` - geometry transformations. This is low-risk (in the sense that I'm very confident this will work fine with a sharded Mesh), but the stakes are high: these are *super* critical in datapipes.
- `test/mesh/subdivision` - this involves allocating a new-size Mesh, so potentially some things we'd want to sanity-check here - do this subdivisions happen evenly; do children live on the same device as parents;
- `test/mesh/calculus` - involves message-passing across shard boundaries
- `test/mesh/neighbors` - this is a very important one that is probably the most direct stress-test of contiguous topology across shards
- `test/mesh/smoothing` - like calculus, is also a nonlocal operation
- `test/mesh/repair` - things like "are duplicate vertices correctly detected if they're duplicated across shards"
- `test/mesh/spatial` - something like a BVH tree construction could get dicey with sharded ops?
- `test/mesh/boundaries` - watertightness checks might be interesting on a Sharded mesh?

If it's possible to add coverage for these tests in a relatively low-code-duplication way, that would be great - we definitely don't want a full copy of this test suite that will inevitably fall out of sync. Maybe a `pytest.mark.parameterize` decorator could work here?

test/mesh/test_shard_tensor_mesh_ops.py
```diff
@@ -0,0 +1,191 @@
...
+    assert_adjacency_equal,
+    assert_allclose,
+    assert_equal,
+    assert_shard_tensor,
+    mesh_to_mode,
+    to_mode_tensor,
+)
+
+
+def _surface_mesh() -> Mesh:
```
peterdsharpe: Would recommend using the meshes from `physicsnemo.mesh.primitives` here, which are intended for re-use and in-general can have much more interesting behavior (to more easily catch edge cases)

loliverhennigh:
Closing this stale ShardTensor mesh support PR for now.

PR #1610, opened 2026-05-04
GLOBE DrivAerML Self-Contained Example

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Adds the self-contained training recipe for GLOBE on DrivAerML.

## Checklist

## Review Process


examples/cfd/external_aerodynamics/globe/drivaer/run.sh
```diff
@@ -0,0 +1,111 @@
...
+OUTPUT_NAME="${SLURM_JOB_NAME:-globe_drivaer_local}"
+SCRIPT_DIR="${SLURM_SUBMIT_DIR:-$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)}"
+OUTPUT_DIR="${SCRIPT_DIR}/output/${OUTPUT_NAME}"
+
+TRAIN_ARGS=(
+    --output-name "${OUTPUT_NAME}"
+    --amp
+)
+
+export DRIVAER_DATA_DIR="${HOME}/datasets/drivaer_aws/drivaer_data_full"
```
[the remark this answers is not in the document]
peterdsharpe: Fixed by documenting

examples/cfd/external_aerodynamics/globe/drivaer/train.py
```diff
@@ -0,0 +1,797 @@
...
+    torch_compile_cache_dir = output_dir / "torch_compile_cache"
+    torch_compile_cache = torch_compile_cache_dir / f"rank_{dist.rank}.compile_cache"
+    checkpoint_dir = output_dir / "checkpoints"
+    best_model_path = output_dir / "best_model.mdlus"
+    profiling_dir = output_dir / "profiling"
+    shutdown_file = output_dir / "SHUTDOWN"
+
+    if dist.rank == 0:
+        for directory in (checkpoint_dir, torch_compile_cache_dir, profiling_dir):
+            directory.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
peterdsharpe: super minor, but fixed

examples/cfd/external_aerodynamics/globe/drivaer/train.py
```diff
@@ -0,0 +1,797 @@
...
+            )
+
+        for epoch in count(start=epoch + 1):
+            loss = {}
+            loss_components = {}
+            for split in splits:
+                with record_function(f"epoch_{epoch}_{split}"):
+                    loss[split], loss_components[split] = run_epoch(split)
+
+            scheduler.step(loss["train"])
```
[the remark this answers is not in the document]
peterdsharpe: intentional

examples/cfd/external_aerodynamics/globe/drivaer/dataset.py
```diff
@@ -0,0 +1,805 @@
...
+            randomize_vehicle: If True, sample random points inside each
+                vehicle boundary cell instead of using centroids. Typically
+                enabled during training for data augmentation.
+
+        Returns:
+            This sample, modified in-place and moved to *device*.
+        """
+        n_points = min(n_prediction_points, self.prediction_mesh.n_points)
+        mask = torch.randint(self.prediction_mesh.n_points, (n_points,))
+        self.prediction_mesh = self.prediction_mesh.to_point_cloud().slice_points(mask)
```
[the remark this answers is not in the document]
peterdsharpe: intentional

examples/cfd/external_aerodynamics/globe/drivaer/requirements.txt
```diff
@@ -0,0 +1,9 @@
+tyro~=1.0.8
+pyyaml
+torchinfo~=1.8.0
+mlflow~=3.10.0
+tenacity~=9.1.4
+pyvista~=0.47.0
+matplotlib~=3.10.8
+tensordict~=0.12.2
```
coreyjadams, line 8: Should we consider upgrading the whole repo's requirement on tensordict for v2.1.0?

examples/cfd/external_aerodynamics/globe/drivaer/splits/train.csv
coreyjadams, line 1: Is this something that we should, in principle, get up on the HF repo?

examples/cfd/external_aerodynamics/globe/drivaer/README.md
```diff
@@ -0,0 +1,107 @@
...
+## Preprocessing Pipeline
+
+1. Load VTP car body surface (~8.8M quad-dominant cells) and triangulate
+2. Compute nondimensional fields: C_p (already nondimensional),
+   C_f = wallShearStress / q_inf
+3. Interpolate cell-centered data to mesh vertices
+4. Parse reference lengths and force coefficients from CSV files
+5. Cache preprocessed samples as .pt files for fast subsequent loading
+6. At load time, randomly subsample cells for the GLOBE boundary mesh
+   (default 20K faces, configurable via `--boundary-n-faces`)
```
coreyjadams: Can you add some expectations here about what training looks like, if it's going right?  Maybe some level-setting on performance ballparks and metrics?

Not saying we have to make promises or anything, but some expectations of "Here's what it might look like for you" could be useful for users.

examples/cfd/external_aerodynamics/globe/drivaer/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Dataset
+
+The DrivAerML dataset should be available at a path like:
+
+```text
+drivaer_data_full/
+  run_1/
+    boundary_1.vtp       # Boundary surface mesh (~600 MB)
+    volume_1.vtu         # Volume mesh (~45 GB, not used by this example)
```
coreyjadams, line 39: I am thinking Volume will _never_ be part of this example, right?  We might add a comment here here that volumetric data will be manageable with the curated dataset + the unified recipe and link to that too?

Up to you.

examples/cfd/external_aerodynamics/globe/drivaer/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+- **C_p** - Pressure coefficient (scalar)
+- **C_f** - Skin friction coefficient (3D vector)
+
+These can be integrated over the car body surface to obtain aerodynamic
+force coefficients (Cd, Cl, Cs) for engineering evaluation.
+
+## Dataset
+
+The DrivAerML dataset should be available at a path like:
```
coreyjadams: Do you want to provide a script to make it easier to download the data without the volumetric files?

examples/cfd/external_aerodynamics/globe/drivaer/run.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
+#SBATCH -A accountname
+#SBATCH -J accountname-%u.train_globe_drivaer
+#SBATCH --time=4:00:00
+#SBATCH -p batch
+#SBATCH -N 6
+#SBATCH --ntasks-per-node=1
+#SBATCH --dependency=singleton
```
coreyjadams, line 8: Oh cool TIL about "singleton"

examples/cfd/external_aerodynamics/globe/drivaer/train.py
```diff
@@ -0,0 +1,797 @@
...
+        torch.profiler.profile(
+            schedule=torch.profiler.schedule(wait=5, warmup=1, active=1, repeat=1),
+            on_trace_ready=torch.profiler.tensorboard_trace_handler(
+                str(profiling_dir), worker_name=f"worker_{dist.rank}"
+            ),
+            with_stack=False,
+        )
+        if use_profiler
+        else contextlib.nullcontext()
+    )
```
coreyjadams: any chance I can convince you to just use the physicsnemo profiler?  it wraps all of this simply ...

coreyjadams (approved):
Hi @peterdsharpe - looks reasonable!  I know we have a more ambitious PR coming against the unified recipe so I assume we'll look there for dealing with volumetric training, etc?  Overall I appreciate the mirror of the airfrans usecase for users familiar there.

I've left some comments, here and there, mostly they are for your consideration though I think a couple in the README could help flesh it out.  Up to you!

examples/cfd/external_aerodynamics/globe/drivaer/requirements.txt
peterdsharpe: Good question! Not urgent, as:

a) Everything we're doing in the new version is back-compatible with tensordict 0.11
b) The main benefit that tensordict 0.12 brings is better `torch.compile` coverage, so for end-users, the worst-case scenario is that they get worse `torch.compile` support (and potentially disable this)
c) There were some growing pains with tensordict 0.12. I *think* we have now addressed all these, but it's enough that I'd want to make that version bump a separate PR.

examples/cfd/external_aerodynamics/globe/drivaer/splits/train.csv
peterdsharpe: Yes, absolutely. This will be part of the splits for DrivAerML, when we publish those alongside the -Mesh-based version of that dataset.

examples/cfd/external_aerodynamics/globe/drivaer/README.md
peterdsharpe: Will do that, but will bring that in as part of the unified recipe PR. (On the off chance that the future PR does not merge, I don't want a dangling forward reference in this PR.)

examples/cfd/external_aerodynamics/globe/drivaer/run.sh
peterdsharpe: My preference! :) This is why you see me stacks of thousands of jobs in HSG SLURM queue haha

examples/cfd/external_aerodynamics/globe/drivaer/train.py
peterdsharpe: This is a wise call, let me look into it

examples/cfd/external_aerodynamics/globe/drivaer/README.md
peterdsharpe: I will add a link to the dataset! That said, I would leave an explicit script out-of-scope (since we don't want to manage links to that, and in the future, the unified recipe will supersede this anyway)

examples/cfd/external_aerodynamics/globe/drivaer/dataset.py
```diff
@@ -0,0 +1,800 @@
...
+            device: Target device for the returned sample.
+            randomize_vehicle: If True, sample random points inside each
+                vehicle boundary cell instead of using centroids. Typically
+                enabled during training for data augmentation.
+
+        Returns:
+            This sample, modified in-place and moved to *device*.
+        """
+        n_points = min(n_prediction_points, self.prediction_mesh.n_points)
+        mask = torch.randint(self.prediction_mesh.n_points, (n_points,))
```
mehdiataei, line 184: Can we switch this to without-replacement sampling, e.g. `torch.randperm(..., generator=...)[:n_points]`, and make validation deterministic with a fixed per-sample mask or seed? That would make `best_model.mdlus` reflect model quality instead of sampling noise

examples/cfd/external_aerodynamics/globe/drivaer/dataset.py
```diff
@@ -0,0 +1,800 @@
...
+        self.prediction_mesh = self.prediction_mesh.to_point_cloud().slice_points(mask)
+
+        for bc_type, mesh in self.boundary_meshes.items():
```
mehdiataei, line 187: Can we either keep centroids fixed for the boundary quadrature, or explicitly document augmentation is intended and does not bias the boundary integral? I think with `randomize_vehicle=True` this moves the quadrature point by replacing cached cell centroids, but the cached/derived `cell_areas` and `cell_normals` still describe the original triangle

examples/cfd/external_aerodynamics/globe/drivaer/dataset.py
```diff
@@ -0,0 +1,800 @@
...
+                If ``False``, preserve ``point_data`` and ``cell_data``
+                (used for visualization).
+            voronoi: If ``True``, use Voronoi-corrected areas and normals
+                via :func:`partition_cells`.  If ``False`` (default), use
+                uniform area rescaling (faster, but less accurate).
+
+        Returns:
+            Mesh with ``n_cells`` cells and corrected area cache.
+        """
+        indices = torch.randperm(mesh.n_cells)[:n_cells]
```
mehdiataei, line 273: `n_cells <= 0` returns an empty boundary, and `n_cells > mesh.n_cells` silently returns fewer cells than requested. Maybe some clamping is needed here

examples/cfd/external_aerodynamics/globe/drivaer/dataset.py
```diff
@@ -0,0 +1,800 @@
...
+            data_dir: Root directory containing ``run_N/`` subdirectories.
+            split: ``"train"`` or ``"validation"``.
+
+        Returns:
+            Sorted list of absolute paths to individual run directories.
+        """
+        splits_csv = Path(__file__).parent / "splits" / f"{split}.csv"
+        with open(splits_csv) as f:
+            run_indices = [row["run_idx"] for row in csv.DictReader(f)]
+        return sorted(data_dir / f"run_{idx}" for idx in run_indices)
```
mehdiataei, line 317: This ignores the split CSV order and sorts paths (`run_100` before `run_2`), not numerically. That affects reproducibility and also changes which sample `inference.py` picks as `validation[0]`.

maybe sort numerically with with something like `key=lambda p: int(p.name.removeprefix("run_"))`

examples/cfd/external_aerodynamics/globe/drivaer/inference.py
```diff
@@ -0,0 +1,110 @@
...
+if not best_model_path.exists():
+    raise RuntimeError(f"No best model found at {best_model_path}")
+logger.info(f"Loading best model from {best_model_path.name!r}...")
+model.load(best_model_path)
+
+# %%
+### Run inference
+with torch.no_grad():
+    model.eval()
+    pred_mesh = model(**sample.model_input_kwargs)
```
mehdiataei, line 92: maybe a bit risky to materialize the full prediction if we have millions of cells

examples/cfd/external_aerodynamics/globe/drivaer/inference.py
```diff
@@ -0,0 +1,110 @@
...
+logging.basicConfig(level=logging.INFO)
+logger = PythonLogger("globe.drivaer.inference")
+
+# %%
+### Resolve output directory
+_output_root = Path(__file__).parent / "output"
+if _env_dir := os.environ.get("GLOBE_OUTPUT_DIR"):
+    output_dir = Path(_env_dir)
+else:
+    output_subdirs = [d for d in _output_root.iterdir() if d.is_dir()]
```
mehdiataei, line 47: This will fail if imported in a clean checkout without the output folder. I suggest moving this into `if __name__ == "__main__"`

examples/cfd/external_aerodynamics/globe/drivaer/train.py
```diff
@@ -0,0 +1,797 @@
...
+            )
+
+        for epoch in count(start=epoch + 1):
+            loss = {}
+            loss_components = {}
+            for split in splits:
+                with record_function(f"epoch_{epoch}_{split}"):
+                    loss[split], loss_components[split] = run_epoch(split)
+
+            scheduler.step(loss["train"])
```
mehdiataei, line 609: Is this intentional because validation is currently stochastic? I think `ReduceLROnPlateau` should probably step on `loss["validation"]`

examples/cfd/external_aerodynamics/globe/drivaer/train.py
```diff
@@ -0,0 +1,797 @@
...
+        )
+    else:
+        optimizer = torch.optim.RAdam(
+            model.parameters(),
+            lr=learning_rate,
+            weight_decay=weight_decay,
+            decoupled_weight_decay=True,
+            foreach=True,
+        )
+    patience_epochs = patience_steps // len(dataloaders["train"])
```
mehdiataei: maybe `max(1, math.ceil(patience_steps / len(dataloaders["train"])))` just to be safe if patience steps is set less than number of training batches.

examples/cfd/external_aerodynamics/globe/drivaer/README.md
mehdiataei: I think the default is 80k not 20k and and I think `--n-faces-per-boundary`, not `--boundary-n-faces`

examples/cfd/external_aerodynamics/globe/drivaer/README.md
```diff
@@ -0,0 +1,107 @@
+# GLOBE for DrivAerML (3D Car Aerodynamics)
+
+This example trains a GLOBE model to predict surface pressure coefficient
+(C_p) and skin friction coefficient (C_f) on parametrically-varied 3D car
+body geometries from the DrivAerML dataset.
+
+## Problem Description
+
+The DrivAerML dataset contains ~207 scale-resolving simulations (SRS) of a
```
mehdiataei: These IDs are missing for some reason:
``167, 211, 218, 221, 248, 282, 291, 295, 316, 325, 329, 364, 370, 376, 403, 473`

examples/cfd/external_aerodynamics/globe/airfrans/train.py
```diff
@@ -114,7 +114,7 @@ def main(
     network_type: Literal["pade", "mlp"] = "pade",
     self_regularization_beta: float | None = 0.01,
     latent_compression_scale: float | None = 100.0,
-    expand_far_targets: bool = False,
+    expand_far_targets: bool = True,
```
mehdiataei, line 118: Maybe comment here why this is turned on here for reference?

examples/cfd/external_aerodynamics/globe/drivaer/dataset.py
```diff
@@ -0,0 +1,800 @@
...
+                y_max=y_half_width,
+                n_x=n_x,
+                n_y=n_y,
+            ),
+            target_n_spatial_dims=3,
+        ).translate([0.0, 0.0, ground_z])
+        # Flip cell winding so normals point upward (+z, into the fluid domain)
+        return Mesh(points=mesh.points, cells=mesh.cells[:, [0, 2, 1]])
+
+    return {
```
mehdiataei, line 112: Both floor patches include `x_bl`. Is this OK?

mehdiataei (commented):
Hey @peterdsharpe Great example! I added some comments.

examples/cfd/external_aerodynamics/globe/airfrans/train.py
peterdsharpe: Ah! This is the higher-order kernel function, which improves accuracy. Meant to turn it on in the AirFRANS PR, putting this in here as a rider. (The DrivAerML example uses this too.)

examples/cfd/external_aerodynamics/globe/drivaer/train.py
peterdsharpe: Yeah, the issue is that this dataset is split as train/validation, not train/validation/test (inheriting these from the split names from the DoMINO DrivAerML work) - in essence what we call "validation" is really "test". So, if we `ReduceLROnPlateau` based on `loss["validation"]`, we're (weakly) contaminating the training data with information from the test set that we later use to evaluate. Because of this I think sticking to "train" is less-bad than using "validation" as the LR decay trigger. (In an ideal world, the dataset would have train/val/test labels, and we would use val for this.)

examples/cfd/external_aerodynamics/globe/drivaer/README.md
peterdsharpe: Ah, good catch! These splits are inherited from the DoMINO DrivAerML case (e.g., [here](https://github.com/NVIDIA/physicsnemo-cfd/blob/main/workflows/bench_example/drivaer_ml_files/train.csv)), so we can't really change the splits here without invalidating cross-applicability there. But I will note this in the README.md.

examples/cfd/external_aerodynamics/globe/drivaer/dataset.py
peterdsharpe: Yes, that is actually required. What we're saying here is that the slip floor goes from x_upstream to x_bl, and the no-slip floor goes from x_bl to x_downstream.

These are the x-positions (i.e., streamwise direction) of the inlet, BC switch point, and outlet, respectively.

That BC switch point was set by the DrivAerML CFD problem setup.

examples/cfd/external_aerodynamics/globe/drivaer/train.py
peterdsharpe: Good catch, I will add a guard on this for robustness

examples/cfd/external_aerodynamics/globe/drivaer/README.md
peterdsharpe: Added

examples/cfd/external_aerodynamics/globe/drivaer/README.md
peterdsharpe: Ah, very good catch - it used to be 20k, and the old arg name, but then this got updated. I'll fix this!

examples/cfd/external_aerodynamics/globe/drivaer/train.py
peterdsharpe: Added

examples/cfd/external_aerodynamics/globe/drivaer/README.md
peterdsharpe: Fixed

examples/cfd/external_aerodynamics/globe/drivaer/README.md
peterdsharpe: Documented

examples/cfd/external_aerodynamics/globe/drivaer/dataset.py
peterdsharpe: This is intentional; see reply to greptile thread in this PR on this issue. This approach is significantly faster, as is generally the case when `n_sampled_points << n_points`. (Without this, data subsampling inside the training loop consumes non-negligible time.)

examples/cfd/external_aerodynamics/globe/drivaer/dataset.py
peterdsharpe: This is intended to increase the effective diversity of mesh topologies by simulating the effect of differing surface discretizations. This is in-theory convergent to the fixed-centroid answer, and in testing on AirFRANS, using this dynamic approach results in a smaller difference between train loss and validation loss. (The mechanism here is essentially data augmentation.)

examples/cfd/external_aerodynamics/globe/drivaer/dataset.py
peterdsharpe: Kept here, as this does not affect reproducibility (it's still sorted, just lex-sorted rather than numerically-sorted). This is no inherent ordering to the run IDs (they were generated via Latin hypercube sampling of the space), so there is no reason to preserve that numerical order here (over say, lex order).

examples/cfd/external_aerodynamics/globe/drivaer/dataset.py
peterdsharpe: Added a guard for the nonpositive case. The `n_cells > mesh.n_cells` is actually already safe; this gracefully falls back to no-subsampling.

examples/cfd/external_aerodynamics/globe/drivaer/inference.py
peterdsharpe: GLOBE does internal chunking for large inputs, so actually inference (i.e., under no_grad) is extremely memory-cheap here.

examples/cfd/external_aerodynamics/globe/drivaer/inference.py
peterdsharpe: This script is intended to be run directly (top-level), rather than a module to be imported - leaving this as-is as importing is not the intended use case

PR #1611, opened 2026-05-04
Adds GLOBE to Unified External Aero Recipe

peterdsharpe:
# PhysicsNeMo Pull Request

This PR adds GLOBE to the unified external aero recipe via a DomainMesh-native refactor.

Note: this is a stacked PR on top of https://github.com/NVIDIA/physicsnemo/pull/1621

## Why

GLOBE doesn't fit the unified recipe's existing point-cloud / transformer contract. Its `forward()` consumes `dict[str, Mesh]` boundary inputs, an unbatched `(N_pred, D)` prediction-point tensor, and a `dict[str, scalar]` of per-sample reference lengths; it returns a `Mesh[0, "n"]`. The other recipe models all take batched flat tensors `(B, N, C)` via a central `MODEL_MAPPINGS` registry and return `(B, N, C_out)` tensors. Naive integration would have meant a GLOBE-specific `nn.Module` adapter wedging mesh I/O into the tensor pipeline, plus a parallel dataset YAML that exposed the connectivity GLOBE needs - both leaky and hard to maintain.

We pre-discussed the architecture with the unified recipe author (Corey); the approach we landed on makes the recipe **DomainMesh-native end-to-end**, so:

- GLOBE drops in directly via Hydra with no per-model adapter.
- Existing transformer models are first-class consumers of the same `DomainMesh` via a declarative `forward_kwargs` spec in each model YAML.
- Future mesh-native models follow the same pattern with no recipe-side changes.

## The DomainMesh contract

Every dataset YAML now ends with a `DomainMesh` whose three components have a fixed semantic meaning:

| Field | Semantic | DrivAerML surface | DrivAerML volume |
|---|---|---|---|
| `interior` | "where the predictions live" | `Mesh[0, 3]` point cloud at the (subsampled) cell centroids; targets in `point_data` | volume mesh; targets in `interior.point_data` |
| `boundaries` | "the inputs" | `{"vehicle": Mesh[2, 3]}` car body, with precomputed `cell_data["normals"]` for transformer models | `{"surface": Mesh[2, 3]}` car body |
| `global_data` | freestream metadata | `U_inf`, `rho_inf`, `p_inf`, `nu`, `L_ref` | same |

**Targets always live at `interior.point_data.<name>`** by convention. Loss and metrics extract them by name; model `forward_kwargs` specs should never reference that location (they'd be reading the answer the model is supposed to predict).

## Per-model `forward_kwargs` spec

Each model YAML declares an `input_type` (`mesh` or `tensors`), an `output_type`, and a `forward_kwargs:` block that maps the model's forward kwargs to declarative paths into the DomainMesh. Spec syntax:

| Spec value | Resolves to |
|---|---|
| `"interior.points"` | Tensor at that dotted path (`getattr` then `getitem`) |
| `42` / `1.0` | 0-d `float32` tensor literal |
| `[a, b, c]` | `torch.cat` along last dim, with ndim alignment |
| `{a: ..., b: ...}` | dict-valued kwarg (used for GLOBE's `boundary_meshes` + `reference_lengths`, DoMINO's `data_dict`) |
| `{source: <path>, expand_like: <other_kwarg>}` | Two-pass: resolve `source`, then expand axis -2 to match `<other_kwarg>` |

Two representative examples:

```yaml
# train_geotransolver_automotive_surface.yaml -- tensor I/O
input_type: tensors
output_type: tensors
forward_kwargs:
  geometry: interior.points
  local_embedding:
    - interior.points
    - boundaries.vehicle.cell_data.normals
  local_positions: interior.points
  global_embedding: global_data.U_inf
```

```yaml
# train_globe_automotive_surface.yaml -- mesh I/O
input_type: mesh
output_type: mesh
forward_kwargs:
  prediction_points: interior.points
  boundary_meshes:
    vehicle: boundaries.vehicle
  reference_lengths:
    L_ref: 1.0
    delta_turb: 0.015
```

The `expand_like` modifier replaces the legacy `broadcast_global` flag that previously walked the whole batch and expanded any token-shaped tensor in-place from `train.py`.

### Upstream

- **New `MeshToDomainMesh` transform** in `physicsnemo/datapipes/transforms/mesh/transforms.py`. Wide-input / narrow-output API: `(cell_data_targets, point_data_targets, interior_points, boundary_name)`. v1 implements the two diagonal corners (`cell_data_targets` + `cell_centroids`, `point_data_targets` + `vertices`); cross-corners raise `NotImplementedError` with messages pointing at the right fix. `apply_to_domain` is identity passthrough so the transform may sit harmlessly at the end of any pipeline.
- 20 unit tests in `test/datapipes/transforms/test_mesh_to_domain_mesh.py`.

### Recipe `src/`

- **New `forward_kwargs.py`**: `walk_path` / `resolve_spec` / `resolve_forward_kwargs` / `extract_targets`. Two-pass resolution so `expand_like` can reference any other kwarg regardless of declaration order.
- **Rewritten `collate.py`**: `build_collate_fn(input_type, forward_kwargs_spec, target_config)`. Tensor-input mode pads token features to `(1, 1, D)` and per-element features to `(1, N, C)`, while per-element scalar targets stay `(1, N)` (so they match the model's `(1, N)` split-from-`(B, N, C)` prediction). Mesh-input mode passes Mesh / dict / scalar through unchanged. `len(samples) != 1` raises.
- **Refactored `loss.py` and `metrics.py`** to a dict-of-tensors interface. Added `field_weights: dict[str, float] | None` to `LossCalculator` (replaces the standalone GLOBE recipe's `error_scales` weighting). Bit-exact (within 1e-7) numerical equivalence with the legacy `(B, N, C)`-tensor formula when all weights are 1.
- **`train.py`** changes:
  - `forward_pass` dispatches on `input_type` / `output_type`, normalizes the model output to `dict[str, Tensor]` (extract `Mesh.point_data` for mesh outputs; split tensor by `FieldSpec` for tensor outputs), casts pred / target to float32 right before loss.
  - `_recursive_to_device` and `_recursive_cast_floats` replace the older `_recursive_to`. The cast helper skips integer tensors and Mesh objects so `cells.long()` survives the bf16 autocast.
  - `batch_size > 1` raises `NotImplementedError` early (every model in the recipe assumes B=1; the YAML field is reserved for future use).
  - Drops `broadcast_global`, `data_mapping`, and the old `dataset.targets` plumbing.
- **`datasets.py`**: `_maybe_inject_targets` auto-fills `cell_data_targets` (or `point_data_targets`) into `MeshToDomainMesh` from the dataset YAML's `targets:` block, so users don't list field names twice. `build_dataloaders` reads targets / metrics from the dataset YAML rather than the training YAML.

### Configs

- All **6 dataset YAMLs** migrated: terminal `MeshToTensorDict + ComputeCellCentroids + RestructureTensorDict` removed. Surface YAMLs append `MeshToDomainMesh`; volume YAMLs already produce DomainMesh natively. The dataset YAML's `targets:` block is now the single source of truth.
- All **13 existing training YAMLs** migrated: `input_type` + `output_type` + `forward_kwargs` declared; `data_mapping`, `broadcast_global`, `model_type`, `dataset.targets`, `dataset.nondim_types` removed.
- **2 new GLOBE training YAMLs** added (`train_globe_automotive_surface.yaml`, `train_globe_automotive_volume.yaml`).

### Documentation

- `README.md` updated: GLOBE listed alongside Transolver / GeoTransolver / FLARE; new "DomainMesh contract" section; `forward_kwargs` spec syntax documented; updated configs table; mermaid diagram refreshed for the DomainMesh-native flow; `field_weights` and `batch_size > 1` notes added.

### Tests

98 tests, all green:

| Suite | Count |
|---|---|
| Upstream `MeshToDomainMesh` (`test/datapipes/transforms/`) | 20 |
| Recipe `forward_kwargs` resolver | 33 |
| Recipe `collate` two-mode batching + failure paths | 20 |
| Recipe `loss` dict-vs-legacy equivalence + `field_weights` | 12 |
| Recipe synthetic E2E for each migrated config (incl. GLOBE) | 13 |

Recipe-local tests live under `examples/cfd/external_aerodynamics/unified_external_aero_recipe/tests/` (plural to avoid colliding with the repo's main `test/` package). They use a `conftest.py` that does the same `sys.path.insert(src/)` the production `train.py` does.

All pre-commit hooks (`ruff` / `interrogate` / `markdownlint` / license / import-linter) pass on every changed file.

## Breaking changes

These are intentional and documented in the migrated YAMLs / README:

- **`data_mapping: "..."` removed.** `forward_kwargs:` block in the model YAML replaces it. The `MODEL_MAPPINGS` registry in `src/collate.py` is gone.
- **`broadcast_global: true` removed.** Replaced by the per-spec `{source: ..., expand_like: <other_kwarg>}` modifier.
- **`model_type:` field removed** (was metadata-only; never read by code).
- **`dataset.targets` removed from training YAMLs.** Targets now come from the dataset YAML's `targets:` block (single source of truth).
- **`dataset.nondim_types` removed** (was wired only to dead-code utility `_to_physical`; no longer present).
- **`MeshToTensorDict` + `ComputeCellCentroids` + `RestructureTensorDict` removed from dataset YAMLs.** They're no longer used; the dataset ends with a `DomainMesh` directly.
- **`LossCalculator` / `MetricCalculator` signature change**: now `__call__(pred: dict[str, Tensor], target: dict[str, Tensor])` instead of `(B, N, C)` tensors with `FieldSpec`-based slicing.

`train.py` raises clear errors if a config still uses the old contract (missing `input_type` / `forward_kwargs`, etc.), so silent failures are unlikely.

## Known limitations / out-of-scope

These were deferred; each is a clean follow-up:

- **DoMINO YAMLs are illustrative only.** (pre-existing issue) The legacy mapping referenced ~13 dataset fields that the dataset YAMLs don't expose (DoMINO never trained end-to-end before either). The migrated YAMLs have a placeholder `forward_kwargs.data_dict` block and a comment noting the gap. Wiring DoMINO requires dataset-side changes (per-cell neighbor / grid features) that are out of scope here.
- **GLOBE reference lengths are constants** (`L_ref=1.0`, `delta_turb=0.015`, in nondim space since positions are scaled by `metadata.L_ref=5`). Per-sample reference lengths from `geo_ref_*.csv` are a future PR.
- **GLOBE has vehicle BC only.** The standalone GLOBE drivaer recipe uses three boundaries (`vehicle` + `slip_floor` + `no_slip_floor`); the floor patches are deferred until DomainMesh on disk supports surface-mode multi-boundary curation.
- **GLOBE runs with `compile: false`.** Enabling `torch.compile` requires three Dynamo flags (`force_parameter_static_shapes=False`, `capture_scalar_outputs=True`, `cache_size_limit=64`) the unified recipe doesn't currently set.
- **No real-data verification.** All testing was synthetic (small meshes on CPU). The synthetic suite covers shape / contract correctness; numerical training behavior on real DrivAerML / HighLift / SHIFT SUV data should be confirmed with at least one `benchmark_io=true` run per config and one full step before relying on it for production training.
- **Multi-GPU DDP / mixed-precision perf flags untested**.

## Reviewer notes

- The conceptual review is the **DomainMesh contract** + the `forward_kwargs` spec design. Everything else (YAML migrations, test scaffolding) is mechanical.

## Checklist

## Review Process


coreyjadams (approved):
LGTM

PR #1220, opened 2025-11-09
Feature/edm precond 3d

AbVishwas:
# PhysicsNeMo Pull Request

## Description
- Added 3D-EDMPrecond wrapper class with 3D version of SongUnet for diffusion models:
  - `SongUNet3D`: 3D U-Net diffusion backbone extending DDPM++ and NCSN++ to volumetric data
  -  New test to cover the SongUnet3D
  - `EDMPrecond3D`: 3D preconditioning wrapper for volumetric diffusion models

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [x] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.

## Dependencies

No addition dependencies 

## Review Process

**All PRs are reviewed by the PhysicsNeMo team before merging.**

Depending on which files are changed, GitHub may automatically assign a maintainer for review.

We are also testing AI-based code review tools (e.g., Greptile), which may add automated comments with a confidence score.
This score reflects the AI’s assessment of merge readiness and is **not** a qualitative judgment of your work, nor is
it an indication that the PR will be accepted / rejected.

AI-generated feedback should be reviewed critically for usefulness.
You are not required to respond to every AI comment, but they are intended to help both authors and reviewers.
Please react to Greptile comments with 👍 or 👎 to provide feedback on their accuracy.

PR #1616, opened 2026-05-05
Adds DiffusionUNet3D and its modules to physicsnemo.experimental

CharlelieLrt:
# PhysicsNeMo Pull Request

## Description

Replaces #1220 from @AbVishwas, which had diverged too much from `main`.

This PR enables training and sampling diffusion models on 3D structured domains (volumetric or grid-based scientific data) with the addition of a `DiffusionUNet3D` backbone, which satisfies the `DiffusionModel` protocol and plugs straight into the existing preconditioners, losses, and samplers in `physicsnemo.diffusion`. Conditioning is exposed through a single `TensorDict` argument with optional `"vector"` and `"volume"`  keys. The PR also adds reusable 3D building blocks (`Conv3D`, `GroupNorm3D`, `UNetAttention3D`, `UNetBlock3D`) under `physicsnemo.experimental.nn` that can be composed for other 3D architectures.

## Checklist

## Review Process


test/models/diffusion/conftest.py
```diff
@@ -14,12 +14,83 @@
...
+import torch
+import torch._dynamo
 
 from physicsnemo.core.version_check import check_version_spec
 
 _APEX_AVAILABLE = check_version_spec("apex", hard_fail=False)
 
+_GLOBAL_SEED = 42
+_CPU_TOLERANCES = {"atol": 1e-3, "rtol": 1e-3}
+_GPU_TOLERANCES = {"atol": 1e-2, "rtol": 5e-2}
```
pzharrington: O(1e-2) is pretty loose, no?

pzharrington (approved):
Largely looks good to me. Regarding ShardTensor, I don't see any issues that would present problems, the patterns seem to pretty closely follow the 2D SongUNet patterns and we know that works. For maximum sanity, you could add some domain parallelism tests explicitly targeting this architecture.

test/models/diffusion/conftest.py
CharlelieLrt: It is. I've had lots of problems with this (on not only in this PR). Here it comes from the SDPA, which gives very different results on CPU and GPU (and even on different GPU arch). It also comes from the fact that everything is initialized with purely random weights and input tensors. Using something more deterministic would probably help.

pzharrington: Hmm ok. I'm generally cautious with that bc I've had scenarios where tests erroneously pass even with tolerances much lower than this, e.g. 1e-4. Maybe instead of setting it across the board you could default to lower and then override it for the flaky tests?

CharlelieLrt: Fair point. Moved the loose tolerances out of `conftest.py` and into the two test files that actually need them in bdcfe6c3, so the looseness won't leak to sibling tests

CharlelieLrt:
> Regarding ShardTensor, I don't see any issues that would present problems, the patterns seem to pretty closely follow the 2D SongUNet patterns and we know that works. For maximum sanity, you could add some domain parallelism tests explicitly targeting this architecture.

Good idea, added two sanity tests in 477acd93 under `test/domain_parallel/models/test_diffusion_unet_3d.py` (unconditional and conditional, both sharded along H), gated by `@pytest.mark.multigpu_static`.

test/domain_parallel/models/test_diffusion_unet_3d.py
```diff
@@ -0,0 +1,109 @@
...
+
+    assert out.shape == (B, C, D, H, W)
+    assert out._spec.placements == x_sharded._spec.placements
+
+
+@pytest.mark.multigpu_static
+def test_diffusion_unet_3d_conditional_distributed(distributed_mesh):
+    """Forward with vector + volume conditioning, sharded along H."""
+    dm = DistributedManager()
+    B, C, D, H, W = 2, 2, 8, 16, 16
```
coreyjadams: Not sure how deep the model is, but when we these test up at 8 GPUs the H dimension will locally be 2.  will that be a problem?

test/domain_parallel/models/test_diffusion_unet_3d.py
```diff
@@ -0,0 +1,109 @@
...
+        attention_levels=[1],
+        dropout=0.0,
+    ).to(device)
+
+
+@pytest.mark.multigpu_static
+def test_diffusion_unet_3d_distributed(distributed_mesh):
+    """Unconditional forward with `x` sharded along the H axis."""
+    dm = DistributedManager()
+    B, C, D, H, W = 2, 2, 8, 16, 16
```
coreyjadams: Same comment here as below.

coreyjadams (approved):
Approving for the domain parallelism tests, thanks for adding those.  Recommend you increase the size along the domain parallel dimension a bit but otherwise looks good to me.

test/domain_parallel/models/test_diffusion_unet_3d.py
CharlelieLrt: Good catch. Bumped `H` and `W` from 16 to 32 in 008a3c13 so on 8 GPUs each rank holds local H=4 at level 0 and H=2 at the bottleneck, which is the minimum safe size for kernel-3 conv with padding.

test/domain_parallel/models/test_diffusion_unet_3d.py
CharlelieLrt: Good catch. Bumped `H` and `W` from 16 to 32 in 008a3c13 so on 8 GPUs each rank holds local H=4 at level 0 and H=2 at the bottleneck, which is the minimum safe size for kernel-3 conv with padding.

PR #1629, opened 2026-05-07
Updates to Unified External Aero Recipe [Stacked PR]

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

This PR is a subset of https://github.com/NVIDIA/physicsnemo/pull/1611 - in essenece this forms a stacked PR with this PR as the first one to merge.

This PR contains all the non-GLOBE-specific changes of #1611. A full description can be seen in the PR description for 1611 (omitting the GLOBE items, and recipe-specific tests which will be cut before merge).

## Checklist

## Review Process


examples/cfd/external_aerodynamics/unified_external_aero_recipe/src/train.py
```diff
@@ -139,374 +173,487 @@ def get_autocast_context(precision: str):
...
-    return loss, metrics, (outputs, targets)
+        metric_td = metric_calculator(pred_f32, target_f32)
+    ### Move every per-field tensor off-device into a plain Python float
+    ### before returning so the caller can sum / log without holding on
+    ### to autograd graph fragments or device tensors.
+    return (
+        loss,
+        {k: v.item() for k, v in loss_td.items()},
+        {k: v.item() for k, v in metric_td.items()},
+    )
```
coreyjadams: We can let this in for now, but this will need to get fixed I bet.  If you want to detach from autograd just call detach, `item` is a blocking sync.

examples/cfd/external_aerodynamics/unified_external_aero_recipe/src/train.py
```diff
@@ -557,27 +706,230 @@ def benchmark_io_epoch(
...
+) -> "NormalizeMeshFields | None":
+    """Return the first ``NormalizeMeshFields`` instance across *datasets*' pipelines.
 
-    Returns (normalizer, nondim) instances from the first dataset that has them,
-    or (None, None) if not found.
+    Used at checkpoint-save time to persist normalization stats alongside the
+    model weights so inference can re-apply the inverse. Returns ``None`` when
+    no dataset has a ``NormalizeMeshFields`` transform.
     """
     from physicsnemo.datapipes.transforms.mesh import NormalizeMeshFields
```
coreyjadams: Should we move to utils?

examples/cfd/external_aerodynamics/unified_external_aero_recipe/src/train.py
```diff
@@ -557,27 +706,230 @@ def benchmark_io_epoch(
...
+            f"global_data by the dataset builder."
+        )
 
-def build_dataloaders(cfg: DictConfig):
+
+def _resolve_manifest_spec(
+    ds_yaml: DictConfig, ds_cfg_block: DictConfig
+) -> dict | None:
+    """Resolve a `data.<key>` block's manifest config; return ``None`` for directory mode.
+
```
coreyjadams: move to utils?

examples/cfd/external_aerodynamics/unified_external_aero_recipe/src/train.py
```diff
@@ -557,27 +706,230 @@ def benchmark_io_epoch(
...
+def _validate_dataset_consistency(
+    ds_key: str,
+    ds_targets: dict[str, FieldType],
+    ds_metrics: list[MetricName],
+    ds_metadata: dict[str, Any],
+    first_targets: dict[str, FieldType],
+    first_metrics: list[MetricName],
+    first_metadata: dict[str, Any],
+) -> None:
+    """Reject ``targets:`` mismatch across multi-dataset training; warn on softer drift.
```
coreyjadams: move to dataset?

examples/cfd/external_aerodynamics/unified_external_aero_recipe/src/train.py
```diff
@@ -557,27 +706,230 @@ def benchmark_io_epoch(
...
+    """Per-split DistributedSamplers (or `(None, None)` on a single rank)."""
+    if not use_distributed:
+        return None, None
+    train_sampler = torch.utils.data.distributed.DistributedSampler(
+        train_dataset, shuffle=True, drop_last=True, seed=sampler_seed
+    )
+    val_sampler = torch.utils.data.distributed.DistributedSampler(
+        val_dataset, shuffle=False, drop_last=False
+    )
+    return train_sampler, val_sampler
```
coreyjadams, line 828: Are these for non-manifest samplers?

physicsnemo/datapipes/transforms/mesh/transforms.py
```diff
@@ -831,3 +834,246 @@ def extra_repr(self) -> str:
...
+
+    The output ``DomainMesh`` follows a semantic contract that separates
+    *where the predictions live* from *what the inputs are*:
+
+    - ``interior``: a :class:`Mesh` (typically a point cloud
+      :class:`Mesh[0, n_spatial_dims]`) whose ``points`` are the prediction
+      locations and whose ``point_data`` carries the prediction targets.
+    - ``boundaries``: a single named entry mapping ``boundary_name`` to the
+      input mesh, with target fields removed (so consumers cannot accidentally
+      read targets through the boundary).
```
coreyjadams, line 880: Do I understand properly that you take an input mesh, and split the inputs to the "interior" and the rest of the data to the "boundary" as a domain mesh?

coreyjadams (approved):
I'm going to approve this to unblock, because I know we need to get this in and merge.  I have a couple questions I left through out.  We're gonna stack, I know, so some of this can fix in the next drop.

Where did `_to_physical` go?  I know it _looks_ like dead code but it's not: It's absolutely intended to re-dimensionalize the data for the inference scripts.  Is it possible to readd it?  Otherwise, we can recreate that functionality when the dust settles I guess.

coreyjadams (approved):

coreyjadams:
@peterdsharpe I approved so you wouldn't be blocked, but I'd love to learn a little more about two pieces of this:
- how does the input mapping work, which was previously these bespoke functions and now is just yanking off a mesh in a helter-skelter set of python functions?  It seems a bit hard to trace in the code I'll admit, though I think I know what you're trying to do?
- I want to make sure we have a way to _re-dimensionalize_ the data after predicting on it.
- I couldn't see how you are handling scalar vs vector inputs and targets here, can you elaborate for me?  Is it implicit off the mesh output now?

examples/cfd/external_aerodynamics/unified_external_aero_recipe/src/train.py
peterdsharpe: I'll fix it, good call - whoops

examples/cfd/external_aerodynamics/unified_external_aero_recipe/src/output_normalize.py
```diff
@@ -0,0 +1,166 @@
...
+        if missing:
+            raise KeyError(
+                f"Mesh output is missing target fields {missing!r}; "
+                f"available: {sorted(available)!r}"
+            )
+        return output.point_data.select(*target_config)
+
+    if output_type == "tensors":
+        if isinstance(output, tuple):
+            output = next(o for o in output if o is not None)
```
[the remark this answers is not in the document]
peterdsharpe: fixed

examples/cfd/external_aerodynamics/unified_external_aero_recipe/src/forward_kwargs.py
```diff
@@ -0,0 +1,322 @@
...
+    )
+
+
+def resolve_spec(spec: Any, source: Any) -> Any:
+    """Resolve a single spec value against `source`.
+
+    See module docstring for spec semantics. This function does NOT handle
+    modifier specs (`{source: ..., expand_like: ...}`); modifiers are deferred
+    by :func:`resolve_forward_kwargs` to a second pass that has access to
+    other already-resolved kwargs.
```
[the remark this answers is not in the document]
peterdsharpe: fixed

examples/cfd/external_aerodynamics/unified_external_aero_recipe/src/train.py
peterdsharpe: done

examples/cfd/external_aerodynamics/unified_external_aero_recipe/src/train.py
peterdsharpe: done

examples/cfd/external_aerodynamics/unified_external_aero_recipe/src/train.py
peterdsharpe: done

physicsnemo/datapipes/transforms/mesh/transforms.py
peterdsharpe: Exactly - one input `Mesh` becomes a `DomainMesh` with two
named slots:

- `interior`: a fresh point-cloud `Mesh` whose `points` are the
  *prediction locations* (cell centroids by default, or the input
  vertices when `interior_points='vertices'`) and whose `point_data`
  carries *only* the target fields.
- `boundaries[boundary_name]` (default key `"vehicle"`): the input mesh
  itself, with the target fields *removed* from its `cell_data` /
  `point_data`. Stripping the targets is important for avoiding contamination -
  consumers can't accidentally read targets through the boundary path
  and short-circuit the loss.
- `global_data`: passed through unchanged.

The "Wide inputs / narrow outputs" section of the docstring lays out the
contract, and the worked `C_p` example at the bottom shows the
target-stripping in action:

```python
>>> "C_p" in domain.interior.point_data.keys()           # target moved here
True
>>> "C_p" in domain.boundaries["vehicle"].cell_data.keys()  # removed from boundary
False
>>> "normals" in domain.boundaries["vehicle"].cell_data.keys()  # non-target stays
True
```

examples/cfd/external_aerodynamics/unified_external_aero_recipe/src/train.py
peterdsharpe: Yep! There are two split strategies - directory mode and manifest mode. This supports the former.

I tightened the `_build_directory_samplers` docstring in this push to
say so explicitly so the next reader doesn't have to triangulate
between the two helpers.

examples/cfd/external_aerodynamics/unified_external_aero_recipe/src/train.py
```diff
@@ -557,27 +706,230 @@ def benchmark_io_epoch(
...
+
+
+def _build_manifest_samplers(
+    train_indices: list[int],
```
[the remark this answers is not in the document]
peterdsharpe: Confirmed - the inline comment was wrong, and the val loader was
inheriting the train sampler's `shuffle=True` / `drop_last=True`
when no explicit val split was configured. Fixed by always building a
real val sampler (over the train indices when no val split is
given), with `shuffle=False, drop_last=False`:

```python
if val_indices is None:
    val_indices = train_indices
val_sampler = ManifestSampler(
    val_indices,
    shuffle=False,
    seed=sampler_seed,
    rank=rank,
    world_size=world_size,
    drop_last=False,
)
return train_sampler, val_sampler
```

physicsnemo/datapipes/transforms/mesh/transforms.py
```diff
@@ -31,6 +31,13 @@
 from physicsnemo.datapipes.transforms.subsample import poisson_sample_indices_fixed
 from physicsnemo.mesh import DomainMesh, Mesh
 
+### Local mirror of `Mesh`'s three public data sections, kept in this
+### module so transforms don't take a typing dependency on
+### `physicsnemo.mesh`. The runtime tuple is derived from the `Literal`
+### via `get_args` so the two stay in lockstep.
+_MeshSection: TypeAlias = Literal["point_data", "cell_data", "global_data"]
+_MESH_SECTION_NAMES: tuple[_MeshSection, ...] = get_args(_MeshSection)
+
```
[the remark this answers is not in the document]
peterdsharpe: Refactored into `mesh.py`

physicsnemo/datapipes/transforms/mesh/transforms.py
```diff
@@ -552,15 +557,13 @@ def __call__(self, mesh: Mesh) -> Mesh:
...
-            "cells": mesh.cells,
-            "point_data": mesh.point_data,
-            "cell_data": mesh.cell_data,
-            "global_data": mesh.global_data,
-        }
-        kwargs[self._section] = new_td
-        return Mesh(**kwargs)
+        ### `Mesh.copy` is a tensorclass-provided shallow copy: `points`,
+        ### `cells`, the untouched data sections, and the geometric `_cache`
+        ### (centroids / areas / normals) are all shared with `mesh`, then
```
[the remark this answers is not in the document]
peterdsharpe: Confirmed empirically that this isn't the case: `Mesh` is decorated
with `@tensorclass(tensor_only=True, shadow=True)` (no `frozen=True`),
and a minimal repro of `Demo.copy() + setattr` succeeds with the
original instance untouched:

```python
from tensordict import tensorclass
import torch

@tensorclass(tensor_only=True, shadow=True)
class Demo:
    a: torch.Tensor
    b: torch.Tensor

d = Demo(a=torch.tensor(1.0), b=torch.tensor(2.0), batch_size=[])
d2 = d.copy()
setattr(d2, "b", torch.tensor(99.0))
# d.b is still tensor(2.0); d2.b is tensor(99.0) - no FrozenInstanceError
```

The `mesh.copy() + setattr(...)` pattern is intentional: an explicit
`Mesh(points=..., cells=..., **{association: new_td})` constructor
would discard the geometric `_cache` (centroids / areas / normals),
forcing
a recompute on the next access. The shallow copy preserves the cache
because `points` and `cells` are unchanged. The `# ty: ignore` is
purely a static-typing accommodation (the tensorclass-injected
`copy()` isn't visible to the type checker, but it's part of the
public tensorclass API).

A nice follow-up for a future PR would be `Mesh.replace(**fields)` to
factor out this pattern and remove the `# ty: ignore`, but that's
orthogonal to this change.

peterdsharpe:
@coreyjadams answers:

> * how does the input mapping work, which was previously these bespoke functions and now is just yanking off a mesh in a helter-skelter set of python functions?  It seems a bit hard to trace in the code I'll admit, though I think I know what you're trying to do?

This input mapping uses a resolver that lives in `forward_kwargs.py` This basically replaces the `MODEL_MAPPINGS` registry. What's happening here is that the model YAML declares its `forward(self, **kwargs)``, as a function of the fields in the `DomainMesh`.

So, for GeoTransolver, that's:
```yaml
# train_geotransolver_automotive_surface.yaml
# -- Data contract -----------------------------------------------------------
# This model consumes batched flat tensors (B=1) and emits a (1, N, C) output
# tensor with channels concatenated in the dataset YAML's `targets:` order.
input_type: tensors
output_type: tensors

# Maps DomainMesh paths to GeoTransolver.forward kwargs. The dataset YAML
# (drivaer_ml_surface.yaml) produces a DomainMesh whose `interior` is a
# Mesh[0, 3] point cloud at the (subsampled) surface cell centroids and
# whose `boundaries.vehicle` is the original triangulated surface with
# precomputed cell normals.
forward_kwargs:
  geometry: interior.points
  local_embedding:
    - interior.points
    - boundaries.vehicle.cell_data.normals
  local_positions: interior.points
  global_embedding: global_data.U_inf
```

> * I want to make sure we have a way to _re-dimensionalize_ the data after predicting on it.

Re-added this back in in a recent commit, with `NormalizeMeshFields.inverse_td()` and `NonDimensionalizeByMetadata.inverse_td()` inside `nondim.py`. These also have `inverse_tensor` forms for callers that want to use that.

And we have `to_physical_units()`, which is a convenience wrapper that we can use in inference scripts :) 
 
> * I couldn't see how you are handling scalar vs vector inputs and targets here, can you elaborate for me?  Is it implicit off the mesh output now?

It's implicit, based on the dataset YAML, not off of the mesh output:
```yaml
# drivaer_ml_volume.yaml
targets:
  velocity: vector   # n_spatial_dims channels
  pressure: scalar   # 1 channel
  nut: scalar
```

peterdsharpe:
FYI, verified that training runs and converges well with default GeoT on DrivAerML surface.

coreyjadams:
OK, I don't *love* the design of using DomainMesh to hold the targets and inputs on interior and boundary, that feels ... I dunno.  Confusing.  But I understand the principle.  Perhaps we can add a note to the docstring that we've built this as a way to support the unified recipe for now, and users are encouraged to use the unified recipe, but MeshToDomainMesh should be considered "under-the-hood" functionality and not something to incorporate into downstream datapipes?  Or maybe we move it out of core and into the example?

I just get the feeling we will figure out a nicer way to do that as soon as we're not in a time crunch.  Otherwise, looks good to me.  Merge it when tests pass!

PR #1633, opened 2026-05-11
GLOBE Performance Improvements

peterdsharpe:
# PhysicsNeMo Pull Request

This PR is a focused performance pass on the experimental GLOBE model and its training scripts. No model architecture changes, no checkpoint compat changes; all numerics are unchanged at fp32-epsilon precision (a couple of internal aggregations now happen in fp64 - details in §5).

**Net results on 16x B200s, DrivAerML training with standalone recipe: per-sample time (load, forward, backward, step) goes from ~7.8 seconds to ~5.5 seconds. (-30%)**

## TL;DR

- **CPU↔GPU synchronization is the dominant win.** The hot path used to issue ~60+ implicit syncs per training step (`mem_get_info`, `.item()`, `if X.any():`, `tensor[bool_mask]`, …). Most have been eliminated.
- **Two `O(tree_depth)` Python loops are gone.** Per-node source aggregation and per-node strength accumulation are now single `cumsum` + range-subtract ops, exploiting the morton-sorted source ordering.
- **The four-phase `BarnesHutKernel` scatter is packed.** One `index_add_` per phase instead of one `scatter_add_` per phase per output field. This more than halves the backward-pass scatter cost on DrivAerML, where `indexing_backward` was ~23 % of total GPU time per step.
- **`GradScaler` is gone from the training scripts** (it's a no-op with bf16 autocast and forces a per-step inf/NaN-check sync).
- **CUDA allocator is now `expandable_segments:True`** to avoid synchronizing `cudaMalloc` / `cudaFree` round-trips under the chunked kernel workload.
- **Numerical-stability fix:** centroid / node-strength aggregations now cumsum in fp64. Was producing p99 ~100 % wrong leaf centroids at DrivAer scale (`N=1M`, coords ~5 m, `leaf_size=1`).
- **Several `torch.compile` graph breaks are fixed**, mostly as a side-effect of the changes above.
- **New optional `tree_build_device` knob** (`"cpu"` / `"cuda"` / `None`) for the small-problem regime where CUDA-launch latency dominates tree work.

## Why these changes matter (hot path context)

A GLOBE training step is dominated by two things:

1. **Per-sample geometry preprocessing.** Build one `ClusterTree` per BC type, then run a dual-tree Barnes-Hut traversal for every (src BC, dst BC) pair to produce a `DualInteractionPlan` describing which target/source point pairs interact in the near, far, and mixed-rank phases.
2. **`BarnesHutKernel` evaluation.** Apply MLPs to those pairs and `scatter_add` the results back to per-target buffers, four times (one for each near/far phase), once per multiscale branch, once per communication hyperlayer.

Profiling showed both phases were CPU-bound on launch/sync overhead rather than GPU-bound on the actual MLP and indexing work. The changes below attack that bottleneck.

---

## 1. Eliminating GPU↔CPU synchronization

Each entry below removes a sync point that profiling identified as significant. These are the same change in spirit (avoid materializing `int(tensor)` or branching on tensor contents in the hot path), applied to several different call sites.

### 1.1 `_auto_chunk_size`: `mem_get_info` → cached `total_memory`

`BarnesHutKernel._auto_chunk_size` ran on every kernel evaluation and called `torch.cuda.mem_get_info(device)`, which is a synchronizing driver query. Across the four phases × multiple branches × multiple hyperlayers, this was **~60 syncs per training step**. Replaced with a cached lookup of total device memory and a fixed-fraction budget (25 %):

```76:123:physicsnemo/experimental/models/globe/field_kernel.py
_CHUNK_MEMORY_BUDGET_PERCENT: Final[int] = 25


def _ceil_div(a: int, b: int) -> int:
    # ... pure-integer ceiling division ...
    return -(-a // b)


@torch.compiler.disable
@cache
def _device_total_memory_bytes(device: torch.device) -> int:
    # CUDA: torch.cuda.get_device_properties(device).total_memory
    # CPU:  psutil.virtual_memory().total
    ...

@torch.compiler.disable
def _device_chunk_budget_bytes(device: torch.device) -> int:
    return _device_total_memory_bytes(device) * _CHUNK_MEMORY_BUDGET_PERCENT // 100
```

Same swap is applied to `MultiscaleKernel.forward`'s branch-checkpoint decision.

The trade is "exactly fits in currently-free memory" → "fits in 25 % of total device memory". Fine on 100+ GB devices (e.g. GB200) where the non-kernel resident state is well under the remaining 75 %.

### 1.2 Smoothing radius: Python float → registered buffer

`Kernel._evaluate_interactions` was rebuilding a tensor on every forward pass:

```python
smoothing_radius = torch.tensor(self.smoothing_radius, device=device, dtype=dtype)
... (vectors * vectors).sum(dim=-1).apply(lambda x: x + smoothing_radius**2)
```

Now stored as a registered buffer and used directly with TensorDict broadcasting:

```238:252:physicsnemo/experimental/models/globe/field_kernel.py
        ### Pre-squared smoothing radius as a registered tensor buffer.
        ### A buffer (rather than a Python float) is required because
        ### ``_evaluate_interactions`` adds this scalar to a TensorDict
        ### inside a checkpoint sub-graph; Python free variables cannot
        ### be lifted across nested Dynamo SubgraphTracers (the lift
        ### chain bottoms out at the root tracer with no parent and
        ### asserts ``lift_tracked_freevar_to_input should not be
        ### called on root SubgraphTracer``).  As a buffer, the
        ### tensor is a tracked module attribute that Dynamo treats as
        ### a graph leaf, so no lift is needed.
        self.register_buffer(
            "_smoothing_radius_sq",
            torch.tensor(smoothing_radius**2, dtype=torch.float32),
            persistent=False,
        )
```

Eager: skips a tensor allocation and a `.apply(lambda)` per forward. Compile: closes the `torch.compile` graph break.

### 1.3 Cluster tree construction: per-level sync → end-of-loop compaction

`ClusterTree.from_points` used to do, on every level of every tree,

```python
leaf_indices = torch.where(is_leaf_seg)[0]
if len(leaf_indices) > 0:
    ...
```

That's a CPU-GPU sync per level × per tree (the in-code comment notes ~16 levels × 4 trees × 28 samples for DrivAerML). The loop now accumulates `(seg_node_ids, seg_starts, seg_sizes, is_leaf_seg)` per level and pays **one** boolean compaction after the loop:

```549:560:physicsnemo/experimental/models/globe/cluster_tree.py
            ### Defer leaf-segment processing to a single end-of-loop pass.
            # Per-iter ``torch.where(is_leaf_seg)[0]`` was a CPU-GPU sync
            # point on every level of every tree (~16 levels x 4 trees x
            # 28 samples).  We instead accumulate (node_id, start, size,
            # validity) for each segment seen during the loop and pay one
            # boolean compaction at the end.  ``torch.where(is_internal_seg)``
            # remains in-loop because the next iteration's active segments
            # are derived from this iteration's internals.
            leaf_seg_node_ids: list[torch.Tensor] = []
            leaf_seg_starts: list[torch.Tensor] = []
            leaf_seg_sizes: list[torch.Tensor] = []
            leaf_seg_validity: list[torch.Tensor] = []
```

The previous helpers `_fill_leaf_aabbs` and `_fill_leaf_total_areas` also each called `int(leaf_sizes.sum())`; they're merged into a single `_fill_leaf_aggregates` that shares the `_ragged_arange` mapping between AABB and area aggregation, removing the redundant `.sum()` sync.

### 1.4 Dual-tree traversal: ~12 `.any()`-gated indexings → one compaction

`ClusterTree.find_dual_interaction_pairs` is the hot loop of geometry prep. The previous body did 12 separate `if X.any():` early-exits and `tensor[bool_mask]` selections per iteration - each one a sync.

The new body classifies every active node-pair into one of three cases (`case_T_only`, `case_S_only`, `case_both`), computes all eight potential child slots with validity masks, stacks them, and pays **one** boolean compaction per iteration:

```981:1041:physicsnemo/experimental/models/globe/cluster_tree.py
                ### 3. Generate next iteration's active set.
                # We compute children over the FULL active set (n_active
                # entries) and use validity masks per (T,S) child slot to
                # encode the case-A / case-B / case-C splitting rules from
                # the original implementation.  After unioning the eight
                # potential child slots we pay ONE boolean compaction
                # instead of the original ~12 ``.any()``-gated indexings.
                ...
                slot_t = torch.stack([...])
                slot_s = torch.stack([...])
                slot_v = torch.stack([...])
                ...
                flat_v = slot_v.reshape(-1)
                active_tgt_nodes = slot_t.reshape(-1)[flat_v]
                active_src_nodes = slot_s.reshape(-1)[flat_v]
```

Far-field outputs are similarly accumulated unfiltered into per-iteration lists with a parallel `is_far` validity mask, then compacted ONCE at the very end of the traversal.

### 1.5 Leaf-pair expansion: dedup `nonzero` calls

In `_expand_dual_leaf_hits`, `tensor[bool_mask]` lowers to a synchronizing `aten::nonzero` that sizes the output. When several indexings share the same mask, the previous code paid one sync per indexing. They're now deduped:

```263:269:physicsnemo/experimental/models/globe/cluster_tree.py
    ### (near, far) output.  ``target_is_far`` is consumed by two
    ### indexings; doing one ``nonzero`` and reusing the integer index
    ### saves one sync (each ``tensor[bool_mask]`` lowers to a
    ### synchronizing ``aten::nonzero`` to size the output).
    far_idx_t = target_is_far.nonzero(as_tuple=True)[0]
    nf_target_ids = target_point_ids[far_idx_t]
    nf_source_node_ids = src_leaf_per_target[far_idx_t]
```

The function also drops a pile of "early return on empty input" branches that were each a sync; the downstream ops (`scatter`, `bincount`, `repeat_interleave`, `_ragged_arange`) all handle zero-element inputs correctly.

### 1.6 Traversal loop bound: `max_depth.item()` → `bit_length`

The dual-traversal loop bound was

```python
max_iters = int(target_tree.max_depth.item()) + int(source_tree.max_depth.item()) + 1
```

i.e. two `.item()` calls *before the loop starts*. Replaced with a static upper bound that needs no GPU read:

```885:892:physicsnemo/experimental/models/globe/cluster_tree.py
            ### Loop bound: every iteration descends at least one tree level
            ### on at least one side, so ``2 * total_levels + safety`` is a
            ### hard upper bound that requires no GPU->CPU read.  Using
            ### ``int(max_depth.item())`` as before would force two syncs
            ### per call before we even start the loop.
            n_src_levels = max(1, int(source_tree.n_sources).bit_length())
            n_tgt_levels = max(1, int(target_tree.n_sources).bit_length())
            max_iters = 2 * (n_src_levels + n_tgt_levels) + 4
```

`n_sources` is a Python `int` (a `.shape[0]` lookup), not a tensor.

### 1.7 Tree-build wrapper: `no_grad` + minor cleanups

`GLOBE._build_trees_and_plans` and `_build_prediction_plans` now run inside `torch.no_grad()` (the outputs are integer indices and a non-grad-tracked area divisor, so autograd bookkeeping on dozens of tensor ops was wasted) and inline the per-plan debug logging into the build loop instead of re-iterating afterward.

---

## 2. Prefix-sum aggregation: replacing Python loops over tree levels

A cluster tree's morton-sorted source ordering has a key property: every node covers a *contiguous range* of sources, `[node_range_start, node_range_start + node_range_count)`. So any node-subtree sum is a prefix-sum range subtract: `prefix[end] - prefix[start]`.

This kills two `O(tree_depth)` Python loops that were previously the dominant cost in their respective functions.

### 2.1 `ClusterTree.compute_source_aggregates`

The previous implementation did:
- Per-leaf weighted-sum via segmented scatter.
- A Python-level `for level_ids in reversed(depth_levels):` propagating centroids and per-source features bottom-up.

Combined cost in profiling: **~2 s per training step** of mixed CPU (level-loop overhead) and GPU (kernel launches per level).

Replaced with a single `cumsum` + range-subtract over the morton-sorted sources:

```718:752:physicsnemo/experimental/models/globe/cluster_tree.py
        ### Range-sum aggregation via morton-sorted prefix sums.
        # Each node covers a contiguous range
        # [node_range_start, node_range_start + node_range_count) in
        # morton-sorted source order, so any node-subtree sum is just
        # ``prefix[end] - prefix[start]``.  This replaces the old
        # leaf-aggregation + bottom-up Python loop, which were the
        # dominant CPU + GPU costs in ``compute_source_aggregates``
        # (~2 s combined per training step in profiling).
        ...
        sorted_points = source_points[self.sorted_source_order]
        sorted_areas = areas[self.sorted_source_order]
        weighted_points_64 = (sorted_points * sorted_areas.unsqueeze(-1)).double()
        cumsum_weighted_points = torch.nn.functional.pad(
            torch.cumsum(weighted_points_64, dim=0), (0, 0, 1, 0)
        )
        ...
        node_total_weighted_pts = (
            cumsum_weighted_points[ends] - cumsum_weighted_points[starts]
        )
```

Per-source feature aggregation gets the same treatment via an inner `_aggregate_via_prefix_sum`, applied across TensorDict leaves. Trailing feature dims are flattened so cumsum sees a single feature axis (avoiding a per-feature kernel chain inside `cumsum`).

The `internal_level_ids` / `internal_level_offsets` / `leaf_node_ids` / `leaf_seg_ids` tensors stored on `ClusterTree` purely to support the old bottom-up loop are dropped.

### 2.2 `BarnesHutKernel._compute_node_strengths`

Same story, on a single scalar field (`source_strengths`). The old implementation did leaf-summing via scatter then a bottom-up Python loop; the new one is one `cumsum` + range-subtract:

```1356:1369:physicsnemo/experimental/models/globe/field_kernel.py
        ### Cumsum and range-subtract in fp64 to avoid catastrophic
        ### cancellation when ``cumsum_total >> range_sum`` - the regime
        ### of small leaves in a large tree built over offset coordinates.
        ### See the matching note in :meth:`ClusterTree.compute_source_aggregates`.
        sorted_strengths_64 = source_strengths[tree.sorted_source_order].double()
        ### Pad with a leading zero so that ``cumsum[i]`` is the sum of
        ### sorted_strengths[:i] - both endpoints index identically.
        prefix_sum = torch.nn.functional.pad(
            torch.cumsum(sorted_strengths_64, dim=0), (1, 0)
        )

        starts = tree.node_range_start
        ends = starts + tree.node_range_count
        return (prefix_sum[ends] - prefix_sum[starts]).to(source_strengths.dtype)
```

---

## 3. Packed scatter for backward-pass efficiency

`BarnesHutKernel` scatters into per-target output buffers four times per forward (near, far-node, near-far, far-near), once per output field. In profiling, `indexing_backward_kernel_stride_1` (the backward of `scatter_add_`) was **~1 s out of ~4.5 s total GPU time per training step (~23 %)** - the largest single GPU kernel by time.

Two changes in `BarnesHutKernel.forward`:

1. **Pack output fields** into a single `(n_targets, total_features)` buffer instead of per-field buffers.
2. **Use `index_add_` instead of `scatter_add_`** for the per-target accumulation.

```1032:1056:physicsnemo/experimental/models/globe/field_kernel.py
        ### Packed output buffer.  All four phases scatter into this single
        ### tensor; the per-phase loops over output keys (one ``scatter_add_``
        ### per key) are replaced with one packed ``scatter_add_`` per phase.
        ### ``indexing_backward`` was the top GPU kernel by time in profiling
        ### (~23% of total GPU time); packing cuts this by the number of
        ### output fields (~4x for ``C_p`` + ``C_f`` on DrivAerML).
        ...
        packed_buf = torch.zeros(
            (n_targets, total_features), dtype=buffer_dtype, device=device
        )
```

The pack/scatter helper is shared across all four phases:

```1232:1290:physicsnemo/experimental/models/globe/field_kernel.py
    def _pack_and_scatter(
        self,
        chunk_result: TensorDict,
        weights: Float[torch.Tensor, " n_pairs"],
        tgt_ids: Int[torch.Tensor, " n_pairs_or_expanded"],
        packed_buf: Float[torch.Tensor, "n_targets total_features"],
        *,
        broadcast_pair_ids: Int[torch.Tensor, " n_pairs_or_expanded"] | None = None,
    ) -> None:
        ...
        ### ``index_add_`` rather than ``scatter_add_`` with broadcasted
        ### indices: equivalent semantics, but ``index_add_`` takes a 1-D
        ### ``index`` and avoids the ``unsqueeze`` + ``expand_as`` overhead
        ### that ``scatter_add_`` requires.  In addition to the direct
        ### speedup, ``index_add_`` has a more compact backward (no index
        ### broadcasting in the saved tensors), which compounds with the
        ### packing optimization above.
        packed_buf.index_add_(0, tgt_ids, weighted)
```

Counterpart `_unpack_buf` slices back into a per-field TensorDict. The canonical-ordered field list and per-key feature widths are cached as `Kernel._output_packing`.

Net effect on DrivAerML (`C_p` 1ch + `C_f` 3ch = 4 fields, 4 phases): 16 `scatter_add_` calls → 4 `index_add_` calls per kernel evaluation.

---

### 4.1 CUDA allocator: `expandable_segments:True`

Both `run.sh` scripts now export

```bash
export PYTORCH_CUDA_ALLOC_CONF="expandable_segments:True"
```

The chunked kernel evaluations stress the cache enough that the default segment allocator falls back to synchronous `cudaMalloc` / `cudaFree` round-trips. `expandable_segments` lets the allocator grow existing segments instead.

### 4.2 Gradient checkpointing applies in both train and eval

The four `if self.training and self.use_gradient_checkpointing` branches across `BarnesHutKernel.forward` are hoisted into one helper:

```1195:1230:physicsnemo/experimental/models/globe/field_kernel.py
    def _maybe_checkpointed_evaluate(self, *args: object) -> TensorDict:
        """Run :meth:`_gather_and_evaluate` with optional gradient checkpointing.

        ...

        The wrap fires in **both** training and eval, not just training,
        for two reasons:

        1. Memory savings (the original purpose).  In eval no autograd
           tape is active (``train.py`` wraps validation in
           ``torch.no_grad()``), so ``checkpoint(use_reentrant=False)``
           degenerates to a near-no-op forward call - no recompute, no
           extra memory, no extra compute.
        2. Workaround for a Dynamo+CUDA TensorDict bug.  When the same
           ``_gather_and_evaluate`` body is inlined into the parent
           graph (i.e. without the checkpoint wrapper), FakeTensor
           tracing fails to propagate ``TensorDict.batch_size`` through
           ``(vectors * vectors).sum(dim=-1)`` ...
        """
        if self.use_gradient_checkpointing:
            return checkpoint(
                self._gather_and_evaluate, *args, use_reentrant=False
            )
        return self._gather_and_evaluate(*args)
```

The `self.use_gradient_checkpointing` flag is also piped through `MultiscaleKernel` → `Kernel` → `BarnesHutKernel` and exposed as a constructor kwarg on `GLOBE` (default `True`).

### 4.3 Packed output buffer

`BarnesHutKernel.forward` used to lazily allocate per-field, per-phase output buffers (`output_bufs[k] = torch.zeros(...)` if-key-not-seen pattern). Now allocated up-front:

```1039:1056:physicsnemo/experimental/models/globe/field_kernel.py
        ### Buffer dtype must match the dtype of ``weighted = chunk * weights``
        ### that the four phases will scatter in (``index_add_`` requires
        ### exact dtype match).  ``chunk`` comes from the MLP at the
        ### autocast dtype (or ``source_points.dtype`` outside autocast);
        ### ``weights`` carries ``source_strengths.dtype``.  Their product
        ### is the type-promoted dtype, which is what the previous
        ### lazy-allocated buffers used to capture - we now compute it
        ### eagerly so a single buffer can be allocated up front.
        ...
        packed_buf = torch.zeros(
            (n_targets, total_features), dtype=buffer_dtype, device=device
        )
```

Composes with §3 (packed scatter) and lets us drop the `if n_near == 0 and n_nf == 0 and n_fn == 0 and n_far_nodes == 0` empty-result short-circuit at the top of `forward`: an all-zero `packed_buf` unpacks to all-zero per-field tensors, which is the correct result anyway.

---

## 5. Numerical stability via fp64 cumsum

The cumsum-based aggregations in §2 are done in `float64` and cast back to `source_points.dtype` at the end. Reason from the code:

```727:737:physicsnemo/experimental/models/globe/cluster_tree.py
        # The cumsum and the range subtract are done in fp64 because fp32
        # suffers catastrophic cancellation when ``range_sum << cumsum_total``,
        # which is the regime of small leaves (``leaf_size=1``) in a large
        # tree built over offset (e.g. all-positive) coordinates.  At
        # drivaer scale (``N=1M``, coords ~5 m), fp32 leaf centroids had
        # median ~2 % relative error and p99 ~100 % wrong.  Lifting the
        # cumsum to fp64 brings this back to fp32 epsilon (~1e-7) and adds
        # <1 % wall-clock to the training step (cumsum is ~2.3x slower in
        # fp64, but cumsum is a tiny fraction of step time).  CUDA fp32
        # cumsum is also non-deterministic across runs (pytorch#75240);
        # fp64 cumsum is much less affected.
```

Listed under "performance" because it's a prerequisite for §2 - without it the prefix-sum trick can't be used.

---

## 6. `torch.compile` compatibility

Most of the syncs above (§1) and the packed-scatter rewrite (§3) also fix `torch.compile` graph breaks. A few changes here are specifically for `torch.compile`:

- **`_ceil_div` for chunk sizing** instead of `math.ceil(a / b)`. Under `specialize_float=False`, a Python float gets traced as an unbacked symbolic float that propagates into the chunk size and crashes Dynamo at the `checkpoint` boundary in `_gather_and_evaluate`. Pure-integer ceiling division (`-(-a // b)`) sidesteps this.
- **`_CHUNK_MEMORY_BUDGET_PERCENT: Final[int] = 25`** instead of e.g. `0.25` - same reason.
- **`@torch.compiler.disable` on the device-memory helpers and `_auto_chunk_size`**, so Dynamo doesn't try to evaluate `torch.cuda.get_device_properties` as a graph constant (which crashes on CPU-only hosts and produces unbacked symbolic ints on CUDA hosts).
- **The smoothing-radius buffer** from §1.2 also fixes the lift-across- SubgraphTracers issue.
- **`view(-1, 1)` → `reshape(-1, 1)`** in `GLOBE.forward`'s output calibration loop, for robustness when upstream ops produce non-contiguous tensors.

The `hierarchical_acceleration.md` design doc is updated to reflect the new "static fraction of total device memory" chunk-size strategy.

---

## 7. Training scripts

Both `examples/cfd/external_aerodynamics/globe/airfrans/train.py` and `.../drivaer/train.py` have the same three changes.

### 7.1 Removed `GradScaler`

GradScaler is only useful for fp16 autocast (to combat gradient underflow). The training scripts use **`torch.bfloat16` autocast**, which has the same dynamic range as fp32 - there's no underflow to protect against. When `enabled=amp`, `GradScaler` still forces a per-step inf/NaN check sync on the gradients, for nothing.

```diff
-    scaler = torch.amp.GradScaler(device=device.type, enabled=amp)
-    ...
-    scaler.scale(batch_loss).backward()
+    batch_loss.backward()
     if gradient_clip_norm is not None:
-        scaler.unscale_(optimizer)
         torch.nn.utils.clip_grad_norm_(
             model.parameters(), max_norm=gradient_clip_norm
         )
-    scaler.step(optimizer)
-    scaler.update()
+    optimizer.step()
```

(Removed from checkpoint save/load plumbing too.)

### 7.2 New `tree_build_device` knob

Exposed as a `Literal["cpu", "cuda"] | None` argument:

```281:288:examples/cfd/external_aerodynamics/globe/drivaer/train.py
        leaf_size=leaf_size,
        network_type=network_type,
        self_regularization_beta=self_regularization_beta,
        latent_compression_scale=latent_compression_scale,
        expand_far_targets=expand_far_targets,
        tree_build_device=tree_build_device,
    ).to(device)
```

Forwarded to `GLOBE.__init__`, which transfers boundary centroids/areas to `build_device` for `from_points` + `find_dual_interaction_pairs`, then moves the resulting `ClusterTree` and `DualInteractionPlan` (both `@tensorclass`, so one `.to()` call each) back to the input's device at the end of `_build_trees_and_plans` / `_build_prediction_plans`.

Default `None` preserves prior behavior (build on the input's device). `"cpu"` can win for small problems (~a few thousand boundary cells) where CUDA launch latency + `cudaStreamSynchronize` round-trips dominate the actual tree work.

### 7.3 CUDA allocator env var (`run.sh`)

See §4.1.

---

## 8. Dependencies (`pyproject.toml`)

```diff
-    "tensordict>=0.10.0",
+    "tensordict>=0.12.2",
     "omegaconf>=2.3.0",
     "importlib-metadata>=8.7.1",
+    "psutil>=6.0.0",
```

- **`tensordict>=0.12.2`** (was `>=0.10.0`) - bumped alongside the `BarnesHutKernel` refactor; the codebase exercises TensorDict broadcasting and batch-size propagation more heavily than before (see §4.2 for one such case in `_maybe_checkpointed_evaluate`), and this update brings better `torch.compile` support.
- **`psutil>=6.0.0`** - used by `_device_total_memory_bytes` for the CPU branch (`psutil.virtual_memory().total`). The CPU build path is debug-only; production runs on CUDA, where the CUDA branch is used.

---

## What's intentionally not in this PR

- No model-architecture changes. No new layers, no new losses, no hyperparameter changes.
- No checkpoint format changes - existing `.mdlus` checkpoints load unmodified.
- The two new constructor kwargs on `GLOBE` (`use_gradient_checkpointing`, `tree_build_device`) both have defaults matching the prior behavior, so this is a non-breaking change for downstream users.

## Checklist

## Review Process


examples/cfd/external_aerodynamics/globe/airfrans/train.py
```diff
@@ -347,16 +351,13 @@ def main(
         min_lr=learning_rate / 64,
         threshold=1e-3,
     )
-    scaler = torch.amp.GradScaler(device=device.type, enabled=amp)
-
```
coreyjadams, line 351: Are you dropping grad scaling?

physicsnemo/experimental/models/globe/cluster_tree.py
```diff
@@ -271,12 +231,13 @@ def _expand_dual_leaf_hits(
...
-
-    if n_pairs == 0:
-        return _empty_result()
+    ### This function is intentionally written without ``if X.any():`` /
+    ### ``int(X.sum())`` early-exit branches.  Each such branch was a
+    ### CPU-GPU sync point in the dual-traversal hot loop; the sync count
+    ### was the dominant CPU stall in profiling.  All downstream operations
+    ### (boolean indexing, ``_ragged_arange``, ``argsort``, ``scatter``,
+    ### ``torch.bincount``) handle zero-element inputs correctly, so we
+    ### let empty intermediate tensors flow through unconditionally.
```
coreyjadams: This is a good GPU optimization, generally.

physicsnemo/experimental/models/globe/field_kernel.py
```diff
@@ -1138,35 +1185,156 @@ def forward(
...
+           batch size, and any downstream op that consults
+           ``batch_size`` (``.unsqueeze(-1)``, ``concatenate_leaves``'s
+           reshape) trips on the inconsistency.  Wrapping the call in
+           ``checkpoint`` gives Dynamo a fresh sub-tracer scope that
+           tracks the post-reduction batch size correctly, so the
+           identical eager-mode-correct body now also traces correctly.
+        """
+        if self.use_gradient_checkpointing:
+            return checkpoint(
+                self._gather_and_evaluate, *args, use_reentrant=False
```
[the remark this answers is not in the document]
peterdsharpe: I don't think an API note is necessary - nothing changes from an external user's perspective.

physicsnemo/experimental/models/globe/field_kernel.py
```diff
@@ -57,6 +58,71 @@
...
+### Static, per-device memory-budget helpers used by chunk-size sizing.
+
+# Fraction of total device memory we are willing to spend on a single
+# kernel evaluation's chunked autograd state (peak intermediate footprint
+# for one (Phase A | B | C | D) call).  Conservative because parameters,
+# DDP buckets, and activations from other kernels coexist on-device.
+#
+# Expressed as an integer percent (rather than e.g. ``0.25``) so that
+# ``_device_chunk_budget_bytes`` stays in pure-integer arithmetic.  Under
+# ``torch.compile`` with the default ``specialize_float=False``, a
```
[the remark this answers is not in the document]
peterdsharpe: I think 25% is low enough that this is a non-issue. We can modify later if it becomes one.

physicsnemo/experimental/models/globe/cluster_tree.py
```diff
@@ -341,13 +307,15 @@ def _empty_result():
...
     has_fn_source = torch.zeros(n_pairs, dtype=torch.bool, device=device)
-    if fn_lp_ids.numel() > 0:
-        has_fn_source[fn_lp_ids] = True
+    has_fn_source[fn_lp_ids] = True
     fn_active_mask = has_fn_source[surv_lp_ids]
 
-    active_surv_ids = surv_point_ids[fn_active_mask]
-    active_surv_lp_ids = surv_lp_ids[fn_active_mask]
+    ### Shared mask -> dedup the boolean indexing (saves one sync).
+    fn_active_idx = fn_active_mask.nonzero(as_tuple=True)[0]
```
coreyjadams: You have a blocking `nonzero` here, and another just a few lines later.  However, if the masks are relatively small, you could concat the `fn_active_mask` and `~src_is_far` tensors, `nonzero` the whole thing, and only introduce one block.

physicsnemo/experimental/models/globe/cluster_tree.py
```diff
@@ -647,30 +573,11 @@ def from_points(
...
-
-                    # Compute leaf total areas
-                    _fill_leaf_total_areas(
-                        leaf_nids, l_starts, l_sizes, sorted_areas, total_area_buf
-                    )
+                ### Defer leaf processing: accumulate, compact once at end.
+                leaf_seg_node_ids.append(seg_node_ids)
+                leaf_seg_starts.append(seg_starts)
+                leaf_seg_sizes.append(seg_sizes)
+                leaf_seg_validity.append(is_leaf_seg)
```
coreyjadams, line 756: Excellent.

test/models/globe/test_barnes_hut_kernel.py
```diff
@@ -1548,5 +1547,211 @@ def test_four_quadrant_per_category_accuracy(
...
+
+    Profiling identified ``cudaStreamSynchronize`` as the dominant CPU
+    stall in GLOBE training (9919 calls / ~3 s in a 10.5 s training step,
+    most of them from ``find_dual_interaction_pairs``).  The perf-pass
+    moved every output stream (near/far/(near,far)/(far,near)/broadcast)
+    to a deferred-compaction protocol where the per-iteration tensors
+    are accumulated unfiltered with a validity mask, and a single
+    boolean compaction is paid per stream at the end of the traversal.
+
+    With those changes the structural syncs remaining per call are:
```
coreyjadams: How long does this test take to run?  And, it should probably guard against no cuda right?  it can just pass `device` (it's a fixture) and skip if device is cpu.

test/models/globe/test_barnes_hut_kernel.py
```diff
@@ -1548,5 +1547,211 @@ def test_four_quadrant_per_category_accuracy(
...
+        record_shapes=False,
+    ) as prof:
+        tree = ClusterTree.from_points(points, leaf_size=leaf_size)
+        tree.find_dual_interaction_pairs(target_tree=tree, theta=theta)
+        torch.cuda.synchronize()
+
+    sync_count = sum(1 for ev in prof.events() if ev.name == "cudaStreamSynchronize")
+
+    ### Observed at time of test authorship on an RTX 4090 with
+    ### PyTorch 2.11: 59 / 67 / 83 syncs for the three parametrisations
```
coreyjadams: This just seems really fragile, since isn't it dynamic with the data values and the depth of traversal it has to do?

coreyjadams (changes requested):
Hey @peterdsharpe glad to see GLOBE marching towards better and better performance.  My hang up here is actually not in the model but in the tests.  I honestly have not-warm-fuzzy feelings about building tests that rely on profiling and counting string-based occurrences.  I think it's probably fragile and slow, too.  Can we drop that test or find another path?

In general we don't have strict performance requirements in the test suite on our other models, not sure why we would enforce it on GLOBE: it's better to use the tests for integration testing on correctness IMO.  What do you think?

examples/cfd/external_aerodynamics/globe/airfrans/train.py
peterdsharpe: Yes! It costs a surprisingly-large 60 milliseconds per iter, and has no effect at bf16 precision (since the exponent is unchanged) - it only matters at fp8 or fp16, neither of which we use.

physicsnemo/experimental/models/globe/cluster_tree.py
peterdsharpe: Superseded by 3838eca6, which now has zero `nonzero` calls in this section

peterdsharpe:
Dropping the sync-counting test!

coreyjadams (approved):

peterdsharpe:
Note: to be merged after `rc-2.1.0` back-merges into `main`; depends on torch 2.12 support.

ktangsali (approved):
Changes to TOML look good

peterdsharpe:
/ok to test [f5de94e](https://github.com/NVIDIA/physicsnemo/pull/1633/commits/f5de94ea564aa743ddf06e1e2743128f94a7e964)

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1633/commits/3e8e4242d18344e0979bb0e9ced0f7e09aa7ada5

peterdsharpe:
/ok to test [661a851](https://github.com/NVIDIA/physicsnemo/pull/1633/commits/661a851b38cc316e72f2c6daf2abdf9891aaad10)

PR #1645, opened 2026-05-13
Centralize cache delete-and-push mechanism to one place

coreyjadams:
# PhysicsNeMo Pull Request

## Checklist

## Review Process


.github/actions/setup-uv-env/action.yml
```diff
@@ -212,16 +234,19 @@ runs:
...
-          "multi-storage-client[boto3]>=0.33.0" \
-          "tensorstore" \
-          "pyarrow"
+          "moto[s3]==5.2.1" \
+          "numpy-stl==3.2.0" \
+          "scikit-image==0.26.0" \
+          "shapely==2.1.2" \
+          "multi-storage-client[boto3]==0.48.0" \
+          "tensorstore==0.1.83" \
+          "pyarrow==24.0.0"
```
coreyjadams: @NickGeneva @laserkelvin and also @peterdsharpe there has been a little discussion about pinning here, vs. pinning in pyproject.toml.  Summarizing some pros and cons.

Why pin?  If we don't pin, and one of these updates, the nightly build will get out of sync with the PR venv and it will trigger a rebuild of the environment (slow on the PR on the GPU nodes) and trigger ALL tests to run (also slow) because the testmon DB requires the venv to match.  So, pinning is a good idea IMO.

We can pin here, and that is nice because it's not disruptive to pyproject.toml, and can control our CI system independently.  We already have that in blossom since we run in a container, and the installed packages are not necessarily aligned with what's in uv.lock.  I contend that is OK.  On the other hand, we might want to be able to control the CI env tightly against the `uv.lock` file for some reason?

We can pin in `pyproject.toml` by creating all of these deps in `ci-deps` development group with specified numbers.  That's an update to pyproject.toml (no big deal) and extra lock resolution (no big deal) but any changes to CI env will have to go through that instead.  And changes to pyproject.toml are meant, deliberately, to invalidate the `testmon` db and trigger all tests to rerun, for what it's worth (and I like that design, updating pyproject.toml in physicsnemo should be painful).

A middle ground might be to put these in a ci-requirements.txt or similar that is contained?

coreyjadams:
This PR effectively is doing two things:

1. Consolidate the logic of delete-then-upload to refresh immutable caches for the various caching mechanisms I've set up to accelerate our CI.  Since there are now several, it made sense to turn it into a custom action that can be reused.

2. Pin CI dependencies to specific versions.

If needed I'll split these up.

.github/actions/replace-cache/action.yml
```diff
@@ -0,0 +1,122 @@
...
+      env:
+        GH_TOKEN: ${{ inputs.github-token }}
+        CACHE_KEY: ${{ inputs.key }}
+        CACHE_DESC: ${{ inputs.description }}
+        REPO: ${{ github.repository }}
+      run: |
+        set -euo pipefail
+        if ! command -v gh >/dev/null 2>&1; then
+          echo "::error::gh CLI not on PATH; cannot manage ${CACHE_DESC} slot."
+          exit 1
```
peterdsharpe, line 73: The guard is probably not necessary if you have `-euo pipefail` already, but this also doesn't hurt

.github/actions/replace-cache/action.yml
```diff
@@ -0,0 +1,122 @@
...
+        # seconds before failing.
+        for attempt in 1 2 3 4 5; do
+          if gh cache list --repo "$REPO" --key "$CACHE_KEY" --json key --jq '.[].key' \
+              | grep -Fxq "$CACHE_KEY"; then
+            echo "${CACHE_DESC} present: $CACHE_KEY"
+            exit 0
+          fi
+          echo "attempt $attempt: ${CACHE_DESC} not yet visible, sleeping..."
+          sleep 5
+        done
```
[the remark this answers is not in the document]
peterdsharpe: Agree, recursive retry with exponential backoff is probably mildly better here

.github/actions/setup-uv-env/action.yml
peterdsharpe: This is a good analysis!

I agree that pinning overall is a good idea.

As for how to implement it, I think all three strategies are viable (`pyproject.toml`, here, or pulled out into a `ci-requirements.txt`) and I'd approve of any of them. I've been mulling over all three options for the past ~5 mins in my head and struggle to come up with any truly airtight ideas for why one is better than the others.

peterdsharpe (approved):
Looks good! Interesting discussion about how to implement version-pinning; I think any of the three presented options are perfectly fine (including as-is).

PR #1682, opened 2026-05-28
Enable Compile for Shard Tensor

coreyjadams:
# PhysicsNeMo Pull Request

This PR enables compilation for ShardTensor.  It depends on the refactor of shard tensor #1556 still - that has to merge first - so the diff will be a little smaller in the end, since that PR is included here too for now.

The main requirements for enabling torch.compile are:
- ensuring we use the newer `setup_context` style autograd functions everywhere
- avoiding `torch.Size` in favor of `int`.  This primarily impacts the ShardTensorSpec objects.
- ensuring `torch_flatten` and it's inverse are available for tracing
- making available a function for coercing the tangent objects for the backward pass to the same shape as the objects themselves.  For ShardTensor, care is taken to avoid non-traceable redistributions here.  We also needed to cache shard hints for the backward pass in a way that was compiler friendly.

There is also a minor bug fix here in the knn op that shows up in shard tensor spec determining the size of sharding shapes automatically.

Two test suites are added: one to test compilation "in general" and one to verify that each op we're wrapping is tracable via the compiler.  It's implicit that DTensor fallback ops are compilable, though verification of that is not really done by us ... 

## Checklist

## Review Process


coreyjadams:
One other notable thing here: our ring attention implementation is not traceable at the moment.  streams and overlapped communication / computation is not traceable.  I think we can resolve that though it's not here quite yet.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -962,9 +1257,48 @@ def backward(self, *args, **kwargs):
...
+    device_mesh: DeviceMesh,
+    partition_fn: (Callable[[str, nn.Module, DeviceMesh], None] | None) = None,
+) -> nn.Module:
+    """Distribute a module over a domain mesh and adapt outputs for FSDP."""
+    distributed_module = distribute_module(
+        module,
+        device_mesh=device_mesh,
+        partition_fn=partition_fn,
+    )
+    return wrap_for_fsdp(distributed_module)
```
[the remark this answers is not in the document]
coreyjadams: This is getting purged.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -448,15 +463,25 @@ def forward(
         torch.Tensor
             Local tensor representation of the ShardTensor.
         """
-        ctx.shard_tensor_spec = input._spec
-        ctx.grad_placements = grad_placements
+        # # JUST LIKE DTENSOR:
+        # # We need to return a fresh Tensor object there as autograd metadata
+        # # will be inplaced into it. So we don't want to pollute the Tensor
+        # # object stored in the _local_tensor of this ShardTensor.
+        # return local_tensor.view_as(local_tensor)
```
pzharrington: nit: double commented block here / dangling `return ...` commented out

physicsnemo/domain_parallel/shard_utils/attention_patches.py
```diff
@@ -724,6 +743,30 @@ def ring_sdpa(
...
+
+    Eager (no ``torch.compile``) usage works as designed: this is the
+    overlap-K/V-with-compute attention kernel used by sharded models in
+    production. Compile support for sharded attention requires a separate
+    refactor of the ring (drop ``record_stream``, switch
+    ``perform_ring_iteration`` to functional p2p collectives, replace the
+    explicit ``cuda.stream`` overlap with implicit collective overlap).
+    Until then, callers that need ``torch.compile`` must either keep
+    sharded attention outside the compiled region or avoid sharded
+    attention entirely.
```
pzharrington, line 768: Dang ☹️

coreyjadams: We'll get it in the next update!

physicsnemo/domain_parallel/shard_tensor.py
coreyjadams: I swear I've killed this twice and it keeps coming back via git merge.

coreyjadams:
/ok to test [742bed3](https://github.com/NVIDIA/physicsnemo/pull/1682/commits/742bed37f8584ead9b5b993a46eea72e75d71456)

examples/minimal/ShardTensorExamples/5_vit_training_loop/training_script.py
```diff
@@ -29,44 +30,88 @@
...
-from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
-from torch.distributed.tensor.placement_types import (  # noqa: E402
+from torch.distributed.tensor import DTensor, distribute_tensor
+from torch.distributed.tensor.placement_types import (
     Replicate,
     Shard,
 )
 
 
-def partition_model(name, submodule, device_mesh):
```
pzharrington, line 47: I would argue that the dropping of `distribute_module` is a backward-compat breaking change (if that pattern no longer works), as this will require modifications to existing training loops. It seems like the combination of `broadcast_plain_params` + `shard_<my_sharded_params>` is the new recommended pattern?

pzharrington: If so, we should mark the changelog to warn about breaking changes and try to communicate more loudly in docs IMO

examples/minimal/ShardTensorExamples/5_vit_training_loop/training_script.py
```diff
@@ -29,44 +30,88 @@
...
+    under FSDP2: all FSDP-managed submodules share one dynamo wrapper frame,
+    so the norm/MLP/head compilations look like one function called with
+    changing shapes, and automatic-dynamic would promote sizes to SymInts.
+    """
+    model.patch_embed = torch.compile(model.patch_embed, dynamic=False)
+    for block in model.stages:
+        block.norm1 = torch.compile(block.norm1, dynamic=False)
+        block.norm2 = torch.compile(block.norm2, dynamic=False)
+        block.mlp = torch.compile(block.mlp, dynamic=False)
+    model.head = torch.compile(model.head, dynamic=False)
```
pzharrington: I suggest pivoting to something compileable like natten to keep the example clean...when the updates come in for the sharded attn we can add it back in?

examples/minimal/ShardTensorExamples/5_vit_training_loop/training_script.py
pzharrington: And maybe provide some boilerplate helpers to import so people don't need to redefine these things every train loop?

coreyjadams:
/ok to test [8c91e84](https://github.com/NVIDIA/physicsnemo/pull/1682/commits/8c91e845ee0620680a3e64dfb5d422edea9bc213)

examples/minimal/ShardTensorExamples/5_vit_training_loop/training_script.py
coreyjadams: We've got natten and conv available now.  ViT is also still there.

examples/minimal/ShardTensorExamples/5_vit_training_loop/training_script.py
coreyjadams: Boilerplate has been added; note in the api added; full docs update will come in a PR to the docs repo.

coreyjadams:
/ok to test [01197c9](https://github.com/NVIDIA/physicsnemo/pull/1682/commits/01197c9fddb47dad8c8c6ae19cebc719287999e9)

docs/api/physicsnemo.domain_parallel.rst
```diff
@@ -2,17 +2,30 @@
...
-``ShardTensor`` provides a distributed tensor implementation that supports uneven sharding across devices. 
-It builds on PyTorch's DTensor while adding flexibility for cases where different ranks may have 
-different local tensor sizes.
+``ShardTensor`` provides a distributed tensor implementation that supports uneven sharding across devices.
+It is a subclass of ``torch.Tensor`` that interoperates with PyTorch's DTensor and plain tensors,
+while adding flexibility for cases where different ranks may have different local tensor sizes.
 
-.. autosummary::
-   :toctree: generated
+A key feature of ``ShardTensor`` is automatic promotion: when a plain ``nn.Module`` weight meets a
```
megnvidia:
```suggestion
A key feature of ``ShardTensor`` is automatic promotion. When a plain ``nn.Module`` weight meets a
```

docs/api/physicsnemo.domain_parallel.rst
```diff
@@ -2,17 +2,30 @@
...
+sharded activation, the weight is automatically promoted to a replicated distributed tensor, and its
```
megnvidia:
```suggestion
sharded activation, ``ShardTensor`` promotes the weight to a replicated distributed tensor and
```

docs/api/physicsnemo.domain_parallel.rst
```diff
@@ -2,17 +2,30 @@
...
+gradient is reduced over the domain mesh in the backward pass.  This means completely vanilla
```
megnvidia:
```suggestion
reduces its gradient over the domain mesh in the backward pass. Standard ``nn.Module`` models
```

docs/api/physicsnemo.domain_parallel.rst
```diff
@@ -2,17 +2,30 @@
...
+``nn.Module`` models work unmodified on sharded inputs - ``distribute_module`` is no longer needed
+or recommended.  For an additional data parallel axis, use DDP when parameters are plain tensors,
```
megnvidia:
```suggestion
therefore work unmodified on sharded inputs. ``distribute_module`` is no longer needed or```

File: docs/api/physicsnemo.domain_parallel.rst [hunk 10]
```diff
@@ -2,17 +2,30 @@
...
+or FSDP2 (``torch.distributed.fsdp.fully_shard``) when parameters are sharded.  ``torch.compile``
```
megnvidia:
```suggestion
recommended. For an additional data parallel axis, use Distributed Data Parallel (DDP) when
```

docs/api/physicsnemo.domain_parallel.rst
```diff
@@ -2,17 +2,30 @@
...
+is supported, with the caveat that sequence-sharded ring attention must stay outside compiled regions.
```
megnvidia, line 22:
```suggestion
parameters are plain tensors, or Fully Sharded Data Parallel 2 (FSDP2)
(``torch.distributed.fsdp.fully_shard``) when parameters are sharded. ``torch.compile``
is supported, with the caveat that sequence-sharded ring attention must stay outside compiled regions.
```

docs/api/physicsnemo.domain_parallel.rst
```diff
@@ -2,17 +2,30 @@
...
+
+.. note::
+
+    **This is a breaking change in PhysicsNeMo 2.3:** Prior to version 2.3,
```
megnvidia:
```suggestion
    **PhysicsNeMo 2.3 changes the ``ShardTensor`` input contract.** Prior to version 2.3, 
```

docs/api/physicsnemo.domain_parallel.rst
```diff
@@ -21,9 +34,108 @@ different local tensor sizes.
...
+    :show-inheritance:
+
 Utility Functions
 -----------------
 
 .. autofunction:: physicsnemo.domain_parallel.scatter_tensor
 
+.. autofunction:: physicsnemo.domain_parallel.sync_module_over_mesh
+
+Synchronization responsibilities
```
megnvidia:
```suggestion
Synchronization Responsibilities
```

docs/api/physicsnemo.domain_parallel.rst
```diff
@@ -21,9 +34,108 @@ different local tensor sizes.
...
+already present on each process.  When using ``from_local``, you are
+responsible for providing the correct local piece on every process.
+
+The startup synchronization happens only once.  It does not keep buffers in
+sync if they change during training.  Synchronize changing buffers separately
+when your model requires it.  Checkpoint loading is also separate: use the
+distributed checkpoint utilities for distributed tensors, and synchronize
+plain model state if only one process loaded it.
+
+Example: FSDP2 with domain parallelism
```
megnvidia:
```suggestion
Example: FSDP2 with Domain Parallelism
```

docs/api/physicsnemo.domain_parallel.rst
```diff
@@ -21,9 +34,108 @@ different local tensor sizes.
...
 -----------------
 
 .. autofunction:: physicsnemo.domain_parallel.scatter_tensor
 
+.. autofunction:: physicsnemo.domain_parallel.sync_module_over_mesh
+
+Synchronization responsibilities
+--------------------------------
+
+In a training script, typically ``DDP`` will sycnronize weights
```
megnvidia:
```suggestion
In a training script, ``DDP`` synchronizes weights across its entire process
```

docs/api/physicsnemo.domain_parallel.rst
```diff
@@ -21,9 +34,108 @@ different local tensor sizes.
...
+across it's entire process group - but with a 2D mesh, that does not include
```
megnvidia:
```suggestion
group. With a 2D mesh, that process group does not include the domain axis of
the mesh.
```

docs/api/physicsnemo.domain_parallel.rst
```diff
@@ -21,9 +34,108 @@ different local tensor sizes.
...
+the domain axis of the mesh.
```
megnvidia:
```suggestion
```

docs/api/physicsnemo.domain_parallel.rst
```diff
@@ -21,9 +34,108 @@ different local tensor sizes.
...
+
+``sync_module_over_mesh`` skips ``DTensor`` and ``ShardTensor`` values.
+``scatter_tensor`` and ``distribute_tensor`` create consistent distributed
+tensors from one source tensor.  In contrast, ``from_local`` uses the value
+already present on each process.  When using ``from_local``, you are
+responsible for providing the correct local piece on every process.
+
+The startup synchronization happens only once.  It does not keep buffers in
+sync if they change during training.  Synchronize changing buffers separately
+when your model requires it.  Checkpoint loading is also separate: use the
```
megnvidia:
```suggestion
when your model requires it.  Checkpoint loading is also separate. Use the
```

physicsnemo/domain_parallel/_shard_redistribute.py
```diff
@@ -687,6 +712,36 @@ def forward(
...
+                if not isinstance(placement, Shard):
+                    continue
+                tensor_dim = placement.dim
+                if tensor_dim in target_sharding_shapes:
+                    mesh_size = device_mesh.size(mesh_dim)
+                    per_rank_sizes = target_sharding_shapes[tensor_dim]
+                    if len(per_rank_sizes) == mesh_size:
+                        overridden = []
+                        for rank_size in per_rank_sizes:
+                            rank_shape = list(global_shape)
```
negin513: I think this writes incorrect `_sharding_shapes` on multi-dim meshes. Starting from `global_shape` here contradicts the invariant documented in `compute_sharding_shapes_from_chunking_global_shape`: `_sharding_shapes[m][r]` is rank r's cross-section *through this rank's coordinates on every other mesh dim*, so tensor dims sharded by other mesh dims must carry this rank's local chunk, not the global size.

Concrete example on a 2x2 mesh, global tensor `(4, 100, 64)`:
- current `(Shard(1), Replicate())` with uneven X (70/30), redistribute to `(Shard(1), Shard(2))`
- dim 1 is preserved, so the override fires for mesh dim 0
- rank (0,0)'s actual local tensor is `(4, 70, 32)`, but this records mesh dim 0's shapes as `[(4,70,64), (4,30,64)]` - Y stamped at the global 64 instead of 32
- meanwhile mesh dim 1's entries keep the even-chunk X (50) instead of the actual 70, since the override never touches the cross terms

The fast path in `_to_new_shard_dim` asserts recorded shape == local shape and raises "stale or corrupt", so the next redistribute on this tensor hard-crashes. 1D meshes are unaffected (no cross terms), and the only 2D compile test redistributes to `(Replicate(), Replicate())` which skips this loop entirely - so nothing catches it.

Suggested fix: start from `chunk_shapes[mesh_dim][r]` and replace only `tensor_dim`, and fix the cross terms on the other mesh dims to use the preserved per-rank size. A 2D test with an uneven preserved shard would be great; I'd expect it to fail today.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -409,12 +417,23 @@ def _convert_results_to_shard_tensor(
...
         )
 
-    if isinstance(result, Iterable) and not isinstance(result, (str, bytes)):
-        return type(result)(
+    # Explicit allowlist mirroring _convert_args_to_dtensor: only walk into
+    # plain tuple / list containers. A generic Iterable check would crash on
+    # things like torch.UntypedStorage (iterable over bytes) or torch.Tensor
+    # because their constructors don't accept a generator. Note: namedtuples
+    # degrade to plain tuple here, same as in the args walker.
+    if isinstance(result, tuple):
```
negin513, line 433: This branch flattens `torch.return_types.*` (max, topk, sort, ...) to plain tuples - they're tuple subclasses, so they match here. That means `torch.max(shard_tensor, dim=1).indices` now raises `AttributeError`. As far as I can tell the old `type(result)(generator)` code preserved these (structseq constructors accept an iterable), so this is a regression rather than a pre-existing gap. The comment about namedtuples degrading "same as in the args walker" doesn't quite apply - arguments are never structseqs, but results routinely are.

A `hasattr(type(result), '_fields')` check (or handling `torch.return_types` explicitly) before this generic tuple branch would fix it. A test asserting `torch.max(st, dim=...).indices` works would lock it in.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -866,6 +937,133 @@ def __tensor_unflatten__(inner_tensors, flatten_spec, outer_size, outer_stride):
...
+        if not any(isinstance(p, Partial) for p in self.placements):
+            return self
+        new_placements = [
+            Replicate() if isinstance(p, Partial) else p for p in self.placements
+        ]
+        return self.redistribute(
+            device_mesh=self.device_mesh, placements=new_placements
+        )
+
+    def __coerce_same_metadata_as_tangent__(
```
negin513, line 1041: Every compiled backward now depends on the invariant "a Partial label on a tangent is always replicate-valued data, never a pending reduction." If any op ever produces a genuinely partial tangent, this relabels instead of reducing and gradients are silently off by mesh_size - no error, no warning. The inline comment explaining the convention is good, but for something this consequential I'd want it stated in the module-level docs, and ideally a debug-mode assertion guarding it.

physicsnemo/domain_parallel/shard_utils/conv_patches.py
```diff
@@ -452,8 +454,12 @@ def backward(
         """
         for mesh_dim in range(ctx.spec.mesh.ndim):
             if ctx.spec.placements[mesh_dim].is_shard():
-                group = ctx.spec.mesh.get_group(mesh_dim)
-                dist.all_reduce(grad_weight_or_bias, group=group)
+                # funcol.all_reduce returns a new tensor (AsyncCollectiveTensor)
+                # that auto-waits when used; assigning back into the loop var
+                # serializes the iterations correctly.
+                grad_weight_or_bias = funcol.all_reduce(
```
negin513, line 460: This returns the raw `AsyncCollectiveTensor` into the autograd engine, so it can land in `param.grad`. It usually auto-waits on first use, but ACT-in-`.grad` is a known footgun with grad hooks and anything touching `.grad` without going through dispatch. `_shard_redistribute.py` explicitly `.wait()`s every funcol result, so the asymmetry reads like an oversight - if it's a deliberate overlap optimization, a comment saying so would help. (Same pattern in `point_cloud_ops.py` `GradReducer.backward` and the grad_weight/grad_bias returns in `normalization_patches.py`.)

physicsnemo/domain_parallel/_shard_redistribute.py
```diff
@@ -330,7 +360,7 @@ def redistribute_local_shard_tensor(
     *,
     async_op: bool = False,
     is_backward: bool = False,
-    target_sharding_shapes: dict[int, tuple[torch.Size, ...]] | None = None,
+    target_sharding_shapes: dict[int, tuple[tuple[int, ...], ...]] | None = None,
```
negin513: This annotation doesn't match the runtime type: `get_tensor_sharding_shapes_by_dim` returns `dict[int, list[int]]` (per-rank sizes along one tensor dim), and that's what every call site passes. The `int(rank_size)` in the forward override loop only works because each element is already an int, not a tuple. Should be `dict[int, Sequence[int]]` (docstring too).

test/domain_parallel/ops/test_compile_ops.py
```diff
@@ -0,0 +1,456 @@
...
+    )
+    v = torch.randn(
+        batch_size, num_heads, seq_len, head_dim, device=dm.device, requires_grad=True
+    )
+
+    q_s = scatter_tensor(q, 0, distributed_mesh, (Shard(2),), requires_grad=True)
+    k_s = scatter_tensor(k, 0, distributed_mesh, (Shard(2),), requires_grad=True)
+    v_s = scatter_tensor(v, 0, distributed_mesh, (Shard(2),), requires_grad=True)
+
+    with pytest.raises(Exception):
```
negin513: `pytest.raises(Exception)` passes for any failure at all - a broken distributed setup, an import error, anything. If someone breaks ring SDPA in an unrelated way, this still passes for the wrong reason. Matching on the message would make it a real regression guard, e.g. `pytest.raises(Exception, match="record_stream|functionaliz")` - or at least a comment documenting what exception the current CI torch version raises.

test/domain_parallel/ops/test_compile_ops.py
```diff
@@ -0,0 +1,456 @@
...
+    backward: bool = True,
+    fullgraph: bool = True,
+) -> Any:
+    r"""Compile ``module`` with ``aot_eager`` and run forward (+ optional bwd).
+
+    Smoke check that the autograd Function dispatched inside ``module`` is
+    AOT-traceable. Resets dynamo first so each test is independent.
+    """
+    torch._dynamo.reset()
+    compiled = torch.compile(module, backend="aot_eager", fullgraph=fullgraph)
```
negin513, line 72: Everything in this file (and test_compile.py) uses `aot_eager`, which validates AOT traceability but never runs Inductor. That's the right first gate, but Inductor codegen of the funcol ops is exactly where I'd expect the next class of surprises, and the PR title says "Enable Compile". One Inductor smoke test on the multi-GPU CI would go a long way - or an honest note in the docs about what's actually validated.

Also noting two ops with no compile coverage: `ShardedIndexSelect` and `ConvGradReducer` (the latter matters more given its backward now returns a funcol result).

physicsnemo/domain_parallel/_shard_tensor_spec.py
```diff
@@ -99,9 +110,28 @@ def __hash__(self) -> int:
...
+        ``_sharding_shapes``, so equal placements with different uneven
+        layouts don't collide. Consumed by the AOT autograd cache through
+        ``ShardTensor._stable_hash_for_caching``.
+
+        Returns
+        -------
+        str
+            Deterministic hex digest identifying this spec.
+        """
+        stable_key = (self.mesh._stable_hash(),) + self._hash_key()[1:]
```
negin513: The `[1:]` slice assumes `_hash_key()`'s first element is the mesh - that's a private `DTensorSpec` method with no contract. If a torch upgrade restructures it, this produces silent hash collisions (wrong AOT cache hits) rather than an error, which would be miserable to debug. A comment documenting the assumed structure would help the next person; even better, a small unit test that two specs differing only in placements get different stable hashes.

physicsnemo/domain_parallel/custom_ops/__init__.py
```diff
@@ -22,3 +22,4 @@
 if ST_AVAILABLE:
     from . import _tensor_ops  # noqa: F401  # registers unbind handlers
     from ._reductions import mean_wrapper, sum_wrapper
+    from ._tensor_ops import unbind_wrapper
```
negin513, line 25: nit: `from . import _tensor_ops` on the line above already runs the registration side effect, so this import is redundant unless you want `unbind_wrapper` re-exported from this namespace - and if so, `_reductions`' wrappers are exported on the previous line, so maybe merge them.

negin513 (commented):
Really nice work overall - the comments explaining *why* each change is needed (the torch.Size/SymInt fakeification writeup, the ProcessGroup deepcopy rationale) are excellent. I left inline comments; the two I'd consider blocking are the multi-dim mesh `_sharding_shapes` override in `ShardRedistribute.forward` and the `torch.return_types` flattening in the result walker. Also worth noting ~50 of the changed files are the #1556 dependency, so this needs a rebase once that merges before the real diff is reviewable.

docs/api/physicsnemo.domain_parallel.rst
```diff
@@ -21,9 +34,108 @@ different local tensor sizes.
...
+
+Example: FSDP2 with domain parallelism
+--------------------------------------
+
+Create a two-dimensional mesh.  The ``"ddp"`` dimension holds different
+training samples and shards the model weights.  The ``"domain"`` dimension
+splits one sample across multiple GPUs.  The product of the two dimensions
+must equal the number of distributed processes.
+
+For example, with 8 GPUs, ``data_parallel_size=4`` and ``domain_size=2``
```
megnvidia:
```suggestion
For example, with eight GPUs, ``data_parallel_size=4`` and ``domain_size=2``
```

docs/api/physicsnemo.domain_parallel.rst
```diff
@@ -21,9 +34,108 @@ different local tensor sizes.
...
+
+    # FSDP2 shards weights only across the data-parallel dimension.
+    fully_shard(model, mesh=data_mesh)
+
+Call ``sync_module_over_mesh`` before ``fully_shard``.  If the model has
+parameters that must themselves be sharded over the domain, synchronize the
+plain model first, then convert those selected parameters with
+``distribute_tensor`` before calling ``fully_shard``.
+
+Example: DDP with domain parallelism
```
megnvidia:
```suggestion
Example: DDP with Domain Parallelism
```

examples/cfd/external_aerodynamics/domino/README.md
```diff
@@ -196,11 +196,12 @@ To train and test the DoMINO model on AWS dataset, follow these steps:
 ### Training with Domain Parallelism
 
 DoMINO has support for training and inference using domain parallelism in PhysicsNeMo,
-via the `ShardTensor` mechanisms and pytorch's FSDP tools.  `ShardTensor`, built on
-PyTorch's `DTensor` object, is a domain-parallel-aware tensor that can live on multiple
-GPUs and perform operations in a numerically consistent way.  For more information
-about the techniques of domain parallelism and `ShardTensor`, refer to PhysicsNeMo
-tutorials such as [`ShardTensor`](https://docs.nvidia.com/deeplearning/physicsnemo/physicsnemo-core/api/physicsnemo.distributed.shardtensor.html).
+via the `ShardTensor` mechanisms and PyTorch's FSDP2 (`fully_shard`) tools.
```
megnvidia:
```suggestion
using the `ShardTensor` mechanisms and PyTorch's FSDP2 (`fully_shard`) tools.
```

examples/cfd/external_aerodynamics/domino/README.md
```diff
@@ -196,11 +196,12 @@ To train and test the DoMINO model on AWS dataset, follow these steps:
...
+`ShardTensor` is a `torch.Tensor` subclass that is domain-parallel aware: it can live
```
megnvidia:
```suggestion
`ShardTensor` is a `torch.Tensor` subclass that is domain-parallel aware. It can live
```

examples/cfd/external_aerodynamics/transformer_models/README.md
```diff
@@ -72,7 +72,7 @@ Several other important configuration settings are available:
 
 > **Note** Like other parameters of the model, changing the value of `model.use_te` will make checkpoints incompatible.
 
-The training script supports data-parallel training via PyTorch DDP. In a future update, we may enable domain parallelism via FSDP and ShardTensor.
+The training script supports data-parallel training via PyTorch DDP. In a future update, we may enable domain parallelism via ShardTensor (with DDP or FSDP2 `fully_shard` on the data-parallel axis).
```
megnvidia:
```suggestion
The training script supports data-parallel training via PyTorch DDP. A possible future change might enable domain parallelism usingShardTensor (with DDP or FSDP2 `fully_shard` on the data-parallel axis).
```

physicsnemo/domain_parallel/_shard_redistribute.py
negin513: Yes, this seems to be incorrect to me. At least inefficient?

examples/weather/stormcast/README.md
```diff
@@ -353,7 +365,7 @@ If training seems slow, check your GPU utilization. If it's low, the problem is
   * Data compression, if bandwidth from storage is the limiting factor
   * Caching data, particularly if your dataset is small enough to be preloaded to memory
 
-Enabling BF16 (`training.perf.fp_optimizations=amp-bf16`) speeds up training considerably but is only recommended for U-Nets. Compiling the training loss with ``torch.compile`` (`training.perf.torch_compile=True`), which will also compile the model forward pass, also typically improves performance but is not compatible with domain parallelism.
+Enabling BF16 (`training.perf.fp_optimizations=amp-bf16`) speeds up training considerably but is only recommended for U-Nets. Compiling the training loss with ``torch.compile`` (`training.perf.torch_compile=True`), which will also compile the model forward pass, also typically improves performance; this recipe currently disables it under domain parallelism (sharded ring attention must stay outside compiled regions, and regional compilation is not yet wired into this example).
```
megnvidia:
```suggestion
Enabling BF16 (`training.perf.fp_optimizations=amp-bf16`) speeds up training considerably but is only recommended for U-Nets. Compiling the training loss with ``torch.compile`` (`training.perf.torch_compile=True`), which will also compile the model forward pass, also typically improves performance. This recipe disables it under domain parallelism (sharded ring attention must stay outside compiled regions, and regional compilation is not yet wired into this example).
```

megnvidia (approved):
i skipped all the .py files. Rest looks good and would be even better with these changes.

physicsnemo/domain_parallel/_shard_redistribute.py
coreyjadams: OK, test added, and fixed.  It took me some time to understand the bug here.  It wasn't that the multi-dimensional redistribute was wrong; it was that the metadata could come out silently wrong, and if you did the redistribute in multiple passes, it would take a shortcut and get it wrong.  All the tests assume the "right" way to redistribute which is to tell shard tensor your desired final state and let it optimally do it; you only hit this bug when you do it manually.

I've add a test to do it manually, though, and fixed the bug.

physicsnemo/domain_parallel/shard_tensor.py
coreyjadams: Fixed; I directly check against torch.return_types rather than looking for attributes, since attributes might not be the same across them all.

physicsnemo/domain_parallel/_shard_redistribute.py
coreyjadams: Addressed.

physicsnemo/domain_parallel/shard_utils/conv_patches.py
coreyjadams: Fixed.

coreyjadams: Fixed.

physicsnemo/domain_parallel/_shard_tensor_spec.py
coreyjadams: Moved to a more stable method by using local sharding shapes + parents stable hash, so this is resolved, thanks.

physicsnemo/domain_parallel/custom_ops/__init__.py
coreyjadams: I think this is harmless, I will let this pass if thats ok with you.

test/domain_parallel/ops/test_compile_ops.py
coreyjadams: Addressed.

test/domain_parallel/ops/test_compile_ops.py
coreyjadams: In all of physicsnemo, we actually do not test inductor codegen in CI.  That's a responsibility of upstream pytorch and would slow down our CI suite dramatically.  We made a deliberate choice to do this: we ensure traceability, and that we send a graph to pytorch for fusion, etc., but skip that step.  The idea is that PyTorch owns the CI and validation of inductor itself.

I have added ConvGradReducer coverage;  ShardedIndexSelect was already present.

physicsnemo/domain_parallel/shard_tensor.py
coreyjadams: OK, this is resolve.  Every partial placement now means what it says: reduction pending.  No more renaming, that's too fragile.  Thanks for catching this.

@dallasfoster - sorry, this will affect your PR into this branch a bit.  Check `ce326a4cd72d814517694f074f40e7c343f75244` where all the changes were made for this specific issue.

coreyjadams:
/ok to test [ce326a4](https://github.com/NVIDIA/physicsnemo/pull/1682/commits/ce326a4cd72d814517694f074f40e7c343f75244)

coreyjadams:
/ok to test [855b621](https://github.com/NVIDIA/physicsnemo/pull/1682/commits/855b6215cf98105d61b608bce322fd9b4f9c3b80)

coreyjadams:
/ok to test [ac53869](https://github.com/NVIDIA/physicsnemo/pull/1682/commits/ac538692cf3202ba0329594978c76588093de6e6)

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -679,17 +726,22 @@ class ShardTensor(torch.Tensor):
...
+    # makes exactly that grad call on the subclass primals, so intercepting
+    # it made compiled regions return plain-tensor gradients for ShardTensor
+    # inputs. __torch_function__ runs these directly on the real tensors.
     _autograd_passthrough_functions: frozenset = frozenset(
         fn
         for fn in (
             getattr(torch.Tensor, "register_hook", None),
             getattr(torch.Tensor, "register_post_accumulate_grad_hook", None),
             getattr(torch.Tensor, "retain_grad", None),
+            torch.autograd.grad,
```
dallasfoster, line 744: the dispatch at L1258 runs it under `DisableTorchFunctionSubclass` and returns it un-wrapped for **eager** too, not just AOT's internal call. Confirm eager `torch.autograd.grad(loss, [st])` still returns ShardTensor-typed grads (add a test); gate on `_is_tracing` if it regressed.

physicsnemo/domain_parallel/shard_tensor.py
dallasfoster: A ShardTensor *subclass* whose forward output is the subclass but whose tangent arrives as base then hits "guessed its metadata incorrectly" on every op. Rebuild through the expected subclass's `__tensor_unflatten__` when it's a `ShardTensor` subclass; keep `None` only for foreign types.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -866,6 +961,124 @@ def __tensor_unflatten__(inner_tensors, flatten_spec, outer_size, outer_stride):
...
+                "AOTAutograd recorded a Partial tangent target for ShardTensor; "
+                "__coerce_tangent_metadata__ must normalize expected tangent "
+                "placements to Replicate"
+            )
+
+        if (
+            self._spec.placements == spec.placements
+            and self._spec._sharding_shapes == spec._sharding_shapes
+        ):
+            return self
```
dallasfoster: Coerce early-return compares `_sharding_shapes` with strict `==`, so `{}` (op-result re-wrap) vs `None` (fresh spec) are unequal → needless redistribute. Treat `{}`/`None` as equal.

physicsnemo/domain_parallel/_shard_tensor_spec.py
```diff
@@ -99,9 +116,43 @@ def __hash__(self) -> int:
...
+                self.mesh.device_type,
+                tuple(mesh.shape),
+                tuple(mesh.flatten().tolist()),
+                self.mesh.mesh_dim_names,
+                self.placements,
+                self.shard_order,
+                self.tensor_meta,
+                sharding_shapes,
+            )
+        return hashlib.blake2b(repr(stable_key).encode(), digest_size=16).hexdigest()
```
dallasfoster, line 157: _stable_hash` (L119-151) omit `_local_shape` and include `_sharding_shapes` only when non-`None`, so two specs with identical placements + global shape but different *uneven* splits (both `_sharding_shapes=None`) compare equal and hash/stable-hash equal. If `_stable_hash_for_caching` (`shard_tensor.py:964`) gates AOT-cache reuse and Dynamo's inner size guards don't independently discriminate the split, a graph compiled for split A is reused for split B (silent wrong result). Confirm which guard gates reuse; fold `_local_shape` in, or stop wildcarding `None`.

physicsnemo/domain_parallel/_shard_redistribute.py
dallasfoster, line 861: `ShardRedistribute.backward` builds its output spec with only `tensor_meta` + `_local_shape=output.shape`, leaving `_sharding_shapes=None` — even though it now computes `target_sharding_shapes` (L814), that value is threaded only into `redistribute_local_shard_tensor`, not onto the emitted grad spec. So a downstream consumer of the backward grad that calls `sharding_shapes()` re-arms the blocking `_all_gather_shard_shapes` under trace — the same asymmetry the forward path deliberately avoids (forward populates at L763-766).

physicsnemo/domain_parallel/custom_ops/_tensor_ops.py
dallasfoster, line 107: Two `unbind` handlers with divergent autograd. The live path `unbind_wrapper` (function-level, for `torch.unbind`/`Tensor.unbind`) routes through `ShardTensor.from_local` → `_FromTorchTensor.apply` — the AOT-hostile bridge `_reductions.py` deliberately avoids (legacy `next_functions` access). `torch.compile(fullgraph=True)` over an unbind (docstring cites Stormcast attention) graph-breaks/raises, while `_unbind_dispatch` (L107) constructs the subclass directly. Make the function path construct directly like the dispatch path; reconcile the two.

physicsnemo/domain_parallel/shard_utils/halo.py
```diff
@@ -630,45 +647,63 @@ def perform_halo_collective(
...
-            torch.empty(0, dtype=dtype, device=device) for _ in range(local_size)
-        ]
-
-        # Set up send/recv buffers
+        # This has to be funcol collectives, below, to be
+        # compatible with torch.compile.
+        #
+        # Symmetric-halo assumption: what I receive from neighbor R has the
+        # same numel as what I send to R (true for uniform halo sizes used
+        # by conv / natten / pooling / etc).
```
dallasfoster, line 669: The funcol all-to-all rewrite in `perform_halo_collective` (a2a branch) is the biggest compile-enabling change and I don't believe it's tested: the only halo compile test drives `UnHaloPadding`'s local slice/cat path, which never calls the collective. Add a `halo_padding`/sharded-conv compile test on a ≥2-rank mesh dim.

physicsnemo/domain_parallel/shard_utils/halo.py
dallasfoster, line 1: Near L590 -  `p2p` branch still uses `dist.P2POp`/`batch_isend_irecv`/`torch.cuda.set_device` — not FX-representable, no guard when `communication_method="p2p"` is used under compile (contrast `ring_sdpa`, explicitly `@torch.compiler.disable`d). Latent (a2a is default) but p2p is a validated public option. Route through funcol or raise a clear error.

test/domain_parallel/ops/test_async_collective_gradients.py
dallasfoster, line 1: I would add the following tests: 

* Numerical grad-equivalence for sum/mean backward under compile - uneven shards, >=2 mesh dims.
* fwd/bwd/compile + dispath-vs-function-handler consistency for `unbind`.
* funcol a2a halo under compile - halo_padding and sharded conv.
* Feed a backward-produced uneven grad into a shape-reading op under compile (like in _shard_redistribute.py).

dallasfoster (commented):
Just a handful of small comments/ possible bugs? I can't be confident on this code without more testing.

physicsnemo/domain_parallel/shard_tensor.py
coreyjadams: Good call, will fix.

physicsnemo/domain_parallel/shard_tensor.py
coreyjadams: I do not think this is a bug. I don't think I should change this.

This is not unwrapping or re-wrapping objects.  What this does is prevent re-interception of this api call, for torch.autograd.grad, not the actual gradient computation.  Grads get build byt the subclass backwards graph still, via __torch_dispatch__, and come out the other wise as ShardTensor typed.

Gating on _is_tracing would be tricky too: since the thing we're protecting against here is routing to DTensor for backup ops (which is correct for math operations), blocking it for _is_tracing protects during tracing but not anytime else.  We still need to protect ShardTensor from dispatching to DTensor for torch.autograd.grad itself.  Instead, we run it on ShardTensor itself and let torch dispatch the ops directly from these objects.

dallasfoster: Thanks for the elucidation, this makes sense.

physicsnemo/domain_parallel/_shard_tensor_spec.py
coreyjadams: OK, took a second to read and understand this one...

Addressed: local shape is included in the hash now.  When the sharding shapes is None, that's a signal that implicitly this shard tensor should be `torch.chunk`-distributed.  So local_shape in the hash makes sense to catch a mutation there.  Further, I added a check that if shapes are none, and the local shape doesn't agree with torch.chunk expected sizing, that's an exception: something has gone wrong, and the program has to exit.

physicsnemo/domain_parallel/_shard_redistribute.py
coreyjadams: This is resolved.

physicsnemo/domain_parallel/custom_ops/_tensor_ops.py
coreyjadams: These are already basically "reconciled" since they call the same underlying logic (`_unbind_output_metadata`) and the only difference is the dispatch layer.  The worry here is never actually realized, since `next_functions` accessor only shows up with a real backward node.  Under compule, the `from_local` function gets traced symbolically, not executed.  So it's not built, and it doesn't trip.

I've add a new test though, to compile this with `fullgraph=True` and ensure we're handling this correctly always.

physicsnemo/domain_parallel/shard_utils/halo.py
coreyjadams: Tests added for halo_padding under compile.

physicsnemo/domain_parallel/shard_utils/halo.py
coreyjadams: I've opted to just raise an exception if `p2p` is used when compiling, since `a2a` is the default anyways.  It's unlikely users will often hit this, if ever, and since there is no funcol equivalent of p2p ops that I'm aware of, this is a simpler solution to be transparent.

I have also added a test to ensure we raise the exception under `torch.compile`.

test/domain_parallel/ops/test_async_collective_gradients.py
coreyjadams: Addressed.

coreyjadams:
/ok to test [802d47e](https://github.com/NVIDIA/physicsnemo/pull/1682/commits/802d47e4dac582175b0296911825ec9b865c37e4)

PR #1564, opened 2026-04-13
Fix Warp deprecation warnings and broken links in Mesh README

manmeet3591:
This PR addresses two maintenance items:

1. **Warp Deprecation Fix**: Replaces `wp.context.Device` with `wp.Device` in `radius_search/_warp_impl.py` to resolve upcoming namespace removals in NVIDIA Warp.
2. **README Link Fixes**: Corrects several broken relative links in `physicsnemo/mesh/README.md` that were redundantly prefixed with `physicsnemo/mesh/` or had incorrect relative paths to `examples/` and `test/`.

PR #1689, opened 2026-05-29
Fix Warp import crash: use public wp.Device, not wp.context.Device

mehdiataei:
## Description

Recent Warp removed the public `warp.context` namespace (it now lives at
`warp._src.context`). `radius_search/_warp_impl.py` used `wp.context.Device` as a
runtime-evaluated type annotation, so importing it crashes and breaks the whole
`physicsnemo.nn` / `physicsnemo.models` import chain.

This switches to the public `wp.Device` (matching the adjacent `wp.Stream` annotation), as recommended by Warp's
own deprecation message ("Use `warp.Device` instead"). 

Note that PhysicsNeMo's uncapped warp-lang>=1.5.0 dependency already resolves to the current stable Warp 1.13.0, which has removed the public warp.context namespace, so this import crash affects fresh installs today (not just Warp 1.14 nightlies), while the wp.Device replacement is verified to import correctly on both Warp 1.13.0 and 1.14.

Repro (Warp 1.14):

    from physicsnemo.models.mlp.fully_connected import FullyConnected
    # AttributeError: module 'warp' has no attribute 'context'

## Checklist

- [x] I am familiar with the Contributing Guidelines.
- [x] The CHANGELOG.md is up to date with these changes.
- [ ] An issue is linked to this pull request.
- [x] If I am implementing a new model or modifying any existing model, I have followed the Models Implementation Coding Standards.

## Dependencies

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed bug in Pangu, FengWu attention window shift for asymmetric longitudes
 - Fixed a bug in `mesh.sampling.find_nearest_cells`, where a mixup between L2 and L-inf norms
   could cause slightly incorrect nearest-neighbor assignments in highly skewed meshes.
+- Fixed an import crash with recent Warp by using the public `wp.Device` instead of the
+  removed `wp.context.Device` in the `radius_search` Warp interop. This unblocks importing
+  `physicsnemo.nn` and `physicsnemo.models`.
 
 ### Security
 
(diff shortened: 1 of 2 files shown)


peterdsharpe (approved):
Nice, LGTM!

mehdiataei:
Closing as redundant. Already fixed on `main` by #1564 (commit `16fd2e54`, May 5)
with the same `wp.context.Device` → `wp.Device` change. The diff only appears because
this branch was cut from an older commit; it's a no-op against current `main`. My bad
for not syncing first.

PR #1691, opened 2026-06-01
PhysicsNeMo PEFT - LoRA

mnabian:
# PhysicsNeMo Pull Request

## Description

Adds a native, self-contained **Low-Rank Adaptation (LoRA)** subpackage for
parameter-efficient fine-tuning of PhysicsNeMo models, shipped under
`physicsnemo.experimental.peft`, plus an end-to-end example and a user-guide page.

LoRA freezes a pretrained model and trains only small low-rank adapter matrices
injected beside selected layers. This adapts a model to a new dataset at a
fraction of the cost of full fine-tuning, produces a tiny adapter checkpoint, lowers memory (frozen layers drop saved
activations), and reduces overfitting/forgetting in the small-data regime typical
of SciML.

## Motivation

- **Small-data domain adaptation** (the regime ISV end-users actually live in):
  adapt a foundation checkpoint to a new vehicle class / chemistry / operating
  condition from a handful of samples.
- **Deployment story**: one frozen base + N swappable adapters at serve time.

### Package — `physicsnemo/experimental/peft/`

| File | Contents |
|---|---|
| `config.py` | `LoRAConfig` dataclass + validation |
| `lora.py` | `LoRALayer` mixin, `LoRALinear`, `LoRA_te_Linear`, `LoRA_te_LayerNormMLP`, and the type→wrapper registry (`register_lora_wrapper` / `get_wrapper_for`) |
| `apply.py` | `apply_lora`, `resolve_targets`, `ApplyResult`, freeze logic, guards |
| `merge.py` | `merge_lora` |
| `io.py` | `save_adapter` / `load_adapter` (adapter archive) |
| `utils.py` | `split_params_for_optimizer`, `print_trainable_parameters`, `set_adapter_enabled`, `compute_base_fingerprint` |

**Public API:**

```python
from physicsnemo.experimental.peft import (
    LoRAConfig,                  # declare rank/alpha + which layers to adapt
    apply_lora,                  # inject adapters in place, freeze the base
    ApplyResult,                 # report returned by apply_lora
    split_params_for_optimizer, # route adapter params to AdamW
    print_trainable_parameters, # "trainable params: N (X% of M total)"
    save_adapter, load_adapter,  # adapter-only archive I/O
    merge_lora,                  # fold adapters into base for zero-overhead inference
    set_adapter_enabled,         # toggle adapters (base-vs-adapter comparison)
    register_lora_wrapper,       # extension seam for new layer types
)
```

### End-to-end example — `examples/cfd/external_aerodynamics/transformer_models/`

A separate, runnable LoRA recipe living alongside the existing GeoTransolver
training example (`train.py` is **not** modified):

- `src/finetune.py` — load a pretrained base → `apply_lora` → train only the
  adapters (AdamW, `find_unused_parameters=True`, `DistributedSampler` sharding)
  → `save_adapter`.
- `src/deploy.py` — `load_adapter` (adapter-swap) or `merge_lora` (fold in).
- `conf/finetune_lora.yaml` — reuses the example's model/data/training config
  groups + a `peft:` block.
- `FINETUNE_LORA.md` — full walkthrough; the main README cross-references it.

### Documentation

- `physicsnemo-docs/docs/user-guide/peft.rst` — user-guide page (overview, when to
  use, the LoRA math, quickstart, layer targeting, optimizer setup, save/load,
  merge, Transformer Engine support, extensibility, API reference), wired into the
  User Guide toctree.

## Quickstart

```python
import torch
from physicsnemo.experimental.peft import (
    LoRAConfig, apply_lora, split_params_for_optimizer, save_adapter,
)

model = build_model()                     # any torch.nn.Module
model.load_state_dict(pretrained)

apply_lora(model, LoRAConfig(rank=16, alpha=16,
           target_pattern=r"blocks\.\d+\.attn\.(q|k|v|out)_proj"))

groups = split_params_for_optimizer(model)
opt = torch.optim.AdamW(groups["lora"] + groups["extras"], lr=5e-4)
# ... train ...
save_adapter(model, "adapter.lora")
```

## Layer targeting

A `LoRAConfig` sets **exactly one** selector, matched against **fully-qualified**
module names (leaf names are not unique):

- `target_modules` — explicit list of names
- `target_pattern` — regex (`re.search`)
- `target_filter` — predicate `(name, module) -> bool`

Plus two modifiers: `wrap_mlp` (additively adapt the transformer feed-forward
sub-block) and `extras_trainable` (modules to train fully, not low-rank). Only
registered layer types are eligible; a selector matching zero wrappable layers
raises (no silent misses).

## Key design decisions

- **In-place mutation, not a `PeftModel` wrapper.** `apply_lora` swaps matched
  leaves for LoRA wrappers and freezes the base; the model keeps its class and
  identity, so existing `.mdlus` checkpoint/inference tooling still works.
- **Works on any `torch.nn.Module`** — no dependency on `physicsnemo.Module` or
  the `.mdlus` format (a plain-PyTorch user can adapt their own model).
- **Transformer Engine**: `te.Linear` adapts per-matrix; the *fused*
  `te.LayerNormMLP` (no addressable child Linears) adapts via a single rank-r
  **residual** across the FFN sub-block (kept un-mergeable). The `te.LayerNormLinear`
  output head is not wrapped in v1 (documented; future `register_lora_wrapper`).
- **Adapter archive**: a plain ZIP (`adapter_config.json` + `adapter_model.pt` +
  `metadata.json`) holding only the trainable slice. Disambiguated from full model
  checkpoints by `metadata.kind == "lora_adapter"` and a structural
  `base_fingerprint`. Loaded **only** by `load_adapter` (with `weights_only=True`);
  recommended extension `.lora` (any extension works) since it is neither a
  `torch.save` file nor a `Module` checkpoint.
- **Optimizer routing**: adapter params go to **AdamW**, never Muon (Newton–Schulz
  orthogonalization is degenerate on low-rank factors) — via
  `split_params_for_optimizer`.
- **Type→wrapper registry**: the single extension seam — supporting a new layer
  type (equivariant, MoE, …) is one `register_lora_wrapper(type, wrapper)` call,
  with no changes to targeting / apply / save / merge.

### Tests (`test/experimental/peft/`)

42 tests across 7 files, run in-container on GPU + Transformer Engine; the
TE/GeoTransolver tests skip cleanly on CPU-only CI.

| File | Covers |
|---|---|
| `test_config.py` | `LoRAConfig` validation (selectors, rank, dropout, reserved init) |
| `test_lora_linear.py` | B=0 init parity, grad-flow-to-LoRA-only, dtype/device inheritance, merge transpose, disable; TE `te.Linear` + fused `te.LayerNormMLP` residual |
| `test_apply.py` | exact/regex/callable targeting, zero-wrap & double-apply guards, freezing, `extras_trainable` not unfreezing nested base, optimizer split, `wrap_mlp` expansion |
| `test_merge.py` | post-merge forward preserved, idempotency |
| `test_io.py` | round-trip, archive structure, any-extension, missing-parent-dir creation, `kind` check, fingerprint mismatch, **`weights_only` rejection of unsafe pickles**, save-after-merge guard |
| `test_smoke.py` | full recipe via real training: apply → AdamW → loss drops → save → load → merge |
| `test_geotransolver_lora.py` | `apply_lora` + `wrap_mlp` on a real TE GeoTransolver (GPU+TE) |

## Checklist

## Review Process


physicsnemo/experimental/peft/io.py
```diff
@@ -0,0 +1,190 @@
...
+        rank=adapter_config["rank"],
+        alpha=adapter_config["alpha"],
+        lora_dropout=adapter_config.get("lora_dropout", 0.0),
+        target_modules=adapter_config["target_modules"],
+        extras_trainable=adapter_config.get("extras_trainable", []),
+        init=adapter_config.get("init", "default"),
+    )
+    apply_lora(model, config)
+
+    state = torch.load(io.BytesIO(state_bytes), map_location="cpu")
```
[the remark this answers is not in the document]
mnabian: Addressed

physicsnemo/experimental/peft/config.py
```diff
@@ -0,0 +1,111 @@
...
+    target_filter : Callable[[str, nn.Module], bool] | None
+        Predicate ``(name, module) -> bool`` (most flexible selector).
+    lora_dropout : float
+        Dropout on the LoRA input path; ``0.0`` disables it. In ``[0.0, 1.0)``.
+    extras_trainable : list[str]
+        Additional fully-qualified module names to leave fully trainable
+        (not low-rank), e.g. a final head or norm.
+    wrap_mlp : bool
+        Convenience flag to also amend the feed-forward MLP. Not implemented in
+        this version (planned; plan §5.3.2 / §14) — setting it raises at apply.
```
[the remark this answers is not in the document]
mnabian: Older comments which does not apply to the most recent version

ys-teh:
The general implementation structure looks modular and good to me. register_lora_wrappers will be useful. You may want to consider putting lora-related items into a folder called lora in case there will be non-lora peft approaches in the future. Utils can likely stay outside.

The implementation differs in style compared to existing finetuning implementation in Alchemi. The closest one, for example, is patching module and it looks like [this](https://github.com/laserkelvin/nvalchemi-toolkit/blob/feat-finetune-strategy-hooks/nvalchemi/training/hooks/finetune.py#L107). This potentially needs to be resolved on the Alchemi side. In any case, your modular implementation should help.

mnabian:
> The general implementation structure looks modular and good to me. register_lora_wrappers will be useful. You may want to consider putting lora-related items into a folder called lora in case there will be non-lora peft approaches in the future. Utils can likely stay outside.

Thanks for your comment. I would hold the lora/ split until we add a second, non-variant method — at which point the clean factoring is method-agnostic core (targeting, I/O, registry, freeze, utils) vs. tuners/lora, tuners/<next>, not just renaming files. Since the public API is the package `__init__`, that move does not break anything.

mnabian:
Feature request to be added: Combining multiple LoRA adapters

physicsnemo/experimental/peft/lora.py
```diff
@@ -0,0 +1,256 @@
...
+
+try:  # Transformer Engine is optional.
+    import transformer_engine.pytorch as te
+
+    _TE_AVAILABLE = True
+except ImportError:  # pragma: no cover - depends on environment
+    _TE_AVAILABLE = False
+
+
+class LoRALayer:
```
ys-teh, line 64: Would it be useful to specify the compulsory interface in this mixin? E.g., forward, merge_into_base, mergeable, and enabled seem to be required to be compatible with other lora utilities you have.

Also, register_lora_wrapper assumes a common constructor signature: `(base_layer, rank, alpha, dropout=0.0)`. It would be useful to document or specify this contract explicitly.

physicsnemo/experimental/peft/lora.py
```diff
@@ -0,0 +1,256 @@
...
+
+    @staticmethod
+    def _infer_in_out_features(base_layer: nn.Module) -> tuple[int, int]:
+        in_f = getattr(base_layer, "in_features", None)
+        out_f = getattr(base_layer, "out_features", None)
+        if in_f is None or out_f is None:
+            # Fall back to the weight shape (out, in), as for nn.Linear.
+            w = base_layer.weight
+            out_f, in_f = int(w.shape[0]), int(w.shape[1])
+        return int(in_f), int(out_f)
```
ys-teh: These functions feel specific to nn.Linear and te.Linear rather than the generic LoRALayer mixin. Could we move them somewhere so other wrappers don't inherit them?

physicsnemo/experimental/peft/lora.py
```diff
@@ -0,0 +1,256 @@
...
+        )
+        self.lora_B = nn.Parameter(
+            torch.zeros(
+                rank, out_features, device=ref_weight.device, dtype=ref_weight.dtype
+            )
+        )
+        self.lora_dropout: nn.Module = (
+            nn.Dropout(p=dropout) if dropout > 0.0 else nn.Identity()
+        )
+        nn.init.kaiming_uniform_(self.lora_A, a=math.sqrt(5))
```
laserkelvin: Is there a reason for using specifically `kaiming_uniform`?

Can we design this so that users can pass their own init callable instead of hardcoding this? [Here](https://arxiv.org/pdf/2406.08447) is a paper that shows that LoRA performance can be pretty sensitive to initialization method, and it's the combination of `kaiming_uniform` and hardcoded `a` to me seems *could* be suboptimal in some cases

physicsnemo/experimental/peft/lora.py
```diff
@@ -0,0 +1,256 @@
...
+    def _infer_in_out_features(base_layer: nn.Module) -> tuple[int, int]:
+        in_f = getattr(base_layer, "in_features", None)
+        out_f = getattr(base_layer, "out_features", None)
+        if in_f is None or out_f is None:
+            # Fall back to the weight shape (out, in), as for nn.Linear.
+            w = base_layer.weight
+            out_f, in_f = int(w.shape[0]), int(w.shape[1])
+        return int(in_f), int(out_f)
+
+    def lora_delta(self, x: torch.Tensor) -> torch.Tensor:
```
laserkelvin: Public method needs docstring

physicsnemo/experimental/peft/lora.py
```diff
@@ -0,0 +1,256 @@
...
+        (shape ``(out, in)``). Accumulate in fp32 then cast to the base dtype.
+
+        Note the transpose: ``lora_A @ lora_B`` is ``(in, out)``; the weight
+        delta is its transpose. ``B @ A`` would be non-conformant.
+        """
+        delta = (self.lora_A.float() @ self.lora_B.float()).t() * self.scaling
+        self.base_layer.weight.add_(delta.to(self.base_layer.weight.dtype))
+
+
+class LoRALinear(nn.Module, LoRALayer):
```
laserkelvin: Please add docstrings for this class and the methods. The `__init__` arguments do not have descriptions.

physicsnemo/experimental/peft/lora.py
laserkelvin, line 1: Please document arguments in docstrings for public functions at the end of the module

physicsnemo/experimental/peft/merge.py
```diff
@@ -0,0 +1,62 @@
...
+import logging
+
+import torch.nn as nn
+
+from physicsnemo.experimental.peft.lora import is_lora_layer
+
+logger = logging.getLogger("experimental.peft")
+
+
+def merge_lora(model: nn.Module) -> nn.Module:
```
laserkelvin, line 30: I think an alternative, and more modular pattern would be to have this function be applied to an individual layer, as opposed to the full model. Users could then just run `nn.Module.apply` to obtain the same effect, but having it work on a single input layer without recursion would let users select which layers to apply to

physicsnemo/experimental/peft/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    MUST be computed on the pristine base, *before* ``apply_lora`` wraps
+    anything: after wrapping, keys gain ``base_layer.`` prefixes and the
+    original structure can no longer be recovered.
+    """
+    h = hashlib.sha256()
+    for name, module in model.named_modules():
+        h.update(f"{name}:{type(module).__name__}\n".encode())
+    for key, tensor in model.state_dict().items():
+        h.update(f"{key}:{tuple(tensor.shape)}\n".encode())
+    return h.hexdigest()[:16]
```
laserkelvin: If this is entirely just used for determining uniqueness (and not user-facing) maybe we should keep the full hash?

physicsnemo/experimental/peft/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    frozen: list = []
+    lora_ids: set[int] = set()
+    for module in model.modules():
+        if is_lora_layer(module):
+            lora.extend([module.lora_A, module.lora_B])
+            lora_ids.update({id(module.lora_A), id(module.lora_B)})
+    for p in model.parameters():
+        if id(p) in lora_ids:
+            continue
+        (extras if p.requires_grad else frozen).append(p)
```
laserkelvin: IMHO this is a very compact way of routing the append, but not super readable

physicsnemo/experimental/peft/utils.py
```diff
@@ -0,0 +1,90 @@
...
+
+def set_adapter_enabled(model: nn.Module, enabled: bool) -> None:
+    """Enable/disable all LoRA deltas in ``model``.
+
+    Lets you run a base-only forward (e.g. for an adapter-vs-base comparison)
+    without merging or reloading.
+    """
+    for module in model.modules():
+        if is_lora_layer(module):
+            module.enabled = bool(enabled)
```
laserkelvin: Why do you need to cast `bool(enabled)`?

physicsnemo/experimental/peft/lora.py
mnabian: Good call! Made the wrapper contract explicit in the LoRALayer docstring (required attrs/methods + constructor signature) and in register_lora_wrapper, and added a guard in apply_lora that rejects a registered wrapper not subclassing LoRALayer (previously it'd be silently skipped by freeze/save/merge). Left it at docs & runtime check rather than an ABC/Protocol for now, since the contract is small and the package is still experimental. Please let me know if that works or you prefer this to be specified more explicitly. Commit: 350b658cb506599de9a1ad36be02a0b7f6136147

physicsnemo/experimental/peft/lora.py
mnabian: Good point! Moved them onto a new _LinearLoRALayer(LoRALayer) that LoRALinear/LoRA_te_Linear subclass. LoRALayer now keeps only the shape-agnostic core (lora_A/lora_B, lora_delta, _make_lora_params), so non-Linear wrappers like the fused te.LayerNormMLP residual inherit it directly without picking up those assumptions. Also flipped the default mergeable to False on the generic mixin. Commit: 350b658cb506599de9a1ad36be02a0b7f6136147

laserkelvin: I hate how GitHub PR review comments get sent to pending in the middle of a review...

My suggestion would be to actually type hint to make `base_layer: LinearLike`, in which case you can raise an exception if it fails `isinstance(base_layer, LinearLike)`. Otherwise, if you were to apply this to any other layer (without `in_features`, `out_features`, or `weight` attributes) it will fail.

Another thing to keep in mind is that a common pattern outside of just `nn.Linear` is to apply LoRA to embedding tables, see [here](https://github.com/microsoft/LoRA/blob/4c0333854cb905966f8cc4e9a74068c1e507c7b7/loralib/layers.py#L55-L60)

physicsnemo/experimental/peft/lora.py
```diff
@@ -0,0 +1,305 @@
...
+    would otherwise be silently skipped.
+    """
+    _LORA_WRAPPERS[layer_type] = wrapper_factory
+
+
+def get_wrapper_for(module: nn.Module) -> Callable[..., nn.Module] | None:
+    """Return the registered wrapper for ``module`` (walking the MRO so that
+    subclasses of a registered type are handled), or ``None`` if not wrappable.
+    """
+    for klass in type(module).__mro__:
```
laserkelvin: Do you want to just call it `_class` instead...?

laserkelvin (changes requested):
I think on the whole, the general functionality makes sense to me - I've made some suggestions and raised some questions about the scope of the support (i.e. what kind of layers can we apply LoRA to and customizability)

The one other thing I wanted to ask about is user-facing documentation: I understand that the module is in experimental right now, but would be helpful to have something to specify user entrypoints into the interface, how it works, and what is and what isn't supported

physicsnemo/experimental/peft/lora.py
```diff
@@ -0,0 +1,305 @@
...
+        )
+        self.lora_B = nn.Parameter(
+            torch.zeros(
+                rank, out_features, device=ref_weight.device, dtype=ref_weight.dtype
+            )
+        )
+        self.lora_dropout: nn.Module = (
+            nn.Dropout(p=dropout) if dropout > 0.0 else nn.Identity()
+        )
+        nn.init.kaiming_uniform_(self.lora_A, a=math.sqrt(5))
```
ys-teh: Note these may not be compatible with some lora layers as well. Here's an [example](https://gitlab-master.nvidia.com/yteh/nvalchemi-experiments/-/blob/expt/lora-finetuning/experiments/lora_finetuning/nvalchemi/training/lora.py?ref_type=heads#L106) for your context. It is also fine to override the entire method for very specific cases like the equivariant one.

physicsnemo/experimental/peft/lora.py
mnabian: It mirrors HF PEFT's default LoRA init. But I agree it's a reasonable default but not necessarily best, so added support for custom initialization via Callable. Also thanks for the paper, it was an interesting read, and a good validation for our initialization choices for A & B. Please see c90a64e6c1a20205d70c790fdc800b02d5f2f023.

physicsnemo/experimental/peft/apply.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def resolve_targets(model: nn.Module, config: LoRAConfig) -> list[Target]:
+    """Return ``(full_name, parent, child_attr, module)`` for every wrappable,
+    matched layer.
+
+    Matches against *fully-qualified* names. Skips already-wrapped layers
+    (double-apply guard) and module types not in the wrapper registry. When
+    ``config.wrap_mlp`` is set, also matches the default MLP feed-forward
+    patterns (additive to the primary selector).
```
ys-teh: Do you think we can add additional eligibility/compatibility check depending on the lora layer types? E.g., introduce `lora_layer_cls.is_compatible(module)` and call it within `resolve_targets`. In the case of `e3nn.o3.Linear`, we can specify additional checks like having at least one shared input/output irreps. In the simple linear case, `is_compatible` can be empty or include simple weight shape check.

physicsnemo/experimental/peft/lora.py
mnabian: I agree. _make_lora_params is standard-tensor conveniences, not part of the contract. Documented them as optional on LoRALayer: a wrapper is free to create its own params/init/delta (like the equivariant example) as long as it ends up exposing lora_A/lora_B. Thanks for the example! Also please see my answer to Kelvin's comment about supporting other initializations, and the corresponding changes in this commit: https://github.com/NVIDIA/physicsnemo/commit/c90a64e6c1a20205d70c790fdc800b02d5f2f023

physicsnemo/experimental/peft/lora.py
mnabian: Yes, fixed here: d4b5c8c3e28b42077229b67bc939a54933043f3c

physicsnemo/experimental/peft/utils.py
mnabian: You are right. This is not needed and is fixed here: d4b5c8c3e28b42077229b67bc939a54933043f3c

physicsnemo/experimental/peft/utils.py
mnabian: Done - expanded to an explicit if/else: d4b5c8c3e28b42077229b67bc939a54933043f3c

physicsnemo/experimental/peft/utils.py
mnabian: Good point! Fixed here: d4b5c8c3e28b42077229b67bc939a54933043f3c

physicsnemo/experimental/peft/lora.py
mnabian: Added in this commit: d4b5c8c3e28b42077229b67bc939a54933043f3c

physicsnemo/experimental/peft/lora.py
mnabian: Added in this commit: d4b5c8c3e28b42077229b67bc939a54933043f3c

physicsnemo/experimental/peft/lora.py
mnabian: Added in this commit: d4b5c8c3e28b42077229b67bc939a54933043f3c

physicsnemo/experimental/peft/merge.py
mnabian: I understand your point about modularity. But the thing merge_lora really does beyond folding the weights is unwrap the layer out of its parent so you get a clean, zero-overhead model back and that can't go through nn.Module.apply, since fn never sees the parent (it'd merge the weights but leave the wrappers behind). The selective case is already covered though. We can set mergeable = False on the layers we want to keep live and merge_lora only folds the rest:
```
# keep these layers as live adapters; merge everything else
for name, m in model.named_modules():
    if is_lora_layer(m) and should_keep_live(name):
        m.mergeable = False

merge_lora(model)   # folds + unwraps only the still-mergeable ones;
                    # the marked layers stay wrapped and trainable/swappable
```
I can document this, and if users want a nicer interface we can add a filter= to merge_lora later, depending on how this is used.

mnabian:
@laserkelvin @ys-teh Thanks for thoroughly reviewing the PR and for your great comments. All of those are addressed now. The user documentation is also ready and will have its own PR to physicsnemo-docs.

physicsnemo/experimental/peft/lora.py
mnabian: I added a guard and a warning. `_infer_in_out_features` now raises a clear TypeError for a base with no in_features/out_features and no 2-D weight, and resolve_targets now warns (instead of silently skipping) when your selector matches a layer with no registered wrapper (e.g. an nn.Embedding). So embeddings won't silently no-op; you'll get a heads-up pointing at register_lora_wrapper. Full embedding support is a planned registry wrapper rather than the Linear path. I'd add it when a target model needs it, since our current physicsnemo models don't use embedding tables. Please see this commit: 6e66b6dd0b21732aca95d9b4db6867b2ba86403b, and this commit in response to Ying Shi's relevant comment: 350b658cb506599de9a1ad36be02a0b7f6136147

laserkelvin (approved):
LGTM, thanks for working through my comments!

physicsnemo/experimental/peft/apply.py
mnabian: This is a great suggestion and has been implemented in this commit: c122ec65865e3b41e6c0e8513cec4d082058bfc3

RishikeshRanade (approved):

ys-teh (approved):
Thanks for addressing the suggestions. Looks good.

coreyjadams (changes requested):
Hi @mnabian overall I think the PR looks fine.  Please add CODEOWNERS and consider where `peft` will eventually land in `importlinter` hierarchy in a release or two - probably right at the top?

To be concrete, though, the only change I request is codeowners.

coreyjadams (approved):

PR #1700, opened 2026-06-05
PhysicsNeMo and Newton integration for neural dynamics, differentiable simulation, and design optimization

mehdiataei:
# PhysicsNeMo Pull Request

PhysicsNeMo and Newton integration for neural dynamics, differentiable simulation, and design optimization.

## Description

This PR introduces an experimental integration between PhysicsNeMo and the Newton physics engine.

It adds:

- Reusable Newton environment, state, trajectory, distributed-training, and zero-copy Warp-to-Torch utilities.
- NeRD vector and entity-transformer models for learning and deploying Newton dynamics.
- Design-space, surrogate, grouped-optimization, geometry, candidate-selection, and verification tools.
- Complete examples for Cartpole and RJ45 NeRD dynamics, differentiable ball optimization, gripper co-design, and MPM nozzle inverse design.
- Lazy Newton and USD imports so PhysicsNeMo and non-Newton workflows remain unaffected when optional dependencies are absent.
- API documentation, example guides, generated results, tests, and end-to-end smoke coverage.

The integration keeps Newton as the authoritative simulator while PhysicsNeMo provides reusable model training, optimization, and deployment infrastructure.

## Checklist

## Dependencies

Newton support is optional and available through the `newton` dependency extra. The examples also require Newton’s example dependencies and Matplotlib. Python 3.14 conditionally installs `usd-core` until Newton provides it directly for that Python version.

## Review Process


mehdiataei:
https://github.com/newton-physics/newton/pull/3093
There is a workaround right now hardcoded in the integration that will be removed later.

docs/api_index.rst
```diff
@@ -19,6 +19,8 @@ API Reference
    api_diffusion
    api/physicsnemo.distributed.rst
    api/physicsnemo.domain_parallel.rst
+   api/physicsnemo.experimental.integrations.newton.rst
+   api/physicsnemo.experimental.integrations.newton.design_space.rst
```
CharlelieLrt: These items should not appear as separate in the left navbar of the docs. Maybe they should not even appear at all since it's in experimental.

pzharrington (approved):
DiT layer changes LGTM

physicsnemo/mesh/sampling/random_point_sampling.py
```diff
@@ -27,10 +27,45 @@
     from physicsnemo.mesh.mesh import Mesh
 
 
+def normalize_point_clouds(
```
peterdsharpe: A few items about this function:

a) Does this need to be restricted to point clouds (i.e., `Mesh[0, N]`? Seems like this is generic and could apply to arbitrary `Mesh` objects.) In this case, the function would accept a `Mesh` rather than a tensor, and it would probably be named something like `normalize_points()` perhaps

b) This isn't really a sampling op, so probably would be better to have this live somewhere other than this file. I'm thinking `physicsnemo/mesh/transformations/` and probably in a new `normalize.py` file? And it can import from `physicsnemo/mesh/transformations/geometric.py`, since these are fundamentally just scale-and-shift geometry transforms (i.e., affine)

c) This function needs a complete NumPy-style docstring, like other functions in the repo. In particular, the return signature needs to be documented better here.

physicsnemo/mesh/sampling/random_point_sampling.py
```diff
@@ -27,10 +27,45 @@
...
+    points: Float[torch.Tensor, "... n_points n_dims"],
+    *,
+    eps: float = 1.0e-8,
+) -> tuple[
+    Float[torch.Tensor, "... n_points n_dims"],
+    Float[torch.Tensor, "... n_dims"],
+    Float[torch.Tensor, "... 1"],
```
peterdsharpe: Why return a singleton trailing dimension on `safe_radii`?

physicsnemo/mesh/sampling/random_point_sampling.py
```diff
@@ -55,6 +90,9 @@ def sample_random_points_on_cells(
         - alpha = 1.0: Uniform distribution over the simplex (default)
         - alpha > 1.0: Concentrates samples toward the center of each cell
         - alpha < 1.0: Concentrates samples toward vertices and edges
+    generator : torch.Generator, optional
+        Random generator used when ``alpha=1``. This enables reproducible
+        sampling without modifying global Torch random state.
```
peterdsharpe:
```suggestion
        sampling without modifying global PyTorch random state.
```

physicsnemo/mesh/sampling/random_point_sampling.py
```diff
@@ -113,10 +151,18 @@ def sample_random_points_on_cells(
...
     if alpha == 1.0:
-        distribution = torch.distributions.Exponential(
-            rate=torch.ones((), device=mesh.points.device),
+        uniform = torch.rand(
+            (n_samples, mesh.n_manifold_dims + 1),
+            dtype=mesh.points.dtype,
+            device=mesh.points.device,
+            generator=generator,
+        )
+        raw_barycentric_coords = -torch.log(
```
peterdsharpe, line 128:
```suggestion
        raw_barycentric_coords = -torch.log(  # Forming an Exponential distribution, but allowing for Generator usage
```

This is a nice implementation, but it's not obvious to a reader (without historical context) what exactly is happening here, so it earns a comment

physicsnemo/mesh/sampling/random_point_sampling.py
```diff
@@ -113,10 +151,18 @@ def sample_random_points_on_cells(
...
+            uniform.clamp_min(torch.finfo(uniform.dtype).tiny)
         )
     else:
+        if generator is not None:
+            raise ValueError("generator is only supported when alpha=1.0")
```
peterdsharpe:
```suggestion
            raise ValueError("generator is only supported when alpha=1.0.")
```

physicsnemo/mesh/sampling/random_point_sampling.py
```diff
@@ -141,3 +188,52 @@ def sample_random_points_on_cells(
...
+    """
+    if num_samples <= 0:
+        raise ValueError("num_samples must be positive")
+    if mesh.n_cells == 0:
+        raise ValueError("mesh must contain at least one cell")
+    vertices = mesh.points[mesh.cells]
+    relative_vectors = vertices[:, 1:] - vertices[:, :1]
+    from physicsnemo.mesh.geometry import compute_cell_areas  # noqa: PLC0415
+
+    measures = compute_cell_areas(relative_vectors)
```
peterdsharpe: This logic here should use `mesh.cell_areas`, not `compute_cell_areas()`. The former will use caching for better performance, the latter will not.

physicsnemo/mesh/mesh.py
```diff
@@ -1367,6 +1370,22 @@ def sample_random_points_on_cells(
...
             alpha=alpha,
+            generator=generator,
+        )
+
+    def sample_random_points(
+        self,
+        num_samples: int,
+        *,
+        generator: torch.Generator | None = None,
+    ) -> torch.Tensor:
```
peterdsharpe: Let's `jaxtyping` the return here

physicsnemo/mesh/mesh.py
```diff
@@ -1367,6 +1370,22 @@ def sample_random_points_on_cells(
...
+        """Sample uniformly over the length, area, or volume of this mesh."""
```
peterdsharpe: Function is missing proper/complete NumPy-style docstring

physicsnemo/mesh/sampling/__init__.py
```diff
@@ -23,6 +23,8 @@
 """
 
 from physicsnemo.mesh.sampling.random_point_sampling import (
+    normalize_point_clouds,
```
peterdsharpe: See later comment (in the definition itself), but normalize_point_clouds probably doesn't belong in the random sampling module. Maybe `transformations`?

peterdsharpe (changes requested):
Submitting first round of comments, only touching on `Mesh`-related changes. Overall seems fine; though `normalize_point_clouds` probably should be genericized (to `Mesh`-level) and moved (perhaps to `transformations`, rather than `sampling`).

eanswer:
Hi @mehdiataei — I went through the **NeRD integration** and the **gripper co-design** example and ran both locally. Overall this is a clean, well-engineered, and solid reimplementation — nice work. Comments below are mostly high-level (I'm not deeply familiar with the PhysicsNeMo codebase itself). Happy to discuss any of them.

### NeRD — strengths
- **Modularity** — the codec / problem / training / deployment split is clean and easy to follow.
- **Distributed training** — DDP with global (all-reduced) normalization statistics.
- **Entity-token transformer** (`NeRDEntityTransformer`) — a nice architectural extension for the maximal-coordinate RJ45 case.
- **Self-contained** data-generation → training → evaluation flow.
- **Adaptation of NeRD to body-based (maximal-coordinate) state and the contact-rich VBD teacher** — well done.

### NeRD — suggestions / future work
1. A **dataset inspection/visualization** utility would help debugging and onboarding.
2. **Parallelize data generation/evaluation for RJ45 example** would improve the scalability of the training pipeline (only if Newton's VBD simulation allows parallization).
3. **Contact information isn't a model input** — the token is `cat([state, inputs])` only, and contacts are explicitly dropped in the input callbacks ([nerd/inputs.py#L114](https://github.com/mehdiataei/physicsnemo/blob/mehdiataei-newton/physicsnemo/experimental/integrations/newton/nerd/inputs.py#L114), [nerd/problem.py#L126](https://github.com/mehdiataei/physicsnemo/blob/mehdiataei-newton/physicsnemo/experimental/integrations/newton/nerd/problem.py#L126))
4. **No robot-root-centric representation for floating-base systems** ([nerd/codecs/entity.py#L213](https://github.com/mehdiataei/physicsnemo/blob/mehdiataei-newton/physicsnemo/experimental/integrations/newton/nerd/codecs/entity.py#L213))

   → (3) and (4) are the main levers for NeRD's generalization across environments and tasks, so they'd be the highe-value items if/when they're in scope in future.
5. **Rotation-vector (SO(3) log-map) target** — using `log(q_next · q_cur⁻¹)` as a 3-vector is the right minimal tangent-space target for small per-step rotations. In NeRD, we tested this formulation and found it performed slightly worse than the quaternion representation in our examples, but it should be minor issue as the improvements were mostly negligible. 

### Design Optimization (gripper)
Looks good overall. One thing I'd like to clarify: in the demo you showed me, the design was optimized from a **two-finger to a three-finger** gripper, but in the code the finger count looks **fixed** (`FINGER_COUNT = 2` in `gripper_scene.py`, not a design variable). Is finger count optimized in a newer version (i.e., morphology co-design), or were those two separate fixed-count runs? Just want to make sure I understand the intended scope.

loliverhennigh:
@mehdiataei Small docs-style request: could you convert the remaining Google-style docstrings in the new Newton integration to NumPy docstring style?

I noticed several `Args:`, `Returns:`, and `Attributes:` sections in the new experimental Newton package, for example in `adjoint.py`, `env.py`, `nerd/deploy.py`, `nerd/problem.py`, `nerd/spec.py`, `scene.py`, `surrogate.py`, `trajectory.py`, and `worlds.py`. Most of the shared `mesh`/`pointnet` additions already look NumPy-style, so this would make the new public/experimental APIs consistent with the rest of the docs.

One small process note: these experimental paths are excluded from the current docstring coverage/lint checks, so CI may not catch this automatically.

mehdiataei:
@eanswer Hi Jie, thank you for the thoughtful review. I’ve addressed the main points: added many-world RJ45 data generation, contact-conditioned input tooling, and robot-centric codec support. The RJ45 benchmark intentionally remains command-only and world-frame to avoid closed-loop contact distribution shift and preserve the fixed-environment collision context. I also clarified these choices and limitations in the documentation.

mehdiataei:
@loliverhennigh Fixed! Thanks.

loliverhennigh:
mpm_beam_pod_nerd_newton_viewer

loliverhennigh (approved):
So I used this to build out the MPM example. It looks pretty good IMO. There might be changes needed in the API so we should keep it in experimental for a bit longer. Lets move fast on this and not hold this work up too much. Ill get the MPM one ready later today and we can keep going from there.

physicsnemo/mesh/sampling/random_point_sampling.py
```diff
@@ -145,3 +156,67 @@ def sample_random_points_on_cells(
...
+        raise ValueError("num_samples must be positive")
+    if mesh.n_cells == 0:
+        raise ValueError("mesh must contain at least one cell")
+    measures = mesh.cell_areas
+    if not torch.compiler.is_compiling():
+        if not bool(torch.isfinite(measures).all()):
+            raise ValueError("mesh cell measures must be finite")
+        if float(measures.sum()) <= 0.0:
+            raise ValueError("mesh must contain positive-measure cells")
+    cell_indices = torch.multinomial(
```
peterdsharpe: `torch.multinomial` raises `RuntimeError: number of categories cannot exceed 2^24` for meshes with more than ~16.7M cells, which we regularly exceed on large CFD/automotive meshes

docs/img/newton/rj45_vbd_vs_nerd.gif
peterdsharpe, line 1: Can this `*.gif` be compressed (encoding, framerate, resolution) or converted to a static image? The entire repo right now is ~200 MB, so committing this GIF as-is (9.14 MB) would increase repo size by ~5%. There's another GIF that's 5.56 MB that we might also look to compress or make static as well.

physicsnemo/mesh/sampling/random_point_sampling.py
peterdsharpe: Suggested fix:
```python
    if num_samples <= 0:
        raise ValueError("num_samples must be positive")
    if mesh.n_cells == 0:
        raise ValueError("mesh must contain at least one cell")
    measures = mesh.cell_areas
    if not torch.compiler.is_compiling():
        if not bool(torch.isfinite(measures).all()):
            raise ValueError("mesh cell measures must be finite")
        if float(measures.sum()) <= 0.0:
            raise ValueError("mesh must contain positive-measure cells")
    ### Pick cells with probability proportional to their measure.
    # Inverse-CDF sampling (cumulative measure + ``searchsorted``) replaces
    # ``torch.multinomial``, whose 2**24-category cap rejects large meshes.
    # Accumulate in float64: a float32 running total loses resolution once it
    # exceeds ~2**24, rounding small cell measures down to zero probability
    # (~8% of cells on a 20M-cell mesh).
    cumulative_measure = torch.cumsum(measures.double(), dim=0)
    thresholds = cumulative_measure[-1] * torch.rand(
        num_samples,
        dtype=cumulative_measure.dtype,
        device=measures.device,
        generator=generator,
    )
    cell_indices = torch.searchsorted(
        cumulative_measure, thresholds, right=True
    ).clamp_max_(mesh.n_cells - 1)
    return sample_random_points_on_cells(
        mesh,
        cell_indices,
        generator=generator,
    )
```

peterdsharpe (changes requested):
Hi @mehdiataei - I'm realizing that of the two newly-added functions in `physicsnemo.mesh` (`normalize_points()` and `sample_random_points()`) as part of this PR, neither appears to have any downstream callers anywhere else in this PR.

Is there a reason these need to come as part of this PR? If not, could we please excise those from this PR, and discuss adding them in a separate PR? That'll help accelerate merging this Newton PR, and give us more time to evaluate the API surface we want to expose here. Thanks!

EDIT: whoops, looks like `.sample_random_points()` has a caller in `physicsnemo.experimental.integrations.newton.geometry` - let's keep that, but remove `normalize_points()`.

physicsnemo/experimental/integrations/newton/codesign.py
```diff
@@ -0,0 +1,439 @@
...
+        candidate_indices = torch.argmin(losses, dim=-1)
+        order = torch.argsort(final_objective)
+        return GroupedDesignResult(
+            designs=designs.detach()[order].cpu().numpy(),
+            losses=final_objective[order].cpu().numpy(),
+            candidate_indices=candidate_indices[order].cpu().numpy(),
+            history=history[:, order].cpu().numpy(),
+            design_history=design_history[:, order].cpu().numpy(),
+            top_k_history=top_k_history.cpu().numpy(),
+            design_space=design_space,
```
coreyjadams, line 345: Each of these is a blocking call, is this necessary like this?

physicsnemo/experimental/integrations/newton/design.py
```diff
@@ -0,0 +1,337 @@
...
+        return cls(
+            FullyConnected(
+                in_features=design_dim,
+                out_features=1,
+                layer_size=hidden_dim,
+                num_layers=depth,
+                activation_fn="silu",
+            ),
+            device=device,
+        )
```
coreyjadams, line 97: What motivates this factory design here?  Most of our object oriented programming in physicsnemo is direct instantiation.

physicsnemo/experimental/integrations/newton/design.py
```diff
@@ -0,0 +1,337 @@
...
+            optimizer.step()
+        self.eval()
+        # Report the error under the final, eval-mode weights that predict()
+        # uses, rather than the train-mode loss from before the last step.
+        with torch.no_grad():
+            prediction = self.model(x)
+            _validate_score_prediction(prediction, target)
+            final_mse = nn.functional.mse_loss(prediction, target)
+        self._is_fitted.fill_(True)
+        return float(final_mse * std.square().squeeze())
```
coreyjadams, line 158: This looks like a training loop that would eventually be merged into our core repository / framework, when it might be better as an example.  How does this fit in?

physicsnemo/experimental/integrations/newton/design.py
```diff
@@ -0,0 +1,337 @@
...
+                f"round {round_index + 1}/{rounds}: evaluated {len(normalized_designs)}, "
+                f"best score {history[-1]:.4f}"
+            )
+
+    return DesignResult(
+        normalized_designs=normalized_designs,
+        scores=scores,
+        history=history,
+        surrogate=surrogate,
+    )
```
coreyjadams, line 308: Again, this seems better suited to an example workflow rather than a core framework tool.

CHANGELOG.md
```diff
@@ -76,6 +82,58 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   geometry-latent kNN distance as a continuous score for downstream
   consumers (e.g. AL acquisition) without the boolean thresholding /
   warning emission of `OODGuard.check()`.
+- Adds experimental Newton physics-engine integration under
```
mnabian, line 85: Not a big deal but I think these logs can be summarized.

examples/newton/diffsim/example_diffsim_cartpole_bptt.py
mnabian, line 1: Nit: We typically classify the examples by application, so we could rename Newton to Robotics. This is not followed strictly though, so up to you.

physicsnemo/models/pointnet/pointnet_mlp.py
```diff
@@ -0,0 +1,177 @@
...
+        self.point_features = int(point_features)
+
+        self.point_encoder = PointNetEncoder(
+            in_channels=point_channels,
+            out_features=point_features,
+            hidden_channels=point_hidden_channels,
+            activation_fn=activation_fn,
+            pooling=pooling,
+        )
+        self.head = FullyConnected(
```
mnabian, line 138: You can use Physicsnemo MLP instead.

physicsnemo/experimental/models/nerd/nerd.py
```diff
@@ -0,0 +1,261 @@
...
+    func_torch: bool = False
+    auto_grad: bool = False
+
+
+class _Block(nn.Module):
+    """Pre-norm transformer block: attention then a GELU MLP, both residual."""
+
+    def __init__(self, n_embd: int, n_head: int, dropout: float, bias: bool) -> None:
+        super().__init__()
+        self.ln_1 = nn.LayerNorm(n_embd, bias=bias)
```
mnabian, line 84: Use physicsnemo.nn's layernorm to benefit from `te.LayerNorm` when te is available.

physicsnemo/experimental/models/nerd/nerd.py
```diff
@@ -0,0 +1,261 @@
...
+        self.attn = TimmSelfAttention(
+            hidden_size=n_embd,
+            num_heads=n_head,
+            attn_drop_rate=dropout,
+            proj_drop_rate=dropout,
+            qkv_bias=bias,
+            proj_bias=bias,
+        )
+        self.ln_2 = nn.LayerNorm(n_embd, bias=bias)
+        self.mlp = nn.Sequential(
```
mnabian, line 94: Use physicsnemo.nn's MLP

physicsnemo/experimental/models/nerd/nerd.py
```diff
@@ -0,0 +1,261 @@
...
+        self.ln_1 = nn.LayerNorm(n_embd, bias=bias)
+        self.attn = TimmSelfAttention(
+            hidden_size=n_embd,
+            num_heads=n_head,
+            attn_drop_rate=dropout,
+            proj_drop_rate=dropout,
+            qkv_bias=bias,
+            proj_bias=bias,
+        )
+        self.ln_2 = nn.LayerNorm(n_embd, bias=bias)
```
mnabian, line 93: Same here

physicsnemo/experimental/models/nerd/nerd.py
```diff
@@ -0,0 +1,261 @@
...
+
+        # The reference nanoGPT input projection retains its bias even when the
+        # Transformer block bias option is disabled.
+        self.token_embed = nn.Linear(input_dim, n_embd)
+        self.pos_embed = nn.Embedding(block_size, n_embd)
+        self.drop = nn.Dropout(dropout)
+        self.blocks = nn.ModuleList(
+            [_Block(n_embd, n_head, dropout, bias) for _ in range(n_layer)]
+        )
+        self.ln_f = nn.LayerNorm(n_embd, bias=bias)
```
mnabian, line 205: Same here

physicsnemo/experimental/models/nerd/nerd.py
```diff
@@ -0,0 +1,261 @@
...
+        self.feature_proj = nn.Linear(n_embd, n_embd, bias=False)
+        self.head = nn.Sequential(
```
mnabian, line 207: Same here

test/models/pointnet/test_pointnet.py
```diff
@@ -0,0 +1,189 @@
...
+# limitations under the License.
+
+import pytest
+import torch
+
+from physicsnemo.models.pointnet import PointNetEncoder, PointNetMLP
+from test import common
+
+
+def _pointnet_case(device, model_type):
```
mnabian, line 24: There should be a validate_forward_accuracy test. See other models' tests and the coding standards.

docs/img/newton/newton_gripper_before_after.gif
mnabian, line 1: I think we are limiting file sizes to less than 5MB.

docs/img/newton/rj45_vbd_vs_nerd.gif
mnabian, line 1: Same file size comment here

examples/newton/gripper/example_gripper_design.py
```diff
@@ -0,0 +1,72 @@
...
+        description="PhysicsNeMo offline neural geometry co-design of a Newton gripper"
+    )
+    output_dir = Path(__file__).parent / "outputs" / "gripper"
+    parser.add_argument("--output-dir", type=Path, default=output_dir)
+    parser.add_argument(
+        "--dataset",
+        type=Path,
+        default=output_dir / "gripper_offline_dataset.npz",
+    )
+    parser.add_argument("--regenerate-dataset", action="store_true")
```
mnabian, line 40: We use Hydra for configs, not argparse.

physicsnemo/models/pointnet/pointnet.py
```diff
@@ -0,0 +1,143 @@
...
+    torch.Size([8, 64])
+    """
+
+    def __init__(
+        self,
+        in_channels: int = 3,
+        out_features: int = 128,
+        hidden_channels: Sequence[int] = (64, 128),
+        activation_fn: str = "silu",
+        pooling: str = "max",
```
CharlelieLrt, line 96: Prefer `Literal` type hint instead of just `str` (at least when there are only a few accepted values)

physicsnemo/models/pointnet/pointnet.py
CharlelieLrt, line 1: In our usual process, new `Module` classes go to `experimental`. This one is fairly straightforward, so if there is high confidence that the API will remain stable, then we can make an exception and maybe the `experimental` staging can be bypassed.

physicsnemo/nn/module/dit_layers.py
```diff
@@ -203,6 +207,9 @@ class TimmSelfAttention(AttentionModuleBase):
     attn_mask : torch.Tensor, optional
         The attention mask to apply (passed to timm's Attention module).
         If ``None``, no mask is applied. Only supported for timm version 1.0.16 and higher.
+    is_causal : bool, optional, default=False
+        Whether to use timm's native causal-attention path. Only supported for
+        timm version 1.0.16 and higher.
```
CharlelieLrt, line 212: Do we really expect the `is_causal` argument to dynamically change from one call to the model's forward to another? If not, then maybe it would be better to have it as an argument of `__init__` instead of `forward`

coreyjadams (changes requested):
I reviewed some of these files.  It is a very large PR, and a number of the team have looked at pieces too.  I think the consensus is a number of components here are not following repo conventions.  

I think the value of robotics integrations are unquestioned, and there is no doubt that we should merge this PR eventually.  But the review at 30k LOC is untenable, to ensure we're well aligned.  Can we factor this into stacked and separable PRs?

mehdiataei:
> Hi @mehdiataei - I'm realizing that of the two newly-added functions in `physicsnemo.mesh` (`normalize_points()` and `sample_random_points()`) as part of this PR, neither appears to have any downstream callers anywhere else in this PR.
> 
> Is there a reason these need to come as part of this PR? If not, could we please excise those from this PR, and discuss adding them in a separate PR? That'll help accelerate merging this Newton PR, and give us more time to evaluate the API surface we want to expose here. Thanks!
> 
> EDIT: whoops, looks like `.sample_random_points()` has a caller in `physicsnemo.experimental.integrations.newton.geometry` - let's keep that, but remove `normalize_points()`.

Removed. It was a leftover from re-homing normalize_point_clouds during the previous review round. Thee gripper example uses an inline batched equivalent, so nothing depends on it anymore.

I also fixed fixed the sampling limit with a cumsum trick that that halves the transient memory as well from your suggested fix.

physicsnemo/experimental/integrations/newton/design.py
mehdiataei: There are reasons behind this and a few of the related decisions. Unfortunately, without the full context (and without making this PR even larger) they may not be immediately clear. We’re discussing potential ways to address this.

But I agree that we I should probably break this down into several components/files with more generic functionalities.

physicsnemo/experimental/models/nerd/nerd.py
```diff
@@ -0,0 +1,261 @@
...
+
+class NeRDTransformer(Module):
+    r"""Causal transformer that predicts one dynamics step as a per-DoF delta.
+
+    Consumes a window of per-step input tokens (a robot state embedding
+    concatenated with its actuation and any conditioning), runs a causal GPT, and
+    predicts the one-step change of the system. ``forward`` returns all tokens for
+    teacher-forced training; autoregressive deployment uses the final prediction.
+    The model is pure sequence-to-sequence: the delta-to-next-state conversion,
+    robot-centric framing, and normalization live in the caller.
```
CharlelieLrt, line 115: This is our `N`-th transformer-like architecture. What differentiate it from the others? It seems specifically intended for robotics use cases, but looking at the forward I am not seeing anything very specialized for robotics.
Maybe it's just a matter of renaming it, but calling it specifically `NeRDT` makes it sound like it is very specialized for a specific application. If that is the case, then it should probably not live in the core library (only models reusable across multiple applications should live in the core library).

CharlelieLrt:
> This looks like a training loop that would eventually be merged into our core repository / framework, when it might be better as an example. How does this fit in?

To second @coreyjadams , the core codebase/framework (including physicsnemo.experimental) is typically only for code that is reusable as building blocks for multiple applications. It is not 100% clear to me that the few model architecture are:
- different enough from other existing architectures to be worth integrating in the framework
- reusable across multiple applications

I might be missing some context, in which case simple clarifications in the docstrings and PR description may be sufficient.

PR #1721, opened 2026-06-11
Enhance dual graph functionality in PyVista integration with 2D meshes

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Previously, `from_pyvista(..., point_source="cell_centroids")` didn't work for 2D-only meshes (e.g., if you wanted to make a dual-graph of connectivity from a surface mesh). Now, it does.

- Updated terminology in documentation to clarify that adjacent cells share facets instead of faces.
- Introduced a new helper function `_cell_facet_point_ids` to yield point-id lists of a cell's facets, improving dimension-generic handling.
- Modified `_build_dual_graph_edges` to utilize the new facet identification method for better performance and accuracy.
- Added regression tests for dual graph generation in various mesh types, ensuring correct edge adjacency for triangulated surfaces and polylines.

## Checklist

## Review Process


peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1721/commits/dfb60349eeb61ff8d6900b293c501f8f2d5f5f17

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1721/commits/5b10b5f265e237702e237edfe3a03555fe546e5e

physicsnemo/mesh/io/io_pyvista.py
```diff
@@ -481,19 +483,56 @@ def _to_vtk_cell_array(cells_np: np.ndarray) -> np.ndarray:
...
+        A VTK cell.
+
+    Yields
+    ------
+    list[int]
+        Point ids of one (d-1)-facet. Nothing is yielded for 0D cells
+        (isolated points have no facets, hence no adjacency).
+    """
+    dim = cell.GetCellDimension()
+    if dim >= 3:
```
laserkelvin: Consider doing a match/case - I think in this case because `dim` checks are pretty readable already it's fine, but IIRC `match/case` actually es slightly more performant (not that it matters in the grand scheme of things)

physicsnemo/mesh/io/io_pyvista.py
```diff
@@ -481,19 +483,56 @@ def _to_vtk_cell_array(cells_np: np.ndarray) -> np.ndarray:
...
+        subcells = (cell.GetEdge(e) for e in range(cell.GetNumberOfEdges()))
+    elif dim == 1:
+        # A line cell's facets are its two endpoint vertices (0-faces).
+        for p in range(cell.GetNumberOfPoints()):
+            yield [cell.GetPointId(p)]
+        return
+    else:  # 0D: isolated points have no facets.
+        return
+    for sub in subcells:
+        yield [sub.GetPointId(p) for p in range(sub.GetNumberOfPoints())]
```
laserkelvin, line 530: Out of curiousity, does this return something deterministically ordered, and are consumers permutation invariant?

Don't want to invent problems but...

physicsnemo/mesh/io/io_pyvista.py
```diff
@@ -522,14 +561,13 @@ def _build_dual_graph_edges(
 
     for i in range(n_cells):
         cell = pyvista_mesh.GetCell(i)
-        for f in range(cell.GetNumberOfFaces()):
-            face = cell.GetFace(f)
-            face_pt_ids.Reset()
-            for p in range(face.GetNumberOfPoints()):
-                face_pt_ids.InsertNextId(face.GetPointId(p))
+        for facet_ids in _cell_facet_point_ids(cell):
```
laserkelvin, line 580: I don't know the data structure well, but this *feels* like it can get really big, and this loop will take a very long time. Is this true, and would we need to think about lifting this loop somehow out of Python?

laserkelvin (approved):
Generally LGTM, just leaving some minor things for you to consider!

physicsnemo/mesh/io/io_pyvista.py
peterdsharpe: Good call, old habits :D `match/case` is also prettier!

peterdsharpe: Fxied

physicsnemo/mesh/io/io_pyvista.py
peterdsharpe: Yes, this is deterministically-ordered - added a brief `Notes` block to capture

physicsnemo/mesh/io/io_pyvista.py
peterdsharpe: Yeah, this is inherently going to be slow :/

We can make it faster in Python via vectorized numpy facet-hashing (sort each cell's facets, match duplicates) but that only really works for homogeneous (e.g., all-tri), manifold meshes. Anything else, and it gets far uglier. VTK's `GetCellNeighbors` (which is underneath these calls) handles all these cases, albeit slowly, so I'd rather keep it general here. We could (and possibly eventually, should) lift this out of Python, but it's a pretty big undertaking that I'd rather defer for now

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1721/commits/68876382d777032ab9bc9e811f8cabbd7d8f5de7

physicsnemo/mesh/io/io_pyvista.py
laserkelvin: Yeah definitely out of scope for this PR, but just wanted to get a sense of whether or not you would be able to bring some of the required data into NumPy and then handle the bulk of the loops/processing outside of the GIL

Pre-mature optimization :)

PR #1739, opened 2026-06-16
Add torch support to zenith_angle.py

jleinonen:
# PhysicsNeMo Pull Request

## Description

The Earth2Studio PR https://github.com/NVIDIA/earth2studio/pull/919 identified the need for cos zenith calculations using torch. This PR adds the support for that to zenith_angle.py.

Changes:

- Implement generic math functions (`_sin`, `_cos`, `_sqrt`, etc.) to handle both torch and numpy inputs in a way that preserves the precision of the numpy version; route calculations to those.
- Add tests that check that torch calculations on both CPU and GPU return results equal to the numpy version.
- Bug fix: in `_greenwich_mean_sidereal_time`, a constant that is `6.2 * 10^-6` in the original AIAA paper/code (equation 2 in http://celestrak.org/publications/AIAA/2006-6753/AIAA-2006-6753-Rev3.pdf) had at some point been mistyped as `6.2 * 10e-6`, which is too big by a factor of 10. This factor is on a third order term of a polynomial; the term is very small so the effect of this change to calculations is negligible (in tests, the relative error on the final results is on the order of `1e-10`). This was flagged by an AI error analysis and confirmed manually - I think it's worth fixing so it'll stop the AI agents complaining going forward.

## Checklist

## Review Process


physicsnemo/utils/zenith_angle.py
```diff
@@ -100,12 +107,47 @@ def cos_zenith_angle_from_timestamp(
...
-    lat_rad = np.deg2rad(lat, dtype=dtype)
+    lon_rad = _deg2rad(lon, dtype=dtype)
+    lat_rad = _deg2rad(lat, dtype=dtype)
     julian_centuries = _timestamp_to_julian_century(timestamp)
     return _star_cos_zenith(julian_centuries, lon_rad, lat_rad)
 
 
+def _deg2rad(x: T, dtype: np.typing.DTypeLike | torch.dtype | None = None) -> T:
+    if isinstance(x, torch.Tensor):
+        if dtype is not None and not isinstance(dtype, torch.dtype):
```
pzharrington, line 132: `not isinstance(dtype, torch.dtype)` doesn't directly guarantee it's a numpy-compatible type, though I see the type hint in the function signature suggests it. Maybe have it directly check for the `np.typing.DTypeLike` for clarity here?

pzharrington: And though it's an internal helper maybe a brief mention of the intended dtyping behavior between torch/np would be helpful in the docstring

test/utils/test_zenith_angle.py
```diff
@@ -87,3 +88,80 @@ def test_toa():
...
+# high-latitude summer scenario so that both branches of _integrate_abs_cosz are
+# exercised.
+_PARITY_CASES = [
+    datetime(2020, 3, 21, 12, 0, 0, tzinfo=UTC).timestamp(),  # equinox noon
+    datetime(2020, 7, 6, 9, 0, 0, tzinfo=UTC).timestamp(),  # summer morning
+    datetime(2020, 1, 15, 3, 0, 0, tzinfo=UTC).timestamp(),  # nighttime
+]
+
+_PARITY_LONS = np.array([-90.0, 0.0, 40.0, 120.0], dtype=np.float32)
+_PARITY_LATS = np.array([-60.0, 0.0, 40.0, 80.0], dtype=np.float32)
```
pzharrington: At the cost of making these tests slightly more expensive, I suggest using more realistic values for maximum sanity -- maybe a 1 degree 181x360 grid?

test/utils/test_zenith_angle.py
pzharrington, line 1: Is it possible to add some parity tests in torch in float64 precision and see if we can get bit-exact parity? There may be differences due to sin/cos implementations in CUDA torch vs. numpy but just want to see

physicsnemo/utils/zenith_angle.py
jleinonen: `np.typing.DTypeLike` is a type alias rather than a real class so `isinstance(dtype, np.typing.DTypeLike)` doesn't work (it raises a `TypeError`). Did you have some other way of doing that? Otherwise I put a comment here to clarify what the intent is.

jleinonen: And I also added a docstring to explain the dtype behavior

test/utils/test_zenith_angle.py
jleinonen: I added this, it's now 181x360

test/utils/test_zenith_angle.py
jleinonen: The tests now run at both float32 and float64 precision. I couldn't get an exact match but it now uses a lower tolerance for the float64 tests.

physicsnemo/utils/zenith_angle.py
pzharrington: Looks good

pzharrington (approved):

PR #1741, opened 2026-06-22
GeoTransolver Optimization 1 / 4: Warp-Free SDF

coreyjadams:
# PhysicsNeMo Pull Request

For the GeoTransolver Datapipe, it has become necessary to purge warp's SDF from the volumetric datapipe.  The warp stream-ordered destructors have different memory retention behavior than pytorch.  Despite several weeks of effort, there is just no way to reconcile this without changes to the warp core repo, and I don't think we'll make that work on a short time scale.

This PR adds a torch + triton based SDF computation based on morton Z-ordering and physicsnemo mesh.  I put it in the datapipes but we could (should?) move it to the mesh package itself.

## Checklist

## Review Process


physicsnemo/datapipes/_timing.py
```diff
@@ -0,0 +1,117 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
```
peterdsharpe: Discussed offline, plan is to remove this file (debugging-only). Tagging here to track.

physicsnemo/mesh/spatial/sdf.py
```diff
@@ -0,0 +1,694 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
```
peterdsharpe, line 1: Discussed offline; plan is for this to migrate to `physicsnemo.mesh.spatial`. Just dropping a comment to track.

physicsnemo/mesh/spatial/_sdf_triton.py
```diff
@@ -0,0 +1,713 @@
...
+        """One query per lane; bounded-stack near-first DFS for nearest triangle."""
+        pid = tl.program_id(0)
+        off = pid * BLOCK + tl.arange(0, BLOCK)
+        m = off < N
+
+        qx = tl.load(query_ptr + off * 3 + 0, mask=m, other=0.0)
+        qy = tl.load(query_ptr + off * 3 + 1, mask=m, other=0.0)
+        qz = tl.load(query_ptr + off * 3 + 2, mask=m, other=0.0)
+
+        best = tl.zeros((BLOCK,), tl.float32) + max_dist_sq
```
peterdsharpe, line 249: If `max_dist` is too small, this kernel will return `sdf=0`, `hit_point=query_point`. So you could get into a situation where a far point reports as on-surface.

We should at-minimum document this (probably down below in the non-kernel function) since this could be a footgun (I know we default to 1e8 max_dist for now, but still). And maybe consider using a sentinel distance value (NaN?)?

physicsnemo/datapipes/transforms/_sdf_triton.py
```diff
@@ -0,0 +1,713 @@
...
+            active = sp > 0
+
+        tl.store(out_dist_ptr + off, best, mask=m)
+        tl.store(out_face_ptr + off, best_face, mask=m)
+        tl.store(out_pt_ptr + off * 3 + 0, bpx, mask=m)
+        tl.store(out_pt_ptr + off * 3 + 1, bpy, mask=m)
+        tl.store(out_pt_ptr + off * 3 + 2, bpz, mask=m)
+
+    @triton.jit
+    def _winding_kernel(
```
peterdsharpe: Per offline discussion, sounds like we're removing the Triton-backed sign winding number utils; skipping review. If this changes, LMK

physicsnemo/datapipes/transforms/_sdf_triton.py
```diff
@@ -0,0 +1,713 @@
...
+        MAX_LEAF: tl.constexpr,
+    ):
+        """One query per lane; per-thread DFS accumulating the winding number.
+
+        Far nodes (query outside ``beta`` * node radius) contribute a single
+        dipole term; near/leaf nodes are refined / evaluated exactly. The
+        accumulated solid angle is divided by ``4*pi`` by the caller.
+        """
+        pid = tl.program_id(0)
+        off = pid * BLOCK + tl.arange(0, BLOCK)
```
peterdsharpe: Claude finds an interesting overflow bug here (same as in triangle case):

**Triton path is unbounded in query count — int32 offset overflow (silent corruption) + no chunking (OOM).** `physicsnemo/datapipes/transforms/_sdf_triton.py`
- Unlike the torch fallback (which chunks queries at `_NEAREST_QUERY_CHUNK = 1<<18`), the Triton
  entry points launch over **all** `n_queries` in one grid and allocate the per-lane stack
  monolithically: `stack = torch.empty(n_queries, _STACK_SIZE, int32)` (lines 585, 680) = **256
  bytes/query**. At ~30M queries that is ~7.7 GB for the stack alone (plus query/output buffers and
  the Morton argsort temporaries) → OOM in the large-N external-aero regime.
- Worse, all per-lane address math uses an **int32** lane index: `off = pid*BLOCK + tl.arange(0,BLOCK)`
  (line 303/427). `off * STACK_SIZE` (lines 318, 327, 387, 389, 437, 442, 506, 508) is an int32×int32
  product that **wraps at off > 2³¹/64 = 33,554,432**, before being sign-extended onto the 64-bit base
  pointer. Past that, lanes read/write each other's DFS stacks on a `torch.empty` buffer → **silent
  out-of-bounds, wrong nearest-triangle / wrong winding sign, no crash.** (`off*3` for query/output
  wraps at ~715M.) This is the classic Triton large-tensor indexing footgun.
- **Fix:** (a) promote the lane index to int64 once at kernel top —
  `off = pid.to(tl.int64)*BLOCK + tl.arange(0,BLOCK).to(tl.int64)` (covers every derived offset;
  `node`/`cell` indices stay safely int32). AND (b) chunk both Triton launches over `query_s` with a
  single reused fixed-size `(CHUNK, _STACK_SIZE)` stack (chunk the *already-Morton-sorted* queries so
  warp coherence is preserved; write `out_*_s` slices before the final `perm` scatter-back). This caps
  memory and keeps offsets in range. Add a large-N regression test.

physicsnemo/datapipes/transforms/_sdf_triton.py
```diff
@@ -0,0 +1,713 @@
...
+        out_pt_ptr,  # (N, 3) f32 closest point
+        N,
+        max_dist_sq,
+        BLOCK: tl.constexpr,
+        STACK_SIZE: tl.constexpr,
+        MAX_LEAF: tl.constexpr,
+    ):
+        """One query per lane; bounded-stack near-first DFS for nearest triangle."""
+        pid = tl.program_id(0)
+        off = pid * BLOCK + tl.arange(0, BLOCK)
```
peterdsharpe: Claude finds an interesting overflow bug here:

**Triton path is unbounded in query count — int32 offset overflow (silent corruption) + no chunking (OOM).** `physicsnemo/datapipes/transforms/_sdf_triton.py`
- Unlike the torch fallback (which chunks queries at `_NEAREST_QUERY_CHUNK = 1<<18`), the Triton
  entry points launch over **all** `n_queries` in one grid and allocate the per-lane stack
  monolithically: `stack = torch.empty(n_queries, _STACK_SIZE, int32)` (lines 585, 680) = **256
  bytes/query**. At ~30M queries that is ~7.7 GB for the stack alone (plus query/output buffers and
  the Morton argsort temporaries) → OOM in the large-N external-aero regime.
- Worse, all per-lane address math uses an **int32** lane index: `off = pid*BLOCK + tl.arange(0,BLOCK)`
  (line 303/427). `off * STACK_SIZE` (lines 318, 327, 387, 389, 437, 442, 506, 508) is an int32×int32
  product that **wraps at off > 2³¹/64 = 33,554,432**, before being sign-extended onto the 64-bit base
  pointer. Past that, lanes read/write each other's DFS stacks on a `torch.empty` buffer → **silent
  out-of-bounds, wrong nearest-triangle / wrong winding sign, no crash.** (`off*3` for query/output
  wraps at ~715M.) This is the classic Triton large-tensor indexing footgun.
- **Fix:** (a) promote the lane index to int64 once at kernel top —
  `off = pid.to(tl.int64)*BLOCK + tl.arange(0,BLOCK).to(tl.int64)` (covers every derived offset;
  `node`/`cell` indices stay safely int32). AND (b) chunk both Triton launches over `query_s` with a
  single reused fixed-size `(CHUNK, _STACK_SIZE)` stack (chunk the *already-Morton-sorted* queries so
  warp coherence is preserved; write `out_*_s` slices before the final `perm` scatter-back). This caps
  memory and keeps offsets in range. Add a large-N regression test.

peterdsharpe (approved):
Overall looks good! Notes:

- Noting that the plan is to migrate the SDF implementations here into `physicsnemo.mesh.spatial`, which I think is the right call.

**Two major issues that should be resolved before merging**:

## Default sign path isn't Warp-equivalent for nonconvex meshes

`_pseudo_normal_sign` decides inside/outside from the single nearest face normal - but when the closet point lands on an edge/vertex, behavior differs from Warp. Specifically, we should be using the angle-weighted pseudo-normal of incident faces, not just picking one. This manifests in practice as errors in SDF magnitude and even sign (since you can land on a different side of the half-plane). I've added a failing regression test that shows this on an L-shaped geometry: `test_sdf_pseudo_normal_sign_wrong_at_sharp_edges` in `test/datapipes/transforms/test_sdf_torch.py` (~18 misclassified points on a exact L-prism; the winding-number companion `test_sdf_winding_sign_correct_at_sharp_edges` passes). We should implement the true angle-weighted pseudo-normal (accumulate normals at edges/vertices).

## Triton overflow issue:

In two places in the Triton path, there's allocation logic like:
```python
off = pid*BLOCK + tl.arange(0,BLOCK)
```

This defaults to int32 and hence overflows after O(tens of millions) of queries. Since this is an index, you basically get silently-wrong results. We should swap this to int64.
```python
off = pid.to(tl.int64)*BLOCK + tl.arange(0,BLOCK).to(tl.int64)
```

## Misc other

- Please add a CHANGELOG entry.

------


Going to ✅ this PR since this is 95% in great shape - after fixing these two issues and doing the migration I think we're good to merge.

physicsnemo/mesh/spatial/_sdf_triton.py
```diff
@@ -0,0 +1,437 @@
...
+implementation and the parity oracle.
+"""
+
+from __future__ import annotations
+
+import torch
+
+from physicsnemo.core.version_check import OptionalImport
+from physicsnemo.mesh.spatial import BVH
+from physicsnemo.mesh.spatial.bvh import _compute_morton_codes
```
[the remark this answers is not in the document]
coreyjadams: Outdated.

physicsnemo/datapipes/transforms/_sdf_triton.py
coreyjadams: This should be fixed

physicsnemo/datapipes/transforms/_sdf_triton.py
coreyjadams: Totally purged, resolving.

physicsnemo/datapipes/_timing.py
coreyjadams: Removed, resolving.

physicsnemo/mesh/spatial/sdf.py
coreyjadams: Done, resolving.

ktangsali (approved):
Changes to docs look good to me

physicsnemo/mesh/spatial/_sdf_triton.py
coreyjadams: This has been resolved, we no longer need a max_dist in this kernel actually.

PR #1742, opened 2026-06-22
GeoT optimization 2/4: Datapipes producer/consumer refactor + stream overlap

coreyjadams:
# PhysicsNeMo Pull Request

~This PR is stacked 🥞 on #1741. It requires the torch SDF implementation first.~. PR #1741 is merged, so this is ready to go.  

This PR is a rebuild of datapipe components, particularly at the IO level.  It's key changes are the following:
- There is a clear and logical separation now between "threads do IO fetching" and "streams do GPU preprocessing".
- There is an additional dataset type for "streaming" type datasets that are unbounded, so imaging "online simulation" datasets.  There is an example included to show this behavior.
-  The unified exteral aero recipe is updated accordingly.

## What does it mean to separate IO threads from preprocessing Steams?

The PR introduces the concept of an "IO Pump": it's self priming, so to speak.  It takes a few args (queue depth, worker fn), and runs a sentinal thread to keep the queue topped up.  Every time an IO payload is consumed, the pump automatically queues another, ensuring the IO is consistently ready to go.  **The key here is the IO threads can't block the GIL, so async IO is key.**

The `IOPump` is FIFO.  Exceptions are captured and held until attempted consumption of data: if a preload a a batch fails, it doesn't actually crash everything until we go to consume that batch.

While this is based purely on threads today, in principle it's about 80% of the way to enabling processPool concurrency for prefetching too.  That will break memory pinning, but for applications that simply cannot release the GIL, it could become a viable option.

For best performance, we should make sure the queue depth is larger than the batch size (so an entire batch is always ready).  And if IO latency is greater than model step time, we should preload enough batches to compensate.

## GPU Preprocessing with Streams

IO payloads are "consumed" in an N-1 fashion: while batch N is processed, batch N-1 is consumed on the GPU for preprocessing, then collated.   The datasets use two functions to isolate work: `_load_host` and `_consume`.  `_load_host` is what is orchestrated by the `IOPump`, and is CPU only and threaded.  `_consume`, on the other hand, is single threaded (really, the MAIN thread) and handles the queuing of transforms onto the appropriate stream (an arg to `_consume`) as well as a deferred sync.

## Cuda Stream Orchestration

Handling the thread and stream orchestration takes a little care.  Python threading is relatively easy to queue and block on: we keep the queue primed by filling it to `depth`, making sure it has enough work so that data is always ready on the host.  During training / inference, at step `N`, we eagerly pull batch `N+1` from the dataset and start the transfer and transformations, and collation. We apply a gating to batch `N` (which, in this model, was pulled during `N-1` processing) with cuda Events to ensure it completes before proceeding - this prevents the model from consuming it before it's ready.  During the processing of batch `N+1`, the same gating on batch `N+1` is applied right after the processing of `N+2` is triggered.  In this way, we keep preprocessing flowing concurrently with the model computation.

**Any data preprocessing that includes blocking syncs (nonzero, HtD blocking, etc) will cause this model to stumble and introduce bubbles in the GPU flow.**

## RNG Updates

This PR makes some RNG updates to ensure we can have reproducibility regardless of ordering.  The IO pump is FIFO but in general we shouldn't necessarily assume FIFO, and RNG reproducibility should be stable under multi-threading.

## Iterable Style Dataset

There was a request to support "Simulation only" type datasets, since we have this nice prefetching model now.  I included this here, and a tutorial / example as well.  It's like a pytorch "iterable" dataset (so no `__len__`` attribute) but doing on-the-fly generation on the GPU, in side streams, concurrent with model processing.  Limited currently to one-step-ahead processing.

## Misc Updates
- Fixes blocking transfers in radius search Mesh ID (@loliverhennigh to review Please?)
- Fix some small GPU syncs in the unified recipe
- Updated tests

# Doc Updates

I am still review the doc updates and readme updates.  The PR should not merge without those.

## Next Steps

I'd like to get this reviewed now, and think about some follow on PRs:
- I want to include an optional sentinel to watch for GPU syncs in the datapipe so they are easy to catch and fix.
- I would like to build features to compile the GPU preprocessing pipelines, both with torch inductor as well as cuda graphs.  This could reduce launch latency overhead a lot.
- I would eventually include ProcessPool parallelism in the datasets but that is P1.

## Checklist

## Review Process


loliverhennigh (approved):
Fine with me, pinning memory like this makes me a bit worried but its ok

examples/cfd/external_aerodynamics/unified_external_aero_recipe/src/train.py
```diff
@@ -365,6 +365,15 @@ def _run_epoch(
...
     epoch_t0 = time.perf_counter()
+    ### Single pinned scalar buffer reused every step so the loss D2H
+    ### transfer is async (non_blocking=True from device to pinned host
+    ### memory). The copy is issued right after forward_pass and read
+    ### just before the logger line; by then backward + optimizer.step
+    ### have run, giving the GPU time to complete the copy without
+    ### blocking the host.
+    _loss_pinned = (
+        torch.zeros(1, pin_memory=True) if torch.cuda.is_available() else None
+    )
```
peterdsharpe: This is very funky, but makes sense! Cool implementation

peterdsharpe (approved):
Reviewed parts I'm codeowner on: dataset yamls, unified recipe changes, and PhysicsNeMo-Mesh changes. All those look good to me!

negin513 (commented):
Cool to add the iterable and map style datasets. 

From my read through: 
1. The map-style `_iter_prefetch` path has the full N-1 overlap.... This is great!! 
2. But the iterable dataset,  `_iter_iterable`, doesn't have that lookahead. So it should show the same behavior as the unoptimized workflow...

Does it makes sense to split this into two PRs?

physicsnemo/datapipes/readers/zarr.py
```diff
@@ -290,12 +282,14 @@ def _select_random_sections_from_slice(
             slice_start,
             slice_stop - n_points + 1,
             (1,),
-            generator=self._subsample_generator,
+            generator=generator,
         ).item()
         return slice(start, start + n_points)
 
     def _load_sample(self, index: int) -> dict[str, torch.Tensor]:
```
negin513, line 289: Here, it seems like that the `_load_sample` runs on the IOPump's worker threads. This might end up in multiple workers having redundant `zarr.open()` calls for per path

physicsnemo/datapipes/readers/zarr.py
```diff
@@ -206,17 +205,6 @@ def fields(self) -> list[str]:
...
-        self._subsample_generator = generator
-
-    def set_epoch(self, epoch: int) -> None:
-        """Reseed the subsample RNG for a new epoch."""
-        if self._subsample_generator is not None:
-            self._subsample_generator.manual_seed(
-                self._subsample_generator.initial_seed() + epoch
-            )
-
     def _open_zarr_store(self, path: Path) -> Any:
```
negin513, line 208: zarr already guarantees safe concurrent reads from a single opened handle, so there's no need to re-open per call...
So we could open each group once eagerly  (the paths are all known there) and let the workers share the handle... That's trivially thread-safe, and lets the cache_stores flag and _cached_stores dict go away entirely (if I am not wrong). I think zarr won't dedupe open() calls for you.

coreyjadams: I think this is fine but would prefer to optimize the Zarr reader in a separate PR, if that's ok.  all the changes to the readers here are focused on updating reproducibility / randomness objects to maintain that feature in the FIFO queue for the IO.

physicsnemo/datapipes/readers/zarr.py
coreyjadams: It should be the case that nothing runs on IOPump's worker threads: the pump dispatches work only to the dataset's thread pool.

coreyjadams: This is the dispatch logic in `Dataloader.py`: the dispatch function sent to IOPump is a local function that uses a dataset submission.  So IOPump's thread calls this dispatch on it's thread, which immediately puts the real work in the Dataset's threadpool.

```python
        def dispatch(index: int) -> Any:
            stream = (
                self._streams[next(stream_counter) % self.num_streams]
                if use_streams
                else None
            )
            return self.dataset.submit(index, stream=stream)

        # Depth = prefetch_factor batches worth of samples kept in flight
        # (at least the stream count when streams drive the overlap).
        depth = max(self.prefetch_factor * self.batch_size, 1)
        if use_streams:
            depth = max(depth, self.num_streams)

        pump = IOPump(self._work_stream(), dispatch, depth=depth)
````

File: physicsnemo/datapipes/protocols.py [hunk 4]
```diff
@@ -33,94 +43,354 @@
...
+        with self._lock:
+            handle = self._prefetch_handles.pop(index, None)
+        if handle is not None:
+            return self.consume(handle)
         return self._load(index)
 
     def prefetch(
         self,
         index: int,
         stream: Optional[torch.cuda.Stream] = None,
```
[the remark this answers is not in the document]
coreyjadams: This is not how the dataloader works; nevertheless, a comment has been added to the docstring to warn about this.

coreyjadams:
/ok to test [255edce](https://github.com/NVIDIA/physicsnemo/pull/1742/commits/255edcee6cc88bce9b98e2099f59053933164b72)

examples/minimal/datapipes/tutorial_5_iterable_online_simulation.py
```diff
@@ -0,0 +1,196 @@
...
+        for position in range(self._num_batches):
+            # Per-(epoch, position) seeding: the stream is reproducible
+            # across runs and distinct across epochs and positions. There is
+            # no stable sample index for a generator, so we key on the
+            # monotonic emission position instead.
+            seed = np.random.SeedSequence(
+                [self._base_seed, self._epoch, position]
+            ).generate_state(1)[0]
+            np.random.seed(int(seed))
+            yield next(sim_iter)
```
negin513: claude flagged this for me for this example: 

Darcy2D reuses its output buffers:  it keeps`self.output_k` / `self.output_p` around and does `.data.copy_()` into them every step, so every batch this yields is the same two tensors. With `use_streams=True` that's exactly the race the dataloader docstring warns about... ?

physicsnemo/datapipes/protocols.py
```diff
@@ -33,94 +43,354 @@
...
+            Stream the consumer should use for this sample.
         """
-        if index in self._prefetch_futures:
-            return
-        executor = self._ensure_executor()
-        self._prefetch_futures[index] = executor.submit(self._load, index)
+        with self._lock:
+            if index in self._prefetch_handles:
+                return
+        handle = self.submit(index, stream)
```
negin513, line 409: Potential race condition since it is outside the lock?

negin513 (approved):
LGTM!! One thing before merge is the tutorial 5 inline comment with that example. 

Otherwise, look great. Does worth do a profile on it?

peterdsharpe:
Pushed `31b6e897` (as discussed with @coreyjadams): extends this PR's history-independent RNG scheme to the two sites it didn't cover — `transforms/base.py` and `transforms/mesh/base.py` still reseeded with `manual_seed(initial_seed() + epoch)`, which compounds across calls, so checkpoint resume at epoch N silently drew different augmentations than a continuous run. They now capture the base seed in `set_generator` and reseed via `_rng.derive_seed(base_seed, epoch)`, matching the readers' SeedSequence scheme. Also updates `RNG.md`/`DISTRIBUTIONS.md` (both documented the drifting idiom as the contract) and adds two sequential-vs-direct `set_epoch` equivalence tests. Full `test/datapipes/transforms/` suite passes (272 passed, 5 skipped). This supersedes #1817, which I'm closing.

coreyjadams:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1742/commits/b426ce1becdff08cf6a7ac3e551b5ef267115365

examples/minimal/datapipes/tutorial_5_iterable_online_simulation.py
coreyjadams: The example it totally pivoted off darcy to something much simpler, just to show the dataloader.

ktangsali (approved):

physicsnemo/datapipes/protocols.py
coreyjadams: It's mentioned in the doc string, and it's actually deliberate.  Submit is outside the lock so it doesn't block everything since it can be doing real work for a bit.  the index fetch and the handle write are both locked, as protection.  Under the designed conditions, it won't trigger, if someone calls `prefetch(i)` for the same `i` concurrently it can race and produce wasted work but won't crash.  I think it's ok to leave.

PR #1745, opened 2026-06-22
add model builder skill

mnabian:
# PhysicsNeMo Pull Request

## Checklist

## Review Process


CharlelieLrt:
@mnabian just curious: is it a 1-to-1 conversion of the older cursor rules into skills?
If so, does it keep all rules or does it updates on the rules? And is it worth deprecating the rules if they are replaced by these skills?

melo-gonzo (commented):
This looks good overall. You will not need to include `skill-card.md` or `BENCHMARK.md` - these will be generated by the `nvskills-ci` pipeline. Once you sync with the main branch we can invoke the CI pipeline and have a look at the reports generated and make any necessary changes. 

To @CharlelieLrt's point - i think its ok to leave the rules as is, and simply point to them in the skills like you do here. For other skills, i've takne the approach of allowing the agent to self discover the rules by only pointing to the main folder/files. Adding specific references as you do here (e.g, pointing to `MOD-00b` may make the skill more token efficient, but also if we do end up changing the location of these rules in the future we'll have to go back an update all of the skills.

mnabian:
> @mnabian just curious: is it a 1-to-1 conversion of the older cursor rules into skills? If so, does it keep all rules or does it updates on the rules? And is it worth deprecating the rules if they are replaced by these skills?

I agree with Carmelo. I think the rules can stay there for now and point the skills to them (or have them discover the rules).

skills/physicsnemo-model-builder/SKILL.md
```diff
@@ -0,0 +1,223 @@
...
+
+### 3. Reuse audit
+
+Before writing `forward`, enumerate existing primitives the contributor would
+otherwise reinvent — attention bases, embeddings, `Mlp`, neighbor ops
+(`knn`, `radius_search`), the TE-aware `LayerNorm`. Use the live search
+patterns in `references/reuse_map.md`; verify each path before citing it. Say
+explicitly "import `X` from `physicsnemo.nn` instead of writing your own." Keep
+genuinely novel, model-specific pieces local to the model.
+
```
pzharrington, line 125: @mnabian Something that can come up at this step is "layer X from `phyiscsnemo.nn` can help me but it needs a minor change/extension Y -- do I make that change to X or should I make a new module to support my needs?". For example this is an important design decision point in nearly every Earth-2 NVR model integration I've done, and unfortunately it can end up quite subjective on what the best path is.

Should we add some language here guiding that decision making (or redirect decision to user, possibly with some agent-generated proposals), or is this skill intended to be used strictly without making changes to the core library other than adding the new model?

mnabian: This is a really good comment! Adding a decision ladder at this step: default to reusing as-is or subclassing X locally; only extend X in place if the change is additive/back-compat and generally useful (and then as its own reviewed PR); new local module otherwise. Hard rule: never change a shared layer's existing behavior to fit one model. And for the genuinely subjective cases like your Earth-2 ones, the agent stops and hands you two concrete options with tradeoffs instead of deciding on its own. See commit b854abff7eb83296d160f6fc73ad62735a54b128

PR #1750, opened 2026-06-24
Add support for StereoRoPE positional embedding and Strata features

zyhu-hu:
# PhysicsNeMo Pull Request

## Description

Contributes the **Strata** weather-emulator architecture in two layers:

1. **`physicsnemo.nn`** — a **stereographic 2D rotary position embedding** primitive
   (the continuous-coordinate sibling of `physicsnemo.nn`'s `RotaryPositionEmbedding2D`).
2. **`physicsnemo.experimental.models.strata`** — the **`DiT3D`** and **`PixelDiT`**
   regression models that consume it.

> **On the "DiT" name.** "DiT" denotes the *architecture family*, not a training
> objective. `DiT3D` is a **conceptual 3D analog** of `physicsnemo.models.dit.DiT`:
> it follows the same DiT-style template (patch-embed → pre-norm transformer →
> linear decode) but is an **independent reimplementation**, *not* a reuse of that
> class's components — it defines its own 3D blocks and shares only the `na3d`
> functional, `Mlp`, and the RoPE primitives. `PixelDiT` additionally keeps DiT's
> *defining* **adaptive-layer-norm (adaLN)** conditioning — only the conditioning
> *signal* differs (the semantic stage's features vs. a diffusion timestep) — and
> its name also credits the PixelDiT paper (arXiv:2511.20645) it adapts.
>
> Both are **deterministic regression** models — *not* generative diffusion
> models: no diffusion process and no noise / timestep / class-label / text
> conditioning (`DiT3D` carries no adaLN at all; `PixelDiT`'s adaLN is driven by
> the semantic stage, not a timestep).

### Part 1 — StereoRoPE (`physicsnemo/nn/module/rope.py`, re-exported from `physicsnemo.nn`)

Token latitude/longitude are mapped to a local tangent plane via a stereographic
projection; the resulting continuous `(x, y)` coordinates drive a 2D RoPE that
**reuses `physicsnemo.nn`'s `apply_rotary_pos_emb` verbatim** (the per-pair
rotation math is identical — no new rotation code). New public API:

- `StereographicRotaryPositionEmbedding2D` — `project(lat, lon, length_scale, …)`
  maps lat/lon to continuous tangent-plane coords; `build_tables` / `forward(q, k, x_pos, y_pos)`
  rotate the query/key tensors. `length_scale` is a required argument (no sensible
  default can be inferred from data).
- `build_axial_rope_cos_sin_2d_continuous` — continuous-coordinate generalization of
  `build_axial_rope_cos_sin_2d` (the integer grid is its special case, asserted by a
  test), built on a reusable `build_rope_cos_sin_1d_continuous` core.
- `stereographic_projection` — antipode-guarded tangent-plane projection.
- `spherical_centroid` — pole- and seam-robust tile center (3D unit-vector mean),
  used for `project`'s default centering (correct at the poles, where a per-axis
  lat/lon mean is not).

**Placement:** a small, parameter-free primitive sharing the rotation convention
and channel layout of `physicsnemo.nn`'s existing RoPE, so per MOD-002a's maturity
exception it lives in production
`nn/module/rope.py` (MOD-000a reusable layer), not `experimental`. It is a plain
`torch.nn.Module`, not a `physicsnemo.Module` model.

### Part 2 — DiT3D / PixelDiT (`physicsnemo/experimental/models/strata/`)

- **`DiT3D`** — a 3D transformer backbone (its own 3D reimplementation of the
  DiT-style template, not a reuse of `physicsnemo.models.dit.DiT`'s components):
  3D patch embedding, 3D neighborhood attention (reusing `physicsnemo.nn.functional.na3d`)
  with optional depth-axis and gated attention, optional **axial** or **stereographic**
  RoPE, **activation checkpointing**, and bf16 autocast (CUDA). `set_tile_size` re-tiles
  a trained model for inference at a new resolution (rebuilding the cached axial tables;
  the stereographic / none modes build per forward, so only the expected shape updates) —
  supporting regional / tiled-global rollout.
- **`PixelDiT`** — two-stage: a `DiT3D` semantic stage (coarse patches) conditions a
  pixel-resolution stage whose blocks inject the semantic tokens via **pixel-wise
  adaptive layer norm** (`pixel_proj` / `bilinear_dw`). `bilinear_dw` supports any
  semantic vertical patch size (trilinear depth upsample, with an exact 2D-bilinear
  fallback when pixel depth == semantic depth, so the common case is numerically
  unchanged). Pixel-stage and semantic-stage RoPE are independent; pixel blocks also
  support activation checkpointing. Adapts the pixel-wise-AdaLN idea from PixelDiT
  (arXiv:2511.20645) — an adaptation, not a faithful port: deterministic regression
  (no diffusion/timestep/label conditioning), an independent AdaLN reimplementation,
  with the original `bilinear_dw` conditioning path added beyond the paper.

**Decoupled geometry:** lat/lon is an optional `forward` input used only for
stereographic RoPE — no `earth2grid` / grid-library dependency, no hard spherical-grid
assumption. The token-coordinate builders are pure free functions in `coords.py`
shared by both stages (no cross-stage instance coupling). Per MOD-002a, new models
are introduced in `experimental`.

### Tests

- `test/nn/module/test_rope.py` (17 new tests): continuous-builder ≡ integer-builder
  consistency, rotation **handedness** (a flipped `rotate_half` is caught), projection
  geometry + closed-form `2·tan(Δ/2)`, antipode finiteness, `spherical_centroid`
  pole/seam robustness, `project` centering + required `length_scale`, module
  shapes/validation/norm preservation, relative-position invariance, batched-coord
  broadcast, end-to-end lat/lon → `project` → `forward`, exact forward-vs-`build_tables`,
  `physicsnemo.Module` checkpoint round-trip, bf16 dtype preservation, `theta` wiring.
- `test/experimental/models/strata/`: constructor/attribute, validation, forward-shape
  across RoPE / AdaLN modes, non-regression goldens (`{args, state_dict, y}` reload —
  version-robust, CPU-reproducible on the `attn_kernel=-1` SDPA path), `physicsnemo.Module`
  checkpoint round-trip, the NATTEN NA3D path on CUDA, gradient-flow (no dead params;
  PixelDiT semantic stage reached through the pixel stage), `DepthwiseConv` (chunked ≡
  plain, deepcopy uses its own weights, `bias=False` device move), pure-`bfloat16`/`half`
  cast of both models (SDPA path on CPU/CUDA, plus the real NA3D + depth-axis + gated
  attention path on CUDA), `set_tile_size` re-tiling (axial buffer rebuild + non-axial shape
  update, with the new tile accepted and the old one rejected), shape variation (incl.
  vertical patch > 1), `torch.compile`, bf16 autocast, activation checkpointing (DiT3D +
  PixelDiT, train-mode), PixelDiT RoPE in both stages independently, and the `coords.py`
  builders (with value-pinned stereographic geometry).
- **About the CPU skips.** On a GPU runner the strata suite reports `74 passed,
  3 skipped`. The three skips are `test_dit3d_natten_forward[cpu]`,
  `test_pixeldit_natten_forward[cpu]`, and `test_dit3d_natten_bf16_cast[cpu]` — they
  `pytest.skip` *by design* because NATTEN's neighborhood-attention kernels are
  **CUDA-only** (no CPU kernel exists), not because of any missing coverage or
  environment gap. Their `[cuda]` counterparts run and pass on a GPU, and the CPU side
  of these models is exercised by the dense **SDPA fallback** (`attn_kernel=-1`), which
  the goldens use for deterministic CPU reproduction. CI without a GPU will skip the
  NA3D cases identically.
- `ruff`, `interrogate` (100%), license headers, and `import-linter` (10/0) all clean.

## Checklist

- [na] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.
author note: no existing issue is relevant to this PR

## Dependencies

None new. 3D neighborhood attention uses `physicsnemo.nn.functional.na3d` (NATTEN —
already an optional `physicsnemo` extra), with a CPU SDPA fallback (`attn_kernel=-1`).
Geometry is an optional `forward` input, so there is no `earth2grid` / grid-library
dependency. `import-linter`'s "Prevent Non-listed external imports" contract passes
unchanged.

## Review Process


CHANGELOG.md
```diff
@@ -10,6 +10,38 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  model: a `DiT3D` semantic stage (coarse patches) conditions a pixel-resolution
+  stage whose blocks inject the semantic tokens via pixel-wise adaptive layer
+  norm (`pixel_proj` or `bilinear_dw` modes). Adapts the pixel-wise-AdaLN idea
+  from PixelDiT (arXiv:2511.20645) — an adaptation, not a faithful port: a
+  deterministic regression model (no diffusion / timestep / label conditioning),
+  an independent reimplementation of the AdaLN, with the original `bilinear_dw`
+  conditioning path added beyond the paper.
+- The `physicsnemo.experimental.models.strata` package also publicly exports its
+  reusable building-block layers (`DiT3DBlock`, `Natten3DSelfAttention`,
+  `PatchEmbed3D`, `FinalLayer3D`, `PixelDiTBlock`, `PixelDiTLastLayer`).
```
pzharrington: Too much detail, condense all changes in this PR to a single bullet

physicsnemo/experimental/models/strata/dit3d.py
```diff
@@ -0,0 +1,630 @@
...
+            C=self.out_channels,
+            d=depth // pd,
+            h=height // ph,
+            w=width // pw,
+        )
+
+    def forward(
+        self,
+        x: Float[torch.Tensor, "batch in_channels depth height width"],
+        pos: Optional[Float[torch.Tensor, "batch two height width"]] = None,
```
pzharrington: "two" is not conventional jaxtyping naming, ust use `2` or `latlon=2` if you want semantic info too

physicsnemo/experimental/models/strata/transformer.py
```diff
@@ -0,0 +1,630 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+r"""DiT3D: a 3D transformer backbone for regression on spherical fields.
+
+DiT3D reuses the Diffusion-Transformer (DiT) *architecture* but is a
+**deterministic regression** model, not a generative diffusion model: it has no
+diffusion / denoising process and none of the original DiT's diffusion
+conditioning (no noise, timestep, class-label, or text conditioning). The "DiT"
+in the name refers to the architecture lineage only.
```
pzharrington, line 23: I think it's the right call to have this be a standalone architecture rather than try to force it under the existing `DiT` and related components. However, I don't quite agree with the naming of `DiT3D` and think that will confuse users who may be actually trying to build a 3D-capable `DiT` for diffusion.

I suggest choosing a different name for this core model to disambiguate, maybe something like `StrataTransformer3D`. Idk have your agents come up with some suggested names 😅  -- the fact that the architecture is similar or comes from the same "lineage" as the DiT family can just be mentioned once in the docstring or something. Same in general for `DiT3DBlock`.

~~The `PixelDiT` can probably stay as-is since it has more of a connection to existing named models in the literature via the use of AdaLN~~

physicsnemo/experimental/models/strata/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+    "DiT3D",
+    "DiT3DMetaData",
+    "PixelDiT",
+    "PixelDiTMetaData",
+    "Natten3DSelfAttention",
+    "DiT3DBlock",
+    "PatchEmbed3D",
+    "FinalLayer3D",
+    "PixelDiTBlock",
+    "PixelDiTLastLayer",
```
pzharrington: I don't think we need to export all subcomponents publicly via `__init__`, probably just the main model and maybe the transformer backbone

physicsnemo/experimental/models/strata/transformer.py
pzharrington: Actually now that I look at it more, it looks like the (currently named) `PixelDiT` is the full model one would instantiate and call in training/inference. So probably clearest to have something like:

- `Strata` (instead of `PixelDiT`) — the composed model you use for train/inference. Name directly connects to the arxiv, which describes the full architecture. Docstring can describe the architecture as well and mention connections to `DiT`/`PixelDiT`
- `StrataTransformer3D`(instead of `DiT3D`) — the internal coarser trunk /  backbone network...can also be used directly as a model if desired.
- `PixelDiTBlock` — used inside `Strata` where the borrowed PixelDiT mechanism literally lives, carrying the connection to that paper.

pzharrington: Thoughts?

physicsnemo/experimental/models/strata/pixel.py
```diff
@@ -0,0 +1,872 @@
...
+        x = x + gate1 * z
+
+        # MLP sub-layer with modulated pre-norm and gated residual.
+        y = self.norm2(x) * (1 + scale2) + shift2
+        z = self.mlp(y)
+        x = x + gate2 * z
+        return x
+
+
+class PixelDiTLastLayer(nn.Module):
```
pzharrington: This looks like basically the same as `FinalLayer3D` -- both are fp32 `LayerNorm` + `Linear`, differing only in the output dim (`pd*ph*pw*out_chans` vs `out_chans`). Let's consolidate

physicsnemo/experimental/models/strata/transformer.py
```diff
@@ -0,0 +1,630 @@
...
+        if not self.training:
+            return False
+        ratio = self._activation_checkpointing_ratio
+        if ratio <= 0.0:
+            return False
+        if ratio >= 1.0:
+            return True
+        return block_idx < round(ratio * len(self.blocks))
+
+    def initialize_weights(self) -> None:
```
pzharrington, line 389: This looks fine, but `PixelDiT` has no `initialize_weights()` override: its `pixel_patch_embed` conv keeps PyTorch default init (not the xavier-fan `DiT3D` uses), and `pixel_final_layer` (`PixelDiTLastLayer`) is never zero-initialized. Was that intentional?

physicsnemo/experimental/models/strata/pixel.py
```diff
@@ -0,0 +1,872 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+r"""PixelDiT: a two-stage 3D transformer regression model with pixel-wise conditioning.
+
+Stage 1 (semantic) is a :class:`~physicsnemo.experimental.models.strata.DiT3D`
```
pzharrington: nit: personally I find the use of "semantic" throughout this file kind of confusing, since that term is more relevant to computer vision and related tasks and not as meaningful for physics. If I understand your architecture correctly, the "semantic" tokens are really just latent tokens predicted on a coarser grid by the backbone. Any ideas on alternate names? I see it's used a lot so not the worst if we decide to keep as-is

physicsnemo/experimental/models/strata/pixel.py
```diff
@@ -0,0 +1,872 @@
...
+    ):
+        super().__init__(meta=PixelDiTMetaData())
+
+        ### Stage 1: semantic DiT3D, built from the provided config.
+        semantic_config = dict(semantic_config or {})
+        self.semantic = DiT3D(**semantic_config)
+        # The semantic output head is never used here (only forward_tokens is);
+        # drop it so it does not create unused parameters under DDP. Removal is
+        # symmetric across save / load because reconstruction rebuilds then drops.
+        del self.semantic.final_layer
```
pzharrington: I suggest an alternate pattern:

- Create a `include_head` kwarg in the `DiT3D` (or whatever we call it) constructor that will enable/disable the creation of `final_layer`
- Then that component will return the features when headless, or the full field when including the `final_layer`. Docstrings and jaxtyping shape hints for the forward method jusut need to be updated to reflect the behavior

Then we can avoid this nonstandard usage of `del`

physicsnemo/nn/module/rope.py
pzharrington, line 1: I thought about the changes here a bit deeper and have the following recommendation:

1. Drop the `StereographicRotaryPositionEmbedding2D` entirely for now -- it is not used really as a module, it is just used purely as a table builder; keep the four new functional helpers: `stereographic_projection`, `spherical_centroid`, `build_rope_cos_sin_1d_continuous`, `build_axial_rope_cos_sin_2d_continuous`
2. Rewire the Strata models to call `build_axial_rope_cos_sin_2d_continuous` instead of doing `self.rope.build_tables(...)`
3. Move all new additions to `experimental/nn/module/rope.py`. That can import from the public, non-experimental `nn/module/rope.py` as needed, but the new items related to continuous and stereographic RoPE can come in through experimental which gives us some more leeway if we need to modify how things are implemented

physicsnemo/experimental/models/strata/transformer.py
zyhu-hu: `Strata/StrataTransformer3D` sounds good to me! for PIxelDiTBlock, maybe the "DiT" part of the string can still be confusing since this block is also only for regression model instead of diffusion model. what about `StrataPixel3DBlock` but mention it is inspired by PixelDiT and cite reference there? and maybe for the `DiTBlock`, we can change to `StrataTransformer3DBlock`?

pzharrington: That sounds good

physicsnemo/experimental/models/strata/transformer.py
```diff
@@ -0,0 +1,630 @@
...
+        h = height // self.patch_size[1]
+        w = width // self.patch_size[2]
+        coords = build_axial_token_coords(d, h, w)
+        cos, sin = self.rope.build_tables(coords[:, 0], coords[:, 1])
+        self.register_buffer(
+            "_rope_cos", cos.to(self._rope_cos.device), persistent=False
+        )
+        self.register_buffer(
+            "_rope_sin", sin.to(self._rope_sin.device), persistent=False
+        )
```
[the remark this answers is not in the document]
zyhu-hu: Good catch — fixed. The constructor (via PatchEmbed3D) rejects a non-divisible input shape, but set_tile_size didn't apply the same check, so a non-divisible tile was silently floor-truncated by the patch-embed conv. Added the matching height/width divisibility check (same messages as PatchEmbed3D) at the top of set_tile_size so re-tiling fails fast for every rope_mode, plus a regression test.

physicsnemo/experimental/models/strata/depthwise_conv.py
```diff
@@ -0,0 +1,197 @@
...
+
+        # Standard (non-chunked) path: warn if a single conv2d call would exceed
+        # the element-count limit, then defer to nn.Conv2d.
+        size_of_chunk = math.ceil(x.numel() * x.dtype.itemsize)
+        if size_of_chunk > 2**32:
+            warnings.warn(
+                f"Convolution {size_of_chunk=} larger than 2^32 so conv2d will "
+                f"revert to a slow implementation or error out. Set the "
+                f"chunk_size option to enable chunking.",
+                stacklevel=2,
```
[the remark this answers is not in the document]
zyhu-hu: Fixed — the limit is element-count, not bytes: conv2d's 32-bit indexing guard raises when `numel > INT_MAX` (2³¹−1). Verified fp32 and bf16 fail at the same numel (despite 2× the bytes). Switched to `x.numel() > torch.iinfo(torch.int32).max`, corrected the wording (RuntimeError, not slow fallback), and pointed the remedy at chunk_size. (Minor: the 1.5e9-element fp32 example would actually warn — 6e9 bytes > 2³² — but the byte-vs-element point was right.)

CHANGELOG.md
zyhu-hu: Done. condensed changes into a single bullet

physicsnemo/experimental/models/strata/dit3d.py
zyhu-hu: replaced "two" with 2 now

physicsnemo/experimental/models/strata/transformer.py
zyhu-hu: renamed as following:
- `PixelDiT` → `Strata` (the composed train/inference model)
- `DiT3D` → `StrataTransformer3D` (backbone) + `StrataTransformer3DBlock`
- `PixelDiT`/`DiT3D` metadata renamed to match; module files (`dit3d.py`→`transformer.py`, `pixel.py`→`strata.py`) and golden fixtures renamed too.
- `PixelDiTBlock` → `StrataPixel3DBlock`

physicsnemo/experimental/models/strata/__init__.py
zyhu-hu: Done — `__all__` is now just `Strata`, `StrataMetaData`, `StrataTransformer3D`, `StrataTransformer3DMetaData`. The building-block layers stay importable from their submodules but are off the package's public surface.

physicsnemo/experimental/models/strata/pixel.py
zyhu-hu: Done — merged both into one `FinalLayer3D(hidden_size, out_features)` (backbone passes `p_d·p_h·p_w·C_out`, pixel stage passes `C_out`); submodule names unchanged so checkpoints are key-preserving.

physicsnemo/experimental/models/strata/transformer.py
zyhu-hu: Good catch — it was an oversight, now fixed. Added a targeted `initialize_weights()` that xavier-inits the pixel patch-embed and zero-inits the pixel output head (matching the backbone), without touching the AdaLN-zero projections.

physicsnemo/experimental/models/strata/pixel.py
zyhu-hu: Agreed — renamed the semantic* identifiers to backbone (backbone_config, self.backbone, backbone_cond, …), which reads as "tokens from the backbone stage."

physicsnemo/experimental/models/strata/pixel.py
zyhu-hu: Done — added `include_head` to the backbone; the composed model builds it with `include_head=False` (no output head built) and consumes `forward_tokens`, so the `del self.semantic.final_layer` is gone and no unused params are created.

physicsnemo/nn/module/rope.py
zyhu-hu: Done, all three points: deleted `StereographicRotaryPositionEmbedding2D`; kept the four helper functions and moved them to `physicsnemo/experimental/nn/rope.py`; rewired both models to call `build_axial_rope_cos_sin_2d_continuous` directly (removed from the production `physicsnemo.nn` exports). Verified the rewire is numerically identical.

CHANGELOG.md
```diff
@@ -10,6 +10,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  deterministic regression model (not a generative diffusion model) whose
+  `StrataTransformer3D` backbone — 3D neighborhood attention via
+  `physicsnemo.nn.functional.na3d`, optional axial or stereographic RoPE, optional
+  depth-axis / gated attention — conditions a pixel-resolution stage through
+  pixel-wise adaptive layer norm (adapting the PixelDiT idea, arXiv:2511.20645).
+  Geometry (latitude/longitude) is an optional forward input, so there is no hard
+  spherical-grid dependency. Also adds the supporting continuous-coordinate and
+  stereographic RoPE helpers to `physicsnemo.experimental.nn`
+  (`build_rope_cos_sin_1d_continuous`, `build_axial_rope_cos_sin_2d_continuous`,
+  `stereographic_projection`, `spherical_centroid`).
```
pzharrington: Still overly verbose, these details can be put in documentation rather than changelog. Just listing the added public-level modules and functions is fine

physicsnemo/experimental/models/strata/depthwise_conv.py
pzharrington, line 1: No need for a separate file, might as well put this in `layers.py`

test/nn/module/test_rope.py
```diff
@@ -209,11 +235,12 @@ def dot(i, j):
...
-# .save() / Module.from_checkpoint() recipe. Their cos/sin tables are
-# persistent=False buffers, deterministically rebuilt from the __init__ args, so
-# a round-trip must reproduce the forward exactly without the tables appearing in
-# the checkpoint.
+# All three RoPE modules subclass physicsnemo.core.Module, so they must support
+# the .save() / Module.from_checkpoint() recipe. The axial 1D/2D modules cache
+# cos/sin as persistent=False buffers (deterministically rebuilt from the
+# __init__ args); the stereographic module is fully stateless (tables are built
+# per forward from the input coordinates). In every case a round-trip reproduces
+# the forward exactly without any tables appearing in the checkpoint.
```
pzharrington, line 243: Stale references to stereographic module should be removed, and there are only two RoPE modules now without it.

physicsnemo/experimental/models/strata/transformer.py
```diff
@@ -0,0 +1,669 @@
...
+        pos: Optional[Float[torch.Tensor, "batch 2 height width"]] = None,
+    ) -> torch.Tensor:
+        r"""Run the model, returning a decoded field (or features when headless).
+
+        With ``include_head=True`` (default) returns the decoded field of shape
+        :math:`(B, C_{out}, D, H, W)`. With ``include_head=False`` there is no
+        output head, so the post-block tokens of shape :math:`(B, N, E)` are
+        returned instead (equivalently use :meth:`forward_tokens`).
+        """
+        _, _, d_in, h_in, w_in = x.shape
```
pzharrington: This happens before input validation and could lead to "too many values to unpack" error. Since `forward_tokens` guarantees the shape equals the configured one, you could just use `self.depth/height/width` for `unpatchify`, and drop the early unpack here.

physicsnemo/experimental/models/strata/strata.py
```diff
@@ -0,0 +1,867 @@
...
+    ) -> Float[torch.Tensor, "batch out_channels depth height width"]:
+        ### Input validation
+        if not torch.compiler.is_compiling():
+            if x.ndim != 5:
+                raise ValueError(
+                    f"Expected 5D input (B, C, D, H, W), got tensor of shape "
+                    f"{tuple(x.shape)}"
+                )
+            if self.rope_mode_pixel == "stereographic" and pos is None:
+                raise ValueError(
```
pzharrington, line 479: This only checks for `pos is None` but not shape, and the backbone validates `pos.shape` only when the backbone's `rope_mode="stereographic"`. So the `backbone="none"/"axial", pixel="stereographic"` config never shape-checks pos; a wrong `(B,2,H,W)` falls into a cryptic rearrange error in `build_stereographic_token_coords`. Worth a one-line shape check in `Strata.forward`.

physicsnemo/experimental/models/strata/strata.py
```diff
@@ -0,0 +1,872 @@
...
+
+                else:
+
+                    def _run(inp, b=block):
+                        return b(inp, latent_dhw=pixel_dhw, rope_tables=rope_tables)
+
+                if self._should_checkpoint_pixel_block(i):
+                    x_pix = activation_checkpoint(_run, x_pix, use_reentrant=False)
+                else:
+                    x_pix = _run(x_pix)
```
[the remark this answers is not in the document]
pzharrington: @zyhu-hu I think this is reasonable ask, for code clarity

physicsnemo/experimental/models/strata/transformer.py
```diff
@@ -0,0 +1,669 @@
...
+            d_patch=d,
+            length_scale=self.rope_length_scale,
+        )  # (B, N, 2)
+        cos, sin = build_axial_rope_cos_sin_2d_continuous(
+            coords[..., 0], coords[..., 1], self.head_dim, theta=self.rope_base
+        )  # (B, N, head_dim)
+        # Insert a heads axis so the tables broadcast over (B, heads, N, head_dim).
+        return cos.unsqueeze(1), sin.unsqueeze(1)
+
+    def prepare_tokens(
```
pzharrington: Use jaxtyping or make private to the class

physicsnemo/experimental/models/strata/transformer.py
```diff
@@ -0,0 +1,669 @@
...
+
+                    def _run(inp, b=block, r=block_rope):
+                        return b(inp, latent_dhw=latent_dhw, rope_tables=r)
+
+                    x = activation_checkpoint(_run, x, use_reentrant=False)
+                else:
+                    x = block(x, latent_dhw=latent_dhw, rope_tables=block_rope)
+        return x, latent_dhw
+
+    def unpatchify(
```
pzharrington: Use jaxtyping or make private to the class

physicsnemo/experimental/models/strata/strata.py
```diff
@@ -0,0 +1,867 @@
...
+            sh=sh,
+            sw=sw,
+            pv=pv,
+            ph=ph,
+            pw=pw,
+        ).reshape(adaln_raw.shape[0], d * h * w, -1)
+
+    @staticmethod
+    def precompute_bilinear_cond(
+        backbone_cond: torch.Tensor,
```
pzharrington: Use jaxtyping annotations

physicsnemo/experimental/models/strata/strata.py
```diff
@@ -0,0 +1,867 @@
...
+        -------
+        None
+            Modifies parameters in place.
+        """
+        w = self.pixel_patch_embed.proj.weight.data
+        nn.init.xavier_uniform_(w.view([w.shape[0], -1]))
+        nn.init.constant_(self.pixel_patch_embed.proj.bias, 0.0)
+
+        nn.init.constant_(self.pixel_final_layer.linear.weight, 0.0)
+        nn.init.constant_(self.pixel_final_layer.linear.bias, 0.0)
```
pzharrington, line 406: Nothing here ever initializes the pixel blocks' attention and MLP `Linear`s (they keep PyTorch's `nn.Linear` default), intended or no? It sorta contradicts the docstring claim of matching `StrataTransformer3D.initialize_weights`

physicsnemo/experimental/models/strata/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+"""
+
+from .transformer import StrataTransformer3D, StrataTransformer3DMetaData
+from .strata import Strata, StrataMetaData
+
+__all__ = [
+    "Strata",
+    "StrataMetaData",
+    "StrataTransformer3D",
+    "StrataTransformer3DMetaData",
```
pzharrington: No need to export the `MetaData` classes

physicsnemo/experimental/models/strata/strata.py
```diff
@@ -0,0 +1,867 @@
...
+from .layers import (
+    FinalLayer3D,
+    Natten3DSelfAttention,
+    PatchEmbed3D,
+    RopeTables,
+    StrataTransformer3DBlock,
+)
+from .transformer import StrataTransformer3D
+
+__all__ = ["Strata", "StrataMetaData"]
```
pzharrington: Same here and in `transformer.py`, no need to export the `MetaData`, it's just a module-local helper

physicsnemo/experimental/models/strata/strata.py
```diff
@@ -0,0 +1,867 @@
...
+    # Data type
+    bf16: bool = True
+    # Inference
+    onnx: bool = False
+    # Physics informed
+    func_torch: bool = False
+    auto_grad: bool = False
+
+
+class StrataPixel3DBlock(nn.Module):
```
pzharrington: Since it's a block, maybe move it to `layers.py`?

CHANGELOG.md
zyhu-hu: Done — trimmed to a bare list of the added public symbols (Strata, StrataTransformer3D, and the four experimental RoPE helpers); the architecture prose now lives in the docstrings/docs instead.

physicsnemo/experimental/models/strata/depthwise_conv.py
zyhu-hu: Done — folded `DepthwiseConv` (and its `_apply_conv2d` / `_build_chunked_depthwise_conv` helpers) into `layers.py` and deleted the file.

test/nn/module/test_rope.py
zyhu-hu: Fixed — updated to "Both RoPE modules" and dropped the stereographic-module wording (there are only the 1D/2D modules now).

physicsnemo/experimental/models/strata/transformer.py
zyhu-hu: Fixed. `forward_tokens` already validates the input shape, so I now decode with the configured `(depth, height, width)` instead of unpacking `x` before validation — a wrong-ndim input now yields the clean `ValueError` from `forward_tokens` rather than an unpack error. Added a regression test feeding a 4D tensor to both models.

physicsnemo/experimental/models/strata/strata.py
zyhu-hu: Added an explicit shape check for `pos`. Now it raises a clear `ValueError` if `pos` shape check fails, with a regression test.

physicsnemo/experimental/models/strata/transformer.py
zyhu-hu: renamed to `_prepare_tokens` (and updated the one caller)

physicsnemo/experimental/models/strata/transformer.py
zyhu-hu: renamed to `_unpatchify`; updated the caller and the `:meth:` reference in `set_tile_size`'s docstring.

physicsnemo/experimental/models/strata/strata.py
zyhu-hu: Added jaxtyping annotations: input `Float[torch.Tensor, "batch backbone_tokens cond_dim"]` → `Float[torch.Tensor, "batch depth cond_dim height width"]`.

physicsnemo/experimental/models/strata/__init__.py
zyhu-hu: Dropped both from the package `__init__` imports here.

physicsnemo/experimental/models/strata/strata.py
zyhu-hu: Removed `StrataMetaData` and `StrataTransformer3DMetaData` from the module-level `__all__` in `strata.py / transformer.py`, so they're no longer exported anywhere; the classes stay defined in-place for that internal use.

physicsnemo/experimental/models/strata/strata.py
zyhu-hu: moved now.

physicsnemo/experimental/models/strata/strata.py
zyhu-hu: thanks, that was an oversight — the blocks were left at PyTorch's default init while the docstring claimed parity with the backbone. The original code skipped the blanket Xavier pass to avoid clobbering the AdaLN-zero projections, but that over-corrected. Fixed properly: I now apply the same `_basic_init` Xavier pass over `pixel_blocks`, then re-zero the AdaLN projections via a new `StrataPixel3DBlock.reset_adaln_zero()` so each conditioning block still starts as an identity residual (output is exactly zero at init). The depthwise smoothing conv keeps its identity init (it's a `Conv2d`, untouched by the `nn.Linear` pass). The non-regression goldens are unaffected (that test loads a committed `state_dict`, so it doesn't depend on `initialize_weights`), and there's a regression test asserting both the Xavier init and the zero-output invariant.

zyhu-hu:
Thanks for the thorough second pass @pzharrington! All comments addressed in focused commits. Two follow-ups beyond the comments: (1) an internal self-review caught a couple of doc/test issues I also fixed — a self-referential attribution in the `StrataPixel3DBlock` docstring (now correctly credits PixelDiT) and a weak assertion in the init test (now actually distinguishes the Xavier init). (2) The `test_non_regression[strata_bilinear*]` CI failures were a `conv2d` float-tolerance issue, not a regression — the `bilinear_dw` goldens are bit-exact on the generating CPU, but `conv2d` diverges across backends (~1e-3 on a differing CPU build, up to ~5e-3 on GPU); I relaxed the golden tolerance to `2e-3` (the tightest that's robust across CPU/GPU).

pzharrington (approved):

PR #1753, opened 2026-06-25
Unified Aero Recipe Tidying

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

This PR does some cleanup on the Unified External Aero Recipe. Specifically:

- In `README.md`, deduplicates some sections, removes the stale warning header, and links in DoMINO.
- In `metrics.py`:
  - Removes the unused `process_group` in `MetricCalculator`.
  - Removes unused `MetricCalculator.total_channels`
  - Removes the unused `field_dim` import
- In `train.py`:
  - Removes a duplicated `omegaconf` import
- In `nondim.py`:
  - Removes the unused `inverse_tensor()`, which is both a) not used, and b) already superseded by the already-existing `inverse()` and `inverse_td()`.

## Checklist

- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.
- [ ] If I am implementing a new model or modifying any existing model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation).

## Review Process

**All PRs are reviewed by the PhysicsNeMo team before merging.**

Depending on which files are changed, GitHub may automatically assign a maintainer for review.

We are also testing AI-based code review tools (e.g., Greptile), which may add automated comments with a confidence score.
This score reflects the AI’s assessment of merge readiness and is **not** a qualitative judgment of your work, nor is
it an indication that the PR will be accepted / rejected.

AI-generated feedback should be reviewed critically for usefulness.
You are not required to respond to every AI comment, but they are intended to help both authors and reviewers.
Please react to Greptile comments with 👍 or 👎 to provide feedback on their accuracy.

PR #1754, opened 2026-06-25
Unified Aero Recipe: Fix for all_reduce

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Stack on top of #1753

`MetricCalculator` tried to do its own cross-rank reduction with `_all_reduce` method, but this was gated by `process_group` which no caller ever provided. So, it was a no-op.

Meanwhile `_run_epoch` in `train.py` computed metrics locally, also without all-reduce. So, under multi-GPU DDP, the logged train/val loss and every metric reflected only rank 0's shard of the data, not the global average.

No impact on training dynamics, but has a material impact on reporting accuracy. `infer.py` already did this correctly via `_allreduce_sums()`, but `train.py` never got fixed - this PR does that.

------

Summary:

So the core of the fix is in train.py, where:

- Before, the epoch summary that gets printed didn't actually do a reduction:
  - We had (line 404 on main): avg_loss = total_loss / n, 
  - The problem is that this total_loss is a sum over batches, but never gets summed across ranks (i.e., dist.all_reduce).
  - And n here is just the local number of batches.
- So this statistic is purely based on 1 rank.
- There was machinery in metrics.py to do _all_reduce(), but this only did the proper reduction for metrics, not for the reported loss.

After:
- We add _reduce_and_average_epoch(), which actually does the all-reduce on losses. And, it does the metrics reduction there too - so we have 1 place that this happens rather than 2 (where 1 of the two, train.py, was broken).
- metrics.py is now then purely local - any distributed stuff happens in train.py. And it does this reduction (and hence, comms) once at end-of-epoch rather than on every step.
- So, now, you'll get correctly-averaged loss printouts in multi-GPU settings.

## Checklist

## Review Process


peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1754/commits/b34736b99f49bbcd5be88fbfcc6f491e7ef4aebe

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1754/commits/d028742ad18ce1bed7faae8c3baeeb72c53260f7

examples/cfd/external_aerodynamics/unified_external_aero_recipe/src/utils.py
```diff
@@ -46,6 +46,26 @@
 ### its error message for the padding rationale).
 Precision: TypeAlias = Literal["float32", "float16", "bfloat16"]
 
+### Canonical ``phase`` tags for each ``metrics.jsonl`` record, shared by
```
coreyjadams, line 49: I have been pushing claude to stop slop-bombing the comments.  At some point the comments are getting unreadable.

I don't know where it trained that wrote comments like this, but this is almost worse than no comments haha.

coreyjadams (approved):
OK, I'll approve this now.  I feel like we've let claude get sloppy on comments here, but we don't have to fix it all today.

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1754/commits/1c1e8643d87f4f233aebe63bf981920431a334a7

PR #1757, opened 2026-06-25
Tensor-valued parameters in diffusion DPS guidance

CharlelieLrt:
# PhysicsNeMo Pull Request

## Description

Diffusion DPS guidance can now use a **non-uniform guidance strength**: the measurement-noise level `std_y` and the SDA scaling `gamma` of `ModelConsistencyDPSGuidance`, `DataConsistencyDPSGuidance`, and their `physicsnemo.diffusion.multi_diffusion` counterparts accept **tensors** in addition to floats.

**What's new:**

- A tensor `std_y` / `gamma` assigns a different strength to each observation component, either **per-channel** (e.g. `(1, C, 1, 1)`) or fully **pointwise** (the full observation shape, e.g. a distinct noise std at each probe).
- Passing floats keeps the previous uniform behavior, so existing code is unaffected.
- For the multi-diffusion variants the tensors are given at the **global resolution** and patched internally to line up with the patched observations.

```python
# per-channel measurement-noise std (per-pixel tensors also work)
std_y = torch.tensor([0.05, 0.1, 0.2]).reshape(1, 3, 1, 1)
guidance = DataConsistencyDPSGuidance(mask=mask, y=y_obs, std_y=std_y)
```

### ⚠️ Breaking change (custom `norm` only)

To make the per-component weighting possible, a **custom `norm` callback must now return an elementwise (un-reduced) residual** with the same shape as its inputs; the guidance applies the reduction and the per-component strength internally. The built-in integer `norm` (e.g. `norm=2`) is **unaffected**.

Migrating a custom callback is a one-line edit, just drop the reduction:

```python
# before: returned a per-sample scalar of shape (B,)
def norm(y_pred, y_true):
    return (y_pred - y_true).pow(2).flatten(1).sum(dim=1)

# after: return the elementwise residual (same shape as the inputs)
def norm(y_pred, y_true):
    return (y_pred - y_true).pow(2)
```

For `DataConsistencyDPSGuidance` (and its multi-diffusion counterpart) there is a second, related change: the `norm` callback now receives the **unmasked** `(x_0, y)` and the mask is applied to its output (`mask * norm(x_0, y)`), instead of the old pre-masked inputs `(mask * x_0, mask * y)`. This is transparent for the built-in `Lp` norms, but a custom `norm` that assumed unobserved entries were already zeroed (e.g. counting observed pixels, or a loss with a non-zero value at zero) may now behave differently.

CHANGELOG and docstrings are updated; CI tests for the new behavior will be added in a follow-up.

## Checklist

## Review Process


CHANGELOG.md
```diff
@@ -137,6 +145,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
   `physicsnemo.diffusion.multi_diffusion`.
+- &#9888;&#65039; **BC-impact (DPS guidance):** a custom `norm` callback passed to
+  `physicsnemo.diffusion.guidance.ModelConsistencyDPSGuidance` /
+  `DataConsistencyDPSGuidance` (and their `physicsnemo.diffusion.multi_diffusion`
+  counterparts) must now return an **elementwise** loss (same shape as its
+  inputs) instead of a per-batch-element reduced scalar of shape `(B,)`.
+  Migration: drop the reduction from your `norm` — e.g. return
+  `(y_pred - y_true).abs().pow(2)` rather than
+  `(y_pred - y_true).pow(2).reshape(B, -1).sum(dim=1)`. The integer `norm`
+  selector (e.g. `norm=2`) is unaffected.
```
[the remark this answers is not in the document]
CharlelieLrt: Good catch, thanks. Clarified in `e94852bc0`: the CHANGELOG BC note now states that for `DataConsistencyDPSGuidance` (and its `multi_diffusion` counterpart) the `norm` callback receives the unmasked `(x_0, y)` with the mask applied to its output (`mask * norm(x_0, y)`), and that a custom `norm` relying on pre-zeroed unobserved entries may differ. Added the same note to the PR description.

physicsnemo/diffusion/guidance/dps_guidance.py
```diff
@@ -729,25 +760,26 @@ def __call__(
...
 
         with torch.enable_grad():
             y_pred = self.observation_operator(x_0)
             loss = self._loss_fn(y_pred, y)
+            # Guidance strength rho(t), broadcast over the observation.
+            bc_shape = (-1,) + (1,) * (loss.ndim - 1)
+            t_bc = t.reshape(bc_shape)  # (B, 1, ..., 1)
+            sigma_t = self.sigma_fn(t_bc)
+            alpha_t = self.alpha_fn(t_bc)
+            rho = 1.0 / (2.0 * (std_y**2 + gamma * (sigma_t**2) / (alpha_t**2)))
```
jleinonen, line 774: Just to understand - I guess this is the equivalent of the old `variance` calculation, just pre-differentiation?

jleinonen (approved):
Thanks, looks good to me!

physicsnemo/diffusion/guidance/dps_guidance.py
CharlelieLrt: Yes, it is equivalent to the old variance, even when `gamma>0` or when the `loss` is not the L2 norm. The only difference is that now `loss` is element-wise (it does not make any reduction).

The other option I had was to move the variance within the `loss`, but this would lead to non-equivalence (and even possibly quite weird) behaviors when `loss` is not the L2 norm. So I preferred avoiding it...

CharlelieLrt:
/ok to test [1e6522b](https://github.com/NVIDIA/physicsnemo/pull/1757/commits/1e6522b8b48be2fd9bda9968c5014bab25786af7)

PR #1758, opened 2026-06-26
HealDA v2 Architecture

aayushg55:
# PhysicsNeMo Pull Request

## Checklist

Resolves https://github.com/NVIDIA/physicsnemo-roadmap/issues/2707
## Review Process


aayushg55:
@pzharrington Let me know what you think about the VideoDiTBlock/VideoDiT classes. I can restructure the code accordingly.

In particular, would something like this work better?
```
class VideoDiTBlock(DiTBlock):
    def __init__(self, ..., temporal_attention=False, cross_attention=None, is_causal=False, ...):
        super().__init__(hidden_size, num_heads, attention_backend=..., condition_embed_dim=..., mlp_ratio=..., drop_path=..., **attn_kwargs)
        self._is_causal = is_causal
        if temporal_attention: self.temporal_attention = ...; self.temporal_attn_norm = ...
        if cross_attention:    self.cross_attention = cross_attention(); self.cross_attn_norm = ...
    def forward(self, hidden_states, c, cross_attention_context=None, attn_kwargs=None):
        # reuse self.adaptive_modulation(c).chunk(6), self.pre_attention_norm/self.pre_mlp_norm,
        # self.attention, self.linear, self.drop_path; insert obs + temporal sub-layers between.
```

physicsnemo/experimental/models/healda/adaln.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def _broadcast(param: Float[torch.Tensor, "batch channels"], ndim: int) -> torch.Tensor:
+    # (B, C) -> (B, 1, ..., 1, C) so a per-sample modulation broadcasts over a
+    # hidden-state tensor of arbitrary rank (3D (B, L, C), 4D (B, T, X, C), ...).
+    shape = (param.shape[0],) + (1,) * (ndim - 2) + (param.shape[1],)
+    return param.view(shape)
+
+
+class AdaLayerNormZero(nn.Module):
```
aayushg55: The video block has four gated sub-layers (spatial attention, MLP, temporal attention, obs cross-attention). So rather than repeat DiTBlock's inline adaLN pattern — adaptive_modulation = Sequential(SiLU, Linear(cond, 6*hidden)) + separate pre_attention_norm/pre_mlp_norm + a modulation helper — four times, I felt it would be cleaner to wrap the whole adaLN-Zero operation (SiLU + Linear → shift/scale/gate, the affine-free LayerNorm, and applying them) into a single reusable AdaLayerNormZero module.

physicsnemo/experimental/models/healda/sharding.py
```diff
@@ -0,0 +1,138 @@
...
+import torch
+import torch.distributed as dist
+from jaxtyping import Float
+from torch.distributed.nn.functional import all_to_all_single
+
+DATA_DIM = 0
+MODEL_DIM = 1
+
+
+def shard_x(
```
aayushg55, line 45: ## Context parallelism: manual all-to-all vs `ShardTensor.redistribute`

Factorized temporal attention needs a time↔space reshard each block (2 per block:
enter t-sharded → spatial attn → reshard to x-sharded → temporal attn → reshard
back). We implemented it both ways: a manual `all_to_all_single`, and
`ShardTensor.redistribute` (`Shard(time)` ↔ `Shard(space)` on a 1D mesh).

Per-block — 8×H100, dit-5B (C=1536 (`16×96`), X=12288, T=8, bf16 autocast, 2 reshards/block:

| path | eager ms/block | compiled ms/block |
|---|---|---|
| manual all-to-all | 16.5 | **14.5** |
| `ShardTensor.redistribute` | 20.5 | 19.3 |

Just the reshard primitives (t↔x round trip = 2 reshards, same shape, eager):

| path | fwd | fwd+bwd |
|---|---|---|
| manual all-to-all | 0.29 ms | 0.97 ms |
| `ShardTensor.redistribute` | 2.48 ms | 5.45 ms |

aayushg55: <details>
<summary><b>Benchmark code — reshard primitives</b> (<code>reshard_microbench.py</code>)</summary>

Run on 8×H100, dit-5B shape:
`T_FULL=8 X_FULL=12288 EMBED_DIM=1536 torchrun --nproc_per_node=8 reshard_microbench.py`

```python
import os
import time

import torch
import torch.distributed as dist
from torch.distributed.device_mesh import init_device_mesh
from torch.distributed.tensor.placement_types import Shard

from physicsnemo.domain_parallel import ShardTensor
from physicsnemo.experimental.models.healda.sharding import (
    shard_t,
    shard_t_shardtensor,
    shard_x,
    shard_x_shardtensor,
)

dist.init_process_group("nccl")
rank, world = dist.get_rank(), dist.get_world_size()
torch.cuda.set_device(rank)
dev = torch.device("cuda", rank)
mesh = init_device_mesh("cuda", (world,), mesh_dim_names=["cp"])
group = mesh.get_group()

b = 1
t_full = int(os.environ.get("T_FULL", 16))
x_full = int(os.environ.get("X_FULL", 12288))
c = int(os.environ.get("EMBED_DIM", 1536))
t_local = t_full // world
dtype = torch.bfloat16


def manual(local):
    return shard_t(shard_x(local, group), group)


def st(local):
    return shard_t_shardtensor(shard_x_shardtensor(local, mesh), mesh)


def bench(fn, name, backward, iters=30, warmup=10):
    local = torch.randn(
        b, t_local, x_full, c, device=dev, dtype=dtype, requires_grad=backward
    )
    for _ in range(warmup):
        out = fn(local)
        if backward:
            out.float().pow(2).mean().backward()
            local.grad = None
    torch.cuda.synchronize()
    dist.barrier()
    t0 = time.perf_counter()
    for _ in range(iters):
        out = fn(local)
        if backward:
            out.float().pow(2).mean().backward()
            local.grad = None
    torch.cuda.synchronize()
    dist.barrier()
    ms = (time.perf_counter() - t0) / iters * 1e3
    if rank == 0:
        print(f"  {name:12s} {'fwd+bwd' if backward else 'fwd':8s}: {ms:8.3f} ms/iter")


def bench_stage(name, fn, iters=30, warmup=10):
    local = torch.randn(b, t_local, x_full, c, device=dev, dtype=dtype)
    for _ in range(warmup):
        fn(local)
    torch.cuda.synchronize()
    dist.barrier()
    t0 = time.perf_counter()
    for _ in range(iters):
        fn(local)
    torch.cuda.synchronize()
    dist.barrier()
    if rank == 0:
        print(f"    stage {name:24s}: {(time.perf_counter() - t0) / iters * 1e3:8.3f} ms")


if rank == 0:
    print(f"world={world} local=({b},{t_local},{x_full},{c}) {dtype}")
for bwd in (False, True):
    bench(manual, "manual a2a", bwd)
    bench(st, "shardtensor", bwd)

_st = ShardTensor.from_local(
    torch.randn(b, t_local, x_full, c, device=dev, dtype=dtype), mesh, [Shard(1)]
)
bench_stage("from_local", lambda x: ShardTensor.from_local(x, mesh, [Shard(1)]))
bench_stage("redistribute", lambda x: _st.redistribute(placements=[Shard(2)]))
dist.destroy_process_group()
```
</details>

aayushg55: @negin513 for viz

aayushg55:
Our model needs QK-normalization with `affine=False` in the spatial attention, but TE's `MultiheadAttention` only supports affine=True. Options I see:

- **(a)** TE: Keep the affine params but exclude them from the optimizer.
- **(b)** Build a small attention module: affine-free QK-norm + TE `DotProductAttention`.
- **(c)** Use the existing timm backend, which already supports affine=False.

Currently, I went with c (this is also what the v1 ckpt uses), but just wanted to flag this incompatibility with the TE backend as set up now.

physicsnemo/models/dit/dit.py
```diff
@@ -263,10 +263,14 @@ def __init__(
 
         is_natten = attention_backend in ("natten2d", "natten2d_rope")
 
-        # Latent (token) grid size, used by the NATTEN backends.
-        self._latent_h = self.input_size[0] // self.patch_size[0]
-        self._latent_w = self.input_size[1] // self.patch_size[1]
-        latent_hw = (self._latent_h, self._latent_w)
+        # Only NATTEN uses the latent grid; other backends may pass a non-2D input_size
+        if is_natten:
```
aayushg55, line 268: This is a patch to fix a recent DiT / HealDA v1 bug (unrelated to my PR introducing v2)

physicsnemo/experimental/models/healda/sharding.py
pzharrington: I suspect the ShardTensor dispatch overhead is responsible to the differences in eager mode. ShardTensor does not yet support torch compilation, but will very very soon (like a week or two). Could be interesting to look at this again when it lands. It'd be nice to have a fully ShardTensor-based implementation but if this PR is looking ready before then, perhaps we can merge it with both implementations and defer the decision (which should not change the user-facing APIs/functionality

physicsnemo/nn/module/hpx/tokenizer.py
```diff
@@ -60,6 +61,9 @@ class HEALPixPatchTokenizer(nn.Module):
         HEALPix resolution level of input data.
     level_coarse : int
         HEALPix resolution level after patch embedding (model level).
+    separate_time_axis : bool, optional, default=False
+        If ``True``, return tokens as :math:`(B, T, X', D)` instead of the flat
+        :math:`(B, T X', D)` sequence.
```
pzharrington, line 66: Unless there is other time-dimension specific behavior in this module that happens when `separate_time_axis=False`, I suggest documenting that pathway in the standard `(B, L, D)` shapes here (and in jaxtyping annotations etc) for clarity

pzharrington: Actually I take it back, disregard. Given the input to forward has a time dim, I think this makes sense either way

physicsnemo/experimental/models/healda/adaln.py
pzharrington: I like this consolidation, but I find it a bit odd to have the variadic return from `forward` and the fact that e norm/application behavior is applied differently for `n_blocks=1` vs `n_blocks > 1`. What about having a simpler class that just returns the `shift, scale, gate` for `n_blocks`:
```
class AdaLNModulation(nn.Module):
    """adaLN-Zero projection: c -> 3*n_blocks modulation tensors, in regular
    [shift, scale, gate] * n_blocks order. Norms live at the call sites."""
    def forward(self, c) -> tuple[torch.Tensor, ...]:
        return self.proj(c).chunk(3 * self.n_blocks, dim=-1)
```
and then some simple reusable helpers, either as classmethods or standalone functinos:
```
def modulate(x_normed, shift, scale):                 # x * (1 + scale) + shift
    return torch.addcmul(_broadcast(shift, x_normed.ndim),
                         x_normed, 1 + _broadcast(scale, x_normed.ndim))

def gated_residual(residual, branch_out, gate, drop_path=None):
    gate = _broadcast(gate, residual.ndim)
    if drop_path is not None:
        gate = drop_path(gate)
    return torch.addcmul(residual, gate, branch_out)
```
Thoughts?

pzharrington:
@aayushg55 I think `VideoDiTBlock` is fine as a standalone block, subclassing `DiTBlock` seems like it'd get messy. However a more general concern is I thought we were replacing the v1 healda architecture with this update (not supporting the older model/checkpoints), it looks like instead you've added the v2 in alongside and kept v1?

physicsnemo/experimental/models/healda/pixel_attention_utils.py
pzharrington, line 1: Might as well just consolidate this into `pixel_cross_attention.py` to reduce file count

physicsnemo/experimental/models/healda/pixel_cross_attention.py
```diff
@@ -0,0 +1,1080 @@
...
+)
+
+triton = OptionalImport("triton")
+tl = OptionalImport("triton.language")
+
+
+# Directory for the persisted Triton autotune cache. Overridable so deployments
+# can redirect it to a fast/shared location; defaults under the user cache dir.
+_AUTOTUNE_CACHE_DIR = os.environ.get(
+    "PHYSICSNEMO_CACHE_DIR", os.path.expanduser("~/.cache/physicsnemo")
```
pzharrington: This is not a standard/documented env var for the package. Can we just rely on whatever Triton defaults to (both location and env var controlling)?

physicsnemo/experimental/models/healda/temporal_attention.py
```diff
@@ -0,0 +1,145 @@
...
+        self.use_rope = use_rope
+        self.head_dim = embed_dim // num_heads
+        self.linear_attention = linear_attention
+        self.temporal_attn_legacy_scaling_bug = temporal_attn_legacy_scaling_bug
+        self.causal_window = causal_window
+
+        if self.use_rope:
+            self.rope = RotaryPositionEmbedding1D(
+                head_dim=self.head_dim, max_seq_len=max_seq_len, theta=rope_base
+            )
```
pzharrington: Heads up I'm thinking of refactoring these modules slightly so they are better for multi-block models (as most models are) -- instead of the embedding tables being rebuilt per attention block the rope module can be instead be created once as a submodule of the top-level network, and then the same embedding tables there can be reused in all blocks. Should be a small/minor refactor

physicsnemo/experimental/models/healda/temporal_attention.py
```diff
@@ -0,0 +1,145 @@
...
+        rope_base: Base frequency for RoPE.
+        max_seq_len: Maximum sequence length for RoPE cache.
+        linear_attention: If True, skip softmax -- attention weights are raw
+            dot-products (causal masking uses zero-fill instead of -inf).
+        temporal_attn_legacy_scaling_bug: When True *and* linear_attention is True,
+            normalize q.k by sequence length instead of head dim. Reproduces a bug
+            from older checkpoints; set False for correct behaviour.
+        causal_window: When set (and the forward pass is causal), restrict each
+            frame to attend to itself and the previous ``causal_window - 1`` frames.
+    """
```
pzharrington: Make into standard docstring format (see CODING_STANDARDS). Also this can probably be moved into the `video_dit_block.py` file to reduce the file count

physicsnemo/experimental/models/healda/kernels/autotune_cache.py
```diff
@@ -0,0 +1,143 @@
...
+                {"key": [_enc_key_elem(e) for e in key], "config": _enc_config(cfg)}
+            )
+            n += 1
+        g[name] = entries
+    os.makedirs(os.path.dirname(os.path.abspath(path)), exist_ok=True)
+    tmp = path + ".tmp"
+    with open(tmp, "w") as f:
+        json.dump(data, f, indent=2)
+    os.replace(tmp, path)
+    return n
```
pzharrington, line 147: Can we maybe prepend a healda (or configurable) name to the paths to avoid possible mixing/pollution with other similarly named kernels? Not too familiar with triton's behavior here but seems like generally better practice to try and keep things self-contained if we are manipulating things in the cache. And there should be a warning to users in the docstrings of any modules that could trigger this behavior

physicsnemo/experimental/models/healda/kernels/obs_tokenizer_film.py
pzharrington, line 1: I suggest trying to consolidate all triton kernels and related utils into one file, there's a lot of different files here

physicsnemo/experimental/models/healda/video_healda.py
```diff
@@ -0,0 +1,304 @@
...
+    onnx: bool = False
+    func_torch: bool = False
+    auto_grad: bool = False
+
+
+class HealDAv2(Module):
+    r"""
+    Video data-assimilation model combining HEALPix tokenizers, a FiLM observation
+    tokenizer, and a VideoDiT backbone.
+
```
pzharrington, line 103: Description of the model and intended usage should be more complete

negin513:
This PR adds a substantial new v2 model stack but doesn't surface any of the building blocks through the package's public API. I think it might worth exposing these similar to v1 through the public API.

physicsnemo/experimental/models/healda/kernels/autotune_cache.py
aayushg55: Yea agree that there might be a better way to do things. For context, the point of this cache is to cache the best kernel config from the autotuning sweep (as opposed to the generated kernel itself, whose caching triton internally controls and can be set through some env variables). This caching is more relevant for training as due to the dynamic obs count and each rank seeing a different number of obs at any given iter, they will be running the autotune at different points in the training, but all ranks are forced to stall if any stalls. Furthermore, across training runs, the autotuning is rerun, so this cost is normally payed every training job.

In contrast, this tries to just pay this cost upfront once and cache the autotuning result.

aayushg55:
> @aayushg55 I think `VideoDiTBlock` is fine as a standalone block, subclassing `DiTBlock` seems like it'd get messy. However a more general concern is I thought we were replacing the v1 healda architecture with this update (not supporting the older model/checkpoints), it looks like instead you've added the v2 in alongside and kept v1?

Sounds good. I wasn't sure if we wanted to be removing the v1 architecture too. But given that the v2 architecture is strictly better and the old architecture can be found in older PNM / my previous PR branch for reproducibility, keeping v1 would add dead code, so I can delete it.

aayushg55:
> This PR adds a substantial new v2 model stack but doesn't surface any of the building blocks through the package's public API. I think it might worth exposing these similar to v1 through the public API.

What would this look like?

physicsnemo/experimental/models/healda/temporal_attention.py
aayushg55: Thanks, will update to match the coding standards. I'd lean toward keeping temporal attention in its own module since the `TemporalAttention` module is quite separate from the `VideoDitBlock` and could be used as a more general module. But if you'd rather co-locate it, your call.

physicsnemo/experimental/models/healda/kernels/obs_tokenizer_film.py
aayushg55: Agreed there are many files under the `healda/` folder and the grouping may be currently unclear, but there are a few different unrelated triton kernels involved and clumping them into a single multi-thousand line file makes it harder to follow.

I propose creating a a `healda/kernels/` subdir with one file per op, and also moving the autograd glue/host dispatch code out of the current 1k-llne module files such that each module file is just the torch model layer and easier to follow without the implementation complexity.

physicsnemo/experimental/models/healda/pixel_attention_utils.py
aayushg55: Will do

physicsnemo/experimental/models/healda/pixel_cross_attention.py
```diff
@@ -0,0 +1,1080 @@
...
+        # unfold the per-pixel output back to the (B, T, X, C) layout.
+        out = self._forward_impl(
+            hidden_states,
+            context.tokens,
+            total_pixels,
+            context.cu_seqlens_k,
+            context.max_seqlen_k,
+            group_map=context.group_map,
+        )
+        return out.view(b, t, x, ch)
```
negin513: I’m a little worried about the dimensions here. So `hidden_states` enters this module with shape `(B, T, X, input_dim)`, while `_forward_impl()` returns the result of `self.out_proj` in both the empty-token and normal paths, so its last dimension is `self.output_dim`. The final reshape, here, currently uses the input channel count `ch`:

    return out.view(b, t, x, ch)

This works now, in the current `HealDAv2` wiring because `input_dim == output_dim == hidden_size`, but `PixelCrossAttention` exposes `input_dim` and `output_dim` as separate configurable values.....

1. Should this reshape use `self.output_dim` instead, and /or 
2. should we add validation/tests to make the expected relationship between `input_dim`, `output_dim`, and the surrounding block `hidden_size` explicit? 

Maybe a small test with `input_dim != output_di` to make sure this works correctly?

physicsnemo/experimental/models/healda/pixel_cross_attention.py
```diff
@@ -0,0 +1,1080 @@
...
+    n_kv_heads: int,
+    scale: float,
+    B_v: Optional[torch.Tensor] = None,
+) -> torch.Tensor:
+    r"""Pure-PyTorch ragged grouped-query attention; reference for :func:`pixel_attention`.
+
+    Projects each pixel's token slice to keys/values, applies softmax attention
+    from that pixel's query heads, and writes the per-pixel output. The key bias is
+    omitted (softmax cancels it); ``group_map`` does not apply (grouping is a
+    kernel-launch optimization that does not change the result).
```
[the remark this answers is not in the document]
aayushg55: Resolved

physicsnemo/experimental/models/healda/adaln.py
aayushg55: I have changed it to your suggested `modulate` and `gated_residual` helpers

physicsnemo/experimental/models/healda/video_healda.py
aayushg55: Updated with more detail

physicsnemo/experimental/models/healda/kernels/obs_tokenizer_film.py
aayushg55: We could also fold the VideoDitBlock and VideoDiT into the same file to reduce file count

negin513 (commented):
Great job @aayushg55!

I’ll add a few comments inline.

One higher-level thing I noticed: this PR does not add an end-to-end training recipe or example showing how the new `HealDAv2`, `VideoDiT`, `ObsContext`, and observation/pixel-attention pieces are intended to be wired together... 

Does it make sense to add a minimal example in this PR under `examples/weather/healda/` ?

physicsnemo/experimental/models/healda/pixel_cross_attention.py
aayushg55: Good catch. The module is currently over parameterized with both an input_dim and output_dim toggle. This should be collapsed to a single parameter. Matching a standard transformer layer, the hidden dimension should not change after going through the op. Will fix

aayushg55:
> One higher-level thing I noticed: this PR does not add an end-to-end training recipe or example showing how the new `HealDAv2`, `VideoDiT`, `ObsContext`, and observation/pixel-attention pieces are intended to be wired together...
> 
> Does it make sense to add a minimal example in this PR under `examples/weather/healda/` ?

I think the idea was for this PR to only introduce the architecture to get it in earlier rather than later and enable adding inference capability in Earth2Studio before the next release. Further integrating the training loop and making sure it is up-to-standard would likely take significant engineering effort (and closer collaboration with the PNM team), as our internal training loops/dataloaders are not using PNM in any form at the moment and do not follow how existing PNM training loops are set up. Once the dataloader pieces are also added, we can follow up with the training loop.

physicsnemo/experimental/models/healda/pixel_cross_attention.py
aayushg55: Fixed

physicsnemo/experimental/models/healda/sharding.py
negin513: Sorry for missing the context here, @aayushg55. Is this comment regarding the performance issue with ShardTensor? 

If so, +1 to Peter, this should be resolved in a couple of weeks. We have added a bunch of new utilities for this.

aayushg55: Yes, this was just to show how for our particular use case, the more general ShardTensor utility currently has overhead over a more specific manual all-to-all approach.

negin513: Oh! your results are showing fwd in shardTensor is 10X slower. This should'nt be the case for `fwd`, since `redistribute` should ultimately just do the `all_to_all` under the hood.... this is pointing me a performance bug (likely extra copies, extra communication, or something).... 

Let me do some testing here; but we are doing an almost new ShardTensor implementation. 

Great catch @aayushg55!

pzharrington: @negin513 the reason I think it's just dispatch overhead is the timing is very short (0.29 ms for pure torch, 2.5 ms for ShardTensor), so the timescales where CPU launch latency and things like that matter more. But of course profiling is the best way to figure it out

negin513:
> > This PR adds a substantial new v2 model stack but doesn't surface any of the building blocks through the package's public API. I think it might worth exposing these similar to v1 through the public API.
> 
> What would this look like?

I was thinking we could update `physicsnemo/experimental/models/healda/__init__.py` to export the main user-facing v2 pieces, similar to how v1 exports `HealDA`.

At minimum, probably:

```python
from .healda_v2 import HealDAv2, HealDAv2MetaData
from .obs_context import ObsContext, PixelGroupMap
```

Similar to HealDA-v1 stuff.  But I would keep the Triton kernel modules private. The main goal is that a recipe or user can do `from physicsnemo.experimental.models.healda import HealDAv2, ObsContext` without importing deep module paths.

aayushg55:
> > > This PR adds a substantial new v2 model stack but doesn't surface any of the building blocks through the package's public API. I think it might worth exposing these similar to v1 through the public API.
> > 
> > 
> > What would this look like?
> 
> I was thinking we could update `physicsnemo/experimental/models/healda/__init__.py` to export the main user-facing v2 pieces, similar to how v1 exports `HealDA`.
> 
> At minimum, probably:
> 
> ```python
> from .healda_v2 import HealDAv2, HealDAv2MetaData
> from .obs_context import ObsContext, PixelGroupMap
> ```
> 
> Similar to HealDA-v1 stuff. But I would keep the Triton kernel modules private. The main goal is that a recipe or user can do `from physicsnemo.experimental.models.healda import HealDAv2, ObsContext` without importing deep module paths.

Thanks, I have updated the physicsnemo/experimental/models/healda/__init__.py to export the v2 pieces

physicsnemo/experimental/models/healda/__init__.py
```diff
@@ -14,11 +14,40 @@
...
+    "ObsTokenizerFiLM",
+    "ObsContext",
+    "PixelGroupMap",
+    "MultiSensorObsEmbedder",
+    "ObsTokenizer",
+    "SensorEmbedder",
+    "UniformFusion",
+    "ScatterAggregator",
+    "scatter_mean",
+]
```
pzharrington, line 40: We definitely don't need to publicly export all of these (e.g. `MetaData` classes are not typically exported), and I would actually push for the majority of these to _not_ be exported. Advanced users who want to go in and access them can use a direct import but otherwise we only need to export the `HealDA` architecture and maybe any components that could be useful for people working with custom obs data.

pzharrington: And a more major change, but one I think is worth doing, is to **outright remove the healda v1 architecture and replace it with this newer/better version**. We decided we are not going to support the legacy v1 model (users can still access it by installing an older version of phyiscsnemo+earth2studio), and planned for it by version-capping the physicsnemo source for healda in last earth2studio release. So we should be free to remove/replace it here.

Ultimately the "v2" is not very meaningful for most external users and I think it's more clear to have one class for the architecture. We don't need to wait a release cycle to deprecate the older one since it's all in `experimental`

physicsnemo/experimental/models/healda/kernels/obs_tokenizer_film.py
pzharrington: Ok yeah I think a `healda/kernels/`  should help, let's go with that. Consolidating VideoDiTBlock also sounds good

physicsnemo/experimental/models/healda/healda_v2.py
```diff
@@ -0,0 +1,372 @@
...
+    a deterministic regression model, where the ``noise_labels`` are fixed to zero.
+
+    Data flows through four stages:
+
+    1. Static conditioning fields are patch-tokenized from the fine ingest grid
+       (``level_in``) down to the backbone grid (``level_model``) by
+       :class:`physicsnemo.nn.module.hpx.tokenizer.HEALPixPatchTokenizer`.
+    2. A :class:`.video_dit.VideoDiT` backbone processes the token sequence with
+       spatial attention, factorized temporal attention, and adaLN-Zero
+       conditioning built from the EDM noise embedding and the calendar
```
pzharrington: EDM noise embedding? This isn't a diffusion model so that's a bit confusing

aayushg55: The architecture uses all of the DiT architecture conditioning components (AdaLN + noise/condition label). We trained it as a regression model (huber loss) setting noise to always be 0, but it could be trained as a diffusion model too.

physicsnemo/experimental/models/healda/__init__.py
negin513: +1 @pzharrington, thanks for the context.  The `MetaData` and `HealDAv2` was my suggestion. I was mostly thinking about discoverability from the recipe/user side on useful utilities such as `ObsContext`, but I agree we should keep the public API tighter and not export internals unnecessarily. 

Given your point that we do not plan to support the legacy v1 model, I agree that replacing the public HealDA export with the newer implementation is cleaner than exposing a parallel HealDAv2. Then we can keep the API focused on the architecture itself, plus only the obs-data pieces that recipes/custom data users actually need, e.g. `ObsContext` (?)  if users are expected to construct those directly. @pzharrington, I 100% follow your judgment on this. + @aayushg55  maybe you can comment around this what utils from HEALDAv2 will be helpful for future to expose through API.

physicsnemo/experimental/models/healda/temporal_attention.py
pzharrington: Ok, on the consolidation, how about the following:

1. Kernels + related utils in `healda/kernels/` and video_dit_block added to video_dit as per other comment
2. temporal_attention, cross_attention, and pixel_cross_attention merged into one attention_layers.py

Imo a singular ~1k line file with all the attention utils is better than disparate files

physicsnemo/experimental/models/healda/kernels/autotune_cache.py
pzharrington: Ok, how about we just make this behavior clearly documented in docstrings, and make it opt-in? i.e. unless user sets some env var specific to healda, no hidden/automatic cache behavior will happen

physicsnemo/experimental/models/healda/__init__.py
aayushg55: I think the cross attention module (PixelCrossAttention), tokenizer (ObsTokenizerFiLM), the obs wrapper (ObsContext), the combined architecture (HealDAv2 renamed to HealDA), and probably the VideoDiT would be things users might be interested in using.

physicsnemo/experimental/models/healda/temporal_attention.py
aayushg55: Sure, that works

physicsnemo/experimental/models/healda/healda_v2.py
pzharrington: Yeah I suppose that's true. I'm just not sure about advertising that here quite yet -- partly since no one has evaluated/experimented with the model in that realm as far as I know, but mainly because as of now making it a diffusion model would have to be a manual implementation. The `forward` signature expected by all `physicsnemo.diffusion` components is `(x, t, condition: torch.Tensor | TensorDict)` ([see here](https://github.com/NVIDIA/physicsnemo/blob/main/physicsnemo/diffusion/base.py#L79)) so it would take a bit of work to massage the the current `Healdav2` into compliance, or one would have to write their own diffusion loop from scratch (which we don't necessarily want to encourage for obvious reasons)

physicsnemo/experimental/models/healda/__init__.py
pzharrington: Sounds good, let's go with that then

physicsnemo/experimental/models/healda/healda_v2.py
aayushg55: Makes sense, I can revise the docstring

physicsnemo/experimental/models/healda/sharding.py
negin513: Ran and profiled the reshard microbenchmark on 8×H100 and, numbers in the same ballpark as @aayushg55's. 

Consistent with your theory @pzharrington , so `from_local` (pure wrapper construction) accounts for ~a quarter of the cost and is paid twice per block. iThere is an extra `all_gather` there though: `from_local` defaults to `sharding_shapes="infer"`, which does a real collective to learn every rank's shape; even though our shard is evenly divided in this case.... There is also an extra avoidable `all-to-all` there too. 

I think I might have a fix around this -- will share shortly,  and considering the factorized `ShardTensor` this additional overhead should be resolved.

negin513: @aayushg55 I've got a fix for the extra `all-to-all` in sharding: https://github.com/NVIDIA/physicsnemo/pull/1779

Numbers on the reshard primitives, with PR #1758 fix: 
(4×GB200, T=16, X=12288, C=1536, bf16)

| path | fwd | fwd+bwd |
  |---|---|---|
  | manual all-to-all | 0.835 ms | 2.529 ms |
  | ShardTensor.redistribute (before fix) | 2.609 ms | 6.229 ms |
  | ShardTensor.redistribute (after fix, #1779 ) | 1.555 ms | 3.969 ms |

I've got a secondary fix as well, which reduce this further but I am waiting for some ShardTensor refactor.

physicsnemo/experimental/models/healda/__init__.py
nbren12: Removing the healda v1 seems fine eventually, but would be nice to have some period of overlap from v1 to v2 so that users can smoothly transition. pinning dependencies is not a flexible solution (e.g. if a user wants to use some unrelated bug fix or feature in physicsnemo).  I'd lobby for adding a deprecation warning and waiting a month or two to delete. It seems doubtful that in that period of time the old code will become a serious maintenance burden.

nbren12: note how e.g. huggingface is able maintain old models: https://github.com/huggingface/diffusers/tree/main/src/diffusers/models/transformers.

I don't really understand the harm of some effectively static .py files and a few model variants. If these files require frequent updates it suggest some problematic coupling exists in the physicsnemo DiT. I think this insistence on "one DiT" may be the issue. 

Regardless, if healda v1 becomes a real problem in the future maybe just delete then.

pzharrington: > some period of overlap from v1 to v2 so that users can smoothly transition

Do we know of any existing users developing with the current arch though? As far as I know, we're just using it to back the v1 checkpoint served in earth2studio. Regarding warnings, all imports from `phyiscsnemo.experimental` carry a warning that APIs are subject to change so I think there is fair communication on that front. I'm less worried about maintenance burden and more just trying to limit the confusion for users seeing two healda archs, or eventually just one but it's still called "v2" for some reason. IMO it's much more clean to say "we have updated parts of this experimental model to expand its capabilities" in the changelog and move on with our lives (aka updating earth2studio to serve the new & improved checkpoints 😈 )

nbren12: Not sure. I know of some people who have used it. There were 48 downloads last month.

>I'm less worried about maintenance burden and more just trying to limit the confusion for users seeing two healda archs, or eventually just one but it's still called "v2" for some reason. IMO it's much more clean to say "we have updated parts of this experimental model to expand its capabilities" in the changelog and move on with our lives (aka updating earth2studio to serve the new & improved checkpoints 😈 )

If I were using this in production, I would pretty annoyed if the same name `Healda` suddenly resolved to a different architecture that behaves differently. Silently changing the model is probably the worst-case scenario for production users...esp since the v1 benefited from many robustness checks that we haven't done for v2. Reproducibility is important for more than just scientists.

Note how huggingface includes the model version in the user facing api ("Lllama-2-7b-hf"):
Screenshot 2026-07-02 at 2 57 48 PM

nbren12: Anyway, just asking for a deferral on the cleanup so that we can at least test the two implementations against one another in the same code base. I don't see why it has to be an immediate change if the concern isn't maintenance.

pzharrington: To be clear, users of earth2studio will not experience any breaking changes. The previous release and current version on github `main` cap the physicsnemo dependency for `healda` so users can still run the v1 checkpoint for inference (that's where I assume the downloads are coming from) backed by the previous/existing architecture in PNM. We also use versioning in our HF packages so the e2studio model wrapper wont be broken when we upload newer checkpoints (and later on a user could still choose to download the older checkpoints if desired by pulling down a previous commit). When we do update the package, we'll update the model wrapper to use the newer tag and users would get seamlessly upgraded to the new version.

For context, we had a call a week or two ago with product folks and Mike/Ayush and decided this in-place upgrade approach of v1 was the move, I'm not just raising this out of nowhere 🙂 I do think there is a need for E2 to come up with a more robust model versioning practice/strategy in general though

pzharrington: And also for context on this:

> I think this insistence on "one DiT" may be the issue.

PNM has gotten feedback from multiple sources (external users, devtech, SAs, ...) and across domains/applications that duplicated or multiple highly similar model architecture implementations degrades the package quality and user experience. Hence the pivot towards more lower-level reusable ops/layers/blocks, and reusing larger pieces where possible -- it allows existing test coverage and feature completeness on those components to spread better across the package. I would not describe our approach as "everything in one DiT" but rather "use existing components where possible, extend and modify as needed".

physicsnemo/experimental/models/healda/temporal_attention.py
pzharrington: This will be imminently merged, when it does and you update this branch from latest main it should require a small update here (sorry for the churn but this setup should be better in terms of memory among other things). https://github.com/NVIDIA/physicsnemo/pull/1778

physicsnemo/experimental/models/healda/__init__.py
pzharrington: @aayushg55 after some more iteration on Slack, Noah and I have arrived at the following action items:

- You or your agents come up with a more descriptive identifier than "v2" for the HealDAv2 architecture class and related items in this PR here
- Add an explicit deprecation warning to the v1 `HealDA` architecture and a `TODO` to remove it, but otherwise keep it in the repo for now
- Remove all public exports of old components except the top-level architecture
- Make sure those older components related to v1 have some reasonable documentation that they will be dropped in the future.

This should unblock progress on this particular PR while other questions on the release are ironed out

aayushg55: Sounds good, will go with `VideoHealDA`

physicsnemo/experimental/models/healda/healda_v2.py
aayushg55: Updated

physicsnemo/experimental/models/healda/kernels/obs_tokenizer_film.py
aayushg55: Both are done

physicsnemo/experimental/models/healda/kernels/autotune_cache.py
aayushg55: Documented this in the main model file and the pixel cross attention module with a clear opt-in only if the `HEALDA_PIXEL_ATTN_AUTOTUNE_CACHE_DIR` is set. Kept separate from the triton internal cache dirs in case someone doesn't want triton's caching but does want this.

physicsnemo/experimental/models/healda/__init__.py
```diff
@@ -14,11 +14,41 @@
...
+
+:class:`HealDAv2` (current) uses :class:`PixelCrossAttention` and
+:class:`ObsTokenizerFiLM`. :class:`HealDA` (v1) uses the
+:class:`MultiSensorObsEmbedder` stack (:class:`SensorEmbedder`,
+:class:`UniformFusion`, :class:`ObsTokenizer`).
+"""
+
+from .healda import HealDA, HealDAMetaData
+from .healda_v2 import HealDAv2, HealDAv2MetaData
+from .obs_context import ObsContext, prepare_obs_context
```
aayushg55, line 27: While making the HealDAv2 example, I realized constructing `ObsContext` directly required callers to manually call `sort_and_pack`, `counts_to_cu_seqlens`, per-observation tensor reordering, and `build_pixel_group_map`. So I instead added `prepare_obs_context` as the exported public helper for the required preprocessing, so callers can pass raw observation tensors and get a valid `ObsContext` with consistent packing metadata.

physicsnemo/experimental/models/healda/obs_context.py
```diff
@@ -0,0 +1,345 @@
...
+        return PixelGroupMap(
+            program_ptr=self.program_ptr.to(device=device, non_blocking=non_blocking),
+            program_pixels=self.program_pixels.to(
+                device=device, non_blocking=non_blocking
+            ),
+        )
+
+
+@dataclasses.dataclass
+class ObsContext:
```
aayushg55, line 76: ## Refactor obs cross-attention to plain kwargs

The observation cross-attention was getting messy: PixelCrossAttention was handed the whole ObsContext object, and observation tokens were stitched onto that object after the fact. That's an odd, one-off contract compared to every other attention module in the codebase.

Switched to passing explicit cross_attn_kwargs (tokens, cu_seqlens_k, max_seqlen_k, group_map), matching how attn_kwargs / temporal_attn_kwargs are threaded through VideoDiT and the standard DiT. The model now tokenizes locally and forwards the tensors; the cross-attention no longer knows about ObsContext at all.

One thing that remains: the model's forward still takes an ObsContext dataclass. It's now only used to (a) bundle the output of prepare_obs_context and (b) define the forward input contract — the tokenizer and cross-attention both consume raw tensors, not the object. Open question: do we keep ObsContext as the typed input contract, or drop it and take the raw tensors + packing directly in forward, documenting the contract there?

aayushg55:
Thanks @negin513 and @pzharrington for the feedback. I have updated the code to address the comments and also did some refactoring to better align with the rest of the PNM standards/DiT structure. Below is a summary:

### Architecture consolidation & structure
- Folded `VideoDiTBlock` into `video_dit.py` and merged the separate attention modules (`temporal_attention`, `cross_attention`, `pixel_cross_attention`) into a single `attention_layers.py`, ordered temporal-first with section headers.
- Consolidated all Triton kernels under `kernels/`.
### API alignment
- Aligned `VideoHealDA` / `VideoDiT` conditioning API (`t`, `condition`, dim naming) with v1 `HealDA` and the standard PNM `DiT`.
- Made `VideoDiT` / `VideoDiTBlock` attention plumbing consistent with `DiT`: `attn_kwargs`, `temporal_attn_kwargs`, `cross_attn_kwargs` dicts instead of an opaque `ObsContext` passed into cross-attention.
- Inlined the pixel cross-attention `forward` (dropped the `_forward_impl` indirection); added `use_rope` to `VideoHealDA` for symmetry.
- Switched to the new shared-provider RoPE implementation (`RotaryEmbedding1DTables` owned by `VideoDiT`).
### Observation preprocessing
- Added `prepare_obs_context` as the exported public helper so callers pass raw obs tensors and get a valid `ObsContext` with consistent packing (sort/pack, `cu_seqlens`, group map) — no more manual multi-step setup.
- Removed the now-dead `ObsContext.tokens` field left over from the old "tokens attached after the fact" pattern; refreshed docstrings so raw fields → `ObsTokenizerFiLM`, packing → cross-attention.
### Standards & docs
- Added MOD-005 forward shape validation.
- Added Examples, Sphinx cross-references, and tidied validation helpers.
- Made the pixel-attn autotune cache opt-in via a directory env var.
### Rename & v1 deprecation
- Renamed `HealDAv2` → `VideoHealDA` (file, class, metadata, tests).
- Added a deprecation notice + runtime `DeprecationWarning` + removal TODO to v1 `HealDA`, following PNM precedent.
- Dropped public exports of v1-only components (kept top-level architectures), documented them as slated for removal, and pointed v1 tests at submodule imports.
- Properly linted the v1 files using the leading-space single-token jaxtyping convention (matching the integrated DiT layers).

physicsnemo/experimental/models/healda/obs_context.py
pzharrington: While I don't love having a custom dataclass as a required arg in the `forward` of the top-level model (typically custom containers like that are just used inside models to simplify internal plumbing), I'm not sure requiring users to just pass a splatted set of tensors as args to `forward` is better. That would allow more clear shape annotations/jaxtyping hints on the constituent tensors in `ObsContext`, but has poorer ergonomics and would essentially still require the user to use setup functions like `prepare_obs_context` to get things in the right format/shapes. There is at least some precedent for non-Tensor/container-like `forward` args in the GNN models `graph` arguments, but it's true the pattern here is nonstandard for a non-GNN model.

pzharrington: If we stick with the current dataclass route, let's improve the docstring to be a bit more clear add the tensor shapes and descriptions of all attributes. Also, small suggestion -- the `PixelGroupMap` is a pretty thin dataclass with just two int tensor attributes and a modified `.to`. That could probably just be absorbed into the `ObsContext` class, which would in turn make documenting a bit easier too (avoid having to also document `PixelGroupMap` separately). Makes sense to do if it's not really used elsewhere

aayushg55: Currently, PixelGroupMap bundles the CSR pair (program_ptr, program_pixels). It's created from build_pixel_group_map() and consumed by the PixelCrossAttention.forward. The initial reason for keeping `PixelGroupMap` separate was that it's an add-on optimization, not necessary for the obs attention. (There may be possible situations where it is not helpful, although with our current problem size, it delivers performance gains).

If enabled, the 2 PixelGroupMap fields must appear together, whereas all the other ObsContext fields are always necessary. So keeping it a class buys treating it as one nullable unit, as opposed to two separate optionals. If we keep it, we could switch to @tensorclass which auto-provides .to()/device movement.

aayushg55: Will update the docstring

aayushg55: Updated docstring, but I realized tensorclass wouldn't help in this case, as we don't want to accidentally cast the integer type tensors so we'd still need to override its auto-generated .to

physicsnemo/experimental/models/healda/video_healda.py
```diff
@@ -0,0 +1,491 @@
...
+    HEALPixPatchDetokenizer,
+    HEALPixPatchTokenizer,
+)
+
+
+@dataclass
+class VideoHealDAMetaData(ModelMetaData):
+    """Metadata for VideoHealDA model."""
+
+    name: str = "VideoHealDA"
```
pzharrington: No `name` field needed

pzharrington (approved):

physicsnemo/experimental/models/healda/video_healda.py
aayushg55: Fixed

negin513 (approved):

negin513:
/ok to test [dc379aa](https://github.com/NVIDIA/physicsnemo/pull/1758/commits/dc379aa2a342edb6afafe6b606d7bee9c38aa8fa)

loliverhennigh (approved):
I think we ca move the kernels to nn.functionals when this is out of experimental but works for now

negin513:
Great point @loliverhennigh!! At what point, do we typically bring things out of experimental? Do we track such tasks in issues or any format? I created an issue (https://github.com/NVIDIA/physicsnemo/issues/1824)  for this as a tracker for now.

nbren12:
Woohoo. Congrats @aayushg55

PR #1762, opened 2026-06-26
Improve StormScope recipe masking ergonomics

pzharrington:
# PhysicsNeMo Pull Request

### Summary
Reworks spatial masking in the StormCast/StormScope recipe into a single, dataloader-owned interface, and fixes a domain-parallelism bug in the loss-mask construction. Previously masking came from two reconciled-at-train-time sources (a dataloader `mask` key and a NaN-inspection pathway), which was confusing and produced a shape bug under domain parallelism. Changes:

- Single mask source (dataloader-owned). The per-sample `"mask"` served by the dataset (convention `1`=valid, `0`=invalid) is now the only mask source. The NaN-inspection pathway is removed entirely. `Trainer._build_masks(target, batch_mask)` translates that mask into a loss weight and an `invalid_mask`; `use_nan_mask_tokens` now solely controls whether `invalid_mask` is also injected into the DiT forward so it can swap invalid tokens for its learned mask token.
- Token-granularity loss weight for DiT. When the DiT runs with `use_nan_mask_tokens`, the loss weight is pooled to patch granularity (`max_pool2d`) and expanded back (`interpolate`), matching the model's token-level substitution. The patch size is read directly from the constructed DiT (patch_size), and the expansion uses a scalar `scale_factor`. Pixel-level weight is retained otherwise.
- Domain-parallelism fix. Eliminates a mask shape bug that arose from detecting NaNs under sharding; the mask now stays correctly `(B, 1, H, W)` as a height-sharded ShardTensor through pooling and expansion.
- Config / docs. Adds `use_mask: false` (mock dataset) and `use_nan_mask_tokens: false` (stormscope hyperparameters) defaults so the keys exist in the OmegaConf struct, documents the `"mask"` key in the dataset docstring and README, and adds an optional `use_mask` pattern to the mock dataset.

### Tests

`test_masking` now exercises the full masked DiT path across three launch configs — single (1 GPU), domain_parallel (2 GPU, height-sharded), and data_domain_parallel (4 GPU, 2×2 data+domain mesh) — skipping configs that don't match the launched world size.

## Checklist

## Review Process

### DIFF
--- a/examples/weather/stormcast/datasets/dataset.py
+++ b/examples/weather/stormcast/datasets/dataset.py
@@ -39,12 +39,16 @@ class StormCastDataset(torch.utils.data.Dataset, ABC):
             and index 1 containing the target state data
         - `lead_time_label` (optional): this must be returned if lead_time_steps > 0. A single
             integer indicating which lead time embedding should be used
-        - `mask` (optional): a numpy.ndarray or torch.Tensor of shape `(1, height, width)`
-            with values in {0, 1} (or boolean), where 1/True marks valid pixels and
-            0/False marks invalid/excluded pixels (e.g. outside sensor coverage, LAM
-            padding zones, land-sea boundaries).  When provided, the training loop uses
-            the mask as a per-pixel loss weight and, for the DiT architecture with
-            `use_nan_mask_tokens=True`, pools it to token granularity to replace
+        - `mask` (optional): a numpy.ndarray or torch.Tensor with values in {0, 1}
+            (or boolean), where 1/True marks valid pixels and 0/False marks
+            invalid/excluded pixels (e.g. outside sensor coverage, LAM padding zones,
+            land-sea boundaries).  The shape must broadcast with `(num_channels_state,
+            height, width)`: use `(1, height, width)` for a spatial mask shared across
+            all channels, `(num_channels_state, height, width)` for per-channel spatial
+            masks, or `(num_channels_state, 1, 1)` to mark entire channels as invalid.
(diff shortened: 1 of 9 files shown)


examples/weather/stormcast/datasets/dataset.py
```diff
@@ -39,6 +39,16 @@ class StormCastDataset(torch.utils.data.Dataset, ABC):
             and index 1 containing the target state data
         - `lead_time_label` (optional): this must be returned if lead_time_steps > 0. A single
             integer indicating which lead time embedding should be used
+        - `mask` (optional): a numpy.ndarray or torch.Tensor of shape `(1, height, width)`
```
jleinonen: The previous version works with any mask shape that broadcasts with `(C, H, W)`. So for instance you can use a `(C, 1, 1)` mask to include only selected channels in the loss calculation, a `(1, H, W)` mask to use the same spatial mask to all channels, or `(C, H, W)` to combine both. Does this still work?

pzharrington: No, if channel masking in the loss is desired I think that should be controlled separately. It will have to be merged with any incoming invalid spatial mask from the data loaders (which focus on just spatial masking for invalid regions -- if there is a possibility of channel-dependent invalidity then I guess we need to support that there too?)

pzharrington: Ok, updated to support channel-wise masking in the loss in addition to invalid masking from the dataloader as discussed on Slack

jleinonen (approved):
With the channel masking added, looks good!

PR #1767, opened 2026-06-30
Fix for Mesh SDF API

peterdsharpe:
## Summary

This updates the not-yet-released Mesh SDF API so `signed_distance_field_mesh()` accepts a `Mesh` directly:

```python
sdf, hit_points = signed_distance_field_mesh(mesh, query_points)
```

The public function now validates non-empty 3D triangle surfaces, uses the float32 working representation required by the BVH and Triton paths, and preserves the query shape and dtype in its outputs. The docs, data transform, and external-aerodynamics recipe now use the mesh-native interface. A private array wrapper remains for debugging and comparisons with the replaced Warp API.

## Testing

`test/mesh/spatial/test_mesh_sdf.py` covers public/private parity, invalid and empty meshes, batched queries, finite search bands, both sign methods, and CUDA/Triton behavior.

### DIFF
--- a/physicsnemo/mesh/spatial/sdf.py
+++ b/physicsnemo/mesh/spatial/sdf.py
@@ -945,12 +945,9 @@ def signed_distance_field_mesh(
     *,
     winding_backend: str = "clustertree",
 ) -> tuple[torch.Tensor, torch.Tensor]:
-    r"""Signed distance field of a triangle surface mesh.
+    r"""Compute the signed distance to a triangle surface mesh.
 
-    Uses :class:`physicsnemo.mesh.spatial.BVH` for the nearest-triangle query
-    and a :class:`physicsnemo.mesh.spatial.ClusterTree` Barnes-Hut summation for
-    the winding-number sign, all in plain PyTorch. Returns the signed distance
-    and the closest surface point for each query.
+    Returns the signed distance and the closest surface point for each query.
 
     Parameters
     ----------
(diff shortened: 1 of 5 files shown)


peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1767/commits/1b66b00261c8b01e1d6e374bf8c270d352482012

physicsnemo/mesh/spatial/sdf.py
```diff
@@ -1095,6 +1100,77 @@ def signed_distance_field_mesh(
...
+    r"""[INTERNAL - DO NOT USE] Array-based SDF mirroring the replaced Warp API.
+
+    .. warning::
+
+       **DON'T USE THIS ONE.** This is a private, debug-only entry point. Use
+       the public :func:`signed_distance_field_mesh`, which takes a
+       :class:`~physicsnemo.mesh.Mesh`, instead.
+
+       This variant takes raw ``(mesh_vertices, mesh_indices)`` arrays solely to
+       mirror the call signature of the (now replaced) Warp-based
```
coreyjadams: Though this is correct, it would be polite to not call out warp like this - especially because the mesh implementation is actually slower!

coreyjadams: Meaning: Let's focus this warning just on "hey don't use this one it's private please, the api will vanish soon."

physicsnemo/mesh/spatial/sdf.py
```diff
@@ -935,15 +938,14 @@ def _winding_number_sign_clustertree(
...
     max_dist: float | None = None,
     use_sign_winding_number: bool = False,
     *,
     winding_backend: str = "clustertree",
 ) -> tuple[torch.Tensor, torch.Tensor]:
-    """Signed distance field of a triangle surface mesh.
+    r"""Signed distance field of a triangle surface mesh.
 
     Uses :class:`physicsnemo.mesh.spatial.BVH` for the nearest-triangle query
     and a :class:`physicsnemo.mesh.spatial.ClusterTree` Barnes-Hut summation for
```
coreyjadams: I wonder if we should keep this "agnostic" to backend - I know I made this but if/when we do sort out the warp issues, and I'm confident we'll get there, it'd be nice to be able to just say, "hey this does SDF on a mesh, don't worry bout it"

coreyjadams: (no action needed unless you agree)

coreyjadams (approved):

physicsnemo/mesh/spatial/sdf.py
peterdsharpe: good call, addressed

physicsnemo/mesh/spatial/sdf.py
peterdsharpe: good call, addressed

physicsnemo/mesh/spatial/sdf.py
```diff
@@ -84,29 +84,32 @@
...
-        ``(n_faces, 3, 3)`` and ``faces`` has shape ``(n_faces, 3)`` (int64).
+        ``(mesh, face_vertices, faces)`` where ``mesh`` is the float32 working
+        copy, ``face_vertices`` has shape ``(n_faces, 3, 3)`` and ``faces`` has
+        shape ``(n_faces, 3)`` (int64).
     """
-    faces = mesh_indices.reshape(-1, 3).to(torch.long)
-    mesh = Mesh(points=mesh_vertices, cells=faces)
-    face_vertices = mesh_vertices[faces]  # (n_faces, 3, 3)
-    return mesh, face_vertices, faces
+    faces = mesh.cells.reshape(-1, 3).to(torch.long)
```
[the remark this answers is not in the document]
peterdsharpe: added

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1767/commits/ba71567f24c8e525540758f49571252bd868c330

ktangsali (approved):
Changes to the docs look good.

PR #1778, opened 2026-07-01
Rope opt

pzharrington:
# PhysicsNeMo Pull Request

## Summary

Refactor 2D RoPE so the cos/sin embedding tables are owned once by the top-level
model instead of being rebuilt and stored inside every transformer block.

Previously each `RopeNatten2DSelfAttention` block built and held its own
`rope_cos`/`rope_sin` buffers. Since the tables depend only on
`(latent_hw, head_dim, theta)` — identical across all blocks — a `depth`-block
DiT carried `depth` redundant copies (plus redundant table construction and, under
domain parallelism, redundant per-block sharding). This hoists the tables into a
single shared provider.

## Changes

**`physicsnemo/nn/module/rope.py`**
- Removed the previous `RotaryPositionEmbedding1D` / `RotaryPositionEmbedding2D`
  modules. I'm aware this is technically a breaking change, but they are used nowhere in
  the repo, were introduced last week, and have no known external users. Dropping them
  in favor of the below modules, which have a more clear name.
- Added two table-provider modules that own the buffers and only return tables
  (they hold no projections and apply no rotation):
  - `RotaryEmbedding2DTables` — `rope_cos`/`rope_sin` in `(h, w, head_dim)` layout.
  - `RotaryEmbedding1DTables` — `cos`/`sin` in `(max_seq_len, head_dim)` layout.

**`physicsnemo/nn/module/dit_layers.py`**
- `RopeNatten2DSelfAttention` is now stateless w.r.t. the tables: it no longer
  takes `latent_hw`/`rope_theta` or allocates buffers. Its `forward` requires the
  `rope_cos`/`rope_sin` tables (clear `ValueError` if missing).

**`physicsnemo/models/dit/dit.py`**
- DiT constructs a single `RotaryEmbedding2DTables` provider for the
  `natten2d_rope` backend and injects its tables into every block's `forward`
  (via the existing `attn_kwargs` threading).

**Exports & tests** updated accordingly.

## Rationale / design notes

- **Memory**: `depth` copies of the tables → 1.
- **Checkpoint compatibility**: tables remain `persistent=False` buffers
  (deterministically rebuilt from `(latent_hw, head_dim, theta)`), so `state_dict`
  is unchanged — no migration needed.
- **Domain parallelism**: the provider keeps the `rope_cos`/`rope_sin` names and
  the height-first `(h, w, head_dim)` layout, so the StormCast recipe's
  `shard_dim_selector` shards the single provider buffer along height (dim 0)
  with **no changes to `parallel.py`**. Each rank still gets globally-correct rows
  with no explicit rank offset.
- **`torch.compile`**: tables are fetched once from a stable buffer and passed
  down, keeping the block forward graph constant.

## Behavior change

`RopeNatten2DSelfAttention` (and `DiTBlock(attention_backend="natten2d_rope")`)
can no longer be used standalone — the tables must be supplied at `forward` by
the owning model. The DiT `natten2d_rope` backend does this automatically. Since
standalone usage doesn't make sense (they are blocks) and the
`RopeNatten2DSelfAttention` module was added last week, this change should be fine.

## Checklist

## Review Process

### DIFF
--- a/physicsnemo/nn/module/dit_layers.py
+++ b/physicsnemo/nn/module/dit_layers.py
@@ -580,12 +580,12 @@ class RopeNatten2DSelfAttention(Natten2DSelfAttention):
         Input tensor of shape :math:`(B, L, D)`.
     latent_hw : Tuple[int, int]
         The height and width of the 2D latent space for reshaping.
-    invalid_token_mask : torch.Tensor, optional
-        See :class:`Natten2DSelfAttention`.
     rope_cos, rope_sin : torch.Tensor
         Axial 2D RoPE tables of shape :math:`(h, w, head\_dim)` from a
         :class:`~physicsnemo.nn.module.rope.RotaryEmbedding2DTables` provider.
-        Required; a :class:`ValueError` is raised if not supplied.
+        Required.
+    invalid_token_mask : torch.Tensor, optional
+        See :class:`Natten2DSelfAttention`.
 
     Outputs
     -------
@@ -608,23 +608,15 @@ class RopeNatten2DSelfAttention(Natten2DSelfAttention):
         self,
         x: Float[torch.Tensor, "batch sequence hidden_size"],
         latent_hw: Tuple[int, int],
+        rope_cos: Float[torch.Tensor, "h w head_dim"],
+        rope_sin: Float[torch.Tensor, "h w head_dim"],
         invalid_token_mask: Optional[
             Union[
                 Float[torch.Tensor, " sequence"],
                 Float[torch.Tensor, "batch sequence"],
             ]
         ] = None,
-        rope_cos: Optional[Float[torch.Tensor, "h w head_dim"]] = None,
-        rope_sin: Optional[Float[torch.Tensor, "h w head_dim"]] = None,
     ) -> Float[torch.Tensor, "batch sequence hidden_size"]:
-        if rope_cos is None or rope_sin is None:
-            raise ValueError(
-                "RopeNatten2DSelfAttention requires the rope_cos/rope_sin tables "
-                "to be passed to forward; they are owned by the top-level model's "
-                "RotaryEmbedding2DTables provider (the DiT 'natten2d_rope' backend "
-                "supplies them automatically). None were supplied."
-            )
-
         B, N, C = x.shape
         h, w = int(latent_hw[0]), int(latent_hw[1])
         if not torch.compiler.is_compiling():
--- a/physicsnemo/nn/module/dit_layers.py
+++ b/physicsnemo/nn/module/dit_layers.py
@@ -587,7 +587,7 @@ class RopeNatten2DSelfAttention(Natten2DSelfAttention):
         :class:`~physicsnemo.nn.module.rope.RotaryEmbedding2DTables` provider.
         Required; a :class:`ValueError` is raised if not supplied.
 
-    Returns
+    Outputs
     -------
     torch.Tensor
         Output tensor of shape :math:`(B, L, D)`.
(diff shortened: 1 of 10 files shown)


physicsnemo/nn/module/rope.py
```diff
@@ -369,76 +342,62 @@ def __init__(
...
         self,
-        q: Float[torch.Tensor, "*batch seq head_dim"],
-        k: Float[torch.Tensor, "*batch seq head_dim"],
         latent_hw: Optional[Tuple[int, int]] = None,
     ) -> Tuple[
-        Float[torch.Tensor, "*batch seq head_dim"],
-        Float[torch.Tensor, "*batch seq head_dim"],
+        Float[torch.Tensor, "h w head_dim"],
+        Float[torch.Tensor, "h w head_dim"],
     ]:
```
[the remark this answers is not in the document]
pzharrington: DiT.forward now passes `self.rope(attn_kwargs.get("latent_hw"))` instead of `block_attn_kwargs.get("latent_hw")`. Since `attn_kwargs` is the caller-supplied dict (not merged with the fixed construction-time `attn_kwargs_forward`), it's `None` on the normal path — so `_maybe_rebuild` is bypassed entirely and the provider returns its prebuilt tables with zero comparison. A genuine variable-resolution override still flows through `attn_kwargs` and triggers the rebuild. The blocks still receive the concrete `latent_hw` via `block_attn_kwargs` (from `attn_kwargs_forward`), so their `N == h*w` and table-shape checks are unaffected.

physicsnemo/nn/module/rope.py
```diff
@@ -27,29 +27,38 @@
...
+
+  A provider holds *no* projections and applies *no* rotation itself: its
+  ``forward`` simply returns the ``(cos, sin)`` tables. The intended pattern is
+  that a top-level, multi-block model constructs a *single* provider and passes
+  the returned tables into every attention block's ``forward`` (which rotates
+  Q/K with the functional :func:`apply_rotary_pos_emb`), so the tables are
+  built, stored, and — under domain parallelism — sharded exactly once instead
+  of once per block. See
+  :class:`~physicsnemo.nn.module.dit_layers.RopeNatten2DSelfAttention` and
+  :class:`~physicsnemo.models.dit.DiT` for a reference wiring.
```
CharlelieLrt, line 44: Should we move this entire module docstring to the API reference in `docs/api`? I think it's not part of the online docs if we keep it here

physicsnemo/nn/module/dit_layers.py
```diff
@@ -640,15 +614,30 @@ def forward(
                 Float[torch.Tensor, "batch sequence"],
             ]
         ] = None,
+        rope_cos: Optional[Float[torch.Tensor, "h w head_dim"]] = None,
+        rope_sin: Optional[Float[torch.Tensor, "h w head_dim"]] = None,
```
CharlelieLrt: Why making them optional if it's to raise an error when the tables are not provided?

CharlelieLrt (approved):
Just two minor question/suggestion, otherwise LGTM

physicsnemo/nn/module/rope.py
pzharrington: Hmm yeah I could add a new `rope.rst` under `physicsnemo/docs/api/nn/layers`, with this description and then automodule docs for the two modules + functions. Does that sound good? Arguably these are embeddings so they could go into the existing `embeddings.rst` but then there's no clear place to include the text here. The `rope.rst` page could also allow documenting the exported functions as well as the modules

physicsnemo/nn/module/dit_layers.py
pzharrington: Yeah these can be made required

physicsnemo/nn/module/rope.py
pzharrington: Added

loliverhennigh (approved):
LGTM

ktangsali (approved):
Changes to docs look good!

PR #1779, opened 2026-07-02
Avoid shape-negotiation all_to_all in ShardTensor redistribute when shapes are known (ShardTensor Optimization #1)

negin513:
## Summary

TLDR: avoid an extra `all_to_all` when shapes are know! 2.5x speed-up.

While working on merging HealDA, we realized `redistribute`'s `Shard->Shard` transpose does two `all_to_all` collectives instead of one -- the first just asks other ranks "what shape is your chunk?" before receiving anything....  The overhead from the extra `all_to_all` adds up fast for cases where we reshard repeatedly, such as HealDA where we do two reshards per block. 

In most cases it doesn't need to ask at all: each rank's own chunk size is deterministic, and the  sender's shape is already sitting on `current_spec` from when the tensor was constructed.

This PR calculates the recv-buffer shape locally instead of asking over the network: the target-dim chunk size is deterministic (every rank computes its own the same way), and the sender's current-dim extent is already recorded on `current_spec` from when the tensor was constructed. It only falls back to the original negotiation when that info genuinely isn't available.
  
  
## Why the staleness check is added?
`current_spec` is fixed for the whole (possibly multi-hop) `redistribute` call, but a 2D mesh can chain multiple transform steps in one call (e.g. `Shard/Shard -> Shard/Replicate` gathers one mesh dim, then transposes the other). An earlier hop can already have reshaped `local_tensor` by the time a later hop runs, making the recorded shapes describe a state that no longer exists. Added a check that `local_tensor.shape` still matches what `current_spec` has on record before trusting the shortcut -- caught by `test_shard_tensor_redistribute2d`'s `S3+R` case, which failed with a shape mismatch before this guard was added.

## Compatibility

  No caller changes needed -- this helps existing code as-is, including `sharding_shapes="infer"` (its default mode already records the needed shapes eagerly).

## Performance

Isolated the redistribute primitive (4xGB200, T=16, X=12288, C=1536, bf16, even sharding, caller already passing known `sharding_shapes` so this measures only this fix's effect):

| | before | after |
|---|---|---|
| fwd | 1.849 ms/iter | 0.763 ms/iter |
| fwd+bwd | 5.131 ms/iter | 3.131 ms/iter |

## Test plan

- `torchrun --nproc_per_node=4 -m pytest test/domain_parallel/test_redistribute.py --multigpu-static -q` -- 11 passed (1D/2D mesh, even/uneven sharding), 11 skipped (dynamic-mode variants)
- `torchrun --nproc_per_node=4 -m pytest test/domain_parallel/ --multigpu-static -q` -- confirmed identical set of 8 pre-existing, unrelated failures with and without this change
- `pre-commit` (ruff check, ruff format, interrogate, license, large-file check) -- all pass on the changed file

Resolves NVIDIA/physicsnemo-roadmap#2761

## Related
- Found while profiling the context-parallel reshard benchmark in #1758
- Independent of #1556 (ShardTensor refactor for `torch.compile` support) -- that PR doesn't touch `_shard_redistribute.py`/`_shard_tensor_spec.py`, so this applies cleanly on top either way
- Complements a follow-up fix (exposing `global_shape` on `from_local` to unlock its existing no-comm `"chunk"` path) -- that addresses a separate `all_gather` in `from_local`; this PR's fast path only activates once callers can avoid `"infer"`, which that follow-up enables through the public API

## Checklist

## Review Process


coreyjadams:
Nice!

physicsnemo/domain_parallel/_shard_redistribute.py
```diff
@@ -229,21 +232,58 @@ def _to_new_shard_dim(
...
+    recv_shapes = None
+    if (
+        current_spec._sharding_shapes is not None
+        and mesh_dim in current_spec._sharding_shapes
+    ):
+        current_shapes_by_rank = current_spec._sharding_shapes[mesh_dim]
+        my_coord = device_mesh.get_coordinate()[mesh_dim]
+        is_fresh = len(current_shapes_by_rank) == mesh_size and torch.Size(
+            current_shapes_by_rank[my_coord]
+        ) == torch.Size(local_tensor.shape)
```
coreyjadams: Is `is_fresh` assured to always evaluate to the same value on ALL ranks?  Are there scenarios where one or more ranks disagres about what path to take?

coreyjadams (changes requested):
This is, I think, the most complicated of the redistribute paths in ShardTensor.  with dynamic shapes along multiple axes, this can get really complicated.  I 1000% see the performance benefit of the fast path here and you are right to push for it, and I agree. What do you think about getting a little more aggressive on the testing for this particular op, and covering more scenarios?  I think our static tests are completing in only a few minutes currently, we have a ton of runtime budget for more tests in multigpu-static and this is a great place to add logic checking.

physicsnemo/domain_parallel/_shard_redistribute.py
negin513: Good catch!! Fixed in 209b1801.

physicsnemo/domain_parallel/_shard_redistribute.py
```diff
@@ -229,21 +232,58 @@ def _to_new_shard_dim(
...
+        if is_fresh:
+            my_target_chunk_size = chunks[my_coord].shape[target_dim]
+            recv_shapes = []
+            for sender_shape in current_shapes_by_rank:
+                if len(sender_shape) != local_tensor.ndim:
+                    recv_shapes = None
+                    break
+                recv_shape = list(sender_shape)
+                recv_shape[target_dim] = my_target_chunk_size
+                recv_shapes.append(recv_shape)
```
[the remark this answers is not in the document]
negin513: fixed in 209b1801.

negin513:
Re-ran the isolated redistribute benchmark against the final branch state (`209b1801`, after the rank-uniform refactor), true main-vs-branch:

| Platform / benchmark | main | PR #1779 | speedup |
|---|---:|---:|---:|
| **4×GB200, fwd** | 1.849 | 0.763 | 2.42× |
| **4×GB200, fwd+bwd** | 5.131 | 3.131 | 1.64× |
| **8×H100, fwd** | 1.138 | 0.415 | **2.74×** |
| **8×H100, fwd+bwd** | 3.538 | 2.608 | **1.36×** |

negin513:
Resolves NVIDIA/physicsnemo-roadmap#2761 and NVIDIA/physicsnemo-roadmap#828 | relevant for HealDA (NVIDIA/physicsnemo-roadmap#2172)

negin513:
Added more tests in 356c4001. @coreyjadams I think this should be ready for re-review.

negin513:
/ok to test [356c400](https://github.com/NVIDIA/physicsnemo/pull/1779/commits/356c4001f39348b133828ae2067bcec0665dc258)

coreyjadams:
Disabling auto merge exclusively so we can refresh the CI label and run multi-gpu CI on this.  Looks good to go, let's make sure it passes the CI here before it merges. :)

coreyjadams:
Once another PR lands, and we update this one, it will refresh and we can queue it.

coreyjadams (approved):

coreyjadams:
/ok to test [f522d04](https://github.com/NVIDIA/physicsnemo/pull/1779/commits/f522d04544fa6753808cb49ccd09c903582b7056)

PR #1781, opened 2026-07-02
Mesh: preserve geometry precision in PyVista conversion

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

`from_pyvista()` and `to_pyvista()` previously converted all geometry to `float32`. That can silently destroy valid `float64` geometry when coordinates have a large offset.

At `1e8`, adjacent `float32` values are eight units apart. A one-unit triangle therefore collapses during conversion:

```text
float64 input                            after forced float32

C ●                                      ● A = B = C
  |\
  | \
A ●──● B

area = 0.5                               area = 0

A = (1e8,     1e8)
B = (1e8 + 1, 1e8)
C = (1e8,     1e8 + 1)
```

This PR:

- Preserves real `float32` and `float64` coordinate dtypes during PyVista import and export.
- Retains PhysicsNeMo's existing conversion to `float32` for integer, reduced-precision, and complex geometry.
- Applies the same policy when importing cell centroids.
- Adds regression coverage for large-offset round trips, cell areas, centroid conversion, and explicit `float32` normalization.

Preserving `float64` can increase memory use and keep downstream geometry calculations in double precision. Callers that intentionally want the previous normalization can use `.to(torch.float32)`.

## Checklist

- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [x] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.
- [ ] If I am implementing a new model or modifying any existing model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation).

## Dependencies

None.

## Review Process

**All PRs are reviewed by the PhysicsNeMo team before merging.**

Depending on which files are changed, GitHub may automatically assign a maintainer for review.

We are also testing AI-based code review tools (e.g., Greptile), which may add automated comments with a confidence score.
This score reflects the AI’s assessment of merge readiness and is **not** a qualitative judgment of your work, nor is
it an indication that the PR will be accepted / rejected.

AI-generated feedback should be reviewed critically for usefulness.
You are not required to respond to every AI comment, but they are intended to help both authors and reviewers.
[...]

PR #1596, opened 2026-04-27
Adds type hints and jaxtyping throughout PhysicsNeMo-Mesh

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Adds and overhauls type hints (including jaxtyping where possible, and casting for type-checker narrowing) and docstrings (aligning to proper Sphinx style, and fixing some non-NumPy-format docstrings) throughout PhysicsNeMo-Mesh.

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [x] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.
- [x] If I am implementing a new model or modifying any existing model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation).

## Review Process

**All PRs are reviewed by the PhysicsNeMo team before merging.**

Depending on which files are changed, GitHub may automatically assign a maintainer for review.

We are also testing AI-based code review tools (e.g., Greptile), which may add automated comments with a confidence score.
This score reflects the AI’s assessment of merge readiness and is **not** a qualitative judgment of your work, nor is
it an indication that the PR will be accepted / rejected.

AI-generated feedback should be reviewed critically for usefulness.
You are not required to respond to every AI comment, but they are intended to help both authors and reviewers.
Please react to Greptile comments with 👍 or 👎 to provide feedback on their accuracy.

PR #1785, opened 2026-07-02
Mesh: honor force_copy throughout PyVista conversion

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

`from_pyvista(..., force_copy=True)` previously copied geometry but still allowed attached point, cell, and global arrays to share storage with the source PyVista object. Mutating the returned `Mesh` could therefore mutate an input whose caller requested independent ownership. The reverse conversion had no matching opt-in ownership guarantee.

This PR:

- Applies `force_copy` to imported point, cell, and global data, including cell-centroid mode.
- Adds keyword-only `to_pyvista(..., force_copy=True)` for independent exported geometry and attached data.
- Copies only buffers that can actually alias; conversion paths that already materialize independent connectivity avoid another full-size copy.
- Adds mutation-based regression coverage for points, connectivity, and attached data in both directions.

## Checklist

## Dependencies

- This PR should merge after #1781. Once #1781 lands, merge `main` into this branch and retain its precision-preserving geometry conversion while applying the copy-ownership behavior here.

## Review Process


peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1785/commits/a1dd03671f8bf4ac2a978a011e99c5371e5d2356

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1785/commits/e039fd5855e41c9a09db1a428cfc472b8f22dfdc

physicsnemo/mesh/io/io_pyvista.py
```diff
@@ -38,13 +38,18 @@
     vtk = OptionalImport("vtk")
 
 
-def _vtk_data_to_tensor_dict(data) -> dict[str, torch.Tensor]:  # noqa: ANN001
+def _vtk_data_to_tensor_dict(
+    data,
+    force_copy: bool = False,  # noqa: ANN001
```
laserkelvin: Confused about the lint ignore here, why can't we type hint `data`?

physicsnemo/mesh/io/io_pyvista.py
```diff
@@ -38,13 +38,18 @@
...
+) -> dict[str, torch.Tensor]:
     """Convert a PyVista/VTK data container to a plain tensor dictionary."""
     tensor_data: dict[str, torch.Tensor] = {}
     for key, value in dict(data).items():
         array = np.asarray(value)
         if not np.issubdtype(array.dtype, np.number) and array.dtype != np.bool_:
             continue
+        if force_copy:
+            array = array.copy()
```
laserkelvin, line 57: I'm going back and forth whether I think it would be a better idea to bypass `numpy` here entirely, and just do the type checks and whatnot in `torch`...

I don't think there would be a performance gain - the only thing I could think of is if you wanted to expand the API to include GPU transfer as well. The PyVista -> numpy -> torch should all be no overhead for the default, non-copy case right?

physicsnemo/mesh/io/io_pyvista.py
```diff
@@ -116,11 +121,10 @@ def from_pyvista(
...
         is lost when ``point_source="cell_centroids"``.
     force_copy : bool
-        If True, copy point and cell arrays so the returned Mesh owns its
-        memory independently of the source PyVista mesh.  When False
-        (default), the returned tensors may share memory with the source
-        for efficiency; mutating the Mesh's ``points`` or ``cells`` could
-        then also modify the PyVista mesh.
+        If True, copy geometry and attached data arrays so the returned Mesh
+        owns its memory independently of the source PyVista mesh. When False
+        (default), returned tensors may share memory with the source for
```
laserkelvin, line 131: This is only a problem if you're doing things on host, right?

laserkelvin (approved):
I'll conditionally approve - the thing I'm mostly thinking about is this is really only a problem when doing compute on host right? Would we consider this undesirable anyway, and users should be moving the data to device?

physicsnemo/mesh/io/io_pyvista.py
```diff
@@ -38,13 +38,18 @@
     vtk = OptionalImport("vtk")
 
 
-def _vtk_data_to_tensor_dict(data) -> dict[str, torch.Tensor]:  # noqa: ANN001
+def _vtk_data_to_tensor_dict(
+    data,
+    force_copy: bool = False,  # noqa: ANN001
+) -> dict[str, torch.Tensor]:
```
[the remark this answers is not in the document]
peterdsharpe: Resolved in 7e00bbe by annotating `data: "pv.DataSetAttributes"` and removing the `noqa` entirely (the file already uses string-form PyVista annotations under `TYPE_CHECKING`). Note the suppression was vestigial either way: this repo's ruff config selects only `E, F, S, I, PERF`, so `ANN001` is not enforced in CI.

test/mesh/io/io_pyvista/test_data_preservation.py
```diff
@@ -153,6 +153,95 @@ def test_mesh_with_explicit_normals(self):
...
+            np.array(
+                [[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [0.0, 1.0, 0.0]],
+                dtype=np.float32,
+            ),
+            np.array([[0, 1, 2]], dtype=np.int64),
+        )
+        mesh.point_data["point_value"] = np.array([1.0, 2.0, 3.0])
+        mesh.cell_data["cell_value"] = np.array([4.0])
+        mesh.field_data["case_value"] = np.array([5.0])
+        return mesh
```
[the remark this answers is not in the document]
peterdsharpe: Added a 3D UnstructuredGrid mutation test in 7e00bbe, using `cell_connectivity` (a live view of VTK's connectivity storage) as the write-through handle. For context on why 1D/3D connectivity was lower-risk: those paths build connectivity via `_to_vtk_cell_array`, whose `np.column_stack(...).ravel()` always materializes independent storage, and the points/data copying is shared code across all manifold dims. The new test locks in that structural guarantee against future refactors.

physicsnemo/mesh/io/io_pyvista.py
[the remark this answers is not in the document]
peterdsharpe: Fixed in 7e00bbe — annotated `data: "pv.DataSetAttributes"` and deleted the `noqa` rather than moving it. One correction for the record: CI could not have failed either way. This repo's ruff config selects only `E, F, S, I, PERF`, so `ANN001` is never enforced; the suppression was vestigial (a leftover from #1596).

test/mesh/io/io_pyvista/test_data_preservation.py
```diff
@@ -153,6 +153,95 @@ def test_mesh_with_explicit_normals(self):
...
+        np.testing.assert_array_equal(
+            pv_mesh.field_data["case_value"], source_global_data
+        )
+
+    def test_to_pyvista_force_copy_cannot_mutate_source_mesh(self):
+        mesh = Mesh(
+            points=torch.tensor([[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [0.0, 1.0, 0.0]]),
+            cells=torch.tensor([[0, 1, 2]]),
+            point_data={"point_value": torch.tensor([1.0, 2.0, 3.0])},
+            cell_data={"cell_value": torch.tensor([4.0])},
```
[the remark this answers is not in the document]
peterdsharpe: This claim is empirically false on PyVista 0.48.4. `PolyData.regular_faces` returns a reshaped numpy **view** of the VTK connectivity array — derived per access, but a view, not a copy — so the write goes through to VTK storage:

```python
pd = pv.PolyData.from_regular_faces(points, faces)
pd.regular_faces[0, 0] = 99
assert pd.regular_faces[0, 0] == 99  # passes — write-through
```

The test therefore has teeth: with `force_copy=False`, `np.shares_memory(mesh.cells.numpy(), pv_mesh.regular_faces)` is `True` and the mutation corrupts `mesh.cells`, so the assertion fails exactly when `force_copy` is broken.

The suggested alternative has it backwards: `pv_mesh.faces` is the handle that does *not* write back — on VTK 9, `GetPolys().GetData()` generates a fresh legacy-format array, and PyVista returns it read-only, so `pv_mesh.faces[1] = 99` raises `ValueError: assignment destination is read-only`.

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1785/commits/7e00bbea54f7fc3f389fc39ff0d9b3fd72f2e3ff

physicsnemo/mesh/io/io_pyvista.py
peterdsharpe: We can, good call! Just type hinted it as `pv.DataSetAttributes`.

physicsnemo/mesh/io/io_pyvista.py
peterdsharpe: So, we can't entirely bypass NumPy as VTK only exposes these arrays as NumPy views. That said, it's still zero-copy end to end (at least, if you stay CPU side).

I don't think expanding the API to include GPU transfer helps us at all - you still need to do the H2D transfer. So you're zero-copy on the CPU side, but the transfer will still take time (and necessarily make a copy, onto the GPU).

Right now the idiom would be `from_pyvista(mesh).to(device)`, which is (IMO) better than `from_pyvista(mesh, device="cuda")` or something

physicsnemo/mesh/io/io_pyvista.py
peterdsharpe: Yes, true. But you might want to do stuff on host (e.g., dataset preprocessing, dataloader workers, I/O, etc. - all typically CPU ops).

The thing that this fixes is the footgun where you do `to_pyvista(mesh)`, then do some VTK operation via PyVista, but this corrupts the source `Mesh` that you're still doing stuff with. The point of this PR is to properly respect `force_copy=True` in such a situation.

(That said, users should mostly be doing stuff on device, yes)

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1785/commits/e88afd7d9820503596d34a465f97b9e1ec128db7

PR #1782, opened 2026-07-02
Mesh: make cache initialization compatible with torch.compile()

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

`Mesh` previously placed its outer cache on the mesh device while leaving nested empty `TensorDict` containers device-unspecified. TensorDict then moved those containers during construction, which `torch.compile(fullgraph=True)` could not capture; basic properties such as cell areas, normals, and centroids failed while tracing.

This PR:

- Constructs nested cache containers directly on the mesh device.
- Uses the same pattern when slicing, padding, translating, or transforming a mesh rebuilds its cache.
- Adds full-graph regression coverage for local cell geometry and a chained cache-rebuilding workflow.

## Checklist

- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.
- [ ] If I am implementing a new model or modifying any existing model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation).

## Review Process

**All PRs are reviewed by the PhysicsNeMo team before merging.**

Depending on which files are changed, GitHub may automatically assign a maintainer for review.

We are also testing AI-based code review tools (e.g., Greptile), which may add automated comments with a confidence score.
This score reflects the AI’s assessment of merge readiness and is **not** a qualitative judgment of your work, nor is
it an indication that the PR will be accepted / rejected.

AI-generated feedback should be reviewed critically for usefulness.
You are not required to respond to every AI comment, but they are intended to help both authors and reviewers.
Please react to Greptile comments with 👍 or 👎 to provide feedback on their accuracy.

PR #1771, opened 2026-07-01
Move warp scoped stream to a new model that uses warp blocking streams.

coreyjadams:
This prevents premature garbage collection and illegal memory access.

The root cause of our issues:

- Warp streams are blocking cuda streams.  Pytorch streams are non-blocking cuda streams.
- Our data loader in physicsnemo (and all our code) uses pytorch stream infrastructure, so non-blocking streams.
- We interop with Warp on the cuda side: we create streams in pytorch, pass the objects to warp, and the cuda objects are not changed.  So when we created a pytorch stream, we pass warp a non-blocking stream when warp expected a blocking stream.  Warp's behavior is 100% explained by it's design with blocking streams in mind.

The issues we see are things like the BVH structures, which create temporary objects on the GPU, would have components go out of scope in python and be garbage collected before the GPU kernel completes because it's async execution.  In pytorch, this is handled by event syncs: you can keep the GPU objects alive until a certain event has occurred, but warp will automatically handle this with the blocking streams.

The solution from Lukasz on the warp team is so simple we're all kicking ourselves for not thinking of it earlier

```python
nonblocking_stream = wp.stream_from_torch(torch.cuda.Stream(device))  # <--- non-blocking stream
warp_stream = wp.Stream(device)  # <--- any Warp stream on the correct device

# do stuff on non-blocking stream
with wp.ScopedStream(nonblocking_stream):
    a = wp.zeros(100_000_000, dtype=float)
    for _ in range(1000):
        wp.launch(inc_kernel, dim=a.shape, inputs=[a])

warp_stream.wait_stream(nonblocking_stream)  # <--- magic
```

This PR codifies this as a new context scope that wraps Warp's stream scope.  Using this, the SDF now works again in the GeoTransolver Datapipe!

I moved ALL our functionals over to this scope in this PR, which is why many files have been touch, since it should be exclusively safer without a performance impact. (Blocking stream != host blocking.  It's a GPU stream block, not a full GPU sync).

# PhysicsNeMo Pull Request

## Checklist

- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
[...]

PR #1786, opened 2026-07-02
Mesh: differentiable dense displacement and sparse control-point morphing

mehdiataei:
# PhysicsNeMo Pull Request

## Description

Adds differentiable mesh morphing to physicsnemo.nn.functional and the mesh API:

* displace_points / Mesh.displace: dense per-point displacement with an optional scalar amount and per-point bool/float weights.
* morph_points / Mesh.morph / DomainMesh.morph: sparse control-point morphing via a compactly supported Wendland-C2 Shepard field with a stationary background. DomainMesh.morph evaluates one field over the interior and all boundaries so coincident points stay coincident.

Both ops support float32/float64, unbatched and batched inputs, and first-order gradients through all tensor inputs (points, controls, displacements, radii, amount, weights). Auto dispatch uses Torch on CPU and Warp on CUDA. The Warp dense-displacement backend is opt-in only (Torch is default everywhere) and exists for backend parity and benchmarking.

Tests cover analytic derivatives, gradcheck on both backends, opcheck, CUDA-graph capture, and torch.compile fullgraph. Includes API docs, a tutorial section, and benchmark registry entries.

Checklist
## Checklist

## Review Process


physicsnemo/nn/functional/geometry/morphing/_utils.py
```diff
@@ -0,0 +1,287 @@
...
+                "a scalar or shape (C,)"
+                if was_unbatched
+                else (
+                    "a scalar, shape (C,), or aligned shape "
+                    f"(B, C)={(batch_size, num_controls)}"
+                )
+            )
+            raise ValueError(f"radius must be {expected}, got {tuple(radius.shape)}")
+    elif isinstance(radius, Real) and not isinstance(radius, bool):
+        if not torch.compiler.is_compiling() and num_controls > 0:
```
peterdsharpe: **Compiled models silently skip this validation.** Both this check and the `amount` check at L94 are gated on `not torch.compiler.is_compiling()`, so the identical call that raises `ValueError` in eager silently proceeds under `torch.compile`:

- `radius=0.0` → near-identity morph (only exactly-coincident points move)
- negative radius → morphs as `+|r|` (components are divided by the signed radius and only the norm enters `q`)
- `amount=float("nan")` → NaN geometry everywhere

Verified on torch 2.12.1: eager raises `ValueError("radius must be strictly positive")` while the compiled call returns the input points unchanged. A compiled training loop proceeds silently on corrupted meshes where eager fails fast.

I understand the guard exists so `fullgraph=True, dynamic=True` tracing doesn't specialize/graph-break on runtime scalars (`test_torch_compile_fullgraph_dynamic_python_scalars_and_defaults`), but that only requires skipping validation for *symbolic* scalars — a plain Python constant can still be checked at trace time (e.g. skip only for `torch.SymFloat`/`torch.SymInt` inputs, or run the checks under `torch._dynamo` constant-folding).

Failing regression test: `test_torch_compile_rejects_invalid_python_scalars_like_eager`.

physicsnemo/mesh/domain_mesh.py
```diff
@@ -597,6 +597,212 @@ def transform(
...
+
+        interior = output_meshes[0]
+        boundaries = {
+            name: output_meshes[index]
+            for index, name in enumerate(self.boundaries.keys(), start=1)
+        }
+        return DomainMesh(
+            interior=interior,
+            boundaries=boundaries,
+            global_data=self.global_data,
```
peterdsharpe: `morph` is the only `DomainMesh` operation that returns the source's `global_data` without cloning — every other operation delegates to `apply_to_meshes`, which always does `self.global_data.clone()` (L232), and the docstring above promises "unchanged domain data". As written, mutating the result (`out.global_data["step"] = ...`, or in-place edits of an existing entry) silently corrupts the source domain:

```python
domain.morph(...).global_data is domain.global_data  # True (translate: False)
```

Fix is one word: `global_data=self.global_data.clone()`.

Failing regression test: `test_domain_morph_clones_domain_global_data`.

physicsnemo/nn/functional/geometry/deform/_torch_impl.py
```diff
@@ -0,0 +1,505 @@
...
+    normal_radius = torch.where(radii < finfo.tiny, torch.ones_like(radii), radii)
+    ratio_scale = torch.where(
+        radii < finfo.tiny,
+        torch.full_like(radii, finfo.max),
+        normal_radius.reciprocal(),
+    ).detach()
+    safe_difference = torch.where(
+        finite_difference, difference, torch.zeros_like(difference)
+    )
+    normalized = (safe_difference * ratio_scale) / (radii * ratio_scale)
```
peterdsharpe, line 154: **A tensor radius containing `+inf` returns opposite geometry on the two backends.** Here `ratio_scale = 1/inf = 0` (detached), so `normalized = (diff * 0) / (inf * 0) = 0/NaN = NaN`, and the `nan_to_num(nan=1.0)` below maps it to the support boundary → the control has **zero** influence. The Warp kernel (`_normalized_component_*` in `_warp_impl/kernels.py`) computes `delta / inf = 0` → `q = 0` → every point is treated as coincident and receives the **full** control displacement.

Verified on CPU (torch 2.12.1 / warp 1.14.0): points `[[0.5, 0], [3, 4]]`, one control at the origin with displacement `[0, 1]`, `radius=torch.tensor([inf])` → torch backend: no displacement anywhere; warp backend: full displacement everywhere. Since auto-dispatch selects torch on CPU and warp on CUDA, the same model produces different meshes depending on device, with no error or warning.

Warp's answer is also the mathematically correct limit: the torch backend itself already returns full displacement at `radius=1e300` and then jumps to zero influence exactly at `inf`. I realize tensor radii are documented as "not validated at runtime" (avoiding a device sync is sensible), so I don't expect an error here — but the backends should agree on the limit rather than silently diverging.

Failing regression test: `test_infinite_tensor_radius_agrees_across_backends`.

physicsnemo/mesh/transformations/geometric.py
```diff
@@ -816,3 +819,210 @@ def scale(
...
+    """Resolve a raw tensor or nested ``point_data`` key."""
+    if isinstance(value, torch.Tensor):
+        return value
+    if not isinstance(value, (str, tuple)):
+        raise TypeError(
+            f"{argument_name} must be a tensor or point_data key/path, got "
+            f"{type(value).__name__}"
+        )
+    try:
+        return mesh.point_data[value]
```
peterdsharpe: Two gaps in this resolution path (both also affect the inlined copy in `DomainMesh.morph`):

1. **No tensor check on the resolved value.** A key can resolve to a nested sub-TensorDict; a homogeneous-float sub-TD with `batch_size=[n_points]` passes every check in `_normalize_weights` (`.shape`, `.device`, `.dtype` all look right) and `mesh.displace(disp, weights=key)` then **silently returns a Mesh whose `.points` is a TensorDict**, not a tensor (verified — no exception at all). `DomainMesh.morph` guards exactly this with `isinstance(component_weights, torch.Tensor)` (domain_mesh.py L708); the same guard belongs here.
   Failing test: `test_mesh_displace_rejects_weights_resolving_to_tensordict`.

2. **`except KeyError` misses tensordict's other failure modes.** A tuple path that descends *into a leaf tensor* — e.g. `mesh.displace(("temperature", "x"))` where `temperature` is a leaf — raises `ValueError` or `AttributeError: 'Tensor' object has no attribute '_get_tuple'` from tensordict internals (varies by entry point), bypassing the actionable `"Available keys: ..."` diagnostic this code deliberately constructs for exactly this kind of mistake.
   Failing test: `test_mesh_point_data_path_through_leaf_has_actionable_diagnostic`.

physicsnemo/mesh/transformations/geometric.py
```diff
@@ -816,3 +819,210 @@ def scale(
...
+        amount=amount,
+        weights=weights_t,
+        implementation=implementation,
+    )
+    return _mesh_with_morphed_points(mesh, points)
+
+
+def morph(
+    mesh: "Mesh",
+    control_points: torch.Tensor,
```
peterdsharpe: `mesh.morph("handles", ...)` currently dies with `AttributeError: 'str' object has no attribute 'ndim'` deep inside `_utils._validate_points`. Since the sibling `displace()` *does* accept point-data keys for its main argument, a key string here is a very natural user mistake, and `DomainMesh.morph` already raises a descriptive `TypeError` for exactly this case (domain_mesh.py L664-668). Suggest the same explicit `isinstance` guard here for `control_points`/`control_displacements`.

Failing regression test: `test_mesh_morph_rejects_non_tensor_control_points`.

physicsnemo/nn/functional/geometry/deform/deform.py
```diff
@@ -0,0 +1,729 @@
...
+        cls._check_impl(implementation, impls)
+        if implementation is not None:
+            impl = impls[implementation]
+            if not impl.available:
+                raise ImportError(
+                    f"Implementation '{implementation}' is not available "
+                    f"for {cls.__name__}"
+                )
+        else:
+            warp_impl = impls.get("warp")
```
peterdsharpe, line 477: When auto-dispatch sees CUDA points but warp is unavailable (not installed, or older than the pin), this silently falls back to the Torch path. `FunctionSpec.dispatch` emits a one-time `RuntimeWarning` via `_warn_fallback` (function_spec.py L487) in exactly this situation for every other functional; because this override bypasses it, a CUDA user without a working warp gets the much slower O(N·C) eager path and an unexplained order-of-magnitude slowdown with no diagnostic. Suggest calling `cls._warn_fallback(warp_impl, impl)` when the CUDA→torch fallback triggers.

Relatedly, L503-511 duplicate the explicit-implementation branch of `FunctionSpec.dispatch` verbatim — if the base-class branch changes (e.g. error-message format), this copy drifts.

physicsnemo/nn/functional/geometry/deform/_warp_impl/kernels.py
```diff
@@ -0,0 +1,996 @@
...
+        return wp.float32(0.0)
+    norm_squared = wp.float32(0.0)
+    for d in range(n_dims):
+        value = _normalized_component_f32(points[b, i, d], controls[b, j, d], radius)
+        scaled = value / maximum
+        norm_squared = norm_squared + scaled * scaled
+    return maximum * wp.sqrt(norm_squared)
+
+
+@wp.func
```
peterdsharpe, line 30: ~460 of the ~1000 lines in this file are verbatim f32/f64 twins differing only in dtype literals (7 kernel/helper pairs: `_normalized_component`, `_normalized_distance`, `displace_forward/backward`, the masked variants, `shepard_forward/backward/point_backward`). Warp ≥ 1.14 (the repo pin) supports generic kernels via `typing.Any` + `wp.overload`, and closure-based kernel factories parameterized on dtype are also idiomatic. Either would halve the file and remove the failure mode where a numerical fix lands in only one precision — reviewers can't realistically diff ~230-line blocks by eye to catch drift.

physicsnemo/nn/functional/geometry/morphing/_warp_impl/op.py
```diff
@@ -0,0 +1,1042 @@
...
+)
+def compact_shepard_field_warp_forward_only_impl(
+    points: torch.Tensor,
+    controls: torch.Tensor,
+    control_displacements: torch.Tensor,
+    radii: torch.Tensor,
+) -> torch.Tensor:
+    """Evaluate only the field, omitting every backward-only auxiliary."""
+
+    _check_common_dtype(points, controls, control_displacements, radii)
```
peterdsharpe: This validation block (and the launch sequence below) duplicates `compact_shepard_field_warp_impl` (L478-488) line for line. The shepard kernels already take `save_auxiliaries` as an int parameter, so a single custom op with a `save_auxiliaries` flag returning zero-size aux tensors in the no-grad case — or at minimum a shared validation/launch helper — would remove this second ~65-line op and the risk of the inference and training paths accepting different inputs after a one-sided edit.

physicsnemo/nn/functional/geometry/morphing/_warp_impl/op.py
```diff
@@ -0,0 +1,1042 @@
...
+    amount_c = amount.contiguous()
+    weights_c = weights.contiguous() if weights is not None else None
+    output = torch.empty_like(points_c)
+    if output.numel() == 0:
+        return output
+
+    wp_dtype = _wp_dtype(points.dtype)
+    is_mask = weights_c is not None and weights_c.dtype == torch.bool
+    if is_mask:
+        kernel = (
```
peterdsharpe: This `X_f32 if points.dtype == torch.float32 else X_f64` selection repeats 8 times across the file. A module-level dict keyed by dtype (or the generic-kernel refactor suggested in kernels.py) reduces it to one lookup and removes the copy-paste risk of pairing an f32 kernel with the f64 branch — which Python happily accepts and only the Warp launch (or a silent memory reinterpretation via the ctype view) would catch.

physicsnemo/mesh/domain_mesh.py
```diff
@@ -597,6 +597,212 @@ def transform(
...
+                    f"{component.points.device} and {control_points.device}"
+                )
+            if component.points.dtype != control_points.dtype:
+                raise TypeError(
+                    f"{label} and control_points must have the same dtype, got "
+                    f"{component.points.dtype} and {control_points.dtype}"
+                )
+            if weights is not None:
+                try:
+                    component_weights = component.point_data[weights]
```
peterdsharpe: Third copy of the resolve-key-with-"Available keys"-diagnostic idiom: `_resolve_point_field` (transformations/geometric.py, added in this PR) and `_resolve_field` (calculus/integration.py) already implement it. `morph` already imports from `transformations.geometric` below; extending `_resolve_point_field` with an optional owner-label argument would keep diagnostics consistent across `Mesh.morph`/`DomainMesh.morph` — and note this copy shares the `except KeyError` gap flagged on `_resolve_point_field` (a tuple path through a leaf tensor raises a raw tensordict error here too).

physicsnemo/nn/functional/geometry/morphing/_torch_impl.py
```diff
@@ -0,0 +1,505 @@
...
+    amount: torch.Tensor,
+    weights: torch.Tensor | None,
+) -> torch.Tensor:
+    """Morph normalized rank-3 points with a compact Shepard field."""
+
+    if points.shape[0] == 0 or points.shape[1] == 0 or control_points.shape[1] == 0:
+        # The point identity supplies its gradient directly. Connect every other
+        # differentiable input through one scalar zero instead of constructing a
+        # full zero field and running dense displacement over it.
+        zero = (
```
peterdsharpe: Third variation of the "connect every differentiable input through a scalar zero" idiom (also `compact_shepard_field_torch` L388-389, and `morph_points_warp` in `_warp_impl/op.py` ~L993-1004). The three copies already differ subtly (the field version omits `amount`/`weights`; the warp version multiplies the sum by 0, this one multiplies per-term). A small shared helper in `_utils.py` would keep the empty-input autograd semantics identical across backends by construction.

physicsnemo/nn/functional/geometry/morphing/morphing.py
```diff
@@ -0,0 +1,729 @@
...
+
+        torch.testing.assert_close(
+            output,
+            reference,
+            atol=cls._COMPARE_BACKWARD_ATOL,
+            rtol=cls._COMPARE_BACKWARD_RTOL,
+        )
+
+
+def displace_points(
```
peterdsharpe: Nit: these hand-written wrappers plus the manual `__doc__` assignments below re-implement `FunctionSpec.make_function`, which already forwards to `dispatch` (including subclass overrides like `MorphPoints.dispatch`) and copies the docstring — it's what `rfft`, `radius_search`, `ray_mesh_intersect`, etc. use. I can see the draw of a real typed signature including the `implementation` kwarg for docs/IDE support; if that's the motivation, consider teaching `make_function` to synthesize it so every functional benefits and these two bespoke wrappers disappear.

peterdsharpe:
Review in-progress! Found a few edge-case bugs; added failing tests to demonstrate. (Will follow up shortly with comments)

physicsnemo/nn/functional/geometry/morphing/_torch_impl.py
```diff
@@ -0,0 +1,505 @@
...
+        local_q_squared = torch.gather(
+            q_squared, dim=-1, index=local_index.unsqueeze(-1)
+        ).squeeze(-1)
+        minimum_index = torch.where(
+            update_minimum, local_index + control_start, minimum_index
+        )
+        reference_q_squared = torch.where(
+            update_minimum, local_q_squared, reference_q_squared
+        )
+        minimum_q = torch.minimum(minimum_q, local_minimum)
```
peterdsharpe: **Chunked backward returns wrong geometry gradients when the minimum normalized distance is exactly tied across control chunks.** The reference selection above uses strict `<` (L246), so on a tie the first chunk's control keeps `minimum_index`/`reference_q_squared` — but `torch.minimum` here splits the backward gradient 50/50 across tied inputs. Half of the `reference_q` gradient then flows to a control that the compensation algebra below (`background`, `common_log_scale`) assumes has no reference role, breaking its exact cancellation.

Repro (float64, radius 1; query point `(0,0)` exactly equidistant from mirrored controls `(±0.3, 0.4)` forced into different control chunks):

|            | forward | ∂/∂points | ∂/∂controls | ∂/∂radius | ∂/∂disps |
|------------|---------|-----------|-------------|-----------|----------|
| unchunked  | `[0.3, 0.3]` | `[0.0, 2.048]` | ✓ | ✓ | ✓ |
| central FD | — | `[0.0, 2.048]` | — | — | — |
| warp       | agrees | agrees | agrees | agrees | agrees |
| **chunked** | `[0.3, 0.3]` ✓ | **`[-0.192, 2.048]`** | ✗ | ✗ | ✓ |

Control chunking activates once `n_controls` exceeds `pair_budget // query_chunk` — ≈836 (float32, B=1, D=3), ≈591 (float64), down to ≈209 (batched float64) — so dense control lattices (FFD-style grids) and batched training hit it, and exact ties are routine for mirror-symmetric meshes with symmetric control layouts. Everything is silent: forward values and displacement gradients match.

The fix appears to be one line — keep the differentiable minimum consistent with the strict-`<` selection:

```suggestion
        minimum_q = torch.where(update_minimum, local_minimum, minimum_q)
```

Verified: this makes the chunked path match unchunked/FD/warp gradients exactly on the tied case, and is bit-identical on non-tied random chunked inputs. Failing regression test: `test_torch_control_chunked_gradients_match_unchunked_on_ties`.

test/nn/functional/geometry/test_point_deform.py
```diff
@@ -0,0 +1,1437 @@
...
+    assert torch.autograd.gradcheck(
+        operation,
+        (points, controls, displacements, radius, amount, weights),
+        eps=1e-6,
+        atol=2e-5,
+        rtol=2e-4,
+    )
+
+
+def test_torch_chunk_checkpoint_bounds_retained_autograd_storage(monkeypatch):
```
peterdsharpe, line 747: This test can't detect chunked-vs-unchunked gradient divergence: both `run_and_measure()` calls run under the same 32 KiB budget (identical chunk partition) and only the checkpoint wrapper is toggled, so both gradient sets traverse the same cross-chunk reduction. With `torch.randn` inputs, exact distance ties also have probability zero — and the parity/gradcheck tests elsewhere use ≤64 controls under the default 256 MiB budget, so they never enter the multi-control-chunk path at all. The cross-chunk tie bug flagged at `_torch_impl.py` L256 passes this entire suite.

Added `test_torch_control_chunked_gradients_match_unchunked_on_ties` to close the gap: chunking is a byte-budget implementation detail and should be gradient-invariant, ties included.

physicsnemo/mesh/transformations/geometric.py
```diff
@@ -816,3 +819,210 @@ def scale(
...
+        ) from None
+
+
+def _mesh_with_morphed_points(mesh: "Mesh", points: torch.Tensor) -> "Mesh":
+    """Construct a geometry-invalidated mesh while retaining topology caches."""
+    from physicsnemo.mesh.mesh import Mesh
+
+    cache = TensorDict(
+        {
+            "cell": TensorDict({}, batch_size=[mesh.n_cells]),
```
peterdsharpe: Now that the branch has merged #1782 ("initialize cache tensors on device"), this helper is the last remaining instance of the pattern that PR fixed: `transform`/`translate` now pass `device=` to the inner `"cell"`/`"point"` TensorDicts and to the `topology` fallback, while this copy still creates them device-less and relies on the outer `device=points.device`. Worth mirroring the merged fix here (and it's a third copy of the same cache-construction block — a shared helper would keep them from drifting again).

peterdsharpe (commented):
Follow-up to the review above with the chunked-backward analysis (the earlier review went out without its summary, so it's included here).

**Summary of the full review** — 15 inline comments: 6 correctness items, 1 dispatch-diagnostics item, 8 maintainability items. I pushed two commits (ce5463b9, b9d6d5b4) adding **failing regression tests** that pin down the confirmed correctness items; CI is expected to be red on exactly these seven until they're addressed:

- `test_torch_control_chunked_gradients_match_unchunked_on_ties` — cross-chunk tie gradient corruption (comment below; one-line fix suggested)
- `test_torch_compile_rejects_invalid_python_scalars_like_eager`
- `test_infinite_tensor_radius_agrees_across_backends`
- `test_mesh_point_data_path_through_leaf_has_actionable_diagnostic`
- `test_mesh_displace_rejects_weights_resolving_to_tensordict`
- `test_mesh_morph_rejects_non_tensor_control_points`
- `test_domain_morph_clones_domain_global_data`

Overall this is a very carefully engineered piece of numerics — the overflow-safe normalized distances and the detached-reference compensation algebra are impressive. The correctness items are all in the seams (chunk boundaries, compile-vs-eager, backend parity, API-boundary validation) rather than the core math.

coreyjadams:
Heads up @mehdiataei : there is a change coming to how we wrap warp streams in physicsnemo, seen here: https://github.com/NVIDIA/physicsnemo/pull/1771

tl;dr - we will have a small custom wrapper to replace ScopedStream.  If your pr merges before #1771 we may want to patch it after.  Want to make sure you're aware.  You can read the summary in #1771 to understand the motivation but it's stream safety issues.

physicsnemo/nn/functional/__init__.py
```diff
@@ -31,9 +31,13 @@
 )
 from .fourier_spectral import imag, irfft, irfft2, real, rfft, rfft2, view_as_complex
 from .geometry import (
+    DisplacePoints,
+    MorphPoints,
+    displace_points,
```
peterdsharpe: Until this PR, the top-level `physicsnemo.nn.functional` namespace contained only snake_case functions — every CamelCase FunctionSpec class (`KNN`, `RadiusSearch`, `RFFT`, `RayMeshIntersect`, `FarthestPointSampling`, ...) is exported one level down in its subpackage `__init__.py`, and the docs autodoc only the functions. Hoisting `DisplacePoints`/`MorphPoints` to the top level breaks that convention and permanently widens the top-level surface; they'd remain importable from `physicsnemo.nn.functional.geometry`, exactly like their siblings.

```suggestion
    displace_points,
```

physicsnemo/nn/functional/__init__.py
```diff
@@ -57,6 +61,9 @@
 )
 
 __all__ = [
+    "DisplacePoints",
+    "MorphPoints",
+    "displace_points",
```
peterdsharpe: Same as above — no other spec class appears in this `__all__`:

```suggestion
    "displace_points",
```

physicsnemo/mesh/transformations/__init__.py
```diff
@@ -16,5 +16,10 @@
...
 """Geometric transformations for simplicial meshes.
 
-This module provides linear and affine transformations with intelligent cache handling.
+This module provides linear, affine, and nonlinear point transformations with
+intelligent cache handling.
 """
+
+from .geometric import displace, morph, rotate, scale, transform, translate
+
+__all__ = ["displace", "morph", "rotate", "scale", "transform", "translate"]
```
peterdsharpe: Giving this package a real `__init__` matches every sibling subpackage (`calculus`, `subdivision`, `repair`, `sampling`, ... all re-export) — `transformations` was the lone holdout, so 👍. Two curation points while minting the new public import path: `rotation_matrix` and `scale_matrix` are public, documented helpers in `geometric.py` and are picked up by the rst `automodule` — either include them or exclude them deliberately; and every sibling uses absolute imports in `__init__.py` rather than relative.

```suggestion
from physicsnemo.mesh.transformations.geometric import (
    displace,
    morph,
    rotate,
    rotation_matrix,
    scale,
    scale_matrix,
    transform,
    translate,
)

__all__ = [
    "displace",
    "morph",
    "rotate",
    "rotation_matrix",
    "scale",
    "scale_matrix",
    "transform",
    "translate",
]
```

physicsnemo/nn/functional/geometry/morphing/morphing.py
```diff
@@ -0,0 +1,729 @@
...
+    )
+
+
+def morph_points(
+    points: torch.Tensor,
+    control_points: torch.Tensor,
+    control_displacements: torch.Tensor,
+    *,
+    radius: float | torch.Tensor,
+    amount: float | torch.Tensor = 1.0,
```
peterdsharpe: **Recommend dropping `amount` from the public API entirely.** Both operations are exactly linear in the displacements, so it adds no expressive power: `morph_points(..., amount=a)` ≡ `morph_points(..., control_displacements=a * d)` (the blend, the exact-coincidence mean, and `weights` all commute with the scalar), and likewise for `displace_points`. Identical gradients for a learned scalar are available via the multiplication.

What it costs to keep: a fifth of the signature on five public surfaces (`displace_points`, `morph_points`, `Mesh.displace`, `Mesh.morph`, `DomainMesh.morph`), the `_normalize_amount` validation path — one of the two scalar-validation paths implicated in the compile-vs-eager divergence flagged earlier — plus its share of docs, benchmarks, and backward-kernel plumbing (the Warp `amount` gradient even needs its own nondeterministic-reduction tolerance carve-out in `compare_backward`). As a naming footnote if it survives: the repo's existing word for a scalar multiplier is `factor` (`scale(mesh, factor)`); `amount` has no precedent as a parameter name.

physicsnemo/nn/functional/geometry/morphing/morphing.py
```diff
@@ -0,0 +1,729 @@
...
+    weights: torch.Tensor | None = None,
```
peterdsharpe: Naming: in a signature whose other tensor arguments are `control_points`/`control_displacements`, `weights` naturally reads as *per-control* (RBF-style) weights — and the API knows it: three separate docstrings have to disclaim "weights are not per-control values" (here, `geometric.py`, `mesh.py`). When the docs must repeatedly disclaim the natural reading of a name, the name is fighting the reader. `point_weights` (or `query_weights`) self-documents, matches the per-point meaning used elsewhere in the repo (e.g. domino utils), and the rename is a free, mechanical change today — a deprecation cycle after merge.

physicsnemo/nn/functional/geometry/morphing/morphing.py
```diff
@@ -0,0 +1,729 @@
...
+        ("large-b8-n32768-d3-float-weights", 8, 32768, 3, "float"),
+    )
+    _BACKWARD_BENCHMARK_CASES = (
+        ("medium-n16384-d3-displacement-only", 1, 16384, 3, "displacement"),
+        ("medium-n16384-d3-all-gradients", 1, 16384, 3, "all"),
+    )
+    _COMPARE_ATOL = 1.0e-6
+    _COMPARE_RTOL = 1.0e-6
+
+    @FunctionSpec.register(name="warp", required_imports=("warp>=0.6.0",), rank=1)
```
peterdsharpe: **Recommend dropping the Warp backend for `DisplacePoints`.** By this registration's own ranks (torch is `rank=0, baseline=True`), auto-dispatch never selects it on any device — it's reachable only via an explicit `implementation="warp"`. For a memory-bound fused axpy, eager Torch is already near bandwidth and `torch.compile` fuses the intermediates, so there's no workload where a user should ask for it. Meanwhile it carries real permanent surface: the `displace_forward/backward` (+ masked) kernels in both dtypes, their custom ops, fake-tensor registrations, benchmarks, and parity tests. Keeping `DisplacePoints` torch-only loses nothing today and can always be revisited with a motivating benchmark; shipping it makes it public contract.

physicsnemo/nn/functional/geometry/morphing/morphing.py
```diff
@@ -0,0 +1,729 @@
...
+    return DisplacePoints.dispatch(
+        points,
+        displacement,
+        amount=amount,
+        weights=weights,
+        implementation=implementation,
+    )
+
+
+def morph_points(
```
peterdsharpe: **Request: add a keyword-only `kernel: Literal["wendland_c2"] = "wendland_c2"` parameter** (validated, single accepted value for now) to `morph_points` and the mesh-level `morph`s. Rationale: `morph`/`morph_points` claim the most generic possible name for one specific algorithm — the Wendland-C2 compact-Shepard semantics are already contractual in the docstrings, so let's make the choice explicit at call sites. It reserves the extension point so FFD lattices or TPS-RBF morphing can land later as new kernel values instead of awkward sibling functions, and lets callers pin `kernel="wendland_c2"` against any future default drift. Cost today is one validated literal; retrofitting the parameter after other kernels exist is messier.

physicsnemo/mesh/mesh.py
```diff
@@ -2651,6 +2653,129 @@ def translate(
...
+            New mesh with displaced points, unchanged connectivity and fields.
+
+        Notes
+        -----
+        Attached fields are retained as Lagrangian data and are not pushed
+        forward. Geometry caches are invalidated and topology caches are
+        retained. Displacement does not automatically detect inverted,
+        degenerate, or self-intersecting cells. Use :meth:`validate` explicitly
+        if required.
+        """
```
peterdsharpe: Repo convention for `Mesh` transform methods is a short stub — `Mesh.translate`/`rotate`/`scale`/`transform` all say "Convenience wrapper for `physicsnemo.mesh.transformations.<name>()`" and leave the full documentation on the autodoc'd free function. This method (and `morph` below) instead carries a fourth full copy of the operation's documentation (spec class, free function, `Mesh` method, `DomainMesh` method), and the copies are already drifting: "Extrapolation is allowed" exists only in the mesh.py copies, and the tensor-validation caveat is worded three different ways across them.

```suggestion
        r"""Displace points by a dense vector field without changing topology.

        Convenience wrapper for
        :func:`physicsnemo.mesh.transformations.displace`, which documents all
        parameters and numerical behavior.

        Returns
        -------
        Mesh
            New mesh with displaced points, unchanged connectivity and fields.
        """
```

peterdsharpe (commented):
Second batch: API/design review (the first batch covered correctness). The implementation quality here is genuinely high — strict validation with actionable errors, `radius` required with no scale-presuming default, displacements-not-destinations stated everywhere, and the raw Shepard field evaluator deliberately kept private (easy to add later, impossible to remove). The comments below are all about the public contract, since that's the part we maintain forever; five of them carry GitHub suggestions that can be applied directly.

Summary of asks, roughly by weight:
1. Keep `DisplacePoints`/`MorphPoints` out of the top-level `nn.functional` namespace (subpackage export matches every other spec class).
2. Drop `amount` (linearity makes it strictly redundant) and rename `weights` → `point_weights`.
3. Stub the `Mesh.displace`/`Mesh.morph` method docstrings per the `translate` convention — the four full copies are already drifting.
4. Drop the Warp backend for `DisplacePoints` — auto-dispatch never selects it, and a fused axpy doesn't earn ~8 kernels of permanent contract.
5. Curate the new `transformations` package exports deliberately (include or exclude `rotation_matrix`/`scale_matrix`; absolute imports like the siblings).
6. Add `kernel="wendland_c2"` to the morph APIs to name the algorithm at call sites and reserve the extension point.

One open question rather than an ask: `DomainMesh.morph` exists but `DomainMesh.displace` doesn't — the weights-as-common-key mechanism would transfer directly, so if the asymmetry is intentional (dense per-component displacement doesn't share a world-space field), a line in the `DomainMesh` docs saying so would preempt the feature request.

physicsnemo/mesh/mesh.py
```diff
@@ -2651,6 +2653,129 @@ def translate(
...
+
+        Notes
+        -----
+        Attached fields are retained as Lagrangian data and are not pushed
+        forward. Geometry caches are invalidated and topology caches are
+        retained. Parameterize learned radii to remain positive, for example as
+        ``torch.nn.functional.softplus(raw_radius) + eps``. Morphing does not
+        automatically detect inverted, degenerate, or self-intersecting cells.
+        Use :meth:`validate` explicitly if required.
+        """
```
peterdsharpe: Same as `displace` above — keep the full documentation in one canonical place (the free function, which the docs autodoc) and stub the method:

```suggestion
        r"""Morph points from sparse compactly-supported control handles.

        Convenience wrapper for
        :func:`physicsnemo.mesh.transformations.morph`, which documents all
        parameters and numerical behavior.

        Returns
        -------
        Mesh
            New mesh with morphed points, unchanged connectivity and fields.
        """
```

peterdsharpe (changes requested):

docs/api/mesh/transformations.rst
```diff
@@ -6,10 +6,10 @@ Geometric Transformations
...
 .. currentmodule:: physicsnemo.mesh.transformations.geometric
 
-Linear and affine transformations on mesh geometry. Each function returns a new
-:class:`~physicsnemo.mesh.mesh.Mesh` with transformed point coordinates and
-appropriately invalidated caches. (Any cached quantities, such as normals and
-areas, are automatically recomputed on next access.)
+Linear, affine, and nonlinear transformations on mesh geometry. Each function
+returns a new :class:`~physicsnemo.mesh.mesh.Mesh` with transformed point
+coordinates and appropriately invalidated caches. (Any cached quantities, such
+as normals and areas, are automatically recomputed on next access.)
```
megnvidia: when the whole sentence is in parens, I usually just delete the parens. Your choice here.

docs/api/mesh/transformations.rst
```diff
@@ -32,6 +32,177 @@ All transformations are also available as methods on
...
+        control_points,
+        control_displacements,
+        radius=1.0,
+    )
+
+    # Autograd continues through the returned point coordinates.
+    objective = single_morph.points.square().mean()
+    objective.backward()
+
+Because ``control_displacements`` is differentiable, a PyTorch optimizer—or a
```
megnvidia: the dashes here make this look like it was written by AI. I would ask your AI to suggest a simpler sentence construction. 

I mean the point is not that hard to get .....but the tech writer in me wants fewer commas, parens, and dashes in sentences. 

again, your choice here

physicsnemo/nn/functional/__init__.py
mehdiataei: Regarding the open question: Dense displacement remains a per Mesh operation because domain components can have different point counts and fields

physicsnemo/nn/functional/geometry/morphing/morphing.py
mehdiataei: I did this place take a look. Thanks.

mehdiataei:
Thanks @peterdsharpe for the review! All addressed. Please take a look.

physicsnemo/nn/functional/geometry/morphing/_warp_impl/kernels.py
```diff
@@ -0,0 +1,762 @@
...
+        else:
+            # When geometry gradients are unnecessary, field doubles as
+            # per-query scratch so no correction tensor needs to be allocated.
+            field[b, i, d] = value
+
+    for j in range(n_controls):
+        q = _normalized_distance_f32(points, controls, radii, b, i, j, n_dims)
+        if q > wp.float32(0.0) and q < wp.float32(1.0):
+            t = wp.float32(1.0) - q
+            phi = t * t * t * t * (wp.float32(4.0) * q + wp.float32(1.0))
```
peterdsharpe: **The Wendland quartic is hand-expanded 16 times in this file.** Per dtype: φ twice each in `shepard_forward`, `shepard_backward`, and `shepard_point_backward`, plus φ′ in both backward kernels — then doubled by the f32/f64 twins. A numerical edit to the kernel polynomial — exactly the kind of change the new `kernel=` extension point invites — currently requires 16 coordinated edits that can't be reviewed by eye. A `_wendland_phi` / `_wendland_phi_prime` `wp.func` pair would single-source the math today (one pair per dtype, collapsing to one generic pair if the twin dedup lands) — independent of, and composable with, the generic-kernel refactor discussed in the open thread above.

physicsnemo/nn/functional/geometry/morphing/_warp_impl/kernels.py
```diff
@@ -0,0 +1,762 @@
...
+    control_displacements: wp.array3d(dtype=wp.float32),
+    radii: wp.array2d(dtype=wp.float32),
+    min_q: wp.array2d(dtype=wp.float32),
+    denominator: wp.array2d(dtype=wp.float32),
+    exact_count: wp.array2d(dtype=wp.int32),
+    reference_index: wp.array2d(dtype=wp.int32),
+    correction: wp.array3d(dtype=wp.float32),
+    grad_field: wp.array3d(dtype=wp.float32),
+    n_dims: int,
+    need_points: int,
```
peterdsharpe: **`need_points` is dead code.** The only launch site always passes `0` (op.py L519) and computes point gradients with `shepard_point_backward_*` instead (op.py L488-500), feeding this kernel's `grad_points` a zero-size placeholder (op.py L523). So this parameter, the `need_points != 0` branch (L341-342 here, L665-666 in the f64 twin), and the `grad_points` output are unreachable and untested — and an atomics path in a hand-derived pullback is exactly where a future one-sided edit goes wrong silently. Suggest deleting the parameter and branch from both dtypes; it also shrinks the twin surface flagged in the open duplication thread.

physicsnemo/nn/functional/geometry/deform/deform.py
```diff
@@ -0,0 +1,648 @@
...
+            points,
+            control_points,
+            control_displacements,
+            radius=radius,
+            point_weights=point_weights,
+            kernel=kernel,
+        )
+
+    @classmethod
+    def make_inputs_forward(cls, device: torch.device | str = "cpu"):
```
peterdsharpe, line 497: **These benchmark generators are never executed by any test.** `make_inputs_forward`/`make_inputs_backward`/`compare_*` (on both specs) have zero CI coverage, so a drifting case tuple breaks the benchmark registry with no signal. Repo precedent wires them straight into tests: test_mesh_lsq_gradient.py L70-97 runs full torch↔warp forward/backward parity by iterating `make_inputs_*`, and test_sdf.py L112 smoke-tests the generator contract. Doing the same here is ~30 lines and buys meaningful coverage beyond the registry itself: `exact-handles-n2048-c16-d3` is the only input in the entire suite where queries coincide exactly with some controls while sitting inside other controls' supports, so the `has_exact` override (`_torch_impl.py` L352-354) — a real semantic branch of the field definition — currently has no direct assertion anywhere.

test/nn/functional/geometry/test_morphing.py
```diff
@@ -0,0 +1,1414 @@
...
+    if dtype == torch.float32:
+        atol, rtol = 4e-5, 4e-5
+    else:
+        atol, rtol = 2e-9, 2e-8
+    torch.testing.assert_close(morph_warp[0], morph_torch[0], atol=atol, rtol=rtol)
+    for actual, expected in zip(morph_warp[1], morph_torch[1]):
+        torch.testing.assert_close(actual, expected, atol=atol, rtol=rtol)
+
+
+def test_default_dispatch_matches_device_backend(device):
```
peterdsharpe: **This oracle can't catch the failure it targets.** Torch and warp agree within default `assert_close` tolerances on these inputs, so auto-dispatch selecting the *wrong* backend still passes — the test stays green even if `MorphPoints.dispatch` ignores the device entirely. `test_domain_morph_evaluates_combined_components_once` (test_mesh_morphing.py L337) shows the right pattern: monkeypatch a spy around the callable and assert what actually ran. The same trick on the backend entry points (`morph_points_torch` / `morph_points_warp` as bound in morphing.py) would pin the device→backend mapping — the one piece of custom dispatch logic this PR adds.

test/nn/functional/geometry/test_morphing.py
```diff
@@ -0,0 +1,1414 @@
...
+    phi = (1 - q).pow(4) * (4 * q + 1)
+    influence = phi / q.square()
+    return influence / (1 + influence)
+
+
+def test_public_exports_and_function_specs():
+    assert displace_points.__name__ == "displace_points"
+    assert morph_points.__name__ == "morph_points"
+    assert issubclass(DisplacePoints, object)
+    assert issubclass(MorphPoints, object)
```
peterdsharpe: `issubclass(..., object)` is vacuously true for any class — presumably this meant to pin the FunctionSpec contract (needs `from physicsnemo.core.function_spec import FunctionSpec` in the imports):

```suggestion
    assert issubclass(DisplacePoints, FunctionSpec)
    assert issubclass(MorphPoints, FunctionSpec)
```

peterdsharpe (changes requested):
Looking great - thank you for addressing those issues! Just a few last things to flag, in addition to the unresolved threads (one on the f32/f64 kernel deduplication, and one on the dispatch-fallback duplication), adding just a few more cleanup things.

We should probably add a DomainMesh tutorial showing off some of these cool morph features, but let's save that for a follow-up PR - this one is already beefy and lands the numerical core. Looking forward to having this merged!

physicsnemo/nn/functional/geometry/deform/_warp_impl/kernels.py
peterdsharpe: Updating on status: the phi dedup and dead-code removal shrank the twin surface and single-sourced the actual math, but the kernel scaffolding is still duplicated per dtype; no generics/overload refactor.

physicsnemo/nn/functional/geometry/deform/deform.py
peterdsharpe: Update: MorphPoints.dispatch still doesn't call _warn_fallback and still duplicates the explicit-implementation branch. The fallback behavior is now test-pinned, but silently — no warning asserted or emitted. This should hopefully be a quick ~5-line fix!

physicsnemo/nn/functional/geometry/deform/_warp_impl/kernels.py
peterdsharpe: Fixed by 22ef0291d6a9a788f2cfd20c743e154b7076440b

physicsnemo/nn/functional/geometry/deform/deform.py
peterdsharpe: Fixed by 878d9fc9638893ddd4d7a1a38dc8879a51a22caf

peterdsharpe (approved):
Looks good to me! Thanks for working through the review process!

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1786/commits/8e29ea81b5701b3a1f6e0e19ec9427812516a01b

coreyjadams (approved):

loliverhennigh (approved):
Lgtm

CharlelieLrt (approved):

ktangsali (approved):
LGTM

PR #1794, opened 2026-07-06
WARP based radius search kernel

cheersanimesh:
# Radius Search Kernels — PR Notes (`animesh/radius_search_kernel`)

Adds experimental, **GPU-only** ball-query ("radius search") kernels as drop-in
alternatives to the Warp hash-grid backend, selected via **environment variables**.
The public `radius_search(...)` API is unchanged; nothing changes unless you opt in.
Also adds timing/capture tooling and NVTX profiling wired through `BQWarp`,
GeoTransolver, and the unified external aero recipe.

For each `query`, find up to `max_points` `points` within `radius`.

## Modes

Base backend via `implementation=`:

- `torch` — brute-force `cdist` baseline (CPU/CUDA).
- `warp` — Warp `HashGrid`, the default on GPU.

Experimental variants activate **only** when `implementation="warp"` **and**
`max_points` is set **and** inputs are on CUDA. Select with
`PHYSICSNEMO_RADIUS_SEARCH_MORTON`:

| Value | What it does |
| --- | --- |
| `scalar` / `fma` / `gemm` | Morton grid + open-addressed **cell hash**; scalar / tiled-FMA / tiled-GEMM distance eval (`_morton_impl.py`). |
| `dense_fma` | **Dense directly-indexed cell table** (no hash), one warp per query — main dense path (`_morton_dense_impl.py`). |
| `dense_fma_e2e` | Sort-free `dense_fma` (drops both Morton radix sorts). |
| `dense_fma_mem_opt` / `dense_fma_mm` / `dense_gemm` | Memory-opt / matmul / tiled-GEMM benchmark kernels. |
| `pysdf_cuda` | Vendored **OptiX-free software-QBVH** (owl + gequel), JIT-compiled CUDA ext (needs `nvcc`). |

Also `PHYSICSNEMO_RADIUS_SEARCH_BVH=1` → Morton-ordered **tiled LBVH**
(`wp.tile_bvh_query_aabb`).

**Precedence:** `MORTON` > `BVH` > hash grid. `max_points=None` or CPU always uses
the default path. Neighbor order within a query is unspecified.

Debug env vars: `..._TIMING`, `..._CAPTURE_DIR` / `_IMPLS` / `_LIMIT`, `..._DUMP`.

## Run

```python
import os, torch
from physicsnemo.nn.functional import radius_search

os.environ["PHYSICSNEMO_RADIUS_SEARCH_MORTON"] = "dense_fma"  # pick a variant (optional)
idx, pts, d = radius_search(
    torch.rand(8192, 3, device="cuda"), torch.rand(4096, 3, device="cuda"),
    radius=0.05, max_points=32, return_points=True, return_dists=True,
    implementation="warp",
)
```

```bash
# shell equivalent
PHYSICSNEMO_RADIUS_SEARCH_MORTON=fma        python my_script.py
PHYSICSNEMO_RADIUS_SEARCH_BVH=1             python my_script.py

# training recipe (train.py defaults MORTON=fma; override to benchmark others)
cd examples/cfd/external_aerodynamics/unified_external_aero_recipe
torchrun --nproc_per_node=8 src/train.py --model=conf/model/geotransolver_volume.yaml
./profile.sh        # nsys capture (needs profile=true; auto-stops after warmup+profile steps)
```

## Test

```bash
pytest test/nn/functional/neighbors/test_radius_search.py -v        # CUDA tests skip on CPU
pytest test/nn/functional/neighbors/test_radius_search.py -k "morton or dense or pysdf or bvh" -v
```

Every variant has brute-force reference tests plus env-var dispatch-parity tests
(vs the torch backend). `pysdf_cuda` skips if it can't JIT-compile.

## Changed files

- **New kernels:** `_morton_impl.py` + `_morton_kernels.py` (scalar/fma/gemm),
  `_morton_dense_impl.py` + `_morton_dense_kernels.py` + `_morton_dense_gemm_kernels.py`
  (dense_*), `_pysdf_cuda_impl.py` + `_pysdf_cuda_ext/` (pysdf_cuda).
  `_compact_cell_points_impl.py` exists but is **not wired into dispatch**.
- **Core:** `_warp_impl.py` (env-gated dispatch + tiled-BVH path), `kernels.py`
  (Morton + tiled-BVH kernels), `utils.py` (timing/capture helpers),
  `radius_search.py` (registry).
- **Integration:** `ball_query.py`, GeoTransolver `context_projector.py` /
  `geotransolver.py`, recipe `base.yaml` / `train.py` / `profile.sh` (profiling knobs + NVTX).
- **Docs/pkg/tests:** `neighbors.rst`, `CHANGELOG.md`, `pyproject.toml` (ship
  `_pysdf_cuda_ext/**`), `test_radius_search.py`.

## Pre-merge cleanup

- Stray prints: `"******using warp/torch impl"` (`radius_search.py`),
  `"Max points : ..."` (`_warp_impl.py`), `"FORWARD BEGIN ANIMESH"` (`geotransolver.py`).
- Untracked scratch: `TILED_BVH_DEBUG.md`, `Untitled`, `_morton_debug.py`,
  `debug_tiled_bvh.py`, `*.nsys-rep`, `scripts/`.
- All new backends are experimental, CUDA-only, and require `max_points`.


peterdsharpe:
I would strongly recommend that we don't ship any raw CUDA kernels here that require runtime-compilation - that will significantly hamper user adoption. Also, redistributing `gequel` and `owl` in their source is not tenable for us to maintain long-term, and contributes to the large size (>12k LOC) of this PR. We also should absolutely not be setting env vars in our `train.py`, particularly a silent opt-in to an experimental backend; and we should not be hardcoding dataset paths.

PR #1800, opened 2026-07-08
Add exact-boundary quality meshing: tessellation.fill_interior

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

**Part 1 of 3** (part 2: #1801, implicit-domain mesher; part 3: #1802, tutorial). No code dependency on #1801 — land in either order.

Adds exact-boundary quality mesh generation: `tessellation.fill_interior` takes a closed codimension-one boundary `Mesh` and fills the enclosed interior with quality simplices.

- **Mesh-native**: a 2D edge `Mesh` in — loops in any order/orientation; holes, multiple components, and islands-inside-holes resolved automatically by containment — and a volume `Mesh` out, on the input's device/dtype.
- **Exact-boundary contract**: input vertices appear bit-identically (leading rows, input order); boundary facets are only ever subdivided, never moved; deterministic (identical inputs give bitwise-identical outputs).
- **Guaranteed quality**: every output triangle provably meets the requested minimum angle (Ruppert refinement), with an optional cell-size bound and bound-preserving ODT smoothing.
- **Provenance (opt-in)**: with `provenance=True`, `point_data["boundary_marker"]` and `point_data["source_point"]` map output vertices back to input vertices, enabling per-boundary BCs and data propagation. Off by default — the function claims no keys in the user-owned `point_data` namespace unless asked.
- Engine is a from-scratch, dependency-free constrained Delaunay pipeline (Bowyer–Watson + Sloan segment recovery + even-odd hole removal); citations in the module docstring. The contract is dimension-generic; `n=3` raises `NotImplementedError` pending exact 3D boundary recovery.
- 93 tests (engine invariants, the Mesh-native contract, and regression tests from two external/internal review rounds) and API docs. Tutorial coverage lives in the dedicated mesh-generation tutorial (#1802) so tutorial 8 stays scoped to I/O. ~90k guaranteed-quality triangles in ~6 s single-core.

fill_interior: boundary in, quality mesh out, with provenance

the 30 degree guarantee holds at every resolution

## Checklist

## Dependencies

No new dependencies.

## Review Process


test/mesh/tessellation/test_delaunay.py
```diff
@@ -0,0 +1,782 @@
...
+    assert (doubled > 0.0).all()
+    import scipy.spatial
+
+    hull = scipy.spatial.ConvexHull(points)
+    assert triangles.shape[0] == 2 * n - hull.vertices.shape[0] - 2
+    assert 0.5 * doubled.sum() == pytest.approx(hull.volume, rel=1e-12)
+
+    # Empty circumcircle, brute force against every point.
+    denominator = 2.0 * doubled
+    bl = ((b - a) ** 2).sum(1)
```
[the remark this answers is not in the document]
peterdsharpe: fixed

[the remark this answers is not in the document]
peterdsharpe: This comment reviewed a stale revision — the current code already uses `pytest.importorskip("scipy.spatial")` at this site (scipy is absent from the CI test environment, so the test skips rather than fails). Keeping the skip *inside* this one test is deliberate: a module-level skip would disable all 45 engine tests when scipy is missing, but only this single cross-check needs it.

physicsnemo/mesh/tessellation/fill_interior.py
```diff
@@ -0,0 +1,292 @@
...
+    Even nesting depth = a component's outer boundary; its holes are the
+    loops one level deeper that it directly contains. Islands inside holes
+    start new components (any nesting depth is supported).
+    """
+    n = len(loop_polys)
+    contains = [[False] * n for _ in range(n)]
+    for i in range(n):
+        for j in range(n):
+            if i != j:
+                contains[i][j] = _point_in_polygon(loop_polys[j][0], loop_polys[i])
```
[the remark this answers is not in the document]
peterdsharpe: Valid — fixed in 4c3be42c.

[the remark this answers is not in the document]
peterdsharpe: Valid — fixed in 4c3be42c: containment now probes with polygon_interior_point.

melo-gonzo:
Hey @peterdsharpe, this is super cool! I worked through the jupyter notebook, which was very helpful. You could consider purging out the metadata for execution counts, times, and some cached paths.

With the help of some agentic investigation, there are three things that seem worth flagging for this initial PR. Since I'm not an expert in this code and meshing, I'll lay out some cases here that were flagged by the investigation. It would be great to get a read on if these issues would be meaningful in real-world usage. 2 and 3 seem valid, while 1 seems like a more general issue with 'reflex corners' and meshers in general.

---
### 1. Segment recovery fails on simple polygons with reflex-adjacent corners

```python
import torch
from physicsnemo.mesh import Mesh
from physicsnemo.mesh.tessellation import fill_interior

pts = torch.tensor([[-0.6, -0.3], [-0.4, -0.7], [-0.1, -0.3],
                    [0.1, -1.0], [0.3, -0.7], [0.5, -0.2]], dtype=torch.float64)
i = torch.arange(6)
fill_interior(Mesh(points=pts, cells=torch.stack([i, (i + 1) % 6], 1)))
# RuntimeError: segment recovery could not find a starting wedge; input
# geometry is too degenerate for float64 predicates
```

Silent form — another exactly-simple hexagon; the picture shows the
triangulation quietly missing 25% of the domain. With `min_angle_degrees=30`,
this seems to loop indefinitely. With `min_angle_degrees=0.0` we get a 
poorly formed mesh. 

```python
import torch
import matplotlib.pyplot as plt
from physicsnemo.mesh.tessellation.delaunay import delaunay_mesh_2d

pts = torch.tensor([[0.97, 0.00], [0.32, 0.09], [0.43, 0.26],
                    [0.23, 0.44], [0.43, -0.83], [0.84, -0.24]], dtype=torch.float64)
# NOTE: the default min_angle_degrees=30 hangs on this input; 0 isolates the constrained triangulation.
p, t, _, _ = delaunay_mesh_2d([pts], max_area=None, min_angle_degrees=0.0)
x, y = pts[:, 0], pts[:, 1]
exact = float(0.5 * (x * y.roll(-1) - x.roll(-1) * y).sum().abs())
tp = p[t]
meshed = float((0.5 * ((tp[:, 1, 0] - tp[:, 0, 0]) * (tp[:, 2, 1] - tp[:, 0, 1])
       - (tp[:, 2, 0] - tp[:, 0, 0]) * (tp[:, 1, 1] - tp[:, 0, 1]))).abs().sum())
fig, ax = plt.subplots()
ax.fill(x, y, color="tab:red", alpha=0.25, lw=0)  # red = domain the mesh fails to cover
ax.tripcolor(p[:, 0], p[:, 1], t, facecolors=0.0 * t[:, 0], cmap="Greys", vmin=0, vmax=1)
ax.triplot(p[:, 0], p[:, 1], t, lw=0.4, color="tab:blue")
ax.set_aspect("equal")
ax.set_title(f"exactly simple hexagon: area in {exact:.3f}, meshed {meshed:.3f}, no error")
plt.show()
```

Root cause traced:
`_first_crossing` picks the starting wedge with two cross-sign tests that
never confirm the ray points *into* the wedge, so near-straight reflex
corners select the wedge behind the vertex; the pipe walk then exits the hull
and `_collect_pipe` computes `3*n` with `n = -1` — Python negative indexing
silently reads garbage adjacency. Suggested guards: pick the wedge by proper
edge-vs-segment intersection, assert `n >= 0` in `_collect_pipe`, cap
`_triangles_around`. The current test shapes are all densely sampled smooth
boundaries where recovery is a no-op; both hexagons above were found by
shrinking (delta debugging), so a shrinking property-based test over random
star polygons would lock this down.

### 2. Component grouping misclassifies valid rectilinear nesting (regression in 4c3be42c)

```python
import torch
from physicsnemo.mesh import Mesh
from physicsnemo.mesh.tessellation import fill_interior

# 6x6 plate with a centered 5x5 cutout: a valid square annulus.
points = torch.tensor([[0.0, 0.0], [6.0, 0.0], [6.0, 6.0], [0.0, 6.0],    # outer loop
                       [0.5, 0.5], [5.5, 0.5], [5.5, 5.5], [0.5, 5.5]],   # hole loop
                      dtype=torch.float64)
cells = torch.tensor([[0, 1], [1, 2], [2, 3], [3, 0],
                      [4, 5], [5, 6], [6, 7], [7, 4]])
try:
    fill_interior(Mesh(points=points, cells=cells))
except ValueError as e:
    print("annulus crash:", e)   # torch.cat(): expected a non-empty list of Tensors

# Same plate plus a 1x1 island inside the hole: valid domain, area 36 - 25 + 1 = 12.
points = torch.cat([points, torch.tensor([[2.5, 2.5], [3.5, 2.5], [3.5, 3.5], [2.5, 3.5]],
                                          dtype=torch.float64)])
cells = torch.cat([cells, torch.tensor([[8, 9], [9, 10], [10, 11], [11, 8]])])
m = fill_interior(Mesh(points=points, cells=cells))
print(f"island-in-hole: {m.n_points} points, {m.n_cells} cells")
# island-in-hole: 4 points, 2 cells  <- the island alone; 92% of the domain silently dropped
```

Cause: `_group_components` probes containment with
`polygon_interior_point(loop_j)` (largest-ear centroid), and a square outer
loop's ear centroid legitimately falls inside its own hole, so every loop
lands at odd depth. Not a thin-ring corner case — any centered square hole of
side ≥ 2 in the 6x6 plate fails; ≤ 1.8 works. Containment between *disjoint*
loops should be tested with a point of loop j's *boundary* (any vertex);
touching loops violate the documented disjointness precondition and can just
raise.

### 3. Corners sharper than `min_angle_degrees` silently break the guarantee

```python
import math
import torch
from physicsnemo.mesh.tessellation.delaunay import delaunay_mesh_2d

wedge = torch.tensor([[0., 0.], [1., 0.],
                      [math.cos(math.radians(20)), math.sin(math.radians(20))]],
                     dtype=torch.float64)
p, t, _, _ = delaunay_mesh_2d([wedge], max_area=None, min_angle_degrees=30.0)
tp = p[t]
areas = 0.5 * ((tp[:,1,0]-tp[:,0,0]) * (tp[:,2,1]-tp[:,0,1])
             - (tp[:,2,0]-tp[:,0,0]) * (tp[:,1,1]-tp[:,0,1]))
d = (p - wedge[0]).norm(dim=-1)
print(len(p), "points;", int((d < 1e-9).sum()), "within 1e-9 of the apex")
# 717 points; 597 within 1e-9 of the apex
print("smallest triangle area:", float(areas.abs().min()))
# smallest triangle area: 2.9e-109
t32 = p.float()[t]
a32 = 0.5 * ((t32[:,1,0]-t32[:,0,0]) * (t32[:,2,1]-t32[:,0,1])
           - (t32[:,2,0]-t32[:,0,0]) * (t32[:,1,1]-t32[:,0,1]))
print(int((a32 == 0).sum()), "of", len(t), "triangles have exactly zero area in float32")
# 424 of 715 triangles have exactly zero area in float32
```

The run terminates "successfully" - no `RuntimeError`, though the docs promise
one for inputs outside the ≥60° assumption — because `_is_bad`'s raw product
of squared edge lengths underflows to `0 > 0 = False` before the insertion
budget trips. A 20° corner makes the 30° bound geometrically unsatisfiable at
the apex, so the honest behaviors are reject-loudly or relax-and-document; 
instead the mesh ships with debris that detonates
downstream. Suggested regardless of policy: make `_is_bad` scale-invariant 
so underflow can't mask bad triangles, guard midpoint == endpoint in `_split_segment`, 
and re-scope the ≥60° caveat — it currently reads as applying only to "segments of distinct loops", i.e.
same-loop corners unrestricted, which is exactly the failing case.

peterdsharpe:
Thanks @melo-gonzo — all three cases were valid, and this was exactly the review the PR needed. Fixed in 3d256a7a:

**1. Segment recovery (valid — and worse than it looked).** Your root-cause diagnosis was correct: the starting-wedge test checked only that the wedge's two vertices straddle the ray *line*, which can't distinguish the wedge containing the forward ray from the backward one — your second hexagon showed the walk then exits the hull through Python negative indexing and silently drops coverage. Not a general reflex-corner limitation: a simple polygon always admits a CDT, so this was squarely our bug. Wedge selection now requires a proper segment-segment intersection with the wedge's opposite edge (taking the crossing nearest the segment start, since a segment can properly cross a non-convex link polygon more than once), and the pipe walk hard-fails on hull exit. Your observation about *why* the suite missed it (densely sampled boundaries make recovery a no-op) is now institutionalized: both hexagons are named regression tests, plus a 25-case random-star-polygon property test asserting shoelace-exact area on pure CDTs.

**2. Nesting misclassification (valid — a regression from an earlier review round).** Exactly as you said: the largest-ear interior point of a square outer loop legitimately falls inside its own hole. Containment between disjoint curves is now probed with the loops' boundary vertices — all of them, so touching/crossing loops raise a mixed-membership error instead of classifying arbitrarily. Your square-annulus and island-in-hole cases are regression tests with exact area assertions.

**3. Sharp corners (valid).** The ~60° corner condition was documented as an assumption but never enforced; your 20° wedge shows the failure mode is silent and severe (sub-float32-area triangles). It's now validated: a corner sharper than 60° with an angle bound requested raises with the corner's location and the `min_angle_degrees=0.0` escape hatch (pure CDT still meshes your wedge exactly). A guaranteed-quality fix for sharp corners (Miller–Pav–Walkington-style concentric shells) is possible future work, but loud rejection beats silent garbage today.

Also scrubbed the notebook execution metadata as suggested.

melo-gonzo (approved):

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1800/commits/3d256a7a304eea8e73bb72d8031756c2ce45e447

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1800/commits/d77f87b2e5837611490b59d4bec1ce1d663baee6

docs/api/mesh/tessellation.rst
```diff
@@ -59,6 +59,54 @@ broadcast to the output identically in both paths using the returned
...
+meshes: :func:`fill_interior` takes a closed codimension-one boundary
+``Mesh`` — in 2D, an edge mesh forming one or more loops, in any order and
+orientation, with holes, multiple components, and islands-inside-holes
+resolved automatically by containment — and fills the enclosed interior
+with quality simplices. The engine is constrained Delaunay triangulation
+(Bowyer--Watson insertion with constrained-edge recovery, holes removed
+topologically by even-odd parity flood fill) followed by Ruppert's
+Delaunay refinement, so in 2D every output triangle **provably** satisfies
+the requested minimum-angle bound and, optionally, a maximum cell size.
+Optional ODT smoothing (``smooth_iterations``) improves the *typical*
```
ktangsali: Has ODT been spelled out before?

ktangsali (approved):

docs/api/mesh/tessellation.rst
peterdsharpe: Ah! Yep, good catch :) Optimal Delaunay Triangulation. Will add.

Update: Now spelled out at first use in each document: "optimal-Delaunay-triangulation (ODT) smoothing — each interior vertex moves to the area-weighted average of its incident triangles' circumcenters (Chen & Xu 2004)". Applied to the rst, the `fill_interior` docstring, and (same pattern) the implicit mesher's docs in #1801.

docs/api/mesh/tessellation.rst
```diff
@@ -59,6 +59,56 @@ broadcast to the output identically in both paths using the returned
 A :class:`~physicsnemo.mesh.mesh.Mesh` can also be constructed in one step with
 :meth:`~physicsnemo.mesh.mesh.Mesh.from_polygons`.
 
+Exact-boundary interior filling
```
megnvidia, line 62: for lines R62 - R90:


Exact-Boundary Interior Filling
--------------------------------

Beyond decomposing existing cells, the package also *generates* quality
meshes. :func:`fill_interior` takes a closed codimension-one boundary
``Mesh`` and fills the enclosed interior with quality simplices. In 2D, the
boundary is an edge mesh forming one or more loops, in any order and
orientation. The algorithm resolves holes, multiple components, and
islands-inside-holes automatically through containment. 

The algorithm uses constrained Delaunay triangulation
(Bowyer--Watson insertion with constrained-edge recovery, holes removed
topologically by even-odd parity flood fill) followed by Ruppert's
Delaunay refinement, so that in every 2D output triangle *provably* satisfies
the requested minimum-angle bound and, optionally, a maximum cell size.

Optional optimal-Delaunay-triangulation (ODT) smoothing, controlled by
``smooth_iterations``, moves each interior vertex to the area-weighted
average of its incident triangles' circumcenters (Chen and Xu 2004). This
improves the *typical* angle while preserving both bounds. The exact-boundary contract provides three guarantees:

- Every input vertex appears bit-identically in the output, in leading rows
  and input order.
- The pipeline only ever *subdivides* boundary facets and never moves them.
- The entire process is deterministic. 

You can attach provenance fields (such as ``"boundary_marker"`` and ``"source_point"``) to the
output's ``point_data`` by setting ``provenance=True``. By default, the pipeline claims no keys in the user-owned namespace.

The contract is dimension-generic by design. The ``n = 3`` case (watertight surface ``Mesh[2, 3]`` producing tetrahedra) currently raises :class:`NotImplementedError` because exact 3D boundary recovery is a separate, hard program. .. TODO(author): Clarify "hard program" — do you mean a computationally harder problem, a separate implementation effort, or both? For implicit domains, or approximate volume meshing of a surface through its SDF, refer to ``physicsnemo.mesh.generate.mesh_implicit_domain``.

physicsnemo/mesh/tessellation/__init__.py
```diff
@@ -20,8 +20,20 @@
 cell-to-vertex rings) into a triangle connectivity, staying correct for
 non-convex polygons. It branches on manifold dimension and currently implements
 the 2D (polygon -> triangle) case.
+
+Also exposes exact-boundary quality mesh *generation* (which inserts new
+points, unlike :func:`triangulate`'s pure decomposition):
+:func:`fill_interior` takes a closed codimension-one boundary ``Mesh`` (edge
+loops in 2D) and fills the enclosed interior with quality simplices via
```
megnvidia, line 27:
```suggestion
loops in 2D) and fills the enclosed interior with quality simplices through
```

physicsnemo/mesh/tessellation/__init__.py
```diff
@@ -20,8 +20,20 @@
...
+constrained Delaunay triangulation with Ruppert refinement — every input
```
megnvidia, line 28:
```suggestion
constrained Delaunay triangulation with Ruppert refinement. The algorithm
```

physicsnemo/mesh/tessellation/__init__.py
```diff
@@ -20,8 +20,20 @@
...
+vertex is preserved bit-identically, and in 2D every output angle is
```
megnvidia, line 29:
```suggestion
preserves every input vertex bit-identically, and in 2D every output angle
```

physicsnemo/mesh/tessellation/__init__.py
```diff
@@ -20,8 +20,20 @@
...
+guaranteed to meet the requested bound. :func:`polygon_interior_point`
```
megnvidia, line 30:
```suggestion
meets the requested bound. :func:`polygon_interior_point`
```

physicsnemo/mesh/tessellation/fill_interior.py
```diff
@@ -0,0 +1,377 @@
...
+    boundary: "Mesh",
+    *,
+    max_cell_size: float | None = None,
+    min_angle_degrees: float = 30.0,
+    smooth_iterations: int = 0,
+    provenance: bool = False,
+) -> "Mesh":
+    r"""Fill the interior of a closed boundary mesh with quality simplices.
+
+    Dimension-generic contract: given a closed codimension-one boundary
```
megnvidia, line 202: from 202-217 there are many punctuation things that are inconsistent with most corporate style guides....but this is a py file. Content like this feels like a gray area for me. I often ignore content in py files and only focus on rst, markdown, html, and such. 

Let me know if you want me to come back and edit this.

CHANGELOG.md
```diff
@@ -10,6 +10,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Adds exact-boundary quality mesh generation to
```
megnvidia, line 13: R13 - R24 suggestions:

- Adds exact-boundary quality mesh generation to
  `physicsnemo.mesh.tessellation`. `fill_interior` takes a closed
  codimension-one boundary `Mesh` (currently 2D edge loops, in any order
  and orientation) and fills the interior with quality simplices through
  constrained Delaunay triangulation with Ruppert refinement. The algorithm
  resolves holes, multiple components, and nested islands automatically. It
  preserves every input vertex bit-identically, and every output triangle
  meets the guaranteed minimum-angle bound. Output is a `Mesh` with
  provenance `point_data`, deterministic, with optional bound-preserving
  optimal-Delaunay-triangulation (ODT) smoothing. The contract is
  dimension-generic. `n = 3` raises `NotImplementedError` pending exact
  boundary recovery. Also adds `polygon_interior_point`, which returns a
  point strictly inside a simple polygon.

docs/api/mesh/tessellation.rst
peterdsharpe: Done in #1802 (75880c94), with two small tweaks: fixed a dangling clause ('so that every 2D output triangle *provably* satisfies…') and answered the TODO inline — 'hard program' now reads 'a substantially harder problem that requires its own implementation effort.'

physicsnemo/mesh/tessellation/__init__.py
peterdsharpe: Done in #1802: 75880c94 (this PR already merged).

physicsnemo/mesh/tessellation/__init__.py
peterdsharpe: Done in #1802: 75880c94 (this PR already merged).

physicsnemo/mesh/tessellation/__init__.py
peterdsharpe: Done in #1802: 75880c94 (this PR already merged).

physicsnemo/mesh/tessellation/__init__.py
peterdsharpe: Done in #1802: 75880c94 (this PR already merged).

CHANGELOG.md
peterdsharpe: Done in #1802: 75880c94 (this PR already merged).

physicsnemo/mesh/tessellation/fill_interior.py
peterdsharpe: Good instinct to ask — these docstrings do render into the API reference, so style does surface there. But no need for you to sweep .py files; we'll handle those ourselves.

PR #1814, opened 2026-07-09
Mesh: Add lattice free-form deformation

mehdiataei:
# PhysicsNeMo Pull Request

## Description

Adds differentiable lattice free-form deformation (FFD) for point tensors, meshes, and domain meshes.

The new API includes:

- `ffd_points` with aligned batched and unbatched inputs.
- `Mesh.ffd` and `DomainMesh.ffd` convenience methods.
- Bernstein-basis classic FFD with global support.
- Uniform cubic B-spline FFD with local support.
- PyTorch and NVIDIA Warp backends with first-order autograd support.
- Optional per-point floating weights and boolean masks.
- Automatic Torch/Warp backend selection based on device and availability.

The implementation includes memory-bounded Torch evaluation, stable high-degree Bernstein evaluation in Warp, empty-input handling, `torch.compile` support, CUDA Graph compatibility, input validation, backend parity tests, benchmark registration, and public documentation. The mesh documentation also includes a visualization comparing global Bernstein deformation with a multi-feature local B-spline deformation.

## Checklist

- [x] If I am implementing a new model or modifying any existing model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation). Not applicable; this PR does not add or modify a model.

## Dependencies

No new dependencies are required.

## Review Process


loliverhennigh (approved):
LGTM, no comment on this one

physicsnemo/nn/functional/geometry/deform/ffd.py
```diff
@@ -0,0 +1,718 @@
...
+            ``points``.
+
+        Raises
+        ------
+        TypeError
+            If tensor dtypes or Python argument types are unsupported.
+        ValueError
+            If shapes, devices, lattice configuration, weights, or ``basis``
+            are invalid.
+        """
```
peterdsharpe: **Consolidate internal docstrings (~110 lines).** `make_function` publishes the *class* docstring as the public `ffd_points` docs (`function_spec.py:477`), so the full Parameters/Returns/Raises sections here on `warp_forward`, `torch_forward`, and `dispatch` are internal-only — and they triplicate content the class docstring already owns. The sibling `MorphPoints` in `deform.py` keeps these to one line each (e.g. `"""Apply compact Shepard morphing with the Warp backend."""`). Cutting to that convention removes ~110 lines and, more importantly, removes three copies that can drift from the canonical class docstring.

physicsnemo/nn/functional/geometry/deform/ffd.py
```diff
@@ -0,0 +1,718 @@
...
+        ValueError
+            If shapes, devices, lattice configuration, weights, or ``basis``
+            are invalid.
+        KeyError
+            If ``implementation`` does not name a registered backend.
+        ImportError
+            If an explicitly requested backend is unavailable.
+        """
+
+        if implementation is None:
```
peterdsharpe, line 417: This `implementation is None` → warp-on-CUDA-else-torch block is a verbatim copy of `MorphPoints.dispatch` (`deform.py:475-485`), and it's the 4th hand-rolled variant of this selector in the codebase (`knn.py:126`, `farthest_point_sampling.py:106`). Fine to keep for symmetry with morph in this PR, but consider a shared `FunctionSpec` helper (e.g. `cls._dispatch_device_default(...)`) as a follow-up so the fallback-warning behavior stays uniform across all four.

physicsnemo/nn/functional/geometry/deform/ffd.py
```diff
@@ -0,0 +1,718 @@
...
+import torch
+from jaxtyping import Bool, Float
+
+from physicsnemo.core.function_spec import FunctionSpec
+
+from ._torch_impl import ffd_points_torch
+from ._utils import normalize_ffd_inputs, restore_point_rank
+from ._warp_impl import ffd_points_warp
+
+_FFD_BASES = (
```
peterdsharpe: The basis vocabulary is now spelled out in four independent places that must stay in sync: `_FFD_BASES` here, `_FFD_MIN_NODES` in `_utils.py`, `_BASIS_IDS` in `ffd_op.py`, and the `Literal[...]` repeated in six signatures (three here, plus `Mesh.ffd`, `DomainMesh.ffd`, and the mesh-level `ffd`). Two cheap consolidations: `_FFD_BASES = tuple(_FFD_MIN_NODES)` (import from `._utils`), and a shared `FFDBasis` type alias next to `_FFD_MIN_NODES` that all six signatures reference.

physicsnemo/nn/functional/geometry/deform/_warp_impl/ffd_op.py
```diff
@@ -0,0 +1,475 @@
...
+
+    try:
+        return _FFD_KERNELS[dtype]
+    except KeyError:
+        raise TypeError(
+            f"Warp free-form deformation supports float32 and float64, got {dtype}"
+        ) from None
+
+
+def _check_common_dtype(*tensors: torch.Tensor) -> None:
```
peterdsharpe: `_check_common_dtype` is a near-verbatim copy of the same helper in `op.py:72-80` (only the noun in the error message differs), and this module already imports `_empty_contiguous_like` and `_wp_view` from `.op`. Suggest importing this too — either with a generic message ("all floating deformation tensors...") or a `noun` parameter.

physicsnemo/nn/functional/geometry/deform/_warp_impl/ffd_op.py
```diff
@@ -0,0 +1,475 @@
...
+
+
+def _lattice_size(resolution: list[int]) -> int:
+    size = 1
+    for nodes in resolution:
+        size *= int(nodes)
+    return size
+
+
+def _window_total(resolution: list[int], basis: str) -> int:
```
peterdsharpe: Two small dedups in one spot:

1. `_window_total` here duplicates `_ffd_window_size` in `_torch_impl.py:532-542` — identical semantics (4^D for bspline, 2^D for interpolating, prod otherwise) under two names. One copy could live in `deform/_utils.py` next to `_FFD_MIN_NODES`, which both modules already import from.
2. `_lattice_size` (and `_ffd_window_size`'s Bernstein branch) re-implement `math.prod`, which `normalize_ffd_inputs` in `_utils.py:465` already uses for the same quantity.

physicsnemo/mesh/transformations/deform/ffd.py
```diff
@@ -0,0 +1,248 @@
...
+from physicsnemo.mesh.transformations.deform._utils import (
+    _mesh_with_deformed_points,
+    _resolve_point_field,
+)
+
+if TYPE_CHECKING:
+    from physicsnemo.mesh.mesh import Mesh
+
+
+def _origin_tensor_for_extent(
```
peterdsharpe, line 41: `_origin_tensor_for_extent` (~45 lines) re-implements a subset of `_normalize_lattice_box`'s validation from the functional `_utils.py` (exactly-D reals, finiteness, dtype-representability, requires_grad rejection) — and since `_default_lattice_box` returns the *original* `origin` rather than the converted tensor, a sequence origin is then validated and converted a second time inside `ffd_points`. Mesh-level `morph` by contrast delegates all validation to the functional layer.

I see why it exists: deriving `extent` needs a numerically valid origin *before* the functional validation runs, and it buys precise error attribution. Two options if you want to slim it: (a) hoist a shared single-vector validator into the functional `deform/_utils.py` and call it from both layers; or (b) convert loosely here and let `ffd_points` be the sole validator, accepting that a non-finite origin then surfaces as the (still accurate) "derived lattice extent must be finite" message. Also fine as-is if you value the current error messages — but worth a deliberate choice.

physicsnemo/nn/functional/geometry/deform/_torch_impl.py
```diff
@@ -473,6 +473,243 @@ def compact_shepard_field_torch(
...
+            )
+        return displace_points_torch(points, field, point_weights)
+
+    window = _ffd_window_size(resolution, basis)
+    query_chunk = _ffd_chunk_size(
+        batch_size, num_points, window, num_dims, points.element_size()
+    )
+    checkpoint_chunks = needs_grad and query_chunk < num_points
+    chunks: list[torch.Tensor] = []
+    for query_start in range(0, num_points, query_chunk):
```
peterdsharpe, line 685: Test-coverage note (verified correct, so purely insurance): with the 256 MB budget, this chunk loop + checkpoint path only activates above ~16k-70k points, and every test uses ≤257 — so the `torch.cat`/checkpoint plumbing ships untested. I forced chunking by shrinking `_FFD_TEMPORARY_BYTE_BUDGET` and confirmed forward + all three gradients match the unchunked path exactly (even at chunk size 1). A test that monkeypatches the budget would pin this; the Shepard path has the same structure and same gap.

test/nn/functional/geometry/test_point_ffd.py
```diff
@@ -0,0 +1,1303 @@
...
+        torch.library.opcheck(
+            ffd_field_warp_impl,
+            args=(points, control_displacements, origin, extent, [4, 4], basis),
+        )
+
+
+@requires_module("warp")
+@pytest.mark.parametrize("implementation", ["torch", "warp", None])
+@pytest.mark.parametrize("basis", _BASES)
+def test_torch_compile_fullgraph(implementation, basis):
```
peterdsharpe, line 953: Test-coverage note: the compile tests use `requires_grad=False` inputs, so the `checkpoint(...)` branch under `torch.compiler.is_compiling()` and the compiled custom-op backward never run in CI, despite compiled-training support being part of the PR's claims. I probed `fullgraph=True` + `backward()` for both backends on CPU and CUDA — all match eager — so it works today; adding a backward pass to one parametrization of this test would keep it that way.

physicsnemo/nn/functional/geometry/deform/ffd.py
```diff
@@ -0,0 +1,718 @@
...
+    "bernstein",
+    "bspline",
+    "linear",
```
peterdsharpe: Scope observation, explicitly not a request: of the five bases, `linear`/`smoothstep`/`smootherstep` are the marginal additions beyond classic FFD (Bernstein) and local control (B-spline) — they add a third basis family to every signature, docstring, kernel branch, and the test matrix. They're cheap given the shared window machinery and they're well-tested, so keeping them is defensible; but if the maintainers want a slimmer first landing, this is the natural seam to split along.

physicsnemo/mesh/domain_mesh.py
```diff
@@ -685,113 +686,207 @@ def morph(
...
+                point_weights=combined_point_weights,
+                kernel=kernel,
+                implementation=implementation,
+            )
+
+        return self._deform_components(components, resolved_point_weights, apply_field)
+
+    def ffd(
+        self,
+        control_displacements: torch.Tensor,
```
peterdsharpe: This needs jaxtyping

physicsnemo/mesh/transformations/deform/ffd.py
```diff
@@ -0,0 +1,248 @@
...
+                    "along every axis; the point bounds are degenerate or non-finite, "
+                    "or the supplied origin is not below their maximum. Supply an "
+                    "explicit extent."
+                )
+    return origin, extent
+
+
+def ffd(
+    mesh: "Mesh",
+    control_displacements: torch.Tensor,
```
peterdsharpe: jaxtyping needed

physicsnemo/mesh/mesh.py
```diff
@@ -2716,6 +2716,39 @@ def morph(
             implementation=implementation,
         )
 
+    def ffd(
+        self,
+        control_displacements: torch.Tensor,
```
peterdsharpe: jaxtyping needed

physicsnemo/mesh/transformations/deform/_utils.py
```diff
@@ -59,6 +59,84 @@ def _resolve_point_field(
     return resolved
 
 
+def _resolve_domain_point_weights(
+    components: "list[tuple[str, Mesh]]",
```
peterdsharpe:
```suggestion
    components: list[tuple[str, "Mesh"]],
```

physicsnemo/mesh/transformations/deform/_utils.py
```diff
@@ -59,6 +59,84 @@ def _resolve_point_field(
...
+                    f"point_weights field {point_weights!r} must have one "
+                    f"common dtype across all components; {label}.point_data "
+                    f"has {component_point_weights.dtype}, expected "
+                    f"{resolved_point_weights[0].dtype}"
+                )
+            resolved_point_weights.append(component_point_weights)
+    return resolved_point_weights
+
+
 def _mesh_with_deformed_points(mesh: "Mesh", points: torch.Tensor) -> "Mesh":
```
peterdsharpe: Can something like this use `dataclasses.replace()` instead?

physicsnemo/mesh/transformations/deform/_utils.py
peterdsharpe, line 155: Probably best to use `Mesh.strip_caches()` here, so that if new keys are added, this doesn't drift

docs/api/mesh/transformations.rst
```diff
@@ -154,6 +156,58 @@ and magnitudes.
...
    :width: 100%
 
+Lattice Free-Form Deformation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:meth:`~physicsnemo.mesh.mesh.Mesh.ffd` defines a regular array of control
+displacements over an axis-aligned evaluation box and deforms every point
+inside the box by tensor-product basis interpolation.
+Compared with sparse morphing, the design parameters form a structured grid of
+fixed size, which suits parametric shape optimization: a lattice of zeros is
```
megnvidia:
```suggestion
fixed size, which suits parametric shape optimization. A lattice of zeros is
```

physicsnemo/mesh/transformations/deform/ffd.py
```diff
@@ -0,0 +1,248 @@
...
+        lower-dimensional geometry embedded in a higher-dimensional space.
+        Default is ``None``.
+    basis : {"bernstein", "bspline", "linear", "smoothstep", "smootherstep"}, optional
+        Per-axis basis family. ``"bernstein"`` is classic global-support
+        free-form deformation for coarse lattices; ``"bspline"`` gives local
+        four-node-per-axis support and scales to fine lattices. For B-splines,
+        coefficient index ``i`` is associated with local coordinate
+        ``(i - 1) / (n - 3)``, so the first and last coefficient planes lie
+        outside the evaluation box. ``"linear"``, ``"smoothstep"``, and
+        ``"smootherstep"`` use the two neighboring nodes per axis and
```
peterdsharpe: Sorry, from here it's not clear what the difference between linear, smoothstep, and smootherstep are - can we explain? Or is there a literature reference we can point people to? This arg doc might also be better as a bulleted list

physicsnemo/nn/functional/geometry/deform/_utils.py
```diff
@@ -284,6 +293,187 @@ def normalize_morph_inputs(
...
+                )
+            if require_positive and any(
+                entry < finfo.tiny * finfo.eps for entry in values
+            ):
+                raise ValueError(
+                    f"{name} values must be strictly positive in the points dtype "
+                    f"{points.dtype}, got {value!r}"
+                )
+        return (
+            torch.tensor(values, dtype=points.dtype, device=points.device)
```
peterdsharpe, line 403: **Perf: Python-list `origin`/`extent` pay an uncached H2D copy on every call — ~0.8 ms, which roughly doubles small-batch call time.** Measured on an RTX 4090 (N=4096, bernstein 4^3, warp backend): 0.96 ms/call with device-tensor box vectors vs 1.74 ms/call with `origin=[0.0, 0.0, 0.0], extent=[1.0, 1.0, 1.0]` — the list form the docs examples use. The cost is this `torch.tensor(values, device=...)` running per call for each of origin and extent.

`ffd_op.py` already solves this exact problem for `resolution` with `_RESOLUTION_CACHE`, on the same rationale that lattice configuration is static. Suggest mirroring that here: cache the converted `(1, D)` tensor keyed by `(device, dtype, tuple(values))` before the expand. Hand-written list configs are as static as resolutions, so unbounded growth is a non-concern in practice (and bounded the same way the resolution cache is). Bonus: list-form calls become CUDA-graph-warmable, since after warmup no new H2D happens.

Cheapest alternative if you prefer no cache: a one-line note in the docs steering hot loops toward device tensors for `origin`/`extent`. (Same per-call `torch.tensor` pattern exists for morph's scalar `radius` in `_normalize_radius`, if you generalize the helper.)

peterdsharpe (approved):
This looks good! Nice work on the API design. No correctness bugs that I see. Before we merge: we should a) add jaxtyping to all functions, especially those part of the public API, and b) do a bit of deduplication of helpers; see comments. That said, once these are fixed I think we're good to merge, so approving so as not to block this further.

Also, should we expose as `Mesh.ffd` or `Mesh.free_form_deformation`? I kind of lean towards the latter, which in-general keeps with library precedent (`signed_distance_field`, not `sdf`). What do you think? (Maybe let's discuss offline?)

physicsnemo/mesh/transformations/deform/_utils.py
peterdsharpe: Actually this would need a `keep` kwarg or something, which we don't have... probably something to add in a new PR! Resolving this issue for now

docs/api/mesh/transformations.rst
```diff
@@ -154,6 +156,58 @@ and magnitudes.
...
+exactly the identity, and the same lattice deforms any geometry embedded in
+the box.
+
+When ``origin`` and ``extent`` are omitted, the box spans the mesh bounds. Each
```
megnvidia, line 170: i am trying to leave a comment on line 171

docs/api/mesh/transformations.rst
```diff
@@ -154,6 +156,58 @@ and magnitudes.
...
+coordinate axis must have positive range; planar or linear geometry embedded
+in a higher-dimensional space therefore needs an explicit positive extent.
+``basis="bernstein"`` is classic global-support free-form deformation for
```
megnvidia:
```suggestion
``basis="bernstein"`` provides classic global-support free-form deformation for
```

docs/api/mesh/transformations.rst
```diff
@@ -154,6 +156,58 @@ and magnitudes.
...
+coarse lattices. ``basis="bspline"`` gives local four-node-per-axis support
+and scales to fine lattices for local sculpting; its first and last coefficient
```
megnvidia:
```suggestion
and scales to fine lattices for local sculpting. Its first and last coefficient
```

docs/api/mesh/transformations.rst
```diff
@@ -154,6 +156,58 @@ and magnitudes.
...
+    deformed = mesh.ffd(control_displacements)
+
+    # Autograd continues through the returned point coordinates.
+    objective = deformed.points.square().mean()
+    objective.backward()
+
+Points outside the lattice box are unchanged. The deformation is generally not
+continuous across the box boundary. A sufficient condition for a fixed
+exterior is to zero the outermost coefficient plane on every Bernstein or
+node-interpolating face, or the first and last three coefficient planes on
```
megnvidia:
```suggestion
node-interpolating face. For cubic B-splines, zero the first and last three coefficient planes on
```

docs/api/mesh/transformations.rst
```diff
@@ -154,6 +156,58 @@ and magnitudes.
...
+every axis for cubic B-splines. ``origin`` and ``extent`` are
+non-differentiable lattice configuration; optimize ``control_displacements``
```
megnvidia:
```suggestion
non-differentiable lattice parameters. Optimize ``control_displacements``
```

docs/api/mesh/transformations.rst
```diff
@@ -154,6 +156,58 @@ and magnitudes.
...
+
+    # Autograd continues through the returned point coordinates.
+    objective = deformed.points.square().mean()
+    objective.backward()
+
+Points outside the lattice box are unchanged. The deformation is generally not
+continuous across the box boundary. A sufficient condition for a fixed
+exterior is to zero the outermost coefficient plane on every Bernstein or
+node-interpolating face, or the first and last three coefficient planes on
+every axis for cubic B-splines. ``origin`` and ``extent`` are
```
megnvidia:
```suggestion
every axis.. ``origin`` and ``extent`` are
```

docs/api/mesh/transformations.rst
```diff
@@ -200,21 +254,29 @@ resolved values also match.
         morphed_domain.boundaries["wall"].points,
     )
 
-Morphing preserves connectivity and attached point, cell, global, and domain
-data. Attached vector and tensor fields are treated as Lagrangian data and are
-not pushed forward. Geometry-dependent caches are discarded and recomputed
-lazily; topology caches are retained.
+:meth:`~physicsnemo.mesh.domain_mesh.DomainMesh.ffd` follows the same pattern
+for lattice free-form deformation: one lattice field is evaluated over the
```
megnvidia:
```suggestion
for lattice free-form deformation: 
```

docs/api/mesh/transformations.rst
```diff
@@ -200,21 +254,29 @@ resolved values also match.
...
+combined interior and boundary points, and the default box spans the combined
```
megnvidia:
```suggestion

- One lattice field is evaluated over the combined interior and boundary points.
- The default box spans the combined component bounds.
```

docs/api/mesh/transformations.rst
```diff
@@ -200,21 +254,29 @@ resolved values also match.
...
+component bounds. The combined bounds must have positive range on every
```
megnvidia:
```suggestion

```

docs/api/mesh/transformations.rst
```diff
@@ -200,21 +254,29 @@ resolved values also match.
...
+coordinate axis unless an explicit extent is supplied.
```
megnvidia:
```suggestion
The combined bounds must have positive range on every coordinate axis unless an
explicit extent is supplied.
```

docs/api/mesh/transformations.rst
```diff
@@ -200,21 +254,29 @@ resolved values also match.
...
+
+Every deformation preserves connectivity and attached point, cell, global, and
+domain data. Attached vector and tensor fields are treated as Lagrangian data
```
megnvidia:
```suggestion
domain data. The mesh treats attached vector and tensor fields as Lagrangian
```

docs/api/mesh/transformations.rst
```diff
@@ -200,21 +254,29 @@ resolved values also match.
...
+and are not pushed forward. Geometry-dependent caches are discarded and
```
megnvidia:
```suggestion
data and does not push them forward. Geometry-dependent caches are discarded and
```

docs/api/mesh/transformations.rst
```diff
@@ -200,21 +254,29 @@ resolved values also match.
...
+recomputed lazily; topology caches are retained.
```
megnvidia:
```suggestion
recomputed lazily. Topology caches are retained.
```

docs/api/nn/functionals/geometry.rst
```diff
@@ -80,6 +80,69 @@ For connectivity-preserving object APIs, use
...
+    deformed = ffd_points(
+        points,
+        control_displacements,
+        origin=[0.0, 0.0, 0.0],
+        extent=[1.0, 1.0, 1.0],
+        basis="bernstein",
+    )
+    deformed.square().mean().backward()
+
+With zero control displacements the operation is exactly the identity, so a
```
megnvidia:
```suggestion
With zero control displacements, the operation is exactly the identity, so a
```

docs/api/nn/functionals/geometry.rst
```diff
@@ -80,6 +80,69 @@ For connectivity-preserving object APIs, use
...
+lattice initialized at zero is a well-behaved starting point for shape
+optimization: an optimizer—or a model producing the lattice
```
megnvidia:
```suggestion
optimization. An optimizer, or a model that produces the lattice
```

docs/api/nn/functionals/geometry.rst
```diff
@@ -80,6 +80,69 @@ For connectivity-preserving object APIs, use
...
+displacements—learns the deformation from a differentiable objective on
```
megnvidia:
```suggestion
displacements, learns the deformation from a differentiable objective on
```

megnvidia (approved):

physicsnemo/_typing.py
```diff
@@ -0,0 +1,25 @@
...
+
+"""Dependency-neutral shared type aliases."""
+
+from typing import Literal, TypeAlias
+
+FFDBasis: TypeAlias = Literal[
+    "bernstein", "bspline", "linear", "cubic_hermite", "quintic_hermite"
+]
+
+__all__ = ["FFDBasis"]
```
peterdsharpe: Can we not have this top-level? If that means keeping the duplicate `Literal` labels (rather than a imported-in-both-`mesh`-and-`nn`), that's ok - preferable to exposing this top-level

PR #1817, opened 2026-07-09
fix(datapipes): resume-reproducible set_epoch across readers and transforms

peterdsharpe:
## Description

Every `set_epoch` implementation in the datapipes stack reseeded its generator with `initial_seed() + epoch`. That idiom compounds: `manual_seed` updates `initial_seed`, so the epoch→seed mapping depends on call history. With base seed 42, sequential `set_epoch(1..5)` produces seeds 43, 45, 48, 52, 57 — while a run resumed from a checkpoint that calls `set_epoch(5)` directly gets 47. A resumed run therefore silently draws **different subsample blocks and augmentations** than an uninterrupted run, breaking seed-matched A/B protocols. Within a single uninterrupted run nothing is wrong (still deterministic), which is why this went unnoticed.

**Fix**: capture the base seed once at `set_generator` time and reseed with `base_seed + epoch`. Applied to all sites:

- `readers/mesh.py` (`MeshReader`, `DomainMeshReader`) — cherry-picked byte-identical from #1798 (which applies the same fix to its new zarr mesh readers), so the two PRs merge cleanly in either order
- `readers/numpy.py`, `readers/zarr.py`, `readers/tensorstore_zarr.py`
- `transforms/base.py` (`Transform`), `transforms/mesh/base.py` (`MeshTransform`) — with a lazy fallback for generators declared by subclasses without going through `set_generator`

`RNG.md` and `DISTRIBUTIONS.md` documented the drifting idiom as the contract; both are updated to the resume-reproducible semantics. `Compose` only delegates and needed no change.

**Behavior note**: epoch ≥ 2 now draws different (finally history-independent) blocks/augmentations than before the fix, so bitwise loss-curve comparisons across this change will differ; training statistics are unaffected.

**Tests**: sequential-vs-direct `set_epoch` equivalence (`block([1,2,3]) == block([3])`) pinned for `MeshReader`, `ZarrReader` coordinated subsampling, and the `MeshDataset` transform pipeline. Full `test/datapipes/readers/` + `test_mesh_augmentations.py` suites pass (114 passed).

## Checklist

## Dependencies

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -197,6 +197,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed `set_epoch` seed drift in datapipes readers and transforms:
+  reseeding used `initial_seed() + epoch`, which compounds across calls
+  (manual_seed updates initial_seed), so resuming from a checkpoint drew
+  different subsample blocks and augmentations than an uninterrupted run.
+  Epoch reseeding now uses a base seed captured at `set_generator` time.
 - `physicsnemo.mesh.sampling.sample_data_at_points` now handles integer and
   boolean fields by returning `float64`, so NaN sentinels and non-integral
   interpolation or multi-cell means are representable (subject to the usual
--- a/physicsnemo/datapipes/RNG.md
+++ b/physicsnemo/datapipes/RNG.md
@@ -70,9 +70,11 @@ sub-dataset.
 ### Epoch reseeding
 
 `DataLoader.set_epoch(epoch)` propagates to the sampler and dataset.
-Each component with a generator reseeds it with
-`initial_seed() + epoch`, producing a different but deterministic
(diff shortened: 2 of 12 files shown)


peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1817/commits/268aff026bcf7499b5dd2002789a990a326ad2ab

peterdsharpe:
Closing in favor of #1742, which fixes the reader half of this more thoroughly (per-`(epoch, index)` SeedSequence-derived generators — history-independent *and* order-independent) and now, as of `31b6e897` pushed there, also covers the transforms half plus the doc updates from this PR. One residual note: NVIDIA/physicsnemo#1798 carries a copy of this PR's `readers/mesh.py` fix (`79d25079`) that will conflict with #1742 once it merges to main — the plan is to drop that commit from #1798 at its next update, taking #1742's version.

PR #1820, opened 2026-07-09
fix(datapipes): write domain-level global_data in SetGlobalField

peterdsharpe:
## What

`SetGlobalField.apply_to_domain` broadcast the injected fields to every sub-mesh's `global_data`, but never wrote them to the `DomainMesh`'s own domain-level `global_data`. Any code reading `DomainMesh.global_data` never saw the injected fields.

- The domain path now writes both records.
- Sub-mesh behavior is unchanged.
- Plain `Mesh` behavior is unchanged.

## Tests

New `test/datapipes/transforms/test_set_global_field.py`:

- domain-level `global_data` receives the field
- the sub-mesh broadcast still happens
- existing domain-level fields are preserved
- injecting an existing key overwrites it, matching the documented contract
- plain `Mesh` path unchanged

Full transforms suite passes; ruff check/format clean.

## Notes

- Found while integrating a DomainMesh-native model into the unified external-aero recipe, which currently carries a local workaround (`SetDomainGlobalField`) that this fix makes unnecessary.
- A companion transform (`ComputeUnitGlobalVector`) that was previously bundled in this PR is now split out into draft PR #1825, since it is new API rather than a bug fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -306,6 +306,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and interior elements equal inclusion probability.
 - Cell-subsampled GLOBE inputs now retain their effective integration measure,
   preventing area-weighted outputs and gradients from collapsing.
+- `SetGlobalField` now writes injected fields to the domain-level `global_data`
+  of a `DomainMesh`, in addition to the existing per-sub-mesh broadcast.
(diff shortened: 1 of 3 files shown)


coreyjadams (approved):
LGTM

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1820/commits/240edf87d49a0b5accfd18fc871a37da7f6fdc06

PR #1822, opened 2026-07-09
Mesh: GPU-Accelerated Remesh

mehdiataei:
# PhysicsNeMo Pull Request

## Description

This PR adds mesh remeshing to PhysicsNeMo using Warp.

The public API is available through `physicsnemo.mesh.remeshing.remesh` and `Mesh.remesh`. It supports CPU and CUDA meshes, preserves the input device, and provides controls for target vertex count, clustering, projection, topology reconstruction, and cleanup.

The implementation includes:

- A Warp-based remeshing functional and associated kernels.
- CPU and CUDA execution support.
- Public functional and object-oriented mesh APIs.
- Dynamic-shape and `torch.compile` support through `FunctionSpec`.
- Unit tests covering the public API, functional API, CPU and CUDA execution, validation, compilation, and topology.
- Documentation, examples, and ASV benchmarks.
- Updated comparison and scaling figures.
- PyACVD is removed as a dependency.

CUDA remeshing can be up to 300× faster than PyACVD on CPU.

## Checklist

- [x] If I am implementing a new model or modifying any existing model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation). (Not applicable; this PR does not add or modify a model.)

## Dependencies

No new dependencies are required. The implementation uses the existing Warp dependency, and PyACVD is not required.

## Review Process


physicsnemo/mesh/mesh.py
```diff
@@ -3394,6 +3396,63 @@ def statistics(self):
 
         return compute_mesh_statistics(self)
 
+    def remesh(
+        self,
+        n_clusters: builtins.int,
```
peterdsharpe, line 3501: `builtins.int`? Should be `int` (same on other args)

docs/api/mesh/remeshing.rst
```diff
@@ -3,33 +3,107 @@ Remeshing
...
+Advanced users can tune the search and initialization policy without changing
+the remeshing kernels:
 
 .. code:: python
 
-    from physicsnemo.mesh.remeshing import remesh
-    from physicsnemo.mesh.primitives.surfaces import sphere_icosahedral
+   from physicsnemo.mesh.remeshing import WarpRemeshOptions
+
+   options = WarpRemeshOptions(
```
peterdsharpe: Why have this interface, not just kwargs? In general, prefer kwargs to option containers.

docs/img/mesh/remeshing_performance.png
peterdsharpe, line 1: "Latency" -> "Runtime"

physicsnemo/utils/_index_tuple_ops.py
```diff
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Fast operations for small bounded integer index tuples."""
+"""Fast tensor operations for small bounded integer index tuples."""
```
peterdsharpe, line 17: This file makes direct references in comments/docstrings to mesh; if we're going to migrate this outside the mesh module, then this file should be genericized

physicsnemo/nn/functional/geometry/remeshing/_config.py
```diff
@@ -0,0 +1,96 @@
...
+"""Configuration for the Warp remeshing functional."""
+
+import math
+from dataclasses import dataclass
+
+_MAX_HASH_GRID_RESOLUTION = 256
+
+
+@dataclass(frozen=True, slots=True)
+class WarpRemeshOptions:
```
peterdsharpe: I'd push back pretty strongly against having an options container, and not just taking these as direct kwargs into the remeshing function. Options containers add a lot of API bloat, and it's had for end-users to understand the API by looking at the signature. If someone inspects `remesh` and sees that they need to build this custom WarpRemeshOptions to tweak a setting, we're needlessly redirecting them here

peterdsharpe: I think it's safe to delete this file and inline this into the kwargs

physicsnemo/nn/functional/geometry/remeshing/remeshing.py
```diff
@@ -0,0 +1,275 @@
...
+            )
+
+    if warp_options is not None and not isinstance(warp_options, WarpRemeshOptions):
+        raise TypeError(
+            "warp_options must be a WarpRemeshOptions instance or None, got "
+            f"{type(warp_options).__name__}"
+        )
+
+
+def _make_uv_sphere(
```
peterdsharpe, line 121: This should use the primitives in `mesh.primitives`, not duplicate it.

physicsnemo/nn/functional/geometry/remeshing/remeshing.py
```diff
@@ -0,0 +1,275 @@
...
+        cls,
+        mesh_vertices: torch.Tensor,
+        mesh_indices: torch.Tensor,
+        n_clusters: int,
+        *,
+        max_iterations: int | None = None,
+        warp_options: WarpRemeshOptions | None = None,
+        implementation: Literal["warp"] | None = None,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """Validate inputs and dispatch the Warp implementation."""
```
peterdsharpe: All public functions need docstrings

physicsnemo/nn/functional/geometry/remeshing/remeshing.py
```diff
@@ -0,0 +1,275 @@
...
+            float(options.voxel_width_scale),
+            options.hash_grid_resolution,
+            options.farthest_point_threshold,
+            options.farthest_point_oversampling,
+        )
+
+    @classmethod
+    def dispatch(
+        cls,
+        mesh_vertices: torch.Tensor,
```
peterdsharpe: missing jaxtyping here

physicsnemo/nn/functional/geometry/remeshing/remeshing.py
```diff
@@ -0,0 +1,275 @@
...
+    )
+
+    @FunctionSpec.register(
+        name="warp",
+        required_imports=("warp>=1.14.0",),
+        rank=0,
+        baseline=True,
+    )
+    def warp_forward(
+        mesh_vertices: torch.Tensor,
```
peterdsharpe: missing jaxtyping

peterdsharpe (changes requested):
Submitting an initial partial review! This PR seems to have good bones, but a *lot* of new API surface here violates repo rules by not including jaxtyping on tensor inputs. Many functions are also missing complete (i.e., conforming to repo-wide style rules, with explanations for parameters) docstrings. Would you mind adding these, to aid in the review? Thanks!

physicsnemo/mesh/remeshing/_remeshing.py
```diff
@@ -14,108 +14,139 @@
...
-        Input mesh to remesh
+        Input triangle surface. Only 2D triangle manifolds embedded in 3D are
+        supported.
     n_clusters : int
-        Target number of output vertices. The actual number may vary
-        slightly depending on mesh topology.
+        Target output vertex count. Cleanup can produce slightly fewer vertices.
+        Must be between 3 and the input point count, inclusive.
+    max_iterations : int | None, optional
+        Maximum centroid-relaxation iterations. ``None`` uses four iterations.
```
peterdsharpe: Why make `None` default to 4, rather than just having this be 4 in the kwarg default?

physicsnemo/mesh/remeshing/_remeshing.py
```diff
@@ -14,108 +14,139 @@
...
+    if warp_options is not None and not isinstance(warp_options, WarpRemeshOptions):
+        raise TypeError(
+            "warp_options must be a WarpRemeshOptions instance or None, got "
+            f"{type(warp_options).__name__}"
         )
 
-    clustering = pyacvd.Clustering(to_pyvista(mesh))
-    clustering.cluster(n_clusters)
-    new_mesh, _stats = repair_mesh(from_pyvista(clustering.create_mesh()))
+    _validate_remesh_inputs(mesh, n_clusters, max_iterations)
```
peterdsharpe: Why not inline this, if it's only called in one place here? Avoids passing scope back and forth

mehdiataei:
@peterdsharpe Regarding the use of bultins.int:

This is intentional. tensorclass generates methods like `.int()`, `.bool()`, and `.float()`. With Python 3.14’s lazy annotation lookup, a bare `int | None` inside the class can resolve to that generated method instead of the builtin type and break signature inspection. Using `builtins.int` avoids the collision. e.g.

```python
@tensorclass
class Example:
    value: torch.Tensor

    # Bare `int | None` resolves to the generated `Example.int` method on Python 3.14.
    def method(self, count: builtins.int | None = None):
        pass
```

mehdiataei:
For future reference based on our conversations:

I dug into this a bit more and the behavior is specific to Python 3.14, where annotations are evaluated lazily.

`print(int)` inside the class body still shows the builtin because it runs before `@tensorclass`. However, the method annotation is evaluated later, after `tensorclass` has added the `.int()` method. At that point, bare `int` resolves to the generated method.

Here’s a minimal repro:

```python
import builtins
from typing import get_type_hints

import torch
from tensordict import tensorclass


@tensorclass
class Bare:
    value: torch.Tensor

    def method(self, count: int | None = None):
        pass


@tensorclass
class Qualified:
    value: torch.Tensor

    def method(self, count: builtins.int | None = None):
        pass


try:
    print(get_type_hints(Bare.method))
except TypeError as error:
    print("Bare annotation failed:", error)

print("Qualified annotation:", get_type_hints(Qualified.method))
```

On Python 3.14, this prints:

```text
Bare annotation failed: unsupported operand type(s) for |: 'function' and 'NoneType'
Qualified annotation: {'count': int | None}
```

So I think keeping `builtins.int` is the safest option here. It explicitly identifies the builtin and avoids depending on when or where the annotation gets evaluated. While using string types is an alternative my understanding based on [Python doc](https://docs.python.org/3/howto/annotations.html#manually-un-stringizing-stringized-annotations) is that this is more future proof and safer.

physicsnemo/nn/functional/geometry/remeshing/_warp_impl/launch_forward.py
```diff
@@ -0,0 +1,560 @@
...
+    generator = torch.Generator(device=points.device).manual_seed(0)
+    # A tiny positive floor keeps isolated input vertices selectable only when
+    # the requested candidate count leaves no alternative.
+    sampling_weights = vertex_areas.clamp_min(torch.finfo(torch.float32).tiny)
+    candidate_indices = torch.multinomial(
+        sampling_weights,
+        candidate_count,
+        replacement=False,
+        generator=generator,
+    )
```
peterdsharpe: **[high] `torch.multinomial` caps at 2^24 categories — remeshing crashes for inputs above ~16.7M vertices.**

Both seeding paths sample over all `n_points` categories, and `torch.multinomial` rejects more than 2^24 categories on both CPU and CUDA (verified on torch 2.12). This path (`_select_fps_centroids`) runs whenever `n_clusters <= farthest_point_threshold` (default 256) — exactly the "decimate a huge scan to a coarse mesh" use case. Repro:

```python
pts = torch.rand(2**24 + 100, 3, device="cuda")
areas = torch.rand(2**24 + 100, device="cuda")
_select_fps_centroids(pts, areas, 128, 4)
# RuntimeError: number of categories cannot exceed 2^24
```

The stratified fill path (line 165) has the same cap, though it only triggers on sparse voxelizations.

Weighted sampling without replacement has a cap-free standard formulation (exponential race / Gumbel top-k) that would drop in here:

```python
keys = torch.empty_like(sampling_weights).exponential_(generator=generator).div_(sampling_weights)
candidate_indices = torch.topk(keys, candidate_count, largest=False).indices
```

If you'd rather defer, at minimum raise a clear `ValueError` up front and document the limit — the internal error surfacing today ("number of categories cannot exceed 2^24") is confusing.

physicsnemo/nn/functional/geometry/remeshing/_warp_impl/launch_forward.py
```diff
@@ -0,0 +1,560 @@
...
+
+def _select_fps_centroids(
+    points: torch.Tensor,
+    vertex_areas: torch.Tensor,
+    n_clusters: int,
+    farthest_point_oversampling: int,
+) -> torch.Tensor:
+    """Select high-quality seeds with FPS over an area-weighted candidate set."""
+    candidate_count = min(points.shape[0], farthest_point_oversampling * n_clusters)
+    generator = torch.Generator(device=points.device).manual_seed(0)
```
peterdsharpe: **[nit]** Seeding is hard-wired to `manual_seed(0)` (also line 152). Deterministic-by-default is a good choice, but it means a user facing a pathological seeding can never get an independent re-draw, and the docs don't mention that the sampling is fixed-seed. Worth a one-line note in the docstring, or a `seed` kwarg on the functional.

physicsnemo/nn/functional/geometry/remeshing/_warp_impl/launch_forward.py
```diff
@@ -0,0 +1,560 @@
...
+
+    used_centroids, compact_cells = torch.unique(
+        output_cells.reshape(-1),
+        sorted=True,
+        return_inverse=True,
+    )
+    output_points = centroids[used_centroids].to(dtype=normalization_center.dtype)
+    output_points = output_points * normalization_scale + normalization_center
+    output_points = output_points.to(dtype=source_points.dtype)
+    output_cells = compact_cells.reshape(-1, 3).to(dtype=torch.int64)
```
peterdsharpe, line 391: **[nit]** Two distinct centroids can project to the *same* surface point, so the output can contain coincident (duplicate-position) vertices — compaction here only dedups by index. `validate()` passes today, but a note in the docstring (or a cheap positional weld) would make the output contract crisper.

physicsnemo/nn/functional/geometry/remeshing/_warp_impl/launch_forward.py
```diff
@@ -0,0 +1,560 @@
...
+    normalized = ((working_points - center) / safe_scale).to(torch.float32)
+    normalized_lower = ((lower_bound - center) / safe_scale).to(torch.float32)
+    normalized_upper = ((upper_bound - center) / safe_scale).to(torch.float32)
+    return (
+        normalized.contiguous(),
+        center,
+        safe_scale,
+        normalized_lower,
+        normalized_upper,
+        2.0,
```
peterdsharpe: **[nit]** `_normalize_points_for_warp` returns the constant `2.0` through a 6-tuple, which makes call sites harder to read (`max_extent` looks data-dependent but isn't). Consider a small `NamedTuple`, or dropping the constant and letting the voxel code use the known `[-1, 1]` bound directly.

physicsnemo/mesh/mesh.py
```diff
@@ -3495,6 +3496,86 @@ def statistics(self):
...
+        *,
+        max_iterations: builtins.int = 4,
+        search_radius_scale: builtins.float = 1.6,
+        voxel_width_scale: builtins.float = 1.15,
+        hash_grid_resolution: builtins.int = 128,
+        farthest_point_threshold: builtins.int = 256,
+        farthest_point_oversampling: builtins.int = 4,
```
peterdsharpe: **[design] Five backend-specific tuning knobs are hard-committed into the top-level `Mesh` API.**

`search_radius_scale`, `voxel_width_scale`, `hash_grid_resolution`, `farthest_point_threshold`, and `farthest_point_oversampling` all encode *this* backend's data structures (hash grid, FPS seeding, voxel stratification). `max_iterations` is the only knob here with backend-independent meaning.

The PR already draws the right line for backend *selection*: `implementation=` exists on the functional but is deliberately not exposed at the `Mesh` level. Backend *tuning* belongs on the same side of that line. Suggestion: trim `Mesh.remesh` / `mesh.remeshing.remesh` to `(n_clusters, *, max_iterations=4)`, keep all six knobs on the tensor functional, and soften that package's "internal implementation" disclaimer to "advanced tensor-level API; backend-specific parameters subject to change", with a pointer from this docstring.

The asymmetry favors starting narrow: these parameters are new in this PR, so trimming them now costs nothing and re-adding them later is a compatible change — but once they ship in a release, removing them requires a deprecation cycle. If you expect quality tuning to be a routine need on real geometry, the fallback is to keep them but explicitly document them as backend-specific and unstable. (This would also dissolve most of the docstring duplication flagged on `_remeshing.py`.)

docs/api/mesh/remeshing.rst
```diff
@@ -3,33 +3,106 @@ Remeshing
...
-    remeshed = remesh(mesh, n_clusters=100)
-    print(remeshed.n_points)  # approximately 100 cluster centroids
+* Remeshing is non-differentiable. The implementation centers and scales
+  geometry before computing in ``float32``, then restores the input coordinate
+  frame and point dtype on return.
+* Warp floating-point atomics can introduce small run-to-run differences in
+  vertex positions and, near assignment ties, topology. Do not rely on bitwise
+  reproducibility.
+* Because Warp clusters by spatial distance rather than mesh connectivity,
+  extremely close disconnected sheets can share a cluster.
```
peterdsharpe: **[docs] This limitation understates the thin-feature failure mode.**

Ambient-distance clustering equally welds *connected* thin features — trailing edges, thin walls, narrow gaps — which is exactly the geometry this library's CFD audience remeshes. ACVD's connectivity-constrained region growing (plus its isolated-cluster-region removal) avoided this by construction; the new pipeline drops cluster-connectivity enforcement entirely and relies on post-hoc non-manifold face removal.

Worth one honest sentence here (e.g. "thin features separated by less than the mean cluster spacing can be welded together"), and cluster-connectivity cleanup would be a good listed item for future work.

physicsnemo/mesh/remeshing/_remeshing.py
```diff
@@ -14,108 +14,141 @@
...
+        raise TypeError(
+            f"max_iterations must be an integer, got {type(max_iterations).__name__}"
+        )
+    if max_iterations < 0:
+        raise ValueError(f"max_iterations must be non-negative, got {max_iterations}")
+    if mesh.n_cells == 0:
+        raise ValueError("remesh requires at least one triangle")
+    if not torch.is_floating_point(mesh.points):
+        raise TypeError(
+            f"mesh points must use a floating-point dtype, got {mesh.points.dtype}"
```
peterdsharpe: **[maintainability]** This block re-implements nearly all of the functional's `_validate_inputs` (n_clusters type/range, max_iterations, nonempty cells, floating dtype) with slightly different wording ("input point count" vs "input vertex count"). Together with the six tuning params documented across five docstrings (`Mesh.remesh`, this function, `Remeshing`, `warp_forward`, `dispatch`), the sync burden is real — and the signature test only guards names, not text.

Suggest the mesh layer keep only what's mesh-specific — the `n_manifold_dims` / `n_spatial_dims` gate — and let the functional own the scalar validation.

physicsnemo/nn/functional/geometry/remeshing/_warp_impl/_kernels/project_centroids_to_surface.py
```diff
@@ -0,0 +1,40 @@
...
+        mesh_id, centroids[centroid_index], max_distance
+    )
+    if query.result:
+        mesh = wp.mesh_get(mesh_id)
+        p0 = mesh.points[mesh.indices[3 * query.face + 0]]
+        p1 = mesh.points[mesh.indices[3 * query.face + 1]]
+        p2 = mesh.points[mesh.indices[3 * query.face + 2]]
+        centroids[centroid_index] = (
+            query.u * p0 + query.v * p1 + (float(1.0) - query.u - query.v) * p2
+        )
```
peterdsharpe: **[suggestion]** Two simplifications available here:

1. The barycentric reconstruction reimplements `wp.mesh_eval_position(mesh_id, query.face, query.u, query.v)` — the builtin is convention-proof against any future change in Warp's u/v ordering.
2. The sign from `mesh_query_point_sign_normal` is unused — `wp.mesh_query_point_no_sign` computes the same closest point more cheaply and sidesteps sign edge cases on open surfaces. (The repo's other uses of `sign_normal`, e.g. in `sdf`, actually need the sign; this one doesn't.)

CHANGELOG.md
```diff
@@ -37,6 +37,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Adds `rectilinear_grid_divergence`, `rectilinear_grid_curl`, and
   `rectilinear_grid_laplacian` to `physicsnemo.nn.functional`, with Torch and
   fused Warp implementations for periodic, nonuniform rectilinear grids.
+- Adds uniform triangle-surface remeshing with NVIDIA Warp on CPU and CUDA,
+  including `remesh`, `Mesh.remesh`, topology cleanup, and direct keyword
+  controls for runtime tuning.
```
peterdsharpe: **[docs]** The pyacvd removal from `mesh-extras` isn't recorded — it belongs under `### Dependencies` (or `### Removed`) alongside the existing warp-lang 1.14.0 entry, since it changes what `pip install physicsnemo[mesh-extras]` pulls in.

examples/minimal/mesh/assets/make_bunny.py
```diff
@@ -22,10 +22,13 @@
...
 from physicsnemo.mesh.io.io_pyvista import from_pyvista
-from physicsnemo.mesh.remeshing._remeshing import remesh
+from physicsnemo.mesh.remeshing import remesh
 
 mesh = from_pyvista(pv.examples.download_bunny_coarse())
-mesh = remesh(mesh.clean().subdivide(levels=3, filter="linear"), 400)
+mesh = remesh(
+    mesh.clean().subdivide(levels=3, filter="linear"),
+    400,
+)
```
peterdsharpe, line 31: **[important]** This script now produces a different `bunny.pt` than the checked-in asset (which was generated with pyacvd). Please regenerate the asset with the new remesher — as-is, re-running the script silently changes the tutorials' geometry.

physicsnemo/nn/functional/geometry/remeshing/remeshing.py
```diff
@@ -0,0 +1,449 @@
...
+from physicsnemo.core.function_spec import FunctionSpec
+
+_INTEGER_DTYPES = {
+    torch.int8,
+    torch.int16,
+    torch.int32,
+    torch.int64,
+    torch.uint8,
+}
+_MAX_HASH_GRID_RESOLUTION = 256
```
peterdsharpe, line 37: **[nit]** The 256 cap is asserted but never motivated. A short clause in the docstring ("bounds hash-grid memory to ~135 MB at 256³", or similar) would preempt the obvious questions of why the limit exists and whether it's safe to raise.

physicsnemo/nn/functional/geometry/remeshing/_warp_impl/_kernels/assign_vertices.py
```diff
@@ -0,0 +1,68 @@
...
+        if candidate_index < centroids.shape[0]:
+            delta = point - centroids[candidate_index]
+            distance_sq = wp.dot(delta, delta)
+            if distance_sq <= radius_sq and distance_sq < best_distance_sq:
+                best_distance_sq = distance_sq
+                best_index = candidate_index
+
+    # A well-spaced initialization makes this path rare. Retain it for
+    # disconnected inputs and highly nonuniform meshes.
+    if best_index < 0:
```
peterdsharpe, line 56: **[test-coverage]** This brute-force fallback (no centroid within `search_radius`) is never deliberately exercised by the test suite — the separated-components test likely keeps every vertex within radius of a same-sphere centroid. A tiny mesh with an outlier vertex and a `search_radius_scale` small enough to miss it would pin this path down.

test/mesh/remeshing/test_remesh_warp.py
```diff
@@ -0,0 +1,325 @@
...
+    source = sphere_icosahedral.load(subdivisions=2, device="cuda")
+    direct = remesh(source, 48)
+    method = source.remesh(48)
+
+    torch.testing.assert_close(direct.cells, method.cells)
+    _assert_clean_topology(direct)
+    _assert_clean_topology(method)
+
+
+def test_warp_remesh_preserves_dtype_and_accepts_noncontiguous_geometry():
```
peterdsharpe: **[test-coverage]** dtype coverage includes float64 but not reduced precision — the normalize/restore path advertises handling any floating dtype (`working_dtype` falls back to float32). A float16 or bfloat16 case would be cheap here and lock in the dtype-restore contract.

peterdsharpe (changes requested):
This looks good — nice work. I checked out the branch and ran the test suites locally (109 passed, CPU + CUDA).

Key comments to fix below:

- The 2^24 `torch.multinomial` limit: remeshing currently crashes above ~16.7M input vertices (repro and a drop-in fix in the inline comment).
- The tuning knobs (`search_radius_scale`, `hash_grid_resolution`, etc.) might be best left out of the `Mesh.remesh` signature and kept on the tensor functional — if you agree? Once these ship in a release we are committed to them.
- Validation logic should be deduplicated between `mesh.remeshing.remesh` and the functional's `_validate_inputs`.
- `bunny.pt` should get regenerated so the checked-in asset matches the new remesher.

Everything else below is a suggestion / nit / optional.

peterdsharpe:
Following up on the 2^24 `torch.multinomial` finding from my review: rather than patching the sampler in place, it might be cleaner to replace both seeding paths (`_select_fps_centroids` and `_select_stratified_centroids`) with the existing `mesh_poisson_disk_sample` functional — "K well-spaced, area-weighted points on the surface" is exactly its output contract, and it's on the same Warp stack.

That would:

- Fix the crash as a side effect: its candidate draw is an area-weighted triangle CDF in Warp (no `torch.multinomial`, so no 2^24 cap).
- Delete ~120 lines here, plus three of the five tuning knobs (`voxel_width_scale`, `farthest_point_threshold`, and `farthest_point_oversampling` exist only to configure the hand-rolled seeding) — which also largely resolves my comment on the `Mesh.remesh` signature.
- Probably improve seeding: blue-noise beats voxel-stratification-with-random-fill as a CVT initialization, and the seeds would lie on the surface rather than at input vertices — the input vertex distribution is the thing remeshing is trying to fix.

One mode note: `weighted_sample_elimination` gives exactly K but has a host-side elimination loop, so it may regress wall-clock at large K. Since `n_clusters` is a ceiling rather than a guarantee anyway, GPU-native dart throwing with `max_points=n_clusters` and radius ≈ `0.7 * sqrt(area / n_clusters)` should preserve the `output <= n_clusters` contract — worth benchmarking both. (It also takes a `random_seed` arg, which would address the fixed-seed nit for free.)

physicsnemo/nn/functional/geometry/remeshing/_warp_impl/launch_forward.py
peterdsharpe: Proposed a concrete alternative fix for this that reuses the existing `mesh_poisson_disk_sample` functional for seeding instead of patching the sampler in place — see https://github.com/NVIDIA/physicsnemo/pull/1822#issuecomment-4996946512. It would remove this code path (and three of the tuning knobs) entirely.

mehdiataei:
Thanks for the suggestion! I benchmarked both Poisson modes against the current initializer. The current approach generates exactly $K$ area-weighted vertex seeds: FPS for small targets and voxel stratification for larger ones, using a chunked sampler that avoids the `torch.multinomial` limit. The CVT/Lloyd remeshing pipeline itself is still Warp-based and runs on either CPU or CUDA.

Weighted elimination does produce an exact $K$, but it scales poorly because of the 5× candidate pool and the host-side elimination loop. Dart throwing is faster, but it only guarantees up to $K$ points, so it needs additional fill/retry logic. Both also duplicate mesh preprocessing. On a 10.5M-vertex mesh, that added about 2.24 GiB of memory and 25–27 ms just for seeding.

There's also a compatibility issue with this CVT formulation. Since Lloyd assignment integrates over the weighted input vertices, triangle-interior Poisson seeds can start out without owning any input vertex. With $K=N=642$, the current initializer retained all 642 clusters, while dart throwing retained 557 and weighted elimination retained 569.

The replacement sampler has been tested with 100 million candidate entries, and the full Warp remeshing pipeline has been validated on a 41.9M-vertex mesh, so I think the current initializer is the safer choice for this PR. That said, dart throwing is still worth revisiting later once preprocessing is shared and there's an exact-K fallback.

peterdsharpe (approved):
LGTM, thank you!

ktangsali (approved):

physicsnemo/nn/functional/geometry/remeshing/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Advanced tensor-level surface-remeshing API.
+
+This package is intentionally not re-exported from ``physicsnemo.nn.functional``.
```
loliverhennigh: Optional / non-blocking: Should we re-export `Remeshing` and `remeshing` from `physicsnemo.nn.functional.geometry`, and expose `remeshing` from `physicsnemo.nn.functional`, like the other geometry functionals? I see the omission is intentional because this is an advanced API, but requiring the deeper import path feels inconsistent with the rest of the functional APIs.

physicsnemo/nn/functional/geometry/remeshing/remeshing.py
```diff
@@ -0,0 +1,456 @@
...
+            n_clusters,
+            max_iterations,
+            float(search_radius_scale),
+            float(voxel_width_scale),
+            hash_grid_resolution,
+            farthest_point_threshold,
+            farthest_point_oversampling,
+        )
+
+    @classmethod
```
loliverhennigh: Optional / non-blocking: Is there a reason to reimplement `FunctionSpec.dispatch` here? We typically do the validation inside `warp_forward` or the Warp implementation and rely on the inherited dispatch for backend selection. Could we move `_validate_inputs` and the device check into `warp_forward`, then remove this custom dispatch? The data-dependent checks should remain inside the custom op for FakeTensor / `torch.compile` support.

loliverhennigh (commented):
Two optional, non-blocking consistency suggestions.

loliverhennigh (approved):

physicsnemo/nn/functional/geometry/remeshing/_warp_impl/_kernels/assign_vertices.py
coreyjadams, line 56: How often is this path hit on realistic meshes and how bad is it for performance?

physicsnemo/nn/functional/geometry/remeshing/_warp_impl/launch_forward.py
```diff
@@ -0,0 +1,591 @@
...
+            _wp_view(vertex_areas, wp.float32),
+            _wp_view(labels, wp.int32),
+            _wp_view(centroid_sums, wp.float32),
+            _wp_view(centroid_areas, wp.float32),
+            search_radius,
+        ]
+
+        for _ in range(max_iterations):
+            centroid_sums.zero_()
+            centroid_areas.zero_()
```
coreyjadams: nit: you can fuse these into the project_centroids kernel I think.

coreyjadams (approved):
OK.  We should wait for Meg to review the docs or you should split it off into a stacked PR on top of this.

docs/api/mesh/remeshing.rst
```diff
@@ -3,33 +3,115 @@ Remeshing
...
+PhysicsNeMo provides Warp-based uniform remeshing on CPU and CUDA for 2D
+triangle manifolds embedded in 3D. ``n_clusters`` is the target number of
+output vertices, not triangles; cleanup can produce slightly fewer vertices.
+Point and cell data are discarded because their associations no longer match
+the reconstructed topology. Global data, point dtype, and device are
+preserved.
 
-The current implementation wraps ``pyacvd`` and therefore supports only
-triangle surfaces (2D manifolds) embedded in 3D:
+CPU and CUDA example
```
megnvidia:
```suggestion
CPU and CUDA Example
```

docs/api/mesh/remeshing.rst
```diff
@@ -3,33 +3,115 @@ Remeshing
...
+   from physicsnemo.mesh.remeshing import remesh
+
+   device = "cuda" if torch.cuda.is_available() else "cpu"
+   dense = sphere_icosahedral.load(subdivisions=6, device=device)
+   coarse = remesh(dense, n_clusters=4_096)
+
+   assert coarse.points.device == dense.points.device
+   assert 0 < coarse.n_points <= 4_096
+
+Warp tuning
```
megnvidia:
```suggestion
Warp Tuning
```

docs/api/mesh/remeshing.rst
```diff
@@ -3,33 +3,115 @@ Remeshing
...
+   )
+
+These values are host-side controls or runtime kernel arguments. Changing them
+reuses the compiled Warp kernels rather than triggering JIT recompilation.
+
+The Warp implementation uses area-weighted centroidal relaxation with a hash
+grid, projects the relaxed vertices onto the source surface using a bounding
+volume hierarchy (BVH), removes collapsed and duplicate faces, and compacts
+unused vertices. Small targets use farthest-point initialization for mesh
+quality; large targets use a linearithmic spatially stratified initializer to
```
megnvidia:
```suggestion
quality. Large targets use a linearithmic spatially stratified initializer to
```

docs/api/mesh/remeshing.rst
```diff
@@ -3,33 +3,115 @@ Remeshing
...
+.. image:: /img/mesh/remeshing_comparison.png
+   :alt: Dense Stanford bunny beside its Warp-remeshed result
+   :align: center
+   :width: 72%
+
+Performance
+-----------
+
+The checked-in ASV benchmark measures warmed, end-to-end GPU execution:
+clustering, surface projection, topology reconstruction, and cleanup. Timing
```
megnvidia:
```suggestion

- clustering
- surface projection
- topology reconstruction
- cleanup

```

docs/api/mesh/remeshing.rst
```diff
@@ -3,33 +3,115 @@ Remeshing
...
+includes an explicit CUDA synchronization.
```
megnvidia:
```suggestion
Timing includes an explicit CUDA synchronization.
```

docs/api/mesh/remeshing.rst
```diff
@@ -3,33 +3,115 @@ Remeshing
...
+baseline.
+
+.. code:: console
+
+   ./benchmarks/run_benchmarks.sh -b remesh
+
+The figure below is a representative run of
+``docs/img/mesh/remeshing_performance.py`` on an NVIDIA RTX PRO 6000 Blackwell
+Server Edition MIG 1g.24GB partition using Warp 1.14.0. Absolute timings depend
+on hardware and software versions; use the ASV benchmark above for measurements
```
megnvidia:
```suggestion
on hardware and software versions. Use the ASV benchmark above for measurements
```

docs/api/mesh/remeshing.rst
```diff
@@ -3,33 +3,115 @@ Remeshing
...
+in another environment.
+
+.. image:: /img/mesh/remeshing_performance.png
+   :alt: GPU remeshing runtime plot across increasing input sizes
+   :align: center
+   :width: 65%
+
+Behavior and limitations
```
megnvidia:
```suggestion
Behavior and Limitations
```

docs/api/mesh/remeshing.rst
```diff
@@ -3,33 +3,115 @@ Remeshing
...
+* Because clustering uses spatial distance rather than mesh connectivity,
+  sheets or thin features separated by less than the mean cluster spacing can
+  be assigned to a common cluster and welded together.
+* Projection can map distinct cluster centroids to the same surface position.
+  Output vertices are compacted by connectivity but are not welded by
+  position.
+* The optional ``max_iterations`` argument defaults to four centroid updates.
 
-API Reference
+API reference
```
megnvidia:
```suggestion
API Reference
```

megnvidia (approved):
approved with changes.

megnvidia (approved):

PR #1825, opened 2026-07-10
feat(datapipes): add ComputeUnitGlobalVector transform

peterdsharpe:
## What

A small opt-in transform, `ComputeUnitGlobalVector`:

- Computes the unit direction `v / |v|` of a vector stored in `global_data`.
- Stores it under a new key; the source vector is not modified.
- Works on `Mesh` and, at the domain level, on `DomainMesh`.
- Rejects bad input loudly: missing field, zero or non-finite vector, integer dtype, batched `(B, 3)` input, and `output_field == vector_field` all raise clear errors.

## Why

- Pipelines often nondimensionalize fields and geometry but leave `global_data` in physical units, because the physical values are needed later to convert predictions back (e.g. force coefficients).
- A model that takes a freestream velocity as input then sees a large dimensional vector (tens of m/s), which can destabilize scale-sensitive models at initialization.
- The unit direction is the scale-free input such models should see. Placed before a rotation augmentation with `transform_global_data: true`, it rotates together with the source vector and the geometry.

## Why a draft

This is new API rather than a bug fix (the bug fix it was originally bundled with is #1820), so a few design questions are worth settling first:

- On `DomainMesh`, should the direction also be broadcast to sub-mesh `global_data` (as `SetGlobalField` does), or stay domain-level only as implemented?
- For datasets that vary inflow speed, should this ship with a companion transform for a nondimensional magnitude (e.g. `|v| / v_ref`), so speed information is not silently dropped?
- Should batched `(B, 3)` vectors be normalized per row? (Currently rejected with an error; transforms run per-sample, where vectors are `(3,)`.)
- Does this belong in core, or should it stay recipe-local until there is a second consumer?

## Tests

New `test/datapipes/transforms/test_compute_unit_global_vector.py` (10 tests):

- direction math on `Mesh` and `DomainMesh`
- source vector and sub-mesh `global_data` unchanged
- output keeps the source dtype, so a later rotation composes without mixed-dtype errors
- clear errors on missing field, zero vector, integer dtype, batched input, and same input/output field
- resolvable via the `${dp:...}` config resolver

Full transforms suite passes (287); ruff check/format clean.


peterdsharpe:
On second thought, thinking of keeping this recipe-local until a second use case emerges

Issue #1796, opened 2026-07-06
ShardTensor is not torch.compile-able: graph break in from_local resumes mid-construction before _spec is attached


negin513:
## Summary

Attempting to `torch.compile` a function that constructs a `ShardTensor` via `from_local` and calls `redistribute` fails on current `main` (post-#1556). Two distinct symptoms, one root cause: dynamo graph-breaks inside the ShardTensor construction path, and the resume machinery can't recover.

## Repro (4 GPUs, `torchrun --nproc_per_node=4`)

```python
import torch
import torch.distributed as dist
from torch.distributed.device_mesh import init_device_mesh
from torch.distributed.tensor.placement_types import Shard
from physicsnemo.domain_parallel import ShardTensor

dist.init_process_group("nccl")
rank, world = dist.get_rank(), dist.get_world_size()
torch.cuda.set_device(rank)
mesh = init_device_mesh("cuda", (world,), mesh_dim_names=["cp"])
S1, S2 = [Shard(1)], [Shard(2)]

def st_flat(local, g1):
    a = ShardTensor.from_local(local, mesh, S1, sharding_shapes="chunk", global_shape=g1)
    return a.redistribute(placements=S2).to_local()

local = torch.randn(1, 2, 12288, 1536, device=f"cuda:{rank}", dtype=torch.bfloat16)
torch.compile(st_flat)(local, (1, 8, 12288, 1536))
```

(The `global_shape` kwarg is from #1791; the same failure occurs with default `sharding_shapes="infer"` on unmodified `main` -- the graph break just happens slightly earlier in spec inference.)

## Failure mode 1 -- flat call: `InternalTorchDynamoError: AttributeError: 'ShardTensor' object has no attribute '_spec'`

Traceback sequence (PyTorch 2.12 nightly, `e53ae1a6`):

1. Dynamo traces `from_local` -> `_FromTorchTensor.apply` -> `forward`, and **graph-breaks** at the `_infer_shard_tensor_spec_from_local_chunks` call (Python-heavy spec construction).
2. It synthesizes a resume frame (`torch_dynamo_resume_in_forward_at_562`) and continues at `shard_tensor = ShardTensor(...)`.
3. Tracing resumes **inside `ShardTensor.__new__`, mid-construction**: `torch.Tensor._make_wrapper_subclass(...)` has returned, but `_spec` has not been assigned yet.
[...]

PR #1838, opened 2026-07-14
fix: make ShardTensor construction opaque to dynamo to prevent graph …

KhushviB:
### Description (#1796)
Fixes `torch.compile` failures that occur when constructing a `ShardTensor` (e.g. via `from_local` and `redistribute`).

**The Bug:**
Dynamo was graph-breaking inside the `ShardTensor` construction path during spec inference (inside `_FromTorchTensor.forward`). When Dynamo attempted to synthesize a resume frame mid-construction (inside `ShardTensor.__new__`), it introspected a half-constructed wrapper subclass. This prematurely triggered `__tensor_flatten__`, which tried to access `self._spec` before it was assigned, causing an `AttributeError`. Additionally, when wrapped in nested helper functions, the graph break produced a resume frame that mis-resolved enclosing-scope names (yielding a `NameError`).

**The Fix:**
Added the `@torch._dynamo.disable` decorator to `_FromTorchTensor.forward` and `ShardTensor.__new__`. 
This explicitly marks the construction path as opaque, instructing Dynamo to graph-break *before* the function, execute the construction eagerly as a black box, and safely resume tracing *after* the `ShardTensor` is returned.

This eliminates the `AttributeError` and nested scope mis-resolutions, closing the eager-mode gap for `ShardTensor` resharding and enabling the removal of manual `@torch._dynamo.disable` wrappers from reshard call sites (like in the HealDA architecture).

### DIFF
--- a/physicsnemo/domain_parallel/shard_tensor.py
+++ b/physicsnemo/domain_parallel/shard_tensor.py
@@ -516,6 +516,7 @@ class _FromTorchTensor(torch.autograd.Function):
     """
 
     @staticmethod
+    @torch._dynamo.disable
     def forward(
         ctx: torch.autograd.function.FunctionCtx,
         local_input: torch.Tensor,
@@ -783,6 +784,7 @@ class ShardTensor(torch.Tensor):
         cls._named_function_registry[func_name] = handler
 
     @staticmethod
+    @torch._dynamo.disable
     def __new__(
         cls,
         local_tensor: torch.Tensor,


coreyjadams:
Hi @KhushviB - thanks for your contribution and interest in ShardTensor.  It is not, as of today, officially supporting torch.compile.  We have a PR we are actively working on a reviewing to enable this (#1682).  

I don't want to disable dynamo through ShardTensor, actually, we want to enable it fully through the whole execution chain :).  That PR is most stable and functional, we're ironing out some final bugs, and should merge it soon.

I'll close this PR, since disabling is not our goal, but please go ahead and try on the other PR and let me know how badly things break (or work) for you?  We would love to get feedback and reproducers of edge cases that we haven't seen yet.

Thanks!
Corey

PR #1844, opened 2026-07-16
Unified aero recipe: correct GLOBE training guidance

peterdsharpe:
## Summary

Makes `model=globe_surface` / `model=globe_volume` train correctly out of the box in the unified recipe, and removes a harmful legacy loss weight.

**Config: model templates now carry known-good training overrides.** `train.yaml`'s defaults list previously merged `_self_` last, so model templates could not override the centralized training schedule. `_self_` now merges before the model template; CLI overrides still win over both. All other model templates resolve **identically** under the new order (verified by composing every template before and after the change). The GLOBE gallery example now uses the recipe defaults for `compile` and learning rate: both former overrides were measured against the defaults on the DrivAerML surface case. `compile=true` works (GLOBE already excludes its untraceable tree phases via `@torch.compiler.disable`) and is ~4% faster in steady state (~17% in validation) after ~90 s of one-time warmup; the default lr (3e-3 + StepLR) matches lr=1e-2 within single-seed spread by epoch 15 (epoch-20 val WSS L2 0.274 vs 0.269, val pressure L2 0.236 vs 0.237).

**Removes `training.field_weights={pressure: 1.0, wss: 100.0}` from the GLOBE example.** The surface dataset YAMLs already bring every target to unit scale via `NormalizeMeshFields`, so the extra ×100 starved the pressure field of gradient signal: on the DrivAerML surface case (50K sampled cells), dropping it improved converged pressure L2 from ~0.30 to ~0.14 at equal WSS L2 (~0.16), and makes the logged `loss/wss` directly comparable across recipe models. The `field_weights` documentation now states the general rule (normalized targets ⇒ leave unset; logged `loss/<field>` includes the weight).

Also adds a note on GLOBE's memory envelope at `sampling_resolution=50000` to the recipe gallery.

## Testing

- Composed all 11 model templates before/after the defaults reorder: all resolve identically (the reorder is a provable no-op today; it enables measured per-model defaults later).
- Recipe test suite (155 tests) passes with the new configs.
- The recommended configuration is validated by a 200+-epoch DrivAerML surface training run on 4×GB200 reaching val pressure L2 0.141 / WSS L2 0.160, vs 0.30 / 0.162 for an otherwise-identical run with the removed weight.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -274,10 +274,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Unified external aerodynamics recipe: model templates can now carry
   known-good training overrides (`train.yaml`'s `_self_` merges before the
   model template; all existing templates resolve identically). The GLOBE
-  example no longer sets `training.field_weights={pressure: 1.0, wss:
-  100.0}` — the weight was redundant with `NormalizeMeshFields`
-  normalization and starved the pressure field of gradient signal (~2x
-  worse converged pressure L2 on DrivAerML surface at equal WSS L2).
+  example now uses the recipe's default `compile` and learning-rate
+  settings — both measured equivalent or better than the previously
+  documented overrides on the DrivAerML surface case — and no longer sets
+  `training.field_weights={pressure: 1.0, wss: 100.0}`, which was redundant
+  with `NormalizeMeshFields` normalization and starved the pressure field
+  of gradient signal (~2x worse converged pressure L2 at equal WSS L2).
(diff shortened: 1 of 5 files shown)


coreyjadams (approved):

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1844/commits/1868949e69c01e8b912f6d65f24cbe5b58bac36e

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1844/commits/57558eaef594b5dcd3872cb5472ad07574d4e7dc

PR #1854, opened 2026-07-17
Unified Aero Recipe: fix direction-blind aggregate vector metrics

peterdsharpe:
## Description

Fixes a direction-blindness defect in the unified external aero recipe's shared metric path (`src/metrics.py`, `MetricCalculator`), which affects every model and dataset evaluated through the recipe.

**Defect.** For vector fields, the aggregate metrics reported under the bare field name (e.g. `wss_l2`, likewise `_l1` / `_mae`) were computed on per-point vector *magnitudes*, so direction errors were invisible: a prediction with the correct magnitude but wrong direction at every point scored 0 ("perfect") on the headline aggregate. Per-component metrics (`wss_x_l2`, ...) were always computed componentwise and are direction-sensitive, so tables reporting components alongside the aggregate contain honest columns — the misleading number is specifically the bare-field-name aggregate.

**Fix.**

- The bare-name aggregate now computes each metric over all components jointly, flattened so the relative norms reduce over the whole field (Frobenius for `l2`), making direction errors count.
- The broken magnitude-only aggregate is removed rather than retained under a separate metric key.
- Existing metric key names are preserved, but their vector aggregate semantics are corrected.

**Tests.** An adversarial regression uses `pred = -target` (correct magnitude, maximally wrong direction everywhere): the bare-name relative l1/l2 metrics must equal 2, and no magnitude-only keys may be emitted. A second test pins aggregate `l2` to the whole-field Frobenius ratio — not a mean of per-point ratios, which explodes wherever the target magnitude is near zero (e.g. WSS at stagnation points).

**Scope of impact.** Any vector field evaluated through the recipe's metric path — training/validation metric tables and `infer.py` summaries alike — for surface (wall shear stress) and volume (velocity/momentum-type) fields, across all model configs run through the recipe. Training and checkpoints are unaffected: the training/validation objective goes through `LossCalculator`, not this metric path. Only reported aggregate vector metrics flattered direction errors.

Existing historical bare-name vector metrics are not directly comparable with metrics produced after this fix because the meaning of the existing key has intentionally changed.

## Checklist

## Dependencies

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -276,12 +276,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `_mae`) were computed on per-point vector magnitudes, so direction errors
   were invisible — a prediction with the correct magnitude but wrong direction
   at every point scored 0. The bare-name aggregate is now computed over all
-  components jointly (whole-field relative norms, Frobenius for `l2`), and the
-  direction-blind magnitude comparison moved to an explicit
-  `<field>_mag_<metric>` key. Per-component metrics (`wss_x_l2`, ...) were
-  always direction-sensitive and are unchanged, and training/checkpoints are
-  unaffected (the training objective goes through `LossCalculator`, not this
-  metric path) — only reported aggregate vector metrics were misleading.
+  components jointly (whole-field relative norms, Frobenius for `l2`). The
+  broken magnitude-only aggregate is not retained under a separate key.
+  Per-component metrics (`wss_x_l2`, ...) were always direction-sensitive and
+  are unchanged, and training/checkpoints are unaffected (the training
+  objective goes through `LossCalculator`, not this metric path) — only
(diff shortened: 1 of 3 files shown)


peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1854/commits/77c3efce45384ca55518ebd74084257646c2121a

coreyjadams:
I think we should either:
a) replace the original metrics with the corrected metrics, or
b) update the key or the original metrics with `_wrong` where necessary, or similar.

This is a really substantial bug, I think, that merits a breaking change.

coreyjadams (approved):
I am in favor of more aggressive fixing, here, and just killing broken behavior rather than leaving it hanging.  Approving to unblock though.

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1854/commits/de7a69210ca282a5b9106966751a071ec698091e

PR #1859, opened 2026-07-21
fix(examples): reject multi-dataset manifest sampling

peterdsharpe:
## Summary

- fail loudly when manifest sampling is combined with any additional dataset
- add an integration regression through `build_dataloaders` for mixed manifest/directory mode
- document the single-dataset manifest limitation and remove the advertised DrivAerML + SHIFT command

## Why

Manifest indices are resolved in each dataset's local coordinate system, but `build_dataloaders` stores only one train/validation index pair and later applies it to the concatenated `MultiDataset`. Multiple manifest datasets overwrite earlier indices; mixing manifest and directory datasets can select the wrong prefix or omit a dataset entirely.

Until the loader accumulates every dataset's indices with the appropriate offsets, rejecting this configuration prevents silent mis-sampling.

## Validation

- `pytest -q examples/cfd/external_aerodynamics/unified_external_aero_recipe/tests` (158 passed)
- `ruff format --check` on the changed Python files
- `ruff check` on the changed Python files
- `git diff --check`


coreyjadams (approved):
LGTM

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1859/commits/7519e49a62670a567b0cbc9bf92e8d7f5873bd03

PR #1862, opened 2026-07-21
Docs: fix mesh image paths

mehdiataei:
## Summary

- use document-relative paths for four images in the mesh API documentation
- fix image rendering in GitHub's RST preview
- preserve the existing Sphinx image locations and output

## Root cause

The mesh documentation used source-root paths such as `/img/mesh/sphere_ffd.png`.
Sphinx resolves those paths from the `docs` source root, but GitHub's RST preview
does not. Resolving the paths relative to each RST file works in both renderers.

## Validation

- ran the repository pre-commit hooks on both changed RST files
- built the documentation with `sphinx-build -b html`
- confirmed all four images were copied and referenced in the generated HTML
- confirmed no source-root image directives remain under `docs/api/mesh`

### DIFF
--- a/docs/api/mesh/remeshing.rst
+++ b/docs/api/mesh/remeshing.rst
@@ -64,7 +64,7 @@ unused vertices. Small targets use farthest-point initialization for mesh
 quality. Large targets use a linearithmic spatially stratified initializer to
 avoid quadratic setup cost.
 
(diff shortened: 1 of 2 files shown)


peterdsharpe (approved):
lgtm

PR #1863, opened 2026-07-22
PhysicsNeMo-Mesh: modernize containers and weighted sampling

### LINKED ISSUE #1846: 🐛[BUG]: Weighted sampling of >2^24 points crashes (datapipes) or returns wrong indices (DoMINO); consolidate on the exact chunked sampler from #1822
### Version

main (post-2.1.0, 2026-07-16)

Source (device-independent; the `torch.multinomial` 2^24-category cap applies on both CPU and CUDA, verified on torch 2.12.0)

### Describe the issue

PhysicsNeMo now has three independent implementations of "draw K indices from N points, area/solution-weighted, without replacement, where N can exceed `torch.multinomial`'s 2^24-category cap". Two of them are defective; the third (added in #1822) is exact and should become the shared utility.

**1. `physicsnemo/datapipes/transforms/subsample.py` — weighted path crashes for N > 2^24 (verified).**
`shuffle_array` guards the >2^24 case only on the *uniform* path (via `poisson_sample_indices_fixed`). The weighted path at `subsample.py:187` calls `torch.multinomial` unguarded, so weighted subsampling of more than 16,777,216 points raises `RuntimeError: number of categories cannot exceed 2^24`.

**2. `physicsnemo/models/domino/utils/utils.py` — chunked path returns wrong indices for N > 2^24 (verified).**
`shuffle_array` (`utils.py:498`) splits the weights into ≤2^24-sized chunks and runs `torch.multinomial` per chunk, but the chunk-local indices are concatenated **without adding chunk offsets** (`utils.py:597-601`). For N > 2^24 the function therefore:

- only ever returns indices from the low-index region of the array (all indices < chunk size);
- can return the same global index multiple times despite `replacement=False`;
- effectively ignores the weight distribution across chunks.

Empirically: with 99.9996% of the weight mass placed in the upper region of a 20.9M-point array, 100% of returned samples came from the low-weight head region. `area_weighted_shuffle_array` and `solution_weighted_shuffle_array` delegate to this function, so DoMINO's area- and solution-weighted subsampling both inherit this for large inputs. Additionally, even with offsets fixed, the per-chunk quota is allocated proportional to chunk *size* rather than chunk *weight mass*, which remains statistically biased relative to a true global without-replacement draw, and a chunk whose quota exceeds its number of nonzero weights raises `invalid multinomial distribution` (also verified).

**3. `physicsnemo/nn/functional/geometry/remeshing/_warp_impl/launch_forward.py` (added in #1822) — exact, uncapped.**
`_weighted_sample_without_replacement` implements a chunked exponential race (keys `E_i / w_i`, per-chunk bottom-k, global bottom-k over the union). The per-chunk bottom-k union provably contains the global bottom-k, so chunking changes memory use, not the sampling distribution. It is exact, deterministic under its generator, and has unit tests (including a `torch.multinomial` monkeypatch ban and a forced multi-chunk merge).

Naming note: `datapipes.transforms.subsample.shuffle_array` and `models.domino.utils.utils.shuffle_array` are two different functions with the same name and near-identical purpose — one more reason to consolidate.

### Proposed fix

1. Promote `_weighted_sample_without_replacement` from the remeshing internals to `physicsnemo/utils/` as a shared tensor utility, together with its unit tests (precedent: `_index_tuple_ops.py` moved there in #1822 when `nn.functional` needed shared tensor ops), and re-import it in the remeshing backend.
2. Use it for the weighted path of `datapipes.transforms.subsample.shuffle_array` — a straight bug fix.
3. Replace the chunked-multinomial path in `models.domino.utils.utils.shuffle_array` with it. Note for DoMINO owners: for >2^24 inputs this changes results from wrong-indices to a correct weighted draw, which will change sampling behavior in any pipeline that currently exercises that path.
4. Keep `poisson_sample_indices_fixed` for the *uniform* >2^24 case, where its O(k) memory is cheaper than the race's O(chunk).

```python
import torch

# 1) datapipes: weighted subsampling crashes above 2^24 points
from physicsnemo.datapipes.transforms.subsample import shuffle_array as dp_shuffle
n = 2**24 + 1
dp_shuffle(torch.rand(n, 1), 1000, weights=torch.rand(n))
# RuntimeError: number of categories cannot exceed 2^24 [16777216]

# 2) domino: chunked path returns wrong (un-offset, weight-ignoring) indices
from physicsnemo.models.domino.utils.utils import shuffle_array as domino_shuffle
n2 = 2**24 + 2**22
w = torch.full((n2,), 1e-6)
w[2**24:] = 1.0                      # ~99.9996% of weight mass in the tail
_, idx = domino_shuffle(torch.rand(n2, 1), 10_000, weights=w)
print((idx < 2**24).float().mean())  # 1.0 — all samples from the low-weight head
print(idx.max())                     # < chunk size: offsets were never added

# 2b) domino: zero-weight chunk with nonzero quota crashes
w2 = torch.zeros(n2)
w2[2**24:] = 1.0
domino_shuffle(torch.rand(n2, 1), 10_000, weights=w2)
# RuntimeError: invalid multinomial distribution (sum of probabilities <= 0)
```

### Relevant log output

```shell
RuntimeError: number of categories cannot exceed 2^24 [16777216]
RuntimeError: invalid multinomial distribution (sum of probabilities <= 0)
```

### Environment details

```shell
Bare-metal, source install; torch 2.12.0+cu130; reproduced on CPU (cap and behavior are device-independent)
```

### LINKED ISSUE #1851: 🚀[FEA]: PhysicsNeMo-Mesh Code Cleanups

Improvement

Low (would be nice)

Aggregating a list of improvements to be merged into `physicsnemo.mesh`; holding off on these for now while the new morph-related PRs merge, so as not to create too many merge conflicts.

-----

- Have `Mesh` and `DomainMesh` inherit from `TensorClass`, not `@tensorclass`. Sidesteps the need for `builtin.int` and such typing, due to tensorclass overrides (see https://github.com/pytorch/tensordict/issues/1243)

-----

- Try having many `Mesh` methods be direct assigns, not re-exposures with new docstrings. Specifically, right now:
```python
@tensorclass
class Mesh:
   ...
   def do_thing(self, arg1, arg2,...):
       from mesh.module import do_thing_func
       return do_thing_func(self, arg1, arg2)
```
Proposal:
```python
from mesh.module import do_thing_func
@tensorclass
class Mesh:
   ...
   do_thing = do_thing_func
```
This works for all functions where the first argument is a Mesh, and we're only acting on one mesh. That way, we don't need to have two copies of the docstring (which create two sources of truth which will get out-of-sync)

-----

`Mesh.strip_caches()` should probably get a `keep` arg, in case you want to explicitly opt-in to keeping certain cache keys. This should then get propagated down to all cache manipulators. This lets you assign new cache keys, without having tons of places to re-update throughout the code. Ideally, consumers should never need to directly manipulate `Mesh._cache`.

-----

Also should fix/investigate https://github.com/NVIDIA/physicsnemo/issues/1846

_No response_

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Implements the PhysicsNeMo-Mesh cleanup work tracked in #1851 and consolidates
the uncapped weighted sampler described in #1846.

Closes #1851
Closes #1846

### What changed

- Make `Mesh` and `DomainMesh` inherit directly from `TensorClass`, while
  preserving dimension-parameterized `Mesh[m, s]` behavior.
- Reuse canonical transformation and visualization functions directly instead
  of maintaining forwarding methods and duplicate docstrings.
- Add `strip_caches(keep=...)` and propagate selective cache retention through
  `DomainMesh`.
- Promote the exact chunked exponential-race sampler to a shared utility and use
  it in datapipes, DoMINO, and remeshing. This removes the
  `torch.multinomial` `2^24` category limit and fixes DoMINO's chunk-local index
  and per-chunk quota bugs.
- Add current and decorator-era `.pmsh` / `.pdmsh` golden fixtures, writer
  manifests, and exact reconstructed-type assertions.

### Serialization compatibility

New code reads both the previous decorator-backed layout and the new root-level
`TensorClass` layout, reconstructing exact `Mesh` / `DomainMesh` types including
interior and boundary meshes. Current writers use the new layout. Older releases
may load newly written data as plain `TensorDict` objects; consumers that require
the structured mesh API should update PhysicsNeMo.

### Validation

- `pytest -q -s test/mesh`: 2,982 passed, 18 skipped.
- Post-rebase focused serialization, reader, sampler, DoMINO, datapipes, and
  remeshing tests: 134 passed.
- `ruff check` and `ruff format --check` on all changed Python files.
- `git diff --check`.

## Checklist

- [x] Model implementation coding standards are not applicable to this change.

## Dependencies

None.

## Review Process

This is a draft. No reviewers or assignees have been requested yet.


peterdsharpe:
Superseded by four independent draft PRs, each based directly on `main`:

- #1868 — `Mesh` / `DomainMesh` direct `TensorClass` inheritance, serialization shims, and current/legacy golden fixtures
- #1865 — uncapped weighted sampling and downstream adoption
- #1866 — direct reuse of canonical functional `Mesh` implementations
- #1867 — selective `strip_caches(keep=...)` support

Closing this combined draft so the four orthogonal changes can be reviewed and merged independently.

PR #1864, opened 2026-07-22
test: harden ClusterTree interaction-plan coverage

peterdsharpe:
## Summary

- exercise exact-cover invariants on coincident, degenerate, large-offset, and highly asymmetric point clouds
- independently verify every far-stream multipole acceptance criterion in float64
- cover float64 traversal, zero-survivor broadcasts, and interaction-plan validation failures

## Rationale

The existing tests cover random clouds and broad theta/leaf-size combinations, but do not directly exercise several geometric degeneracies or the zero-survivor broadcast bookkeeping path. This adds focused test-only coverage without expanding the existing parameter matrix or changing runtime behavior.

## Validation

- ruff format --check test/mesh/spatial/test_cluster_tree.py
- ruff check test/mesh/spatial/test_cluster_tree.py
- pytest -q test/mesh/spatial/test_cluster_tree.py (84 passed on CPU and CUDA)

### DIFF
--- a/test/mesh/spatial/test_cluster_tree.py
+++ b/test/mesh/spatial/test_cluster_tree.py
@@ -119,7 +119,7 @@ def _subtree_point_ids(tree, node_id):
 
 def _assert_exact_cover(
     target_points,
-    source_points,
+    source_points=None,
     *,
     theta=1.0,
     leaf_size=4,
@@ -128,7 +128,11 @@ def _assert_exact_cover(
 ):
     """Build a plan and assert that it covers every pair exactly once."""
     target_tree = ClusterTree.from_points(target_points, leaf_size=leaf_size)
-    source_tree = ClusterTree.from_points(source_points, leaf_size=leaf_size)
+    if source_points is None:
+        source_points = target_points
+        source_tree = target_tree
+    else:
+        source_tree = ClusterTree.from_points(source_points, leaf_size=leaf_size)
     plan = source_tree.find_dual_interaction_pairs(
         target_tree,
         theta=theta,
@@ -276,7 +280,7 @@ def test_plan_validates_and_far_field_engages(device):
 def test_adversarial_self_plan_coverage(device, n_dims):
     """Hostile geometries still produce an exact self-interaction cover."""
     for name, points in _adversarial_clouds(device, n_dims).items():
-        _assert_exact_cover(points, points, context=f"{n_dims}D {name}")
+        _assert_exact_cover(points, context=f"{n_dims}D {name}")
 
 
 @pytest.mark.parametrize("expand_far_targets", [False, True])
@@ -305,7 +309,7 @@ def test_adversarial_cross_plan_coverage(device, expand_far_targets):
 def test_float64_plan_coverage(device):
     """Float64 trees preserve the exactly-once plan contract."""
     points = _points(97, 3, device, seed=20, dtype=torch.float64)
-    _assert_exact_cover(points, points)
+    _assert_exact_cover(points)
 
 
 def test_zero_survivor_leaf_pair_is_valid(device):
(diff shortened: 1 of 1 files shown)


test/mesh/spatial/test_cluster_tree.py
```diff
@@ -182,6 +272,112 @@ def test_plan_validates_and_far_field_engages(device):
...
+    assert plan.n_near == 0
+    assert plan.n_nf == target_points.shape[0]
+    assert plan.n_fn == source_points.shape[0]
+    assert plan.fn_broadcast_targets.numel() == 0
+    assert plan.fn_broadcast_counts.eq(0).all()
+
+
+def test_far_admissions_satisfy_mac_in_float64(device):
+    """Independently recompute every far-stream acceptance criterion."""
+    theta = 1.0
```
mehdiataei: Using theta = 1.0 makes the theta-squared multiplier an identity, so this test cannot catch an implementation that drops theta or applies the wrong exponent. Could we use a non-unit value such as 0.7? With the current seeded clouds, 0.7 still produces nonempty far, near-far, and far-near streams on both CPU and CUDA.

test/mesh/spatial/test_cluster_tree.py
```diff
@@ -182,6 +272,112 @@ def test_plan_validates_and_far_field_engages(device):
     assert plan.n_far_nodes + plan.n_nf + plan.n_fn > 0
 
 
+@pytest.mark.parametrize("n_dims", [2, 3])
+def test_adversarial_self_plan_coverage(device, n_dims):
+    """Hostile geometries still produce an exact self-interaction cover."""
+    for name, points in _adversarial_clouds(device, n_dims).items():
+        _assert_exact_cover(points, points, context=f"{n_dims}D {name}")
```
mehdiataei: Passing the same points twice still makes _assert_exact_cover construct two distinct ClusterTree objects, so this does not exercise the documented self-interaction path. Could we reuse one tree as both source and target, or add a helper option that does so? That keeps this adversarial coverage valid if identity-specific self behavior is added later.

test/mesh/spatial/test_cluster_tree.py
```diff
@@ -182,6 +272,112 @@ def test_plan_validates_and_far_field_engages(device):
...
+    plan = _assert_exact_cover(
+        target_points,
+        source_points,
+        leaf_size=target_points.shape[0],
+        context="zero-survivor leaf pair",
+    )
+
+    assert plan.n_near == 0
+    assert plan.n_nf == target_points.shape[0]
+    assert plan.n_fn == source_points.shape[0]
```
mehdiataei: These exact stream counts pin the current traversal decomposition rather than the zero-survivor contract. Could we assert plan.n_fn > 0 to make the path non-vacuous, then retain the empty broadcast-target and all-zero broadcast-count checks? That would preserve the intended coverage without requiring this exact partition.

mehdiataei (approved):
Thanks @peterdsharpe ! Some small changes. I approved it.

test/mesh/spatial/test_cluster_tree.py
peterdsharpe: Addressed in 7714cdfb: `_assert_exact_cover` now treats an omitted source cloud as self-interaction and reuses `target_tree` as `source_tree`; the adversarial and float64 self-coverage tests now use that path.

test/mesh/spatial/test_cluster_tree.py
peterdsharpe: Addressed in 7714cdfb: the MAC soundness test now uses `theta = 0.7` while retaining the nonempty assertions for all three far streams. The full test file passes on CPU (42 passed; 42 CUDA cases skipped on this host).

test/mesh/spatial/test_cluster_tree.py
peterdsharpe: Addressed in 7714cdfb: the test now asserts only `plan.n_fn > 0` to keep the zero-survivor path non-vacuous, while retaining the empty broadcast-target and all-zero broadcast-count checks.

test/mesh/spatial/test_cluster_tree.py
```diff
@@ -182,6 +272,112 @@ def test_plan_validates_and_far_field_engages(device):
...
+    assert plan.fn_broadcast_targets.numel() == 0
+    assert plan.fn_broadcast_counts.eq(0).all()
```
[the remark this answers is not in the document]
peterdsharpe: Addressed in 7714cdfb: the implementation-specific stream counts were relaxed to `plan.n_fn > 0`, while the empty broadcast-target and all-zero broadcast-count checks remain.

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1864/commits/7714cdfb695df123f40e45207dd998c33d6b94f2

PR #1865, opened 2026-07-22
Fix uncapped weighted sampling without replacement

### LINKED ISSUE #1846: 🐛[BUG]: Weighted sampling of >2^24 points crashes (datapipes) or returns wrong indices (DoMINO); consolidate on the exact chunked sampler from #1822

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Add one exact, chunked weighted-sampling-without-replacement utility that has no
`2**24` population cap, then use it consistently in datapipes, DoMINO, and mesh
remeshing. This removes the large-population crash/wrong-index paths while
retaining deterministic generator behavior and avoiding an allocation
proportional to the full population.

This upstreams an improvement [@mehdiataei](https://github.com/mehdiataei) made as part of a recent morphing PR.

Closes #1846. Part of #1851.

Tests:

```text
pytest -q -s \
  test/utils/test_weighted_sampling.py \
  test/datapipes/transforms/test_subsample.py \
  test/models/domino/test_domino_utils.py \
  test/nn/functional/geometry/test_remeshing.py
```

## Checklist

- [x] Model implementation standards are not applicable; no model implementation is changed.

## Dependencies

None.

## Review Process

Ready for review.


test/datapipes/transforms/test_subsample.py
```diff
@@ -213,6 +215,26 @@ def test_subsample_points_weighted():
     assert result["surface_fields"].shape == (100, 2)
 
 
+def test_weighted_shuffle_uses_uncapped_sampler(monkeypatch):
+    def reject_multinomial(*args, **kwargs):
+        raise AssertionError("torch.multinomial must not be used")
```
peterdsharpe: Let's switch these to explain *why* `torch.multinomial` must not be used (i.e., the `>2**24` overflow issue.) I think there's another such AssertionError outside this PR scope we can loop in too.

test/models/domino/test_domino_utils.py
```diff
@@ -159,6 +160,21 @@ def test_shuffle_array():
     assert len(torch.unique(indices)) == 2  # No duplicates
 
 
+def test_shuffle_array_uses_uncapped_sampler(monkeypatch):
+    def reject_multinomial(*args, **kwargs):
+        raise AssertionError("torch.multinomial must not be used")
```
peterdsharpe: ditto https://github.com/NVIDIA/physicsnemo/pull/1865/changes#r3630037403

test/datapipes/transforms/test_subsample.py
peterdsharpe: Updated in fec22890. All three monkeypatched guards now explain the actual regression: `torch.multinomial` rejects inputs with more than `2**24` categories. This includes the analogous guard in `test/utils/test_weighted_sampling.py`. The four focused suites pass: 76 tests.

test/models/domino/test_domino_utils.py
peterdsharpe: Updated here in fec22890 as well. The datapipes, DoMINO, and shared-utility guards now use the same `>2**24` category-limit explanation. The four focused suites pass: 76 tests.

physicsnemo/nn/functional/geometry/remeshing/_warp_impl/launch_forward.py
```diff
@@ -28,6 +28,7 @@
     farthest_point_sampling,
 )
 from physicsnemo.utils._index_tuple_ops import unique_index_tuples
+from physicsnemo.utils._weighted_sampling import weighted_sample_without_replacement
```
loliverhennigh: In general not a big fan of having the functionals need imports from other parts of PhysicsNemo but its fine. I like if they are totally self contained even if it means a bit of duplicated code sometimes.

loliverhennigh (approved):
LGTM, I made a comment but its fine as is

physicsnemo/nn/functional/geometry/remeshing/_warp_impl/launch_forward.py
coreyjadams: I agree with Oliver here.  Functionals is the core of what will become physicsnemo ops so we need to keep that in mind, for our future selves.

Why can't these sampling ops be in functionals instead of utils?  They make more sense there perhaps?

peterdsharpe: Sure, we can put it in functionals instead of utils, I like that too. Will do this

coreyjadams (approved):
I like this refactor and consolidation.  It's in the P1 category as cleanups go, IMO, but it's a nice way to bring this all together and I think it makes a clear picture that this is a useful function.  We should put this one on our list for pnm ops migration and optimization ASAP.

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1865/commits/2328d83d2db3c02c21f477f4241743da1e766383

PR #1866, opened 2026-07-22
PhysicsNeMo-Mesh: reuse functional method implementations

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Replace duplicated `Mesh` convenience-method implementations with shared
canonical functional implementations while preserving the public class-facing
callable contract.

A small `_mesh_method` adapter gives each method its own `Mesh.<name>` identity
for introspection, type-hint resolution, help output, and bound-method pickling,
while forwarding behavior to one implementation. The paired APIs cover:

- topology: `is_manifold`, `is_watertight`
- transformations: `displace`, `free_form_deform`, `morph`,
  `radial_basis_function_deform`, `rotate`, `scale`, `transform`, `translate`
- visualization: `draw`
- calculus: `compute_cell_derivatives`, `compute_point_derivatives`, `integrate`,
  `integrate_flux`, `integrate_moment`
- validation: `validate`

Shared docstrings show both supported call forms—`function(mesh, ...)` and
`mesh.function(...)`—and explain that the bound method supplies `mesh`
automatically.

`draw` and `validate` are the canonical functional names matching the methods.
The previous `draw_mesh` and `validate_mesh` names remain as pending-deprecation
compatibility wrappers. `draw` also exposes the backend-specific axes/plotter
annotation.

`Mesh.validate` and `DomainMesh.validate` share the canonical validation options
and dtype-aware `tolerance=None` default. The historical positional location of
`tolerance` is preserved, while the new, currently unsupported
`check_self_intersection` option is keyword-only and documented to raise
`NotImplementedError` when enabled.

`Mesh.quality_metrics` and `Mesh.statistics` retain explicit property getters so
their class-facing documentation describes argument-free property access.

The quality metric now uses a normalized longest-edge/minimum-altitude aspect
ratio. It is dimensionless and scale-invariant for simplices in every manifold
dimension, and regular simplices have `aspect_ratio=1` and `quality_score=1`.
This intentionally corrects the previous erroneous values.

Sampling and remeshing remain lazy wrappers because they adapt behavior or
dependency loading rather than duplicating a matching canonical implementation.
The import cost of the newly eager topology, calculus, and validation paths
remains an open design question.

Part of #1851.

## Verification

```text
pytest -s -q \
  test/mesh/mesh/test_method_aliases.py \
  test/mesh/boundaries/test_topology.py \
  test/mesh/transformations/test_transformations.py \
  test/mesh/transformations/test_deform.py \
  test/mesh/transformations/test_ffd.py \
  test/mesh/transformations/test_rbf.py \
  test/mesh/calculus/test_calculus.py \
  test/mesh/calculus/test_gradient_layout.py \
  test/mesh/calculus/test_integration.py \
  test/mesh/calculus/test_measure.py \
  test/mesh/validation/test_validation.py \
  test/mesh/test_domain_mesh_transforms.py \
  test/mesh/visualization/test_visualization.py
```

Result: 646 passed, 17 skipped.

The broader mesh regression suite also passes:

```text
pytest -s -q test/mesh
```

Result: 3,024 passed, 18 skipped.

`ruff check`, `ruff format --check`, and `git diff --check` pass for all touched
files.

## Checklist

- [x] Model implementation standards are not applicable; no model is changed.

## Dependencies

None.

## Review Process

Ready for review. The lazy-import design question remains intentionally open.


peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1866/commits/89096cda0abf40f768f7ecd75fbeb625a63676af

test/mesh/mesh/test_method_aliases.py
```diff
@@ -0,0 +1,68 @@
...
+)
+from physicsnemo.mesh.validation import (
+    compute_mesh_statistics,
+    compute_quality_metrics,
+    validate_mesh,
+)
+from physicsnemo.mesh.visualization.draw_mesh import draw_mesh
+
+
+def test_mesh_methods_reuse_function_objects():
```
mehdiataei: Do we really want `Mesh.foo is foo` to be the contract here? It makes an implementation detail part of the public API and leaves no room for method-specific names, docs, annotations, or lazy imports. I’d rather test that the bound signature and behavior stay in sync, then keep a small wrapper where the two entry points are not actually identical.

physicsnemo/mesh/domain_mesh.py
```diff
@@ -1117,7 +1117,8 @@ def validate(
         check_inverted_cells: bool = False,
         check_out_of_bounds: bool = True,
         check_manifoldness: bool = False,
-        tolerance: float = 1e-10,
+        check_self_intersection: bool = False,
```
mehdiataei: Could we avoid inserting `check_self_intersection` before the existing `tolerance` and `raise_on_error` parameters? A call like `mesh.validate(True, True, False, True, False, 1e-6)` used to set the tolerance. It now enables the unsupported self-intersection check and raises `NotImplementedError`. I’d preserve the old positional order and make the new option keyword-only, or add it after the existing parameters. The docstring should also say that this check is not implemented yet.

physicsnemo/mesh/mesh.py
```diff
@@ -34,13 +34,20 @@
 
 import torch
 import torch.nn.functional as F
-from jaxtyping import Bool, Float
+from jaxtyping import Float
 from tensordict import NonTensorData, TensorDict, tensorclass
 
+from physicsnemo.mesh.boundaries import is_manifold, is_watertight
```
mehdiataei, line 41: This looks like it pulls quite a bit more into a basic `Mesh` import than intended. Importing `physicsnemo.mesh.boundaries` reaches the broad `physicsnemo.utils` initializer, including checkpoint, logging, profiling, and optional WandB support. In five clean runs after importing Torch, the Mesh import averaged about 1.13 seconds here versus 0.82 seconds at the merge base. Could we keep these methods lazy or separate the topology helper from that package initialization path?

physicsnemo/mesh/mesh.py
```diff
@@ -2520,697 +2464,37 @@ def next_power_size(current_size: int, base: float) -> int:
...
-        keys: str | tuple[str, ...] | list[str | tuple[str, ...]] | None = None,
-        method: Literal["lsq", "dec"] = "lsq",
-        gradient_type: Literal["intrinsic", "extrinsic", "both"] = "intrinsic",
-    ) -> "Mesh":
-        """Compute gradients of cell_data fields.
+    ### Transformations
 
-        This is a convenience method that delegates to
-        :func:`physicsnemo.mesh.calculus.compute_cell_derivatives`.
+    displace = displace
```
mehdiataei, line 2475: Could we keep the class-facing docs written for the class API? These aliases reuse docstrings that document an explicit `mesh` parameter and show standalone calls such as `compute_point_derivatives(mesh, ...)` and `validate_mesh(mesh)`. Under `help(Mesh)` and the `autoclass` page, that makes the method API look like it needs an explicit receiver. Shared implementation is useful, but the docs should still teach the interface the reader opened.

physicsnemo/mesh/mesh.py
```diff
@@ -2520,697 +2464,37 @@ def next_power_size(current_size: int, base: float) -> int:
...
-        >>> from physicsnemo.mesh import Mesh
-        >>> pts = torch.tensor([[0., 0.], [1., 0.], [0.5, 1.]])
-        >>> cells = torch.tensor([[0, 1, 2]])
-        >>> mesh = Mesh(points=pts, cells=cells)
-        >>> mesh.cell_data["p"] = torch.tensor([3.0])
-        >>> mesh.integrate("p")
-        tensor(1.5000)
-        """
-        from physicsnemo.mesh.calculus.integration import integrate
+    draw = draw_mesh
```
mehdiataei: I think `draw` and `validate` still need small wrappers because their public names differ from the functional names. `help(Mesh)` now shows `draw = draw_mesh(...)` and `validate = validate_mesh(...)`. Pickling `mesh.draw` or `mesh.validate` also fails because Python tries to restore attributes named `draw_mesh` and `validate_mesh`, while both round trips worked before this change. `Mesh.draw` also loses the useful `ax` annotation and exposes it as `Any`.

physicsnemo/mesh/mesh.py
```diff
@@ -3436,107 +2720,13 @@ def laplacian(
...
-        >>> from physicsnemo.mesh.primitives.basic import two_triangles_2d
-        >>> mesh = two_triangles_2d.load()
-        >>> metrics = mesh.quality_metrics
-        >>> assert "quality_score" in metrics.keys()
-        """
-        from physicsnemo.mesh.validation import compute_quality_metrics
+    ### Validation
 
-        return compute_quality_metrics(self)
+    quality_metrics = property(compute_quality_metrics)
```
mehdiataei: Could we keep property-specific getters or at least property-specific docs here? `help(Mesh)` now advertises `mesh` as a parameter for both properties and a configurable `tolerance` for `statistics`, even though property access accepts neither. There is also a correctness issue in the inherited quality-metrics text: it says an equilateral triangle has `aspect_ratio=1` and `quality_score=1`, but the current formula returns about `2.3094` and `0.7565` for a unit equilateral triangle.

mehdiataei (commented):
One broader docs/API note: `docs/api/mesh/core.rst` still says these methods are thin wrappers, so that overview will need to be updated if this design stays. Current `main` also added `radial_basis_function_deform`, so the rebase should make an explicit choice for that method/function pair rather than leaving the deformation APIs inconsistent.

test/mesh/transformations/test_ffd.py
```diff
@@ -71,8 +71,12 @@ def test_mesh_free_form_deform_signatures_and_annotations_are_introspectable():
...
         assert signature.parameters["origin"].default is None
         assert signature.parameters["extent"].default is None
-        assert get_type_hints(deform_method)["basis"] == _EXPECTED_FFD_BASIS
-        assert get_type_hints(deform_method)["return"] is owner
+        hints = get_type_hints(
+            deform_method,
+            localns={"Mesh": Mesh, "DomainMesh": DomainMesh},
+        )
+        assert hints["basis"] == _EXPECTED_FFD_BASIS
+        assert hints["return"] is owner
```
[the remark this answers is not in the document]
mehdiataei: I think this should also be addressed.

physicsnemo/mesh/domain_mesh.py
peterdsharpe: Fair call, I'll switch the order and add the docstring note

physicsnemo/mesh/mesh.py
peterdsharpe: How does this work in-practice? As far as I can see the only way to accomplish this is to explicitly define new wrappers as Mesh methods, which is the main thing this PR is trying to avoid to reduce duplication. These two differing versions of the docstring would only differ in `self` vs. `mesh`, right?

If there's a way to accomplish this proposal without actually `def`ing a wrapper, let me know.

physicsnemo/mesh/mesh.py
peterdsharpe: Why would you pickle `mesh.draw` or `mesh.validate`?

`Mesh.draw()` can get the `ax` annotation back inside `draw_mesh()`; will fix.

peterdsharpe: Perhaps the overarching thing here is that we may want to make the functional interfaces `draw()` and `validate()` instead of `draw_mesh()` and `validate_mesh()`, just for symmetry. I can do that, and leave `PendingDeprecationWarnings` on the old names - my hunch is that most downstream users are using the method interface rather than the functional interface anyway.

physicsnemo/mesh/mesh.py
mehdiataei: I don’t think we should add wrappers just for documentation.
Could we instead make the shared docstring describe both public forms? For example, show both compute_cell_derivatives(mesh, ...) and mesh.compute_cell_derivatives(...), and note that mesh is bound automatically in the method form.

physicsnemo/mesh/mesh.py
mehdiataei: Yes I like that better

physicsnemo/mesh/mesh.py
peterdsharpe: Yes, that's a good solution. Will do

test/mesh/mesh/test_method_aliases.py
peterdsharpe: Agreed, and addressed in 5023f207. The tests no longer make object identity part of the contract. They now require a distinct, correctly named class-facing callable with the synchronized bound signature and docstring, resolvable type hints, bound-method pickle round trips, and representative delegated behavior.

test/mesh/transformations/test_ffd.py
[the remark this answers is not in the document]
peterdsharpe: Addressed in 5023f207. The class-facing adapters now live in mesh.py, copy the canonical annotations, and have Mesh.<name> metadata, so get_type_hints(Mesh.<method>) resolves Mesh without a localns. The contract test covers this for every paired method, including FFD.

physicsnemo/mesh/mesh.py
peterdsharpe: Addressed in 5023f207. draw and validate are now the canonical functional names, with draw_mesh and validate_mesh retained as pending-deprecation compatibility wrappers. The draw annotation uses a shared AxesOrPlotter type, and the method-contract test verifies names, signatures, type-hint resolution, and bound-method pickle round trips for both methods.

physicsnemo/mesh/mesh.py
peterdsharpe: Addressed in 7d85e0b8. Each shared canonical docstring now documents both function(mesh, ...) and mesh.function(...) and states that the bound form supplies mesh automatically. The core API page also explains the shared implementation plus class-facing adapter.

physicsnemo/mesh/domain_mesh.py
peterdsharpe: Addressed in 5023f207. tolerance and raise_on_error retain their historical positions, while check_self_intersection is keyword-only in both Mesh.validate and DomainMesh.validate. The docstrings now state that it is unimplemented, and regression tests cover the positional tolerance call plus exact parameter kinds/defaults.

physicsnemo/mesh/mesh.py
peterdsharpe: This is a valid concern. Workshopped a bunch of different approaches, all of which are ugly in different ways and require coordinated import-boundary work across `boundaries`, `calculus`, and `validation` (and avoiding the broad `physicsnemo.utils` initialization path), with dedicated import-time and public-export compatibility coverage. 

That is broader than this method-reuse refactor, so let's aim to address it in a focused follow-on PR.

physicsnemo/mesh/mesh.py
peterdsharpe: quality-metrics issue fixed; properties will be reverted back to `@property`

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1866/commits/2f1c0a7284e74f25c91e8488710644a6d16795f6

physicsnemo/mesh/domain_mesh.py
```diff
@@ -1259,8 +1259,10 @@ def validate(
         check_inverted_cells: bool = False,
         check_out_of_bounds: bool = True,
         check_manifoldness: bool = False,
-        tolerance: float = 1e-10,
+        tolerance: float | None = None,
```
mehdiataei, line 1262: Small Python 3.14 gotcha here. `tensorclass` adds `DomainMesh.float`, so deferred annotation lookup resolves this `float` to that method and `inspect.signature(DomainMesh.validate)` raises a `TypeError`. This file already imports `builtins` for this exact case. Could this be `builtins.float | None`? I reproduced it with the project's Python 3.14 environment.

physicsnemo/mesh/validation/quality.py
```diff
@@ -82,16 +84,49 @@ def compute_cell_edge_lengths(
...
+        n_vertices_per_cell, n_manifold_dims
+    )
+    facet_vertices = cell_vertices[:, facet_indices]
+    facet_relative_vectors = facet_vertices[..., 1:, :] - facet_vertices[..., :1, :]
+    facet_measures = compute_cell_areas(facet_relative_vectors.flatten(0, 1)).unflatten(
+        0, (mesh.n_cells, n_vertices_per_cell)
+    )
+
+    eps = safe_eps(mesh.points.dtype)
+    return n_manifold_dims * cell_measures[:, None] / facet_measures.clamp(min=eps)
```
mehdiataei: I do not think the scale invariance holds yet in 3D. `facet_measures` has units of `length ** (d - 1)`, but this clamps it with the length-valued `safe_eps`. A regular float32 tetrahedron scaled by `1e-5` reports an aspect ratio of `7.6042` instead of `1`, even though the geometry is still representable. Could we use a dimension-aware or zero-only denominator guard and add a small-scale `d >= 3` test?

physicsnemo/mesh/mesh.py
```diff
@@ -55,15 +65,41 @@
...
+
+    def method(self, *args, **kwargs):
+        return function(self, *args, **kwargs)
+
+    method.__name__ = function.__name__
+    method.__qualname__ = f"Mesh.{function.__name__}"
+    method.__module__ = __name__
+    method.__doc__ = function.__doc__
+    method.__annotations__ = function.__annotations__.copy()
+    setattr(method, "__signature__", inspect.signature(function))
```
mehdiataei: There is a mismatch between introspection and the actual unbound call here. The wrapper really accepts `self`, while `__signature__` and the copied annotations advertise `mesh`. `inspect.signature(Mesh.translate).bind(mesh=m, offset=...)` succeeds, but `Mesh.translate(mesh=m, offset=...)` fails because `self` is missing. Could the generated method expose `self` consistently so the advertised call matches the callable?

physicsnemo/mesh/validation/validate.py
```diff
@@ -67,14 +71,15 @@ def validate_mesh(
...
-        Check for self-intersecting cells (very expensive)
     tolerance : float | None
         Tolerance for geometric checks (areas, distances).
         If ``None`` (default), uses a dtype-aware epsilon via :func:`safe_eps`.
     raise_on_error : bool
         If True, raise ValueError on first error. If False,
         return dict with all validation results.
+    check_self_intersection : bool
+        Request a self-intersection check. This option is keyword-only and not
+        yet implemented; passing ``True`` raises ``NotImplementedError``.
```
mehdiataei, line 82: Could we move the unsupported-option check to the start of `validate`? An out-of-bounds mesh currently returns at the early OOB path before reaching the `NotImplementedError`, so `check_self_intersection=True` does not always behave as documented. With `raise_on_error=True`, an earlier validation error can win too.

physicsnemo/mesh/visualization/__init__.py
```diff
@@ -14,4 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Visualization utilities for physicsnemo.mesh Mesh objects."""
+"""Visualization utilities for :class:`physicsnemo.mesh.Mesh` objects."""
+
+from physicsnemo.mesh.visualization.draw_mesh import draw, draw_mesh
```
mehdiataei, line 19: This export shadows the existing public `draw_mesh` submodule. On `main`, `import physicsnemo.mesh.visualization.draw_mesh as dm` returns the module and exposes things like `dm.BACKENDS_INSTALLED`. On this branch the same import returns the compatibility function instead. Could we keep the legacy function on its submodule path without re-exporting it here? The new `validation.validate` export has the same module-name collision to think through.

mehdiataei (commented):
A few things I found on the latest pass.

CHANGELOG.md
```diff
@@ -253,9 +253,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  `validate_mesh` remain as pending-deprecation compatibility names.
+  `Mesh.validate` and `DomainMesh.validate` share the canonical validation
+  option order, preserve the historical positional `tolerance` argument, and
+  expose the new `check_self_intersection` option as keyword-only.
+- `physicsnemo.mesh`: `validate(check_self_intersection=True)` now raises
   `NotImplementedError` (the check is unimplemented) instead of silently returning a
   `None` sentinel that masquerades as "no self-intersections found".
+- `physicsnemo.mesh` quality metrics now use a normalized
+  longest-edge/minimum-altitude aspect ratio. The metric is dimensionless and
+  scale-invariant for simplices of every manifold dimension, and a regular
```
mehdiataei, line 271: I think "every manifold dimension" is a little too broad here. A valid 0-simplex mesh with cells shaped `(n, 1)` still reaches `max()` on an empty edge dimension and raises `IndexError`. Could we either handle 0D explicitly or narrow this to positive manifold dimensions and add a clear guard or test?

physicsnemo/mesh/mesh.py
```diff
@@ -55,15 +65,41 @@
...
+    normal descriptor binding removes its leading ``mesh`` parameter.
+    """
+
+    def method(self, *args, **kwargs):
+        return function(self, *args, **kwargs)
+
+    method.__name__ = function.__name__
+    method.__qualname__ = f"Mesh.{function.__name__}"
+    method.__module__ = __name__
+    method.__doc__ = function.__doc__
```
mehdiataei: One small docs wrinkle remains with the shared docstring. `help(mesh.translate)` now shows the correct bound signature without `mesh`, but the Parameters section still lists `mesh : Mesh`. The new sentence explains the two call forms, though the parameter list can still make the method look like it takes an extra receiver. Is there a lightweight way for the adapter to drop just that entry for method help without bringing duplicated docstrings back?

physicsnemo/mesh/validation/validate.py
```diff
@@ -37,21 +38,24 @@
...
     raise_on_error: bool = False,
+    *,
+    check_self_intersection: bool = False,  # Very expensive, opt-in
 ) -> Mapping[str, bool | int | torch.Tensor]:
     """Validate mesh integrity and detect common errors.
 
     Performs a comprehensive set of checks to ensure mesh is well-formed
-    and suitable for geometric computations.
+    and suitable for geometric computations. Call it as ``validate(mesh, ...)``
+    or as ``mesh.validate(...)``; the bound method supplies ``mesh``
```
mehdiataei: Could we split this into two sentences? The same semicolon pattern appears in the other new dual-call notes too, and we have been trying to avoid semicolons in these API docs.

docs/api/mesh/validation.rst
```diff
@@ -5,10 +5,11 @@ Validation and Quality
 
 Tools for assessing mesh integrity and element quality.
 
-**Validation** (:func:`validate_mesh`)
+**Validation** (:func:`validate`)
```
mehdiataei: Could we make these proper subheadings or a definition list instead of bold labels? The API docs style pass has been avoiding decorative bold.

physicsnemo/mesh/visualization/draw_mesh.py
```diff
@@ -28,13 +29,17 @@
...
+    AxesOrPlotter: TypeAlias = Any
+
 # Check availability at module load (add new backends here)
 BACKENDS_INSTALLED: dict[str, bool] = {
     name: check_version_spec(name) for name in ["matplotlib", "pyvista"]
 }
 
 
-def draw_mesh(
+def draw(
```
mehdiataei, line 42: The shared example lost the `matplotlib.pyplot` import that used to be in the `Mesh.draw` docstring, so the final `plt.show()` line is not runnable as shown. Could we add `import matplotlib.pyplot as plt` before it?

physicsnemo/mesh/domain_mesh.py
peterdsharpe: Same issue we discussed in the previous PR; this is fixed by #1868. Let's merge this as-is and fix this in 1868

physicsnemo/mesh/mesh.py
peterdsharpe: `Mesh.translate(mesh=m, offset=...)` is this an actual use case? Under what circumstance would a user invoke class method to do an instance-level method call?

physicsnemo/mesh/visualization/__init__.py
peterdsharpe: I'm not too worried about this; I would be surprised if any users are using things like `dm.BACKENDS_INSTALLED` here, as the primary entrypoint to `draw()` is the method interface

physicsnemo/mesh/mesh.py
mehdiataei: Fair. But I think problem is that the function tells Python tools, “I accept an argument named mesh,” but it actually accepts one named self. Tools that read the signature could call it incorrectly. It's a nit so if you can ignore it.

physicsnemo/mesh/mesh.py
peterdsharpe: Non-issue

mehdiataei (approved):
All good on my side so feel free to merge.

mehdiataei (approved):
All good on my side so feel free to merge.

docs/api/mesh/core.rst
```diff
@@ -16,8 +16,9 @@ calls, and expensive geometric quantities -- centroids, normals, areas, curvatur
 
 Most mesh operations (subdivision, derivatives, transformations) are
 available both as ``Mesh`` methods and as standalone functions in the
-corresponding submodules. The methods are thin wrappers that pass ``self`` to
-the standalone functions.
+corresponding submodules. Each pair directly shares one canonical function;
```
megnvidia:
```suggestion
corresponding submodules. Each pair shares one canonical function, and
```

CHANGELOG.md
```diff
@@ -253,9 +253,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   cells, return self), matching its type hint and `slice_points`;
   `gaussian_curvature_cells` reuses the cached `gaussian_curvature_vertices`
   property instead of recomputing it.
-- `physicsnemo.mesh`: `validate_mesh(check_self_intersection=True)` now raises
+- `physicsnemo.mesh.Mesh` convenience methods now directly reuse shared
+  canonical functions, removing duplicate implementation bodies and docstrings.
+  This includes geometric, deformation (including radial-basis-function
+  deformation), calculus, topology, visualization, and validation operations.
+- `physicsnemo.mesh`: `draw` and `validate` are now the canonical standalone
+  names matching `Mesh.draw` and `Mesh.validate`; `draw_mesh` and
```
megnvidia:
```suggestion
  names matching `Mesh.draw` and `Mesh.validate`. The `draw_mesh` and
```

CHANGELOG.md
```diff
@@ -253,9 +253,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
   `NotImplementedError` (the check is unimplemented) instead of silently returning a
   `None` sentinel that masquerades as "no self-intersections found".
+- `physicsnemo.mesh` quality metrics now use a normalized
+  longest-edge/minimum-altitude aspect ratio. The metric is dimensionless and
+  scale-invariant for simplices of every manifold dimension, and a regular
+  simplex now has `aspect_ratio=1` and `quality_score=1`. This intentionally
+  corrects the previous erroneous aspect-ratio and quality-score values.
+- `Mesh.quality_metrics` and `Mesh.statistics` again use explicit property
+  getters so their class-facing documentation describes argument-free property
+  access; configurable statistics tolerance remains available through the
```
megnvidia:
```suggestion
  access. Configurable statistics tolerance remains available through the
```

CHANGELOG.md
```diff
@@ -253,9 +253,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  names matching `Mesh.draw` and `Mesh.validate`; `draw_mesh` and
+  `validate_mesh` remain as pending-deprecation compatibility names.
+  `Mesh.validate` and `DomainMesh.validate` share the canonical validation
+  option order, preserve the historical positional `tolerance` argument, and
+  expose the new `check_self_intersection` option as keyword-only.
+- `physicsnemo.mesh`: `validate(check_self_intersection=True)` now raises
   `NotImplementedError` (the check is unimplemented) instead of silently returning a
   `None` sentinel that masquerades as "no self-intersections found".
+- `physicsnemo.mesh` quality metrics now use a normalized
+  longest-edge/minimum-altitude aspect ratio. The metric is dimensionless and
```
megnvidia:
```suggestion
  aspect ratio of longest edge to minimum altitude. The metric is dimensionless and
```

megnvidia (approved):

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1866/commits/3a0da51a87062e1f372f954942a08e5e7290eb9d

PR #1867, opened 2026-07-22
PhysicsNeMo-Mesh: improve cache handling APIs

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Improve the Mesh cache-handling API and route repository consumers through it so cache validity is explicit, reusable, and consistent.

- Extend `Mesh.strip_caches(keep=...)` and `DomainMesh.strip_caches(keep=...)` to retain selected cache entries or complete cache categories.
- Accept a bare string for one top-level cache key, a bare tuple for one nested key, or a sequence such as a list for multiple keys.
- Add `Mesh.with_points(...)` for coordinate changes that preserve point indexing and connectivity. It invalidates geometry-dependent caches while retaining topology caches by default.
- Add `Mesh.with_cells(...)` for connectivity changes that preserve cell indexing and simplex type. It clears cell, point, and topology caches by default, with an explicit expert `keep=` override.
- Keep derived Mesh data and cache containers independent, while safely sharing tensor leaves under the documented replacement semantics.
- Route geometry-preserving transformations, projections, procedural primitives, face reorientation, and winding changes through the appropriate cache-aware API.
- Route data-only Mesh operations, datapipes, calculus helpers, visualization, GLOBE, and unified aero consumers through `with_data` so valid caches survive.
- Continue using specialized operations or explicit construction when point/cell identity, cardinality, or simplex type changes and associated data must be remapped.
- Document the cache policy for `with_data`, `with_points`, `with_cells`, and `strip_caches` in the Mesh API reference.

The private structural helper is named `_new_with_structure` to avoid TensorClass-generated `replace` and dataclass `__replace__` APIs. Those generic replacements are documented as unsafe for changing `points` or `cells` because they do not invalidate caches.

No-argument `strip_caches()` still removes every cached value. The returned mesh now owns independent data and cache `TensorDict` containers while sharing their tensor leaves, matching the functional-update contract of the other replacement methods.

Part of #1851.

### Validation

```text
pytest -q test/mesh
# 3258 passed, 18 skipped

pytest -q \
  test/mesh/utilities/test_cache.py \
  test/mesh/test_domain_mesh_transforms.py::TestStripCaches \
  test/mesh/mesh/test_compile.py::test_cache_rebuild_paths_under_fullgraph_compile \
  test/mesh/transformations/test_transformations.py
# 174 passed

pre-commit run --files <changed files>
# all hooks passed
```

Additional manual checks cover full-graph `torch.compile` use of `with_points` and `with_cells`. The GLOBE model file retains its existing symbolic-shape `F821` suppressions; all newly changed lines pass Ruff.

## Checklist

## Dependencies

None.

## Review Process

Ready for review.


coreyjadams (approved):
It took me awhile to understand the changes in this PR.  I appreciate the API change though: this is a more natural and understandable way to update a mesh and handle cache invalidation.  I like it.

It'd be nice to ensure we get the API docs updated too.  I know you have a lot of PRs open on Mesh, so I don't want to block this PR on docs (esp because Meg is away right now) so let's track that if it's not already covered?

physicsnemo/mesh/mesh.py
```diff
@@ -1599,6 +1604,181 @@ def sample_data_at_points(
...
+            ("point", torch.Size([self.n_points])),
+            ("topology", torch.Size([])),
+        ):
+            if category not in cache:
+                cache[category] = TensorDict(
+                    {},
+                    batch_size=batch_size,
+                    device=device,
+                )
+        return cache
```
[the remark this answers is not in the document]
peterdsharpe: fixed

physicsnemo/mesh/mesh.py
```diff
@@ -1599,6 +1604,181 @@ def sample_data_at_points(
...
+        self,
+        keep: Sequence[str | tuple[str, ...]] = (),
+    ) -> TensorDict:
+        """Return an independent cache container containing only ``keep``.
+
+        Tensor leaves are intentionally shared, but every retained nested
+        ``TensorDict`` container is shallow-copied so populating a cache on a
+        derived mesh cannot mutate the source mesh's cache structure.
+        """
+        cache = self._cache.select(*keep, strict=False).copy()
```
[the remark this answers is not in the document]
peterdsharpe: fixed

physicsnemo/mesh/transformations/geometric.py
```diff
@@ -623,36 +623,23 @@ def translate(
...
-        cells=mesh.cells,
-        point_data=mesh.point_data,
-        cell_data=mesh.cell_data,
-        global_data=mesh.global_data,
-        _cache=new_cache,
-    )
+    return translated_mesh
 
 
 def rotate(
```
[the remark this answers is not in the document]
peterdsharpe: keeping as-is; this is local

peterdsharpe:
Thanks! Added docs in here too

docs/api/mesh/core.rst
```diff
@@ -42,6 +42,39 @@ polygons -- a "polygon soup" (see :doc:`tessellation`) -- use
     mesh = mesh.compute_point_derivatives(keys="T", method="lsq")
     print(mesh.point_data["T_gradient"])  # shape (3, 2)
 
+Cache-aware functional updates
```
megnvidia:
```suggestion
Cache-Aware Functional Updates
```

docs/api/mesh/core.rst
```diff
@@ -42,6 +42,39 @@ polygons -- a "polygon soup" (see :doc:`tessellation`) -- use
...
+
+   * - Method
+     - Intended change
+     - Default cache policy
+   * - :meth:`Mesh.with_data`
+     - Replace point, cell, or global field data
+     - Retain every geometry and topology cache
+   * - :meth:`Mesh.with_points`
+     - Replace coordinates without changing point indexing or connectivity
+     - Retain topology; clear geometry caches
```
megnvidia:
```suggestion
     - Retain topology and clear geometry caches
```

docs/api/mesh/core.rst
```diff
@@ -42,6 +42,39 @@ polygons -- a "polygon soup" (see :doc:`tessellation`) -- use
...
+   * - :meth:`Mesh.with_cells`
+     - Replace connectivity without changing cell indexing or simplex type
+     - Clear every cache
+   * - :meth:`Mesh.strip_caches`
+     - Remove cached values without changing the mesh
+     - Clear every cache
+
+The ``keep`` override accepts a top-level string such as ``"topology"``, one
```
megnvidia:
```suggestion
The ``keep`` override accepts any of the following:
```

docs/api/mesh/core.rst
```diff
@@ -42,6 +42,39 @@ polygons -- a "polygon soup" (see :doc:`tessellation`) -- use
...
+nested-key tuple such as ``("cell", "areas")``, or a sequence of keys. Use a
```
megnvidia:
```suggestion

- A top-level string such as ``"topology"``
- A single nested-key tuple such as ``("cell", "areas")``
- A sequence of keys

```

docs/api/mesh/core.rst
```diff
@@ -42,6 +42,39 @@ polygons -- a "polygon soup" (see :doc:`tessellation`) -- use
...
+list when retaining multiple keys, for example
```
megnvidia:
```suggestion
Use a list when retaining multiple keys, for example
```

docs/api/mesh/core.rst
```diff
@@ -42,6 +42,39 @@ polygons -- a "polygon soup" (see :doc:`tessellation`) -- use
...
+``["topology", ("cell", "areas")]``. Retaining a cache that depends on replaced
+geometry or connectivity is an expert operation; the caller is responsible for
```
megnvidia:
```suggestion
geometry or connectivity is an expert operation. You are responsible for the
```

docs/api/mesh/core.rst
```diff
@@ -42,6 +42,39 @@ polygons -- a "polygon soup" (see :doc:`tessellation`) -- use
...
+its validity.
```
megnvidia:
```suggestion
validity of the retained cache.
```

megnvidia (approved):
With changes it is good to publish

physicsnemo/mesh/mesh.py
```diff
@@ -1617,6 +1622,192 @@ def sample_data_at_points(
...
+        every retained value remains valid for the replacement coordinates.
+
+        Examples
+        --------
+        >>> moved = mesh.with_points(mesh.points + 1.0)  # doctest: +SKIP
+        >>> embedded = mesh.with_points(  # doctest: +SKIP
+        ...     torch.nn.functional.pad(mesh.points, (0, 1))
+        ... )
+        """
+        if not torch.compiler.is_compiling():
```
mehdiataei: Could we keep these shape checks active under `torch.compile`, or use `torch._check` here? Right now eager mode rejects a changed point count or simplex type, but the compiled path skips the checks and can return an invalid `Mesh`.

```python
def fn(points, cells):
    mesh = Mesh(points=points, cells=cells)
    return mesh.with_points(points[:-1])

fn(points, cells)
# ValueError

torch.compile(fn, backend="eager", fullgraph=True)(points, cells)
# Returns a Mesh with too few points
```

Trying to use something like `cell_areas` afterward fails with an unrelated out-of-bounds error. It would be nice for the compiled path to enforce the same structural contract.

physicsnemo/mesh/mesh.py
```diff
@@ -1617,6 +1622,192 @@ def sample_data_at_points(
...
+
+    def _new_with_structure(
+        self,
+        *,
+        points: torch.Tensor,
+        cells: torch.Tensor,
+        keep: str | tuple[str, ...] | Sequence[str | tuple[str, ...]],
+    ) -> "Mesh":
+        """Build a same-index mesh with selected structural fields replaced."""
+        return Mesh(
```
mehdiataei: Could we preserve the concrete mesh type here instead of always constructing `Mesh`? This changes the existing behavior for custom subclasses.

```python
class TaggedMesh(Mesh):
    def tag(self):
        return "tagged"

mesh = TaggedMesh(points=points, cells=cells)
out = mesh.displace(torch.zeros_like(points))

type(out)
# Mesh

hasattr(out, "tag")
# False
```

On `main`, the deformation path preserves `TaggedMesh` because it uses `dataclasses.replace`. This now affects all the deformation methods, and `with_data` causes the same change in `NormalizeMeshFields`. Maybe an internal cache-aware `dataclasses.replace(self, ...)` would preserve the subtype while still applying the new cache policy.

peterdsharpe: Sure, will test using `dataclasses.replace` or perhaps `self.__class__` constructor

physicsnemo/mesh/mesh.py
peterdsharpe: The motivation for not doing this is that shape-checks tend to result in CPU-GPU-syncs in mesh processing pipelines. This is because things like slice_cells with value-dependent masks yield array shapes that depend on inputs, so any later ops will force sync here.

I know `torch.compile` can do some limited SymInt stuff, so I'll play with it to see if that's possible here - but if not, I would vote to skip shape checks under compile. (This is usually what we do with `torch.compile` for anything that is part of a model forward, which this occasionally is.)

physicsnemo/mesh/mesh.py
peterdsharpe: fixed with `dataclasses.replace`

physicsnemo/mesh/mesh.py
peterdsharpe: Fixed with `torch._check`

mehdiataei (approved):

PR #1868, opened 2026-07-22
PhysicsNeMo-Mesh: inherit directly from TensorClass

### LINKED ISSUE #1851: 🚀[FEA]: PhysicsNeMo-Mesh Code Cleanups

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Replace the `@tensorclass`-decorated `Mesh` and `DomainMesh` containers with
direct `TensorClass` inheritance. `Mesh` retains its tensor-only/shadow behavior
and `Mesh[m, s]` runtime specialization syntax, while `DomainMesh` remains
unsubscriptable.

TensorDict 0.14 writes directly inherited tensorclasses with their type
discriminator and `_tensordict/` payload, so nested `Mesh` and `DomainMesh`
objects preserve their concrete types. Committed decorator-era `.pmsh` and
`.pdmsh` fixtures continue to reconstruct exact mesh types, including nested
interior and boundary meshes. Current writer layouts are pinned by committed
JSON manifests.

The compatibility adapters preserve the inherited public loader signatures and
documentation, support preallocated `out=` storage, reject a `device=` that
conflicts with the output tensors instead of producing a mixed-device result,
and forward TensorDict's explicit `allow_pickle` safety policy for non-tensor
fields.

Field defaults remain available to generated constructors for fieldless
subclasses. Serialization, functional mesh updates, and floating dtype
conversion preserve concrete subclasses and their additional tensor fields.

Closes #1851.

Tests on Python 3.14 with TensorDict 0.14.0:

```text
uv run --no-sync pytest -q test/mesh
# 2795 passed, 31 skipped

uv run --no-sync pre-commit run --all-files
# passed
```

## Checklist

- [x] Model implementation standards are not applicable; no model is changed.

## Dependencies

Satisfied by the stable `tensordict[zarr]>=0.14.0` dependency currently on
`main`. TensorDict 0.14 contains the typed direct-inheritance serialization
support from [pytorch/tensordict#1722](https://github.com/pytorch/tensordict/pull/1722).

## Review Process

Ready for review.


mehdiataei (commented):
I found a few compatibility issues in the inheritance and serialization paths.

physicsnemo/mesh/_serialization.py
```diff
@@ -0,0 +1,152 @@
...
+decorator, whose memmap writer nested the payload one directory down::
+
+    square.pdmsh/
+        meta.json          <- {"_type": "...DomainMesh"}
+        _tensordict/       <- the actual fields
+
+Inheriting from ``TensorClass`` instead writes those fields at the root of the
+directory, so :func:`install_legacy_memmap_reader` teaches a container to
+recognize the old nesting and read it explicitly. Files written by the current
+code need no special handling; they load through ``tensordict``'s own
```
mehdiataei: Current files still need a general type-restoration path here. A `Mesh` nested in a normal `TensorDict` silently reloads as a `TensorDict`, while the base revision preserved `Mesh`. The same happens for `DomainMesh` and for fields on unrelated tensorclasses. Can we preserve the child type discriminator instead of only repairing fields inside `DomainMesh`?

physicsnemo/mesh/mesh.py
```diff
@@ -88,8 +85,19 @@
...
+    def __getitem__(cls, params: Any) -> type:
+        return cls.__class_getitem__(params)
+
+
+class Mesh(
+    TensorClass,
+    tensor_only=True,
+    shadow=True,
+    metaclass=_MeshTensorClassMeta,
+):
```
mehdiataei, line 108: Direct inheritance regenerates constructors for subclasses. Existing code like `class MyMesh(Mesh): pass` can no longer call `MyMesh(points)` because `cells`, the data containers, and `_cache` become required. `DomainMesh` has the same regression. Could we put the existing `None` defaults on the field declarations so generated subclass constructors keep them?

physicsnemo/mesh/domain_mesh.py
```diff
@@ -1614,11 +1621,52 @@ def draw(
...
+        and any(isinstance(mesh, TensorDict) for mesh in boundaries.values())
+    ):
+        tensordict = tensordict.copy()
+        if isinstance(interior, TensorDict):
+            tensordict["interior"] = Mesh._from_tensordict(interior)
+        if isinstance(boundaries, TensorDict):
+            boundaries = boundaries.copy()
+            for name, mesh in boundaries.items():
+                if isinstance(mesh, TensorDict):
+                    boundaries[name] = Mesh._from_tensordict(mesh)
```
mehdiataei: Hardcoding `Mesh._from_tensordict` loses the subtype recorded in the parent metadata. A compatible `Mesh` subclass is downcast after a `DomainMesh` round trip, while one with an added tensor field fails because `Mesh` rejects the extra key. Direct `DomainMesh` subclasses also get a separate `_from_tensordict`, so they bypass this patch and cannot load their own current files. Could this restore through the recorded type and be inherited by subclasses?

physicsnemo/mesh/_serialization.py
```diff
@@ -0,0 +1,152 @@
...
+        if device is not None:
+            result = result.to(device, non_blocking=non_blocking)
+        if out is not None:
+            payload = cast(TensorDictBase, _payload_of(out))
+            payload.update(
+                result._tensordict,
+                inplace=True,
+                non_blocking=non_blocking,
+            )
+            result = cls._from_tensordict(payload)
```
mehdiataei: `device` and `out` conflict silently here. With a CPU `out` and `device="cuda"`, legacy loads stay on CPU, while a current `Mesh` can end up with CUDA geometry and CPU field data. That breaks the all-tensors-move-together invariant. Could we reject mismatched targets or make the behavior consistent across both layouts and container types?

physicsnemo/mesh/_serialization.py
```diff
@@ -0,0 +1,152 @@
...
+    def load_memmap(
+        cls,
+        prefix: str | Path,
+        device: torch.device | None = None,
+        non_blocking: bool = False,
+        *,
+        out: TensorDictBase | None = None,
+        robust_key: bool | None = None,
+    ) -> Any:
+        """Load from a memory-mapped directory tree, in either layout."""
```
mehdiataei: These become the public `Mesh` and `DomainMesh` load APIs, but runtime help now shows an internal one-line description and an `Any` return type. `load_memmap` also loses the inherited docs for `device`, `non_blocking`, `out`, `robust_key`, and meta loading. Since the core docs render class members, could these wrappers preserve the original metadata or provide complete public signatures and docstrings?

CHANGELOG.md
```diff
@@ -210,6 +210,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- `physicsnemo.mesh.Mesh` and `DomainMesh` now inherit directly from
+  `TensorClass`, use the resulting root-level `.pmsh` / `.pdmsh` layout, and
+  continue to load decorator-era serialized data (including honoring
+  `device=`, which was previously ignored on load). Note that the change is
+  not forward-compatible: `.pmsh` / `.pdmsh` files written by this release
+  load as plain `TensorDict` objects, with no error, in PhysicsNeMo 2.1.x and
+  earlier. Readers of newly written files should be on 2.2.0+. This is also a
```
mehdiataei: This warning is important, but users writing files are more likely to read `Mesh.save` or `DomainMesh.save` than the changelog. Could we repeat it in the public serialization docs so people see it before creating files that older releases silently load as `TensorDict`?

mehdiataei:
Overall, I think the direction makes sense, and the normal `Mesh` and `DomainMesh` paths look solid. My hesitation is that the implementation feels a bit too special-cased right now.

The biggest concern is serialization. A `Mesh` or `DomainMesh` stored inside another `TensorClass` can quietly come back as a plain `TensorDict`, while the `DomainMesh` fix only restores the fields it already knows about.

For example, a user might define a cached training sample like this:

```python
class TrainingSample(TensorClass):
    mesh: Mesh
    target: torch.Tensor


sample = TrainingSample(
    mesh=mesh,
    target=torch.tensor(1.0),
)

sample.save(path)
loaded = TrainingSample.load(path)

type(loaded)
# TrainingSample

type(loaded.mesh)
# TensorDict, expected Mesh
```

This round trip preserved the nested `Mesh` type before the change. Loading still succeeds without a warning, so the problem only becomes visible later when the user expects to use it as a mesh. I’m not sure how common this exact pattern is, but nesting a `Mesh` inside another `TensorClass` seems like a reasonable thing to expect once `Mesh` directly inherits from `TensorClass`.

Subclass behavior also feels a bit uneven. Existing subclasses lose constructor defaults, some round trips lose the concrete subtype, and dtype conversions can skip or drop fields added by subclasses. Given that this started as a low-priority cleanup, I don’t think these compatibility edges are quite safe enough to merge yet.

My suggestion would be to move the compatibility behavior into a shared inherited base or mixin instead of patching methods onto the finished classes. It should preserve the concrete type in the saved metadata, reconstruct nested objects using that recorded type, and apply automatically to subclasses. I’d also add round-trip tests for nested `TensorClass` fields, custom subclasses, and `device` combined with `out`.

If nested serialization is intentionally outside the supported API, that limitation should be documented clearly. If solving these cases cleanly makes the change much larger, keeping the decorator for now may be the safer option.

peterdsharpe:
Thanks for the review!

Personally, the subclass/inheritance changes don't worry me (as inheritance on a decorator-@tensorclass was always a bit borked), but I do think that the specific example (TrainingExamples in the public comment) is one we should definitely fix. Looks like this is fixed by upstream https://github.com/pytorch/tensordict/pull/1722, which is merged but not yet released.

Let's pause on this PR until that goes live.

peterdsharpe:
Paused until https://github.com/pytorch/tensordict/pull/1756 merges, at which point we'll use nightly.

physicsnemo/mesh/_serialization.py
peterdsharpe: Resolved now that main pins tensordict-nightly 2026.8.6 with the upstream typed serialization fix. The custom type-restoration path is gone and loading delegates to TensorDict metadata. Regression coverage now includes Mesh and DomainMesh nested in ordinary TensorDict containers, plus the TrainingSample TensorClass example.

physicsnemo/mesh/_serialization.py
peterdsharpe: Resolved by preserving the inherited TensorClass loader metadata and full public signature, including subpath, mode, and num_threads. Runtime tests assert that both Mesh and DomainMesh load and load_memmap have exactly the TensorClass signatures and documentation.

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1868/commits/bbebe08990f0c46701b96c07838e784f3e1257c4

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1868/commits/0379218679ff220af08c7776c48ef00b6fc8dfd8

PR #1874, opened 2026-07-25
Unified Aero Recipe: offset manifest indices into the combined MultiDataset

peterdsharpe:
## What

Follow-up to #1859. Implements the offset-shifted manifest indices that
#1859 deliberately deferred, and removes the `NotImplementedError` guard
it added.

## Why

Manifest split indices are local to a single reader, but the sampler
addresses the concatenated `MultiDataset`. Before #1859, choosing more
than one manifest dataset meant the last one's indices overwrote the
others and were applied to the whole concatenation — so training ran on
the wrong samples, silently.

I reproduced this on the pre-#1859 tree. Two manifest datasets, `a` (4
runs, val = `run_0`/`run_1`) and `b` (2 runs):

```
train: ['a/run_0/domain_0.pdmsh']    <-- a's held-out val run
val:   ['a/run_1/domain_1.pdmsh']
```

Training on a validation run, `b` never sampled at all, and no error.

#1859 correctly converted that into a loud failure. But every dataset
YAML in this recipe is manifest-mode — none defines a `val_datadir` —
so the guard also blocks the multi-dataset recipe the README documents
(Transolver across DrivAerML + SHIFT SUV), leaving `extra_datasets`
unusable for any combination. #1859's docstring assumed the SHIFT SUV
datasets were directory-mode; they auto-discover a `manifest.json` next
to `train_datadir`.

## How

Accumulate offset-shifted indices per dataset. Each dataset's local
train/val indices are shifted by its cumulative offset in the
corresponding `MultiDataset`. Train and val offsets are tracked
separately, because the two concatenations diverge whenever a dataset
contributes to one split but not the other (a directory dataset with no
`val_datadir`, or an `augment=True` run that builds a second
un-augmented reader for validation).

Directory-mode datasets contribute their full local ranges, so the two
modes can be combined in one run. The missing-val-split fallback moves
into the per-dataset loop, ahead of offsetting, so it can no longer
capture a neighbouring dataset's indices.

Same scenario on this branch:

```
train: ['a/run_2', 'a/run_3', 'b/run_0']
val:   ['a/run_0', 'a/run_1', 'b/run_1']
```

## Also

`resolve_manifest_spec` now warns when a dataset falls back to directory
mode while a `manifest.json` sits next to its `train_datadir`. Clearing
the top-level split selectors is how a manifest/directory mix is
configured, so this is easy to reach — and left silent it sweeps the
manifest's held-out val/test runs into training.

## Testing

`examples/.../unified_external_aero_recipe/tests/`: 165 passed (161 + 4).
New coverage: multi-manifest offsets, mixed manifest/directory ordering
in both orders, the per-dataset val fallback, distributed sharding over
the combined index space, and — the property that actually matters —
that indices resolve back through `MultiDataset`'s own global → (dataset,
local) mapping to the runs the manifests name.

Mutation-checked: swapping `val_offset` for `train_offset` fails 6 tests,
including all the new ones.

### DIFF
--- a/examples/cfd/external_aerodynamics/unified_external_aero_recipe/README.md
+++ b/examples/cfd/external_aerodynamics/unified_external_aero_recipe/README.md
@@ -34,10 +34,10 @@ The pipeline non-dimensionalizes raw fields to unitless model inputs
 conditions (`U_inf`, `rho_inf`, `p_inf`, ...) live in each file's
 `global_data` and are read by `MeshReaderWithGlobalData`. Because the
 datasets are non-dimensionalized and loaded through the PhysicsNeMo
-datapipes' `MultiDataset` abstraction, directory-split datasets can be
-merged on the fly for multi-dataset training. Manifest-split datasets
-currently must be trained one at a time because their sampler indices
-are local to one dataset. Non-dimensionalization itself is the
+datapipes' `MultiDataset` abstraction, you can merge directory- and
+manifest-split datasets on the fly for multi-dataset training. Manifest
+indices are resolved within each dataset and shifted into the combined
+index space before sampling. Non-dimensionalization itself is the
 `NonDimensionalizeByMetadata` transform in `src/nondim.py`.
 
 ## Quick start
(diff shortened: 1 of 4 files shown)


peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1874/commits/c5667ec402c776ce70a63cd056bc0f80f763fb20

coreyjadams (approved):
We discussed offline; approving to unblock.  We should consider moving some of this logic onto the helper functions in Multi_Dataset however let's not hold up the bug fix.

PR #1875, opened 2026-07-26
Add differentiable Sobolev mesh deformation

mehdiataei:
## Summary

- Add `sobolev_deform_points`, `SobolevDeformPoints`, and `Mesh.sobolev_deform` as public APIs.
- Implement a matrix-free uniform-mass P1 Helmholtz filter with Jacobi-preconditioned conjugate-gradient solves.
- Add a CUDA Warp backend for segments, triangles, and tetrahedra. CUDA selects Warp by default when available, while CPU and higher-dimensional simplices use Torch.
- Provide first-order implicit differentiation through the displacement and source vertices. The Warp backward performs an adjoint CG solve and an analytic P1 geometry vector-Jacobian product.
- Use deterministic point-incidence gathers, periodic true-residual replacement, and per-batch normalization for reliable float32 and small-magnitude solves.
- Support optional fixed vertices, batched point tensors with shared topology, isolated points, and exact zero-length or empty-topology dense displacement.
- Add API documentation, 2D and 3D optimization examples, reproducible figures, benchmark registration, and focused tests.

## Motivation

Direct optimization of dense mesh vertices can produce vertex-scale oscillations in shape updates and adjoints. This adds a mesh-aware deformation parameterization that filters those high-frequency modes while preserving broad shape changes and fixed-point constraints.

The Warp backend moves P1 assembly, matrix-free operator application, CG, and the analytic geometry pullback onto CUDA. Warp's `optim.linear` module disables generated backward kernels, so the implementation differentiates the converged linear system with an explicit adjoint solve instead of differentiating the CG iteration history.

## User impact

Users can optimize candidate vertex coordinates through `mesh.sobolev_deform(...)` and receive smoother reverse-mode sensitivities. The smoothing length uses the same physical units as the mesh coordinates.

Torch is the default on CPU. Warp is the default on CUDA for segment, triangle, and tetrahedron meshes when it is available. Both backends support first-order gradients and `torch.compile`. CUDA Graph capture and higher-order derivatives at positive smoothing lengths are not supported.

## Validation

- `CUDA_VISIBLE_DEVICES=0 uv run --no-sync pytest -q test/nn/functional/geometry/test_point_sobolev.py test/mesh/transformations/test_sobolev.py`
  - 50 passed
  - 18 expected skips for CUDA-only cases on CPU
- Torch and Warp forward and gradient parity passed for segments, triangles, and tetrahedra in float32 and float64.
- Tiny-field relative-tolerance, true-residual restart, empty-topology identity, fixed and isolated vertices, FakeTensor, CUDA Graph rejection, and full-graph compile tests passed.
- CUDA Inductor forward and backward completed successfully.
- Both standalone Sobolev optimization examples completed on CUDA. The 3D example also completed on CPU.
- Ruff check and format, Interrogate, Markdownlint, license checks, large-file checks, and import-linter passed through pre-commit.


mehdiataei:
@loliverhennigh Can you please review this? Thanks!

loliverhennigh (approved):
LGTM, no big comment but just wanted to confirm there is now warp implementation for this? I only see PyTorch.

mehdiataei:
@loliverhennigh Yeah I didn't attempted the Warp backend due to the matrix-free Jacobi-PCG solves to keep it differentiable. Let me give a try I have an idea.

mehdiataei:
@loliverhennigh I implemented the Warp backend. I'll try to upstream some of these to Warp later. Please take a look.

loliverhennigh (approved):
LGTB

peterdsharpe (approved):
Hey @mehdiataei, thanks for the PR.

This one looks good to me. Approving to unblock, but one change to the examples I'd request before merging:

- I think we discussed offline about this but can we convert `examples/mesh/mesh/*.py` from `*.py` files to `*.ipynb` in the same tutorial style as the existing ones? Perhaps you can tie this + `deformation_energy_optimization.py` into a `tutorial_10_mesh_deformations.ipynb` or similar? In addition to aligning with convention, this'll let us embed in images so that end-users can see usage on GitHub - I think it'll help discoverability of these deformation capabilities. Naming of the file/tutorial up to you.

megnvidia (approved):
lgtm

PR #1887, opened 2026-07-29
Field GP 

ktangsali:
# PhysicsNeMo Pull Request

## Description
Adds GPs for fields 

## Checklist

## Review Process


examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
```diff
@@ -0,0 +1,364 @@
...
+                head = build_field_gp_head(
+                    input_dim=feature_dim,
+                    n_train_points=1,  # unused at inference
+                    num_tasks=num_tasks,
+                    n_inducing=n_inducing,
+                    mlp_hidden=mlp_hidden,
+                    lengthscale_range=ls_range,
+                    lengthscale_prior=ls_prior,
+                    outputscale_prior=os_prior,
+                )
```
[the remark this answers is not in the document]
ktangsali: Stale — this was fixed in https://github.com/NVIDIA/physicsnemo/pull/1887/commits/30dc9bbcb305e6855ba57050b4907f8ffe69d821

laserkelvin:
I've been bogged down with toolkit-ops reviews and whatnot, but I'll review first thing next week.

physicsnemo/experimental/uq/__init__.py
```diff
@@ -17,15 +17,25 @@
 """Uncertainty quantification modules (experimental).
 
 This subpackage provides UQ building blocks that can be attached to encoder
-backbones to produce calibrated uncertainty estimates.  Currently available:
+backbones to produce calibrated uncertainty estimates.  Both heads below are
+variational Gaussian processes with inducing points; they differ in what
+constitutes a data point, and therefore in what they predict:
```
melo-gonzo: I would maybe relax the language around producing calibrated uncertainty estimates. It can be done but i dont think its the default behavior right? Also this reads as if the GP heads are the only exposed API through this file, which is true now but will likely be updates in future as we develop UQ more. Perhaps just a section specific for the GP heads.

examples/cfd/external_aerodynamics/transformer_models/src/conf/geotransolver_surface_field_gp.yaml
```diff
@@ -0,0 +1,159 @@
...
+# what lets an eval run load a training checkpoint. The two arguments not listed
+# here are supplied at runtime: `input_dim` (the backbone feature width, probed
+# with one forward pass) and `n_train` (the ELBO num_data normaliser).
+gp_head:
+  _target_: physicsnemo.experimental.uq.FieldVariationalGPHead
+
+  num_tasks: 4                 # pressure + 3 wall-shear-stress components
+  n_inducing: 1024             # inducing points per task (shared structure)
+
+  # DKL feature extractor: pointwise MLP between the (wide) backbone features
```
melo-gonzo: Maybe spell out some of these acronyms on first use.

physicsnemo/experimental/uq/field_variational_gp_head.py
```diff
@@ -0,0 +1,838 @@
...
+        self.num_tasks = num_tasks
+        self.mean_module = gpytorch.means.ConstantMean(batch_shape=batch_shape)
+
+        ls_constraint = gpytorch.constraints.Interval(*lengthscale_range)
+        ls_prior_obj = None
+        if lengthscale_prior is not None:
+            ls_prior_obj = gpytorch.priors.GammaPrior(*lengthscale_prior)
+
+        base_kernel = gpytorch.kernels.MaternKernel(
+            nu=2.5,
```
melo-gonzo: what is nu here? any reason to have hard coded?

physicsnemo/experimental/uq/field_variational_gp_head.py
```diff
@@ -0,0 +1,838 @@
...
+        input_dim: int,
+        num_tasks: int = 4,
+        n_inducing: int = 256,
+        n_train: int | None = None,
+        inducing_points: torch.Tensor | None = None,
+        lengthscale_range: tuple[float, float] = (0.01, 10.0),
+        lengthscale_prior: tuple[float, float] | None = None,
+        outputscale_prior: tuple[float, float] | None = None,
+        mlp_hidden: list[int] | None = None,
+        feature_norm: str = "none",
```
melo-gonzo: May consider using `Literal['none','l2_radial']` here instead of just `str`.

physicsnemo/experimental/uq/field_variational_gp_head.py
```diff
@@ -0,0 +1,838 @@
...
+    >>> pred.mean.shape
+    torch.Size([1, 4096, 4])
+    """
+
+    def __init__(
+        self,
+        input_dim: int,
+        num_tasks: int = 4,
+        n_inducing: int = 256,
+        n_train: int | None = None,
```
melo-gonzo: L383 saying `None` is not accepted, should be required param.

examples/cfd/external_aerodynamics/transformer_models/src/normalization/drivaerstar/surface_fields_normalization.npz
laserkelvin, line 1: I would prefer we not use `npz` - since this file is pretty small, could we just use JSON instead?

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
```diff
@@ -0,0 +1,371 @@
...
+torch.serialization.add_safe_globals([omegaconf.base.Metadata])
+
+
+@torch.no_grad()
+def field_gp_predict_full_mesh(
+    batch: dict,
+    model: torch.nn.Module,
+    head,
+    chunk_size: int,
+    device: torch.device,
```
melo-gonzo: looks like you might have wanted to use this in L87 but rederive the device.

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
```diff
@@ -0,0 +1,371 @@
...
+torch.serialization.add_safe_globals([int])
+torch.serialization.add_safe_globals([omegaconf.nodes.AnyNode])
+torch.serialization.add_safe_globals([omegaconf.base.Metadata])
+
+
+@torch.no_grad()
+def field_gp_predict_full_mesh(
+    batch: dict,
+    model: torch.nn.Module,
+    head,
```
melo-gonzo: `FieldVariationalGPHead` for type?

examples/cfd/external_aerodynamics/transformer_models/src/train_field_gp.py
```diff
@@ -0,0 +1,945 @@
...
+    _, point_features = model(
+        global_embedding=features,
+        local_embedding=embeddings,
+        geometry=geometry,
+        local_positions=local_positions,
+        return_point_features=True,
+    )
+    dim = int(point_features.shape[-1])
+    logger.info(f"Probed per-point backbone feature dim: {dim}")
+    return dim
```
melo-gonzo: mostly duplicate of the same function in `inference_field_gp` it seems. perhaps consolidate into the `gp_utils.py`.

examples/cfd/external_aerodynamics/transformer_models/src/field_gp_utils.py
laserkelvin, line 1: Public functions need proper NumPy docstrings

examples/cfd/external_aerodynamics/transformer_models/src/field_gp_utils.py
```diff
@@ -0,0 +1,149 @@
...
+            global_embedding=features,
+            local_embedding=embeddings,
+            geometry=geometry,
+            local_positions=local_positions,
+            return_point_features=True,
+        )
+        # point_features: (B, N, D) -> (B*N, D)
+        pf = point_features.reshape(-1, point_features.shape[-1])
+        take = min(pf.shape[0], n_inducing - n_have)
+        idx = torch.randperm(pf.shape[0], device=pf.device)[:take]
```
laserkelvin, line 225: What controls determinism here?

examples/cfd/external_aerodynamics/transformer_models/src/field_gp_utils.py
```diff
@@ -0,0 +1,149 @@
...
+    backbone + GP mean learn a sensible field before the KL regulariser pulls
+    the variational posterior toward the prior.
+    """
+    if warmup_end <= warmup_start:
+        return 1.0
+    if epoch < warmup_start:
+        return 0.0
+    if epoch >= warmup_end:
+        return 1.0
+    return (epoch - warmup_start) / (warmup_end - warmup_start)
```
melo-gonzo: same as in `gp_utils.py`, `def gp_ramp_weight`?

melo-gonzo (approved):
Overall lgtm. Added some consolidation between train/inference and the utils files + a few other minor comments.

examples/cfd/external_aerodynamics/transformer_models/src/field_gp_utils.py
```diff
@@ -0,0 +1,149 @@
...
+            local_embedding=embeddings,
+            geometry=geometry,
+            local_positions=local_positions,
+            return_point_features=True,
+        )
+        # point_features: (B, N, D) -> (B*N, D)
+        pf = point_features.reshape(-1, point_features.shape[-1])
+        take = min(pf.shape[0], n_inducing - n_have)
+        idx = torch.randperm(pf.shape[0], device=pf.device)[:take]
+        collected.append(pf[idx].detach().cpu())
```
laserkelvin: Why do you need to move to CPU?

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
```diff
@@ -0,0 +1,371 @@
...
+from physicsnemo.distributed import DistributedManager
+from physicsnemo.utils import load_checkpoint
+from physicsnemo.utils.logging import PythonLogger, RankZeroLoggingWrapper
+
+from train import cast_precisions, update_model_params_for_fp8, get_autocast_context
+
+# Reuse the VTK I/O + datapipe helpers from the deterministic VTK inference.
+from inference_on_vtk import build_data_dict, create_datapipe
+
+torch.serialization.add_safe_globals([omegaconf.listconfig.ListConfig])
```
laserkelvin, line 83: Why are these safe globals necessary? As in, why do we have to serialize this stuff?

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
```diff
@@ -0,0 +1,371 @@
...
+
+    Everything is written in physical units.  The targets are standardized
+    *physical* fields — the DrivAerStar statistics are Pa-scale — so
+    unstandardizing is the whole conversion and no dynamic-pressure factor
+    applies.  Note the four channels carry very different normalisation scales,
+    so the physical stds are not comparable across channels; divide each by its
+    ``surface_factors["std"]`` entry to recover the dimensionless form when a
+    cross-channel UQ map is what you want.
+    """
+    mesh = pv.read(vtp_path)
```
laserkelvin, line 219: Would be good to do some validation on the path before trying to read, e.g. `Path(vtp_path).exists()`

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
```diff
@@ -0,0 +1,371 @@
...
+            "pointing at a trained run."
+        )
+
+
+def write_field_gp_predictions_to_vtk(
+    vtp_path: str,
+    output_path: str,
+    mean_norm: torch.Tensor,
+    std_norm: torch.Tensor,
+    surface_factors: dict,
```
laserkelvin: Can we type hint this better?

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
```diff
@@ -0,0 +1,371 @@
...
+    unstandardizing is the whole conversion and no dynamic-pressure factor
+    applies.  Note the four channels carry very different normalisation scales,
+    so the physical stds are not comparable across channels; divide each by its
+    ``surface_factors["std"]`` entry to recover the dimensionless form when a
+    cross-channel UQ map is what you want.
+    """
+    mesh = pv.read(vtp_path)
+    output_mesh = mesh.copy()
+
+    mean_np = mean_norm.cpu().numpy()
```
laserkelvin: Why are all these ops done on CPU and with NumPy? Why can't they be done on the GPU?

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
```diff
@@ -0,0 +1,371 @@
...
+        )
+    vtk_cfg = cfg.vtk_inference
+    if not vtk_cfg.get("input_dir", None):
+        raise ValueError("vtk_inference.input_dir is required")
+    if not vtk_cfg.get("output_dir", None):
+        raise ValueError("vtk_inference.output_dir is required")
+
+    input_dir = Path(vtk_cfg.input_dir)
+    output_dir = Path(vtk_cfg.output_dir)
+    air_density = vtk_cfg.get("air_density", 1.2050)
```
laserkelvin, line 267: Comment where the numbers come from?

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
```diff
@@ -0,0 +1,371 @@
...
+
+
+@torch.no_grad()
+def field_gp_predict_full_mesh(
+    batch: dict,
+    model: torch.nn.Module,
+    head,
+    chunk_size: int,
+    device: torch.device,
+    precision: str,
```
laserkelvin: Do we validate that these are valid types at all? Should this really by type hinted as a literal instead?

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
```diff
@@ -0,0 +1,371 @@
...
+            import traceback
+
+            traceback.print_exc()
+            continue
+
+    logger.info("Field-GP inference complete!")
+
+
+@torch.no_grad()
+def _probe_feature_dim(model, batch, precision, device) -> int:
```
laserkelvin: Type hints

examples/cfd/external_aerodynamics/transformer_models/src/train_field_gp.py
```diff
@@ -0,0 +1,945 @@
...
+    """
+    if gp_feats.dim() != 3:
+        gp_feats = gp_feats.reshape(1, -1, gp_feats.shape[-1])
+        target = target.reshape(1, -1, target.shape[-1])
+    b, p, _ = gp_feats.shape
+    if p < 2 or n_pairs < 1:
+        return gp_feats.new_zeros(())
+    target = target.to(gp_feats.dtype)
+    total = gp_feats.new_zeros(())
+    for k in range(b):
```
laserkelvin: Do you have to use a loop for this?

examples/cfd/external_aerodynamics/transformer_models/src/train_field_gp.py
```diff
@@ -0,0 +1,945 @@
...
+        i = torch.randint(0, p, (n_pairs,), device=gp_feats.device)
```
laserkelvin: Determinism - do you have a pre-initialized generator instance or something instead?

examples/cfd/external_aerodynamics/transformer_models/src/train_field_gp.py
```diff
@@ -0,0 +1,945 @@
...
+        dist_manager,
+    )
+
+    # ---- Field-GP config ----
+    # cfg.gp_head is a hydra _target_ block; the eval scripts instantiate the
+    # same block, so the structure baked into the checkpoint cannot drift apart.
+    n_inducing = cfg.gp_head.n_inducing
+
+    # KL annealing window (epochs) and auxiliary mean-MSE weight
+    beta_warmup_start = getattr(cfg, "beta_warmup_start", 0)
```
laserkelvin: I'm not a huge fan of this pattern since if there's a typo in the key, it'll just silently use defaults

examples/cfd/external_aerodynamics/transformer_models/src/train_field_gp.py
```diff
@@ -0,0 +1,945 @@
...
+                    scaled_loss.backward()
+
+            if is_step_boundary:
+                sync_non_ddp_gradients([head], dist_manager.world_size)
+                if grad_clip_norm > 0.0:
+                    if scaler is not None:
+                        # Gradients are still scaled; unscale in place first so
+                        # the clip threshold is in true gradient units.
+                        scaler.unscale_(optimizer)
+                    # Returns the pre-clip norm, which is the diagnostic we
```
laserkelvin, line 708: My issue with this is that it seems to hide some of the "tricks" needed to actually make the model train, which don't exist in the architecture itself and therefore not really standalone.

examples/cfd/external_aerodynamics/transformer_models/src/train_field_gp.py
```diff
@@ -0,0 +1,945 @@
...
+                    # actually want logged -- it shows how often (and by how
+                    # much) the threshold binds.
+                    head_grad_norm = float(
```
laserkelvin: If this is solely used for logging, would it be better to make it optional so there isn't a need for a blocking call?

examples/cfd/external_aerodynamics/transformer_models/README.md
```diff
@@ -398,6 +408,275 @@ pip install gpytorch
...
+- **Large-scale Heteroscedastic Regression via Gaussian Process:** [Liu, Ong & Cai, 2018b](https://arxiv.org/abs/1811.01179) — a sparse, minibatch-capable heteroscedastic ELBO; the closest published objective to ours, with the noise carried by a second GP rather than our noise MLP
+- **What Uncertainties Do We Need in Bayesian Deep Learning for Computer Vision?:** [Kendall & Gal, NeurIPS 2017](https://arxiv.org/abs/1703.04977) — the amortized-network noise head and noise-attenuated likelihood we actually use. We borrow the mechanism, not the "aleatoric" interpretation (see [The variance decomposition](#the-variance-decomposition))
+
+*Interpreting the noise term, and multi-output structure:*
+
+- **Bayesian Calibration of Computer Models:** [Kennedy & O'Hagan, JRSS-B 2001](https://doi.org/10.1111/1467-9868.00294) — model discrepancy: the gap between a deterministic simulator and truth, which is what `sigma^2(x)` absorbs on a deterministic RANS dataset
+- **Remarks on Multi-Output Gaussian Process Regression:** [Liu et al., 2018a](https://doi.org/10.1016/j.knosys.2018.03.022) — multi-output GP structure; this head uses the independent-per-task case (no cross-channel covariance)
```
laserkelvin: This link does not go to the paper stated:

> Graph embedding techniques, applications, and performance: A survey; Palash Goyal, Emilio Ferrara

physicsnemo/experimental/models/geotransolver/geotransolver.py
```diff
@@ -265,7 +265,7 @@ class GeoTransolver(Module):
         training, and emits warnings on out-of-distribution inputs during
         inference. Default is ``None``.
     attention_type : str, optional
-        attention_type is used to choose the attention type (GALE or GALE_FA). 
+        attention_type is used to choose the attention type (GALE or GALE_FA).
```
laserkelvin: If it's just two choices, then `Literal["GALE", "GALE_FA"]` is more accurate?

physicsnemo/experimental/uq/field_variational_gp_head.py
```diff
@@ -0,0 +1,838 @@
...
+    ) -> Float[torch.Tensor, "points tasks"]:
+        """Per-point, per-task observation-noise variance.
+
+        ``sigma_t(x) = exp(log_base_noise_t + clamp(g_t(x), -3, 3))`` — a learned
+        per-task base scale times a bounded multiplicative modulation from the
+        noise MLP.  The clamp keeps the modulation within ~20x either way so a
+        bad step cannot drive the noise to 0 (infinite log-likelihood) or blow
+        it up; the final clamp to ``noise_std_range`` is a hard safety net.
+        """
+        log_mod = self.noise_head(gp_in).clamp(-3.0, 3.0)
```
laserkelvin, line 609: Is -3 sufficiently small? 10^3 is big so I have no qualms with that

physicsnemo/experimental/uq/field_variational_gp_head.py
```diff
@@ -0,0 +1,838 @@
...
+        return features
+
+    def _apply_fe(self, features: torch.Tensor) -> torch.Tensor:
+        """Run feature transform (DKL + norm), then cast to GP precision."""
+        features = self._transform_features(features)
+        if self._use_double:
+            return features.double()
+        return features
+
+    def transform_features(self, features: torch.Tensor) -> torch.Tensor:
```
laserkelvin: Why is this necessary?

laserkelvin (changes requested):
Various comments; one universal theme is please use `jaxtyping` for tensor annotations. I think a lot of them have known shapes but they're just hinted as `torch.Tensor`

physicsnemo/experimental/uq/__init__.py
ktangsali: Agreed! Re-worded it in https://github.com/NVIDIA/physicsnemo/pull/1887/changes/7914a97c643fd6481f7099fc176d50605bf9562b

examples/cfd/external_aerodynamics/transformer_models/src/conf/geotransolver_surface_field_gp.yaml
ktangsali: Done in https://github.com/NVIDIA/physicsnemo/pull/1887/changes/4371600d532588f2726ce259d84dbe3d9f0c3668

physicsnemo/experimental/uq/field_variational_gp_head.py
ktangsali: No good reason, it's configurable now via `matern_nu`, on both the layer and the head, defaulting to 2.5 so the reference recipe is unchanged. See https://github.com/NVIDIA/physicsnemo/pull/1887/changes/7914a97c643fd6481f7099fc176d50605bf9562b https://github.com/NVIDIA/physicsnemo/pull/1887/changes/2f597256a090c8905f8536370bc762f74b407861

physicsnemo/experimental/uq/field_variational_gp_head.py
ktangsali: Done

ktangsali: https://github.com/NVIDIA/physicsnemo/pull/1887/changes/7914a97c643fd6481f7099fc176d50605bf9562b

physicsnemo/experimental/uq/field_variational_gp_head.py
ktangsali: Good catch. Fixed it in both the heads. https://github.com/NVIDIA/physicsnemo/pull/1887/changes/7914a97c643fd6481f7099fc176d50605bf9562b https://github.com/NVIDIA/physicsnemo/pull/1887/changes/2f597256a090c8905f8536370bc762f74b407861

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
ktangsali: Good catch, argument removed, and now the device is derived from `batch["fx"]`

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
ktangsali: Done in https://github.com/NVIDIA/physicsnemo/pull/1887/changes/4371600d532588f2726ce259d84dbe3d9f0c3668

examples/cfd/external_aerodynamics/transformer_models/src/train_field_gp.py
ktangsali: Consolidated to `field_gp_utils.py`. https://github.com/NVIDIA/physicsnemo/pull/1887/changes/4371600d532588f2726ce259d84dbe3d9f0c3668

examples/cfd/external_aerodynamics/transformer_models/src/field_gp_utils.py
ktangsali: Yes, identical except for an extra guard on an empty or inverted warmup window. The guard moved into `gp_ramp_weight` and `beta_ramp_weight` is gone; both call sites in the field-GP recipe use the shared one. https://github.com/NVIDIA/physicsnemo/pull/1887/changes/4371600d532588f2726ce259d84dbe3d9f0c3668

examples/cfd/external_aerodynamics/transformer_models/src/normalization/drivaerstar/surface_fields_normalization.npz
ktangsali: I would defer this to a separate PR. The npz convention is also used in a few other places, and we can tackle this in a separate targeted PR.

examples/cfd/external_aerodynamics/transformer_models/src/field_gp_utils.py
ktangsali: Done, addedd doc strings based on numpy formats. https://github.com/NVIDIA/physicsnemo/pull/1887/changes/4371600d532588f2726ce259d84dbe3d9f0c3668

examples/cfd/external_aerodynamics/transformer_models/src/field_gp_utils.py
ktangsali: No particular reason tbh. Fixed it here and the previous scalar GP one too. https://github.com/NVIDIA/physicsnemo/pull/1887/changes/4371600d532588f2726ce259d84dbe3d9f0c3668 https://github.com/NVIDIA/physicsnemo/pull/1887/changes/2f597256a090c8905f8536370bc762f74b407861

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
ktangsali: It's not our serialisation choice — it's physicsnemo's `.mdlus` format. Those checkpoints store the model's instantiation arguments alongside its weights so a model can be rebuilt from the file alone, and because the args arrive from hydra they're OmegaConf nodes rather than plain containers. `torch.load` runs with `weights_only=True`, which refuses to unpickle any type not on its allowlist, so loading a trained backbone fails on the config until these are registered.

Every training and inference script in this directory carries the same block for that reason, train.py included, and none of them said so.

Added a comment saying so. https://github.com/NVIDIA/physicsnemo/pull/1887/changes/4371600d532588f2726ce259d84dbe3d9f0c3668 https://github.com/NVIDIA/physicsnemo/pull/1887/changes/2f597256a090c8905f8536370bc762f74b407861

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
ktangsali: Added https://github.com/NVIDIA/physicsnemo/pull/1887/changes/4371600d532588f2726ce259d84dbe3d9f0c3668

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
ktangsali: Done in https://github.com/NVIDIA/physicsnemo/pull/1887/changes/4371600d532588f2726ce259d84dbe3d9f0c3668

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
ktangsali: I removed numpy, but it still happens on host. that's primarily because anyways the pyvista arrays are held on host and we produce the predictions in chunk, so having them all on GPU won't be ideal. https://github.com/NVIDIA/physicsnemo/pull/1887/changes/4371600d532588f2726ce259d84dbe3d9f0c3668

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
ktangsali: Added https://github.com/NVIDIA/physicsnemo/pull/1887/changes/4371600d532588f2726ce259d84dbe3d9f0c3668

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
ktangsali: Fixed it. https://github.com/NVIDIA/physicsnemo/pull/1887/changes/4371600d532588f2726ce259d84dbe3d9f0c3668

examples/cfd/external_aerodynamics/transformer_models/src/inference_field_gp.py
ktangsali: Removed this function from here. A type hinted version can now be found in field_gp_utils.py https://github.com/NVIDIA/physicsnemo/pull/1887/changes/4371600d532588f2726ce259d84dbe3d9f0c3668

examples/cfd/external_aerodynamics/transformer_models/src/train_field_gp.py
ktangsali: Added batch support https://github.com/NVIDIA/physicsnemo/pull/1887/changes/4371600d532588f2726ce259d84dbe3d9f0c3668

examples/cfd/external_aerodynamics/transformer_models/src/train_field_gp.py
ktangsali: Added a global seed (`cfg.seed`) to address this - does that work? https://github.com/NVIDIA/physicsnemo/pull/1887/changes/4371600d532588f2726ce259d84dbe3d9f0c3668

examples/cfd/external_aerodynamics/transformer_models/src/train_field_gp.py
ktangsali: All the `getattr(cfg, key, default)` calls are gone from this script. The twenty odd knobs it reads are listed in `_REQUIRED_KNOBS` and checked once at startup,
reporting all the missing ones in a single message pointing at the reference config; after that they're read as plain attributes, so a missing key raises. `cfg.get` survives only where absent genuinely means "derive it", `checkpoint_dir`, `checkpoint_epoch`, `seed`,  and those say so in a comment.

P.S. This was an existing pattern in the transolver recipe, so it was inherited from it, fixing that one can be done in a separate PR though.

examples/cfd/external_aerodynamics/transformer_models/src/field_gp_utils.py
ktangsali: If the question is whether a random draw is acceptable here: the locations are an initialization, not a fixed design. The strategy is built with `learn_inducing_locations=True`, so all 1024 are variational parameters the optimizer moves for the whole run, and the draw only decides where the approximation starts. 

If you are talking about determinism in general, then I have now added a global config for seed.

examples/cfd/external_aerodynamics/transformer_models/src/train_field_gp.py
ktangsali: It is logging-only. It stays a tensor now and `.item()` is called in the f-string that formats the log line, so the device sync happens at the logging cadence rather than every optimizer step. https://github.com/NVIDIA/physicsnemo/pull/1887/changes/4371600d532588f2726ce259d84dbe3d9f0c3668

examples/cfd/external_aerodynamics/transformer_models/README.md
ktangsali: Amazing catch! Fixed here: https://github.com/NVIDIA/physicsnemo/pull/1887/changes/6dcb403750ac408938b4a980e8f22943c5ca8292

physicsnemo/experimental/uq/field_variational_gp_head.py
ktangsali: The clamp bounds the modulation, not the noise itself, which the docstring didn't make clear. sigma_t(x) = exp(log_base_noise_t + clamp(g_t(x), -3, 3)): the base scale per channel is learned and unbounded, and the clamp limits how far any one point may sit from its channel's base scale, a factor of ~20 either way.

ktangsali: Docstring edit in https://github.com/NVIDIA/physicsnemo/pull/1887/changes/7914a97c643fd6481f7099fc176d50605bf9562b

physicsnemo/experimental/uq/field_variational_gp_head.py
ktangsali: It keeps the Cholesky factorisation of K_uu viable. With the short lengthscales this recipe runs (the cap is 2.0 on l2-normalised features) the inducing points are strongly correlated and K_uu is ill-conditioned; in float32 that shows up as either a Cholesky failure or a negative predictive variance, and no amount of jitter fixes it without also flattening the posterior.

Only the GP internals run in float64. Outputs are cast back in `forward_and_loss` and `predict`, so the backbone keeps training in its own precision and the extra cost is confined to the kernel solve. It's behind `use_double` if a caller wants to trade that off.

Updated the docstring to clarify this further. https://github.com/NVIDIA/physicsnemo/pull/1887/changes/7914a97c643fd6481f7099fc176d50605bf9562b

physicsnemo/experimental/models/geotransolver/geotransolver.py
ktangsali: Done

examples/cfd/external_aerodynamics/transformer_models/src/train_field_gp.py
ktangsali: Yes, totally - I have tried to address that in docstrings and in the readme too: https://github.com/NVIDIA/physicsnemo/pull/1887/changes/0dfba9fb68428c93b7a4364fe8f3aea49ed381e8

ktangsali:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1887/commits/f0e5178154f94c2eebed48dbb810a5dfc032bd37

physicsnemo/models/geotransolver/geotransolver.py
```diff
@@ -570,6 +573,7 @@ def forward(
         geometry: Float[torch.Tensor, "batch tokens geometry_dim"] | None = None,
         time: torch.Tensor | None = None,
         return_embedding_states: bool = False,
+        return_point_features: bool = False,
```
coreyjadams, line 577: Could we add a * here to force kwargs only for return embedding states and return point features?  I think that would be a little more user friendly since these have nearly the same API, and the same return signature.

physicsnemo/models/geotransolver/geotransolver.py
```diff
@@ -730,9 +746,15 @@ def forward(
...
         else:
             x = tuple(x)
+            point_features_out = tuple(point_features)
 
+        if return_embedding_states and return_point_features:
+            return x, embedding_states, point_features_out
         if return_embedding_states:
             return x, embedding_states
+        if return_point_features:
+            return x, point_features_out
```
coreyjadams, line 765: Can you add a small test in the geotranslover tests just to exercise these paths and make sure we're checking all these dfifferent return options?  I don't think it needs to go crazy, nor do we necessarily need numerical checks on the returned states and point features.  I am more concerned about having test coverage on API stability, that's all.  I think it would be a straightforward test.

coreyjadams (approved):
Hi @ktangsali I reviewed the GeoTransolver changes with just a few small requests.  No show stoppers so I'm approving the PR.

physicsnemo/models/geotransolver/geotransolver.py
ktangsali: Done! The * sits after time, so the two flags are keyword-only while the four positional arguments stay positional - the shared test helpers (validate_forward_accuracy, validate_cuda_graphs, validate_jit, validate_amp, validate_checkpoint) call the model with a positional 4-tuple, so a * any earlier would have broken them.

https://github.com/NVIDIA/physicsnemo/pull/1887/commits/0bf5f67ce23d450a5efac80bf89b3ad62fb3438f

physicsnemo/models/geotransolver/geotransolver.py
ktangsali: Good point! Added `test_geotransolver_return_options`, `test_geotransolver_return_flags_are_keyword_only`,  `test_geotransolver_point_features_with_local_features` and `test_geotransolver_point_features_tuple_inputs` to test these paths. Same previous .pth files work. 

https://github.com/NVIDIA/physicsnemo/pull/1887/commits/0bf5f67ce23d450a5efac80bf89b3ad62fb3438f

examples/cfd/external_aerodynamics/transformer_models/src/conf/geotransolver_surface_field_gp.yaml
```diff
@@ -23,28 +23,41 @@ output_dir: "runs"
 checkpoint_dir: null
 run_id: "geotransolver/surface/field_gp"
 
-# --- Field GP head ---
+# RNG seed. Everything stochastic in this recipe draws from torch's global RNG:
+# the backbone's initialization, the datapipe's point sampling, the inducing-point
+# draw, the per-step GP subsample and the distance penalty's point pairs. An
+# integer makes the run reproducible; null (the default, and what the results in
+# the README were produced with) leaves the generators entropy-seeded.
```
laserkelvin: What does "entropy-seeded" mean here? If it's set to `None` is it just basically...unseeded?

examples/cfd/external_aerodynamics/transformer_models/src/field_gp_utils.py
```diff
@@ -103,13 +164,41 @@ def collect_inducing_features(
     precision: str,
     device: torch.device,
     logger: logging.Logger | None = None,
-) -> torch.Tensor:
+) -> Float[torch.Tensor, "inducing dim"]:
```
laserkelvin: jaxtyping would interpret the output as 2D, with "inducing" and "dim" as the dimensions - it doesn't sound right to me but I wanted to double check with you

physicsnemo/experimental/uq/__init__.py
```diff
@@ -16,17 +16,29 @@
...
 
-This subpackage provides UQ building blocks that can be attached to encoder
-backbones to produce calibrated uncertainty estimates.  Both heads below are
-variational Gaussian processes with inducing points; they differ in what
-constitutes a data point, and therefore in what they predict:
+This subpackage collects UQ building blocks that attach to existing backbones.
+It holds the Gaussian-process heads below; other UQ methods are expected to join
+them as the area develops.
+
+Gaussian-process heads
```
laserkelvin, line 23: I'm on the fence about whether to gate this right now being in `physicsnemo.experimental.uq`

I know the `uq` namespace wasn't added in this PR, but I'm picking the issue up now before it goes further. I think the namespace needs better design and thought, and I don't want it to just be a dumping ground of GP layers without some time put into thinking about how we want people to design and subsequently use various techniques for UQ.

I think if we were to implement the GP building blocks I wouldn't be as opposed to having the `uq` namespace, because it would mean that we would have a strategy for `physicsnemo` users. They could take these building blocks, and design their own GP kernels, variational inference algorithms, etc.

examples/cfd/external_aerodynamics/transformer_models/src/conf/geotransolver_surface_field_gp.yaml
ktangsali: Yes, essentially that. removed the flowery language 😅

ktangsali: https://github.com/NVIDIA/physicsnemo/pull/1887/commits/ac339fa226edbf53d0a7965c173fa4c84e6f4b6e

examples/cfd/external_aerodynamics/transformer_models/src/field_gp_utils.py
ktangsali: 2D is right. The helper returns the inducing-point matrix itself: each batch's (B, N, D) point features get flattened to (B*N, D), a random subset is taken, and rows accumulate until there are n_inducing of them - so (n_inducing, D) in the backbone's raw feature space. set_inducing_points accepts either (M, D), shared across tasks, or (num_tasks, M, D), and expands the shared case itself after applying the DKL extractor and feature norm. This helper always produces the shared form. Renamed to hopefully "n_inducing input_dim" to clarify better.

ktangsali: https://github.com/NVIDIA/physicsnemo/pull/1887/commits/ac339fa226edbf53d0a7965c173fa4c84e6f4b6e

physicsnemo/experimental/uq/__init__.py
ktangsali: That's exactly why it's in experimental - the namespace buys velocity while the methods are still moving. When we promote it out, we should definitely spend the design effort (some of these things could become modules in physicsnemo.nn). I say we tackle that in a separate PR in subsequent release.

coreyjadams:
@ktangsali @mnabian Where do you envision drawing a line between `uq` and `guardrails`?  They are, conceptually, addressing very, very similar user concerns: is the model making a reliable prediction, and how reliable?  I think Kelvin is raising a good point about the API here.

ktangsali:
> @ktangsali @mnabian Where do you envision drawing a line between `uq` and `guardrails`? They are, conceptually, addressing very, very similar user concerns: is the model making a reliable prediction, and how reliable? I think Kelvin is raising a good point about the API here.

I think both of these are serving different aspects. I like to think of UQ and Guardrails as a Venn diagram - yes, there is an intersection where both _can_ be used for OOD detection, but their use cases and the way they operate differ. Roughly: a guardrail looks at what you feed the model and warns you when it falls outside what the model was trained on; UQ looks at what came out and puts an error bar on it. And a model can be uncertain on a perfectly in-distribution sample, which is something a guardrail has nothing to say about.

I don't think we need to settle this in a PR though. It's worth doing properly as part of the namespace design when this comes out of experimental.

mnabian:
> > @ktangsali @mnabian Where do you envision drawing a line between `uq` and `guardrails`? They are, conceptually, addressing very, very similar user concerns: is the model making a reliable prediction, and how reliable? I think Kelvin is raising a good point about the API here.
> 
> I think both of these are serving different aspects. I like to think of UQ and Guardrails as a Venn diagram - yes, there is an intersection where both _can_ be used for OOD detection, but their use cases and the way they operate differ. Roughly: a guardrail looks at what you feed the model and warns you when it falls outside what the model was trained on; UQ looks at what came out and puts an error bar on it. And a model can be uncertain on a perfectly in-distribution sample, which is something a guardrail has nothing to say about.
> 
> I don't think we need to settle this in a PR though. It's worth doing properly as part of the namespace design when this comes out of experimental.

I agree - guardrails generally can't do UQ, and not all of our UQ tools can do OOD detection. We should clarify the overlaps and differences in the user dcumentation.

coreyjadams:
They are two sides of a venn diagram, as long as that venn diagram is narrowly scoped to "ways to ask if my model is right?" :)

I don't think we want to have too many high level folders that can appear similar to users, I'm imagining a generalized uq/guardrails folder outside of experimental where these might be siblings.

@laserkelvin do you think this API discussion in experimental should hold the PR?  If there are other concerns, that's different.  This isn't moving out of experimental today, I think we should put it as a discussion soon for how this eventual API could look, including CFD, alchemi, and e2s use cases.  But that's not really this PR.

laserkelvin (approved):
I'll approve to unblock the PR, but when I have a second I'll post replies to the discussion

ktangsali:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1887/commits/d37a8d9dcd9f0a7647daed578beb29f28003a275

PR #1889, opened 2026-07-31
Mesh: support generic linear PyVista cells

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

`from_pyvista` previously relied on a small set of regular cell layouts, which
limited conversion of common linear VTK meshes and made mixed-dimensional cell
selection and parent-data alignment difficult to handle safely.

This PR:

- broadens linear-cell support to native lines and polylines, pixels, triangle
  strips, polygons, voxels, wedges, pyramids, pentagonal and hexagonal prisms,
  and polyhedra
- selects an explicitly requested dimension from mixed `UnstructuredGrid`
  inputs, while retaining native 1D cells when present and deriving the unique
  edge graph otherwise
- tracks source parent cells through extraction and simplex conversion so split
  cells retain the correct cell data without mutating the input dataset
- validates attribute tuple counts, VTK cell-array structure, point-index
  bounds, fixed and variable cell arities, cell-type cardinality, and
  polyhedron face/location references before invoking VTK filters
- rejects unsupported higher-order and generic topology pending a globally
  conforming tessellation path, while retaining explicit point-cloud conversion
  that does not interpret cell topology
- adds focused regression coverage for supported cell families, mixed-dimension
  selection, provenance, malformed VTK inputs, data preservation, and warning
  behavior

## Validation

- `uv run pytest -q test/mesh/io/io_pyvista`: 188 passed
- `uv run ruff check` passed for all changed Python files
- `uv run ruff format --check` passed for all changed Python files

## Checklist

- [x] If I am implementing a new model or modifying any existing model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation). Not applicable because this PR does not modify model code.

## Dependencies

No new dependencies.

## Review Process


physicsnemo/mesh/io/io_pyvista.py
```diff
@@ -187,62 +1011,106 @@ def from_pyvista(
...
             )
-        if not all_triangles:
-            pyvista_mesh = pyvista_mesh.triangulate()
-
-    elif manifold_dim == 3:
-        if not isinstance(pyvista_mesh, pv.UnstructuredGrid):
-            raise ValueError(
-                f"Expected an UnstructuredGrid with volume cells for 3D meshes, "
-                f"but got {type(pyvista_mesh)=}."
+        if not pyvista_mesh.is_all_triangles:
```
mehdiataei: `is_all_triangles` is a little misleading here because PyVista only checks polygon faces and ignores strips. A `PolyData` with a triangle face plus a valid strip takes this fast path, then `regular_faces` only reads the polygon face. The strip and its cell data disappear without a warning. I reproduced a two-cell input that returns just the triangle and no cell data.

Could we handle polygon faces and strips separately, then combine their connectivity and parent maps? Sending the mixed dataset through the current helper is not enough on VTK 9.6 because provenance for the passthrough triangle comes back corrupted.

physicsnemo/mesh/io/io_pyvista.py
```diff
@@ -83,6 +89,721 @@ def _geometry_to_vtk_numpy(tensor: torch.Tensor) -> np.ndarray:
...
+        ]
+        referenced_face_point_ids = np.unique(
+            np.concatenate(
+                [
+                    face_point_ids[face_offsets[face_id] : face_offsets[face_id + 1]]
+                    for face_id in parent_face_ids
+                ]
+            )
+        )
+        if not np.array_equal(parent_point_ids, referenced_face_point_ids):
```
mehdiataei: I think this needs a stronger face-complex check, not just a comparison of the point ID union. I tried a tetra parent with two duplicated faces and two missing faces. It passes this check and `IsConvex()`, then `from_pyvista` returns a valid-looking tetra even though `extract_surface()` reports no boundary. A repeated point ID inside one face can also make the 1D path return a self-loop.

Could we reject repeated vertices and duplicate faces, and verify that the faces form one closed shell before calling VTK?

physicsnemo/mesh/io/io_pyvista.py
```diff
@@ -83,6 +89,721 @@ def _geometry_to_vtk_numpy(tensor: torch.Tensor) -> np.ndarray:
...
+        unit_normal = candidate_normals[normal_index] / candidate_norms[normal_index]
+        signed_distances = (points - face_points[0]) @ unit_normal  # (P,)
+        if bool((signed_distances > distance_tolerance).any()) and bool(
+            (signed_distances < -distance_tolerance).any()
+        ):
+            return False
+    return True
+
+
+def _validate_selected_polyhedra_for_tetrahedralization(
```
mehdiataei, line 904: Could we move the parent-connectivity versus face-point consistency part into the common polyhedron validation? Right now it only runs on the 3D tetrahedralization path. The same malformed polyhedron still reaches `cell_centers()` and `extract_all_edges()`, and I could get a wrong centroid and missing edges from it. The convexity check can stay specific to tetrahedralization, but the structural checks seem useful for every path that consumes topology.

mehdiataei (approved):
Hi Peter. I approved the PR with some minor changes reqeusted.

physicsnemo/mesh/io/io_pyvista.py
peterdsharpe: Fixed!

physicsnemo/mesh/io/io_pyvista.py
peterdsharpe: Fixed!

PR #1891, opened 2026-08-04
Pin TensorDict nightly

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Pins the core TensorDict dependency to
`tensordict-nightly[zarr]==2026.8.4`, replacing `tensordict>=0.12.4`. This
provides the latest nightly TensorDict APIs together with its Zarr integration.

The nightly is expected to include Linux aarch64 wheels following
[pytorch/tensordict#1756](https://github.com/pytorch/tensordict/pull/1756). The
2026.8.4 PyPI publish is still incomplete, so `uv.lock` is intentionally left
unchanged in this draft. It will be regenerated once the aarch64 wheels land.

Because the distribution is named `tensordict-nightly` while its Python import
remains `tensordict`, this also teaches the external-import contract that
mapping. Without it, import-linter rejects PhysicsNeMo's existing TensorDict
imports as undeclared dependencies.

- Closes https://github.com/NVIDIA/physicsnemo/issues/1857
- Provides foundation for https://github.com/NVIDIA/physicsnemo/pull/1798
- Unblocks https://github.com/NVIDIA/physicsnemo/pull/1868
- and probably others :)

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [x] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.
- [x] Model implementation standards are not applicable; this PR does not modify a model.

## Dependencies

- Replaces the stable `tensordict` distribution with the pinned
  `tensordict-nightly[zarr]` 2026.8.4 release.
- `uv.lock` remains a follow-up within this draft after the multi-platform
  PyPI upload completes.

## Validation

- `pre-commit run --files pyproject.toml CHANGELOG.md test/ci_tests/prevent_untracked_imports.py`
- `git diff --check`

PR #1894, opened 2026-08-05
feat(mesh/io): zarr save/load for Mesh and DomainMesh via tensordict's zarr backend

negin513:
### Description

Adds zarr save/load for `Mesh` and `DomainMesh` built on **tensordict's upstream zarr storage backend** (pytorch/tensordict#1754), following the discussion on #1798 that physicsnemo should use the upstream support rather than carry its own zarr serialization layer.

Both storage paths are now tensordict backends end to end: `.pmsh`/`.pdmsh` = tensordict-memmap, `.zarr` = tensordict-zarr. physicsnemo contributes only the glue tensordict cannot provide (~250 lines):

- **`physicsnemo/mesh/io/io_zarr.py`** — `to_zarr(mesh, store)` / `from_zarr(store)`.
  - *Layout policy*: chunking aligned to the common training subsample size, plus zstd compression. tensordict's default (one uncompressed chunk per leaf) is a checkpoint layout — partial reads decode entire arrays — so dataset-appropriate chunking is applied here.
  - *Type reconstruction*: `TensorDict.from_zarr` returns a plain `PersistentTensorDict`; a root attr records Mesh vs DomainMesh so `from_zarr` rebuilds the tensorclass. Mesh subgroups inside a DomainMesh store (e.g. `store.zarr/boundaries/vehicle`) load directly.
  - zarr is a lazy `OptionalImport` (same pattern as `io_pyvista`); no new core dependencies.
- **`physicsnemo/mesh/io/__init__.py`** — lazy exports, mirroring `from_pyvista`.
- **`physicsnemo/datapipes/readers/mesh.py`** — `MeshReader` / `DomainMeshReader` / extra-boundary loading route paths containing `zarr.json` to `from_zarr`; discovery, subsampling, pinning, and RNG are untouched. Opt in via `pattern: "run_*/domain_*.zarr"`. Also switches reader discovery from `pathlib.Path.glob` to `glob.glob`: the former silently drops entries under Lustre metadata-server load, so training can proceed silently on a subset of the dataset.

**Upstream issues found while building this:**
1. pytorch/tensordict#1758 — `to_zarr` silently drops `chunks`/`compressors` for nested leaves. Scoped workaround included (`_propagate_nested_create_kwargs`), self-removing once fixed upstream. Without it, `to_zarr` can only produce the checkpoint layout.
2. The tensorclass `to_zarr` proxy re-wraps the returned `PersistentTensorDict` as the tensorclass, which validating tensorclasses reject — avoided by serializing `obj.to_tensordict()`.

**Requires** a tensordict release containing the zarr backend (currently main-only) and `zarr >= 3`. Both remain optional dependencies; the import guard raises a clear error otherwise.

### Scope and trade-offs

`from_zarr` loads eagerly at full resolution; subsampling happens afterwards in the reader, exactly as with memmap. This keeps the loader changes minimal and works well for surface-scale meshes, cold-start-sensitive workloads, and archival/interchange (compressed, self-describing, language-agnostic stores). For very large volume meshes, reading the full interior each draw is wasteful — a chunk-aligned partial-read path over the same stores (read only the subsample window) is the natural follow-up and requires no on-disk format change. Until then, memmap remains the recommended path for volume-scale training.

### Validation

- **Round-trip on real data**: a full DrivAerML DomainMesh case -> `to_zarr` -> stock `DomainMeshReader(pattern="*.zarr")` -> every array bitwise-equal (`torch.equal`) to the memmap source, including all boundaries, interior fields, and global data.
- **Full-scale training exercise**: the complete 484-case DrivAerML dataset converted losslessly and trained end-to-end (GeoTransolver surface and volume, multi-GPU) against the memmap baseline with identical configs; convergence is statistically identical on both stores.

### Checklist

- [ ] The CHANGELOG.md is up to date with these changes.

### Dependencies

Optional: tensordict with the zarr storage backend (> 0.13), `zarr >= 3`.


physicsnemo/mesh/io/io_zarr.py
```diff
@@ -0,0 +1,213 @@
...
+def _mesh_from_group(group: "zarr.Group", device) -> Mesh:
+    """Materialize one mesh group as an in-memory Mesh."""
+
+    def _tensor(arr):
+        return torch.as_tensor(arr[...], device=device)
+
+    def _tree(name):
+        if name not in group:
+            return {}
+        return {k: _tensor(v) for k, v in group[name].arrays()}
```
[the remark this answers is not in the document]
peterdsharpe: @negin513 this seems like a real thing we should fix before merging (probably doable by rewriting this `_tree_` function as a recursive one?)

Also we probably want `_tree` to output a TensorDict, not a `dict[str, Tensor]` (TensorDict is the "canonical" `Mesh` interface; `dict` is just a convenience method)

physicsnemo/mesh/io/io_zarr.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # The memmap format does not persist 0-element tensors but zarr does;
```
peterdsharpe, line 157: FYI, this is fixed upstream in `tensordict` by the already-merged https://github.com/pytorch/tensordict/pull/1750; this will be in the tensordict-nightly build. So probably no special handling needed here for the "no cells" case, after we pin to that?

peterdsharpe (approved):
Overall this looks great @negin513.

Approving to unblock, since we're hoping to land this capability soon. However, please double-check that these prerequisites are satisfied before merging:

- [x] Please add unit tests for `to_zarr` and `from_zarr`, including committing a small golden-file (and checking that it round-trips correctly, including a few edge cases you think are likely).
    - These CI tests will probably fail until `tensordict-nightly[zarr]` is both [released](https://github.com/pytorch/tensordict/actions/workflows/nightly_build.yml) and [pinned](https://github.com/NVIDIA/physicsnemo/pull/1891), no worries.

- [x] Please fix handling for nested TensorDicts in `from_zarr()`, which is a feature we use often. ([Comment](https://github.com/NVIDIA/physicsnemo/pull/1894/changes#r3720594533)) [Details from the TensorDict docs here](https://docs.pytorch.org/tensordict/main/overview.html#nested-tensordicts).

- [x] Please wait for https://github.com/NVIDIA/physicsnemo/pull/1891 to land before merging this PR. (This will essentially be required, as the CI tests requested in the first bullet point here will not pass without this dependency.)

Optionally:

- [x] [This special handling is probably unnecessary](https://github.com/NVIDIA/physicsnemo/pull/1894/changes#r3720626077) (though harmless) once `tensordict-nightly` is pinned, could consider removing for cleanliness.

physicsnemo/mesh/io/io_zarr.py
```diff
@@ -0,0 +1,213 @@
...
+    from tensordict import TensorDict
+
+    if not hasattr(TensorDict, "from_zarr"):
+        import tensordict
+
+        raise ImportError(
+            "This tensordict version has no zarr storage backend "
+            f"(found {tensordict.__version__}; need the release containing "
+            "pytorch/tensordict#1754)."
+        )
```
coreyjadams: nit: Since tensordict is a core dep, we can move the import to the top.  I think we can drop these attribute checks too, since @peterdsharpe 's request to wait for the version of tensordict with zarr to land will make it unnecessary.

physicsnemo/mesh/io/io_zarr.py
```diff
@@ -0,0 +1,213 @@
...
+        return
+    orig = PersistentTensorDict._make_nested
+
+    def _make_nested(self, key, group):
+        out = orig(self, key, group)
+        out.kwargs = self.kwargs
+        return out
+
+    PersistentTensorDict._make_nested = _make_nested
+    PersistentTensorDict._pn_nested_kwargs_patch = True
```
coreyjadams: Looks like the bug closed, ok to drop this?

negin513:
Thanks for the fast review @peterdsharpe! Addressed in 04f5a3c0:

**Your checklist:**
- ✅ **Unit tests** — new `test/mesh/io/io_zarr/` suite: Mesh/DomainMesh round-trips, nested-TensorDict fields, point clouds, boundary-subgroup loads, dtype preservation, error paths, and reader integration. Includes a committed **golden store** (`golden/domain_mesh.zarr`, 377 KB) so on-disk format drift breaks a test instead of users' datasets; expected values regenerate from seeded factories (no pickled refs). The suite skips cleanly (module-level) on tensordict builds without the zarr backend, so CI stays green until the nightly pin lands.
- ✅ **Nested TensorDicts** — `_read_tree` is now recursive and returns `TensorDict` (both your points). Covered by `test_nested_tensordict_fields_round_trip`.
- ⏳ **#1891** — understood, will wait for it before merging.
- 👍 **Empty-cells shim** — agreed; will drop it once we're pinned to a nightly containing pytorch/tensordict#1750.

**Also in this commit:**
- Fixed the recursive-`**` discovery regression Greptile flagged (`glob.glob(..., recursive=True)`).
- **Subsample push-down for zarr stores**: when a reader has `subsample_n_cells/points` set, it now reads *only the selected window* (contiguous row-runs; for cell meshes: cell window → gather referenced points → remap + compose measure weights) instead of materializing the full mesh. Output is bitwise-identical to an eager load followed by in-memory subsampling under the same seed (tested), and it removes the full-resolution-read caveat for large volume meshes noted in the description — verified end-to-end on the full DrivAerML volume training config. `from_zarr` stays the eager full-fidelity loader; sampling policy stays in the readers.
- The nested-kwargs workaround now references the upstream report, pytorch/tensordict#1758.

peterdsharpe:
@negin513 one last thing - does `Mesh._cache` serialize/deserialize too with the Zarr backend? With the memmap backend, it does. This is not necessarily something we *need* to fix here, but it'd be nice to have a consistent behavior here between backends, whether that means "both backends cache, neither does, or both are configurable". Let's discuss this offline and after code freeze; this is non-critical-path and doesn't need to block this PR I think?

tagging @coreyjadams for awareness or suggestions

test/mesh/io/io_zarr/golden/domain_mesh.zarr/zarr.json
```diff
@@ -0,0 +1,1235 @@
+{
```
coreyjadams: Rather than committing many small files, it would be more lightweight for the repo and tests to generate this file as part of CI: just commit the few lines to make the file, rather than the file itself, and put it as a fixture in the pytests?

coreyjadams (approved):
I have a few small changes to request here, but approving to unblock this.

negin513:
Thanks @coreyjadams! All three addressed in afc7a1ab (+ b0ab7a48).

AlsoL  **Golden store** — removed the committed files. Instead of a pure regenerate-and-round-trip fixture (which can't catch format drift, since writer and reader move together), the store is fixture-generated and the test asserts the on-disk *contract* explicitly: group paths, attr names, dtypes, chunking, codecs. Format changes still fail a test and must update the expectations deliberately; no binaries in the repo.

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1894/commits/cce509e76eca647a874c8cb5b75970bf5e93b324

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1894/commits/df8eb4cb226018a823bfe65ba0675525b01693fc

negin513:
/ok to test [bb854d2](https://github.com/NVIDIA/physicsnemo/pull/1894/commits/bb854d29c972cba4f3ecddf08f0b67fbc2f25cff)

PR #1896, opened 2026-08-05
Add DPM-Solver++(2M) diffusion sampler

gertln:
# PhysicsNeMo Pull Request

## Description

Adds `DPMSolverPlusPlus2M` (registry key `"dpmpp_2m"`), a second-order **multistep**
solver for the probability-flow ODE.

`physicsnemo.diffusion.samplers` currently ships no multistep solvers. DPM-Solver++(2M)
reaches second order with **one denoiser evaluation per step** by reusing the previous
step's data prediction, where `HeunSolver` needs two (it computes the second stage
unconditionally and masks it, so its cost is `2 * num_steps` NFE). For a fixed evaluation
budget it can therefore take twice as many steps.

The solver works with **general linear-Gaussian schedules**
`x_t = alpha_t * x_0 + sigma_t * epsilon`. `sample()` reads `alpha`, `sigma`, `alpha_dot`
and `sigma_dot` off its `noise_scheduler` and passes them to the solver, so selecting it
by name is correct for supported linear-Gaussian schedules without any change to the
caller's code.

Reference: [DPM-Solver++: Fast Solver for Guided Sampling of Diffusion Probabilistic
Models](https://arxiv.org/abs/2211.01095), Algorithm 2.

### Motivating use case

Measured with corrdiff-cosmo-era5 with identical initial latents and 10 validation
timestamps, against a high-budget reference:

fig_error_vs_nfe

At matched NFE the advantage runs from 5.9x at 12 NFE to 3.3x at 32 NFE. Practically,
`dpmpp_2m` at 20 NFE is more accurate than `heun` at 36 NFE.

The reference is Heun at 192 NFE. DPM-Solver++(2M) at 192 NFE differs from it by
2.3e-3, an empirical high-budget consistency check between two structurally different
solvers.

### Validation

Closed-form probability-flow ODE on the **EDM** schedule (Gaussian data
distribution, `D(x,t) = x s^2/(s^2+t^2)`, exact solution `x(t) = C sqrt(s^2+t^2)`).
Max-abs error against the exact solution at matched NFE:

| NFE | euler | heun | dpmpp_2m | heun / dpmpp_2m |
|----:|------:|-----:|---------:|----------------:|
| 8   | 8.61e-1 | 2.85e0  | 5.58e-1 | 5.1x |
| 16  | 4.67e-1 | 7.28e-1 | 1.84e-1 | 4.0x |
| 32  | 2.43e-1 | 1.60e-1 | 4.05e-2 | 4.0x |
| 64  | 1.24e-1 | 3.77e-2 | 9.45e-3 | 4.0x |
| 128 | 6.26e-2 | 9.14e-3 | 2.29e-3 | 4.0x |

Across 16 to 128 NFE, DPM-Solver++(2M) produced approximately 4x lower error than Heun
on this benchmark, consistent with the benefit of taking twice as many second-order
steps at matched NFE.

Full-trajectory manufactured-solution tests also verify convergence on VP and VE,
including the zero endpoint.

### Design

- **String dispatch uses the scheduler’s parameterization.** For `dpmpp_2m`,
  `sample()` injects `alpha`, `sigma`, and their derivatives from its noise
  scheduler, unwrapping `DomainParallelNoiseScheduler` when necessary. It
  copies `solver_options` and overrides schedule callbacks so they remain
  consistent with the scheduler’s timesteps. Direct construction defaults to
  EDM, and partial callback sets are rejected.
- **Passed instances remain caller-configured.** `sample()` does not mutate a
  preconstructed solver. Callers using an instance with a non-EDM scheduler
  must provide matching schedule functions when constructing it.
- **Lifecycle capabilities are private opt-ins.** `_requires_schedule_fns` and
  `_requires_state_reset` remain outside the runtime-checkable `Solver`
  protocol, preserving existing duck-typed implementations. Stateful solvers
  are reset before each trajectory and in `finally`; the marker prevents
  unrelated third-party `reset()` methods from being called.

### Numerical edge cases

- Final steps are detected using `sigma_next == 0`, and degenerate steps using
  `sigma_cur == 0`, rather than checking time directly. These conditions
  coincide with `t == 0` only for EDM. Both use tensor masks, preserving
  `fullgraph` tracing.
- At `sigma_cur == 0`, the denoiser is evaluated at a finite surrogate time
  and its result is masked, so the step returns the state unchanged. The ODE
  right-hand side may be singular at this endpoint.
- A repeated timestep, reachable when a fine ladder is rounded to a
  low-precision dtype and produced by the iDDPM ladder even in float64, is an
  exact identity step. The following step falls back to first order rather
  than dividing by zero.
- The data prediction is recovered as
  `(sigma_dot * x - sigma * rhs) /
  (alpha * sigma_dot - sigma * alpha_dot)`, avoiding explicit division by
  `sigma` near the endpoint.
- Steps in the half-log-SNR variable `lambda = log(alpha / sigma)` are
  computed in at least float32, regardless of the latent dtype.
- `step()` contains no tensor-value-dependent Python branches that would
  force host synchronization and prevent `fullgraph` tracing.

### Backward compatibility

Purely additive. The four pre-existing solvers produce bit-identical output to the base
commit (`max|diff| = 0.0`, with and without `time_eval`), and neither new conditional in
`sample()` fires on their path. Under the EDM defaults, the generalized update agrees
with the EDM specialization to floating-point rounding and uses `expm1` to avoid
cancellation.

## Checklist

- [x] If I am implementing a new model or modifying any existing model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation). (N/A: this is a solver, not a model.)

## Dependencies

None.


physicsnemo/diffusion/noise_schedulers/noise_schedulers.py
```diff
@@ -324,6 +324,16 @@ class LinearGaussianNoiseScheduler(ABC, NoiseScheduler):
...
+    Attributes
+    ----------
+    is_edm_parameterization : bool, default=False
+        Whether the schedule satisfies :math:`\sigma(t) = t` and
+        :math:`\alpha(t) = 1`, so that the diffusion time is the noise level.
+        Solvers derived only for this case, such as
+        :class:`~physicsnemo.diffusion.samplers.solvers.DPMSolverPlusPlus2M`,
+        are rejected by :func:`~physicsnemo.diffusion.samplers.sample`
+        otherwise. Subclasses satisfying both identities may set it to ``True``.
+
```
CharlelieLrt: Why is this needed? The new solver should work for any noise schedule that is of the linear-gaussian form `x_t = alpha_t * x_0 + sigma_t * epsilon`. I think it would be better (and not very difficult) to generalize its implementation to make it work for any schedule `(alpha_t, sigma_t)` (and not just `(1, t)`), rather than adding edge cases and checks.

gertln: You are right, the implementation can be generalized using the scheduler’s `alpha`, `sigma` and their derivatives `alpha_dot `and `sigma_dot`. These provide the schedule values needed by the update and allow the solver to recover the data prediction from the denoiser’s ODE output. I can let `sample()` pass them when constructing `dpmpp_2m`, making the string API work with every linear-Gaussian scheduler. I would then remove the EDM-only checks and detect the final step using `sigma_next == 0`.

CharlelieLrt: Yes, I think that would make sense.
For broader context, we have other noise schedules that we might merge soon, so we don't want to have this new solver compatible with EDM but not with those.

Just a detail:

> I can let sample() pass them when constructing dpmpp_2m, making the string API work with every linear-Gaussian scheduler.

This would work, but is a little bit cumbersome in terms of API for `sample()`, because one would have to pass all `sigma`, `alpha` and their derivative to the `sample()` (which itself needs to pass them to the constructor of the solver). But that's more a detail, you can go  for this and I'll try to think of a more elegant way to handle this in the `sample()` API.

gertln: The new commit generalizes the solver using `alpha`, `sigma` and their derivatives from the noise scheduler. When `dpmpp_2m` is selected by name, `sample()` configures it automatically. I have removed the EDM-only attribute and checks.

megnvidia (approved):
lgtm

CharlelieLrt:
Replaced with https://github.com/NVIDIA/physicsnemo/pull/1932 .

PR #1902, opened 2026-08-10
Geotransolver flare domain parallel

coreyjadams:
# PhysicsNeMo Pull Request

This PR adds support for FLARE and GeoTransolver domain parallelism at the model level.  It does not include datapipe support or recipe support yet. Those will come in follow ups.

The models are coming together because FLARE attention is a backend for GeoTransolver; to support GeoT we have to have FLARE anyways so it's cleanest to bring them both together.

There is one tweak to `shard_tensor.py` incorporated into the PR that needs to be addressed, that really only rears its head in FLARE: when using a wrapper in shard tensor, where we have a `torch.Tensor` entering the wrapper, it is currently possible for a plain tensor to enter that wrapper.  FLARE has this as a parameter that enters the attention mechanism, which is unusual compared to all the other models we see.

To address this, in this PR I add a tweak to fix this in shard tensor: plain `torch.Tensor` objects will fire promotion (to replicated DTensor, since it's mathematically the same and the machinery is all there) before going to handlers.

## New wrappers 

There are several new operations in `domain_parallel/shard_utils/`: a `linear_patches.py` is the most consequential, and `attention_patches.py` has substantial additions.   A new file `grad_ops.py` consolidates backward-only operations from several other files (`point_cloud_ops.py`, `conv_patches.py`, `attention_patches.py`) into one logical file and `test_grad_ops.py` introduces some tests.

### Linear Patches

Linear layers are embarrassingly parallel, however pytorch's DTensor op flattens them prior to operating.  For some tensors, the `view` operation itself become problematic especially if the sharded dimension is not the last dimension.
To fix this, the PR adds a custom linear path which routes all linear operations through a local, differentiable op that doesn't hit view issues.

### Attention wrappers

The standard attention path in physicsnemo currently is ring attention.  FLARE, however, has one tensor which is NOT sharded.  This PR adds an attention path for a replicated tensors: it doesn't have to traverse a ring, and so we use a different algorithm.

### Layer Updates

For both `gale.py` and `context_projector.py`, we have to do a reduction after the tokenization/projection.  We duck type, so import shard tensor is unnecessary, and trigger a redistribute.  The sharded placement after projecting back out is automatic, just like with transolver.

## Checklist

## Review Process


physicsnemo/models/geotransolver/context_projector.py
```diff
@@ -1057,4 +1058,12 @@ def build_context(
         # Concatenate all context features along the last dimension
         context = torch.cat(context_parts, dim=-1) if context_parts else None
 
+        # The tokenizers reduce over the (possibly sharded) point axis, so a
+        # distributed context arrives as an unreduced Partial sum, and every
+        # downstream block consumes it nonlinearly. Resolve it once here --
+        # a single differentiable all-reduce reused by all layers. Duck-typed
+        # because models cannot import domain_parallel.
+        if context is not None and hasattr(context, "redistribute"):
+            context = context.redistribute(placements=[Replicate()])
```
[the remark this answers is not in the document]
coreyjadams: This is not a code path that will be likely to be exercised, it's P2 and not needed to fix here.

coreyjadams:
/ok to test [cd2316b](https://github.com/NVIDIA/physicsnemo/pull/1902/commits/cd2316b614f5ec39172f54bc08d1df0d3062d92e)

negin513 (commented):
A few findings from a detailed pass over the domain-parallel changes; otherwise LGTM.

physicsnemo/domain_parallel/shard_utils/attention_patches.py
```diff
@@ -857,53 +991,118 @@ def sdpa_wrapper(
...
-        if attn_mask is not None:
-            local_attn_mask = attn_mask.to_local()
-        else:
-            local_attn_mask = None
+        if q_sharded:
+            # Rule 2 backward clause: dK/dV are partial sums over q's shards.
+            local_k = GradReducer.apply(local_k, k._spec)
+            local_v = GradReducer.apply(local_v, v._spec)
+
         local_output = torch.nn.functional.scaled_dot_product_attention(
```
negin513, line 1092: **Wrong causal masking on sharded q.** This new sharded-q/replicated-KV branch passes `is_causal` and `dropout_p` straight to local SDPA, so ranks > 0 mask against local row indices instead of global ones — silently wrong output (this combination used to raise `MissingShardPatch`). The branch below (~line 1090) rejects both flags; this one should too.

physicsnemo/domain_parallel/shard_utils/attention_patches.py
```diff
@@ -857,53 +991,118 @@ def sdpa_wrapper(
...
+        raise MissingShardPatch("q may only be sharded on the sequence dim (2)")
+    if k_placement.is_shard() and k_placement.dim != 2:
+        raise MissingShardPatch("k/v may only be sharded on the sequence dim (2)")
 
-    # if the placements are replicated (which is what we expect in transolver's
-    # Physics Attention)
-    # then just run locally and convert the output back to a replicated tensor:
+    mesh = q._spec.mesh
+    q_sharded = q_placement.is_shard()
+    kv_sharded = k_placement.is_shard()
```
negin513, line 1047: **`Partial` placements are treated as replicated.** The `is_shard()`-or-else check sends `Partial` tensors down the replicated path, computing attention on unreduced per-rank partial sums. Suggest raising on `is_partial()` (or redistributing) here in the wrapper — then the `redistribute()` workarounds in `gale.py`/`context_projector.py` could be removed.

physicsnemo/domain_parallel/shard_utils/attention_patches.py
```diff
@@ -857,53 +991,118 @@ def sdpa_wrapper(
...
+                q._spec.placements,
+                sharding_shapes=output_shard_shapes,
+            )
+        return ShardTensor.from_local(local_output, mesh, q._spec.placements)
+
+    # --- Replicated K/V: attention is local over the full K/V set. ----------
+    if not kv_sharded:
+        # A mask indexes (..., S_q, S_kv): its rows must be laid out like q.
+        if attn_mask is not None and hasattr(attn_mask, "_spec"):
+            if attn_mask._spec.placements[0] != q_placement:
```
negin513, line 1079: **Mask validation compares shard dims across tensors of different rank.** A correctly row-sharded 2-D mask (`Shard(0)`) is rejected against a 4-D q (`Shard(2)`), while a wrongly column-sharded 3-D mask passes. This needs a rank-aware dim comparison (same pattern at ~line 1081 on the ring branch).

physicsnemo/domain_parallel/shard_utils/linear_patches.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def _replicated_local(tensor: torch.Tensor | None) -> torch.Tensor | None:
+    r"""Return the local view of a fully-replicated tensor, or ``None``.
+
+    ``None`` (no bias) and plain tensors pass through; distributed tensors
+    yield their local view only if fully replicated. Returns
+    ``NotImplemented`` for anything else so the caller can fall back.
+    """
+    if tensor is None or not isinstance(tensor, DTensor):
```
negin513: **`_replicated_local` misses `ShardTensor`.** `isinstance(tensor, DTensor)` does not match `ShardTensor` weights (it subclasses `torch.Tensor`), so a ShardTensor weight passes through un-localized and produces a nested ShardTensor with corrupt spec. Fix: `isinstance(tensor, (ShardTensor, DTensor))` — matching the convention in `conv_patches.py` and `normalization_patches.py`.

physicsnemo/domain_parallel/shard_utils/linear_patches.py
```diff
@@ -0,0 +1,140 @@
...
+    )
+    local_weight = _replicated_local(weight) if local_path else NotImplemented
+    local_bias = _replicated_local(bias) if local_path else NotImplemented
+
+    if local_weight is NotImplemented or local_bias is NotImplemented:
+        return _torch_function_fallback_via_dtensor(func, args, kwargs)
+
+    # Weight/bias grads are partial sums over the input's sharded dims.
+    input_spec = input._spec
+    local_weight = ConvGradReducer.apply(local_weight, input_spec)
```
negin513: **Cleanup: rename `ConvGradReducer`.** Linear grads now flow through a conv-named reducer that is only incidentally shape-agnostic — if it ever specializes for conv, this breaks silently. Since both reducers now live in `grad_ops.py`, suggest a neutral name and possibly merging it with the near-identical `GradReducer`.

physicsnemo/domain_parallel/shard_utils/linear_patches.py
coreyjadams: Already fixed.

coreyjadams: (For what it's worth, the tests were actually catching this one and failing which is nice.)

physicsnemo/domain_parallel/shard_utils/attention_patches.py
coreyjadams: Resolved.

physicsnemo/domain_parallel/shard_utils/attention_patches.py
```diff
@@ -857,53 +992,136 @@ def sdpa_wrapper(
...
+    # 1. The output follows q. q never moves; a sharded q parallelizes
+    #    attention rows across ranks.
+    # 2. The K/V placement decides communication. Sharded K/V blocks must be
+    #    accumulated with the log-sum-exp combine: a ring when q is also
+    #    sharded (each q block must visit every K/V block), a single static
+    #    fold when q is replicated (every rank already holds all of q and
+    #    one K/V block). Replicated K/V needs no forward communication; the
+    #    only distributed concern is backward, where dK/dV are partial sums
+    #    over q's shards and need an all-reduce -- and only when q is
+    #    sharded (with q replicated, every rank computes identical grads).
```
pzharrington: Suggest migrating some of these notes on behavior to the docstring. At the moment the main docstring is not too descriptive, a bit opaque on what types of sharding on q,k,v are supported

physicsnemo/domain_parallel/shard_utils/grad_ops.py
pzharrington, line 1: Nifty utils -- minor nit is the `ContiguousGrad` and `ConvGradReducer` are described and/or named as being specific to attn/conv ops, but they are actually more generally useful and could be rolled into implementations for other ops. Maybe rename the latter to e.g. `WeightGradReducer`?

physicsnemo/domain_parallel/shard_utils/linear_patches.py
```diff
@@ -0,0 +1,144 @@
...
+            # Plain tensor: grads must be reduced eagerly.
+            return param if param is None else ConvGradReducer.apply(param, input_spec)
+        if all(p.is_replicate() for p in param._spec.placements):
+            # No-comm unwrap; the Partial cotangent resolves lazily downstream.
+            return param.to_local(grad_placements=param_grad_placements)
+        # Gather; the redistribute backward reduce-scatters the Partial cotangent.
+        return param.full_tensor(grad_placements=param_grad_placements)
+
+    local_weight = gather_param(weight)
+    local_bias = gather_param(bias)
```
pzharrington: nit: the name `gather` is a bit confusing here since it is really just marking the param to get an appropriate reduction in the backward pass....or at least that's what I think it's doing 😅. Is the last return line ever reached / is this intended to be used with actually sharded weights?

physicsnemo/domain_parallel/shard_utils/linear_patches.py
coreyjadams: Done

physicsnemo/domain_parallel/shard_utils/grad_ops.py
coreyjadams: Went with "GradReducer", but they are merged.

physicsnemo/domain_parallel/shard_utils/attention_patches.py
coreyjadams: Resolved.

physicsnemo/domain_parallel/shard_utils/attention_patches.py
coreyjadams: I'm going to defer this since currently, all our P0 models hit this via standard attention techniques, or FLARE.  aka, they use an MLP on input, unbind along something of size 3, and then reshape and do attention -> always producing the same shape.  We can put checks in if it's truly necessary, but, I think this one is a narrow edge case.

physicsnemo/domain_parallel/shard_utils/linear_patches.py
coreyjadams: This is actually protection against the use case of DTensor if we aren't autopromoting - which, by the way, could be turned off.  So if a sharded DTensor comes in this is, actually, doing a gather.  

Because this is intercepting _above_ Dtensor we don't have the same assurances it will `allgather` before operating so it's good to have this check, I think.

coreyjadams: Regardless, I changed the name to `localize_param` since that's a little bit more descriptive to 99% of the usage...

physicsnemo/domain_parallel/shard_utils/attention_patches.py
coreyjadams: Moved to docstring.

coreyjadams:
/ok to test [44979b6](https://github.com/NVIDIA/physicsnemo/pull/1902/commits/44979b699831e3fd1dfcc681a40d58725a66c566)

physicsnemo/domain_parallel/shard_utils/linear_patches.py
pzharrington: Ah right ok so the last line is covering the case of FSDP

pzharrington (approved):
LGTM!

PR #1907, opened 2026-08-11
fix(test): rename io_zarr test_round_trip.py to avoid pytest module collision

negin513:
## Description

#1894 added `test/mesh/io/io_zarr/test_round_trip.py`, which shares a basename with the pre-existing `test/mesh/io/io_pyvista/test_round_trip.py`. The `test/mesh` tree has no `__init__.py` files, so pytest's default (prepend) import mode maps both files to the bare module name `test_round_trip`, and collection aborts with:

```
import file mismatch:
imported module 'test_round_trip' has this __file__ attribute:
  test/mesh/io/io_pyvista/test_round_trip.py
which is not the same as the test file we want to collect:
  test/mesh/io/io_zarr/test_round_trip.py
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
```

This breaks any CI stream that collects the full test tree — first seen in the `multigpu-static` stream on #1904 ([failing job](https://github.com/NVIDIA/physicsnemo/actions/runs/31499600444/job/93805987739)). It was not caught on #1894 itself because the multi-GPU selection gate skipped that stream there.

Fix: rename the io_zarr file to `test_zarr_round_trip.py` (pure `git mv`, no content changes). Verified this was the only bare-module-name collision among the `__init__.py`-less test directories. A longer-term hardening option is adding `__init__.py` under `test/mesh` or switching to `--import-mode=importlib`, left out of scope here to unblock CI.

## Checklist


coreyjadams (approved):

coreyjadams:
Peter is out on PTO; I'm going to force merge this so we can unblock CI.

coreyjadams:
/ok to test [df6e2cc](https://github.com/NVIDIA/physicsnemo/pull/1907/commits/df6e2ccf880a0b81c244b45bac09230e280682a8)

PR #1914, opened 2026-08-12
Compile-safe ShardTensor subclassing and domain-parallel halo scatter-correction

dallasfoster:
# PhysicsNeMo Pull Request

## Description

Builds on the `torch.compile` support for `domain_parallel.ShardTensor` merged in #1682, in two
related parts. 

### 1. Compile-safe ShardTensor subclassing and AOT tangent coercion

#1682 makes the base `ShardTensor` work under `torch.compile`, but a *subclass* of it still can't
compile without hand-re-implementing PyTorch's internal flatten/unflatten protocol. This part adds
the small set of hooks a subclass needs, and fixes the two places a subclass's autograd otherwise
breaks under the compiler.

**How we use it in the ALCHEMI Toolkit.** ALCHEMI runs domain-parallel atomistic simulations on a
`ShardTensor` subclass that attaches per-instance routing metadata to each tensor — e.g. the packed
halo-exchange plan for a sharded particle system (the same kind of routing produced by part 2
below). To compile the model end-to-end, that metadata has to travel *with* the tensor as a graph
input — so it survives Dynamo graph breaks and can change value each step without forcing a
recompile — and the subclass's `to_local()` and backward pass have to stay correct through
AOTAutograd. The base flatten protocol has no room for a subclass's extra tensors or opaque
metadata, so today such a subclass either can't compile or has to fork torch internals; these hooks
close that gap while leaving base `ShardTensor` untouched.

What's added:

- **Carrying extra state through compile** — `_extra_inner_tensors` (extra always-present inner
  tensors, e.g. a routing plan) and `_subclass_propagated_attrs` (per-instance metadata copied onto
  op results), wired into the flatten protocol via `__subclass_flatten_context__` /
  `__subclass_unflatten__`, `_stable_inner_sentinel`, and a DTensor-style `__metadata_guard__` that
  keeps opaque metadata from triggering spurious recompiles.
- **Keeping subclass autograd correct** — a `grad_dtype` shield and a `__torch_function__` tracing
  passthrough that restore `to_local()` differentiability under compile, plus
  `install_aot_plain_tangent_coercion` for a backward tangent materialized as a plain tensor across
  a graph break.

The subclass-friendly `__coerce_same_metadata_as_tangent__` **extends** the coerce hook from #1682
rather than replacing it: it preserves main's "partial means partial" invariant (it still raises on
a `Partial` tangent target and does not reintroduce normalize-in-coerce) and only adds the subclass
paths — accepting a nested flatten context and rebuilding a base-typed tangent into the requested
`ShardTensor`-subclass type; a foreign `expected_type` keeps the DTensor `None` convention. All
extension points are empty / no-op by default, so base `ShardTensor` behavior is unchanged.

### 2. Halo scatter-correction for domain-parallel `ShardTensor`

A `torch.compile`-safe halo scatter-correction primitive for `Shard(0)` ShardTensors with a
borrowed-ghost overlay:

- `halo_scatter_correct` fuses the fold-to-owner / refresh-ghost exchanges into a self-adjoint
  `torch.library.custom_op` (correct forward and backward under `aot_eager` and inductor). Routing
  rides as a packed graph-input tensor (`pack_halo_routing`, with a fixed-cap option for
  `dynamic=False`), so it survives graph breaks and per-step value changes without recompiling.
- `select_halo_backend` routes by node locality between a portable functional-collective (`funcol`)
  path — the always-available fallback and numerical oracle — and an intra-node symmetric-memory
  (CUDA-IPC) path.
- `register_halo_scatter_handlers` wires the correction onto `ShardTensor.scatter_add` / `index_add`
  and their in-place `scatter_add_` / `index_add_` forms (correct value and gradient, eager and
  compiled).

**Validation.** The full local suite (gloo) and a 4×H100 `multigpu_static` run pass: subclass
extension points (6), AOT tangent coercion (2), the compile tests covering the coerce hook +
`to_local` + subclass paths incl. 2D mesh (41), and halo scatter compile (6) — bitwise-identical to the `funcol` reference where compared.

## Checklist

## Dependencies

No new packaged dependencies. The intra-node halo transport uses PyTorch's symmetric-memory
(CUDA-IPC) and degrades to the `funcol` path where it is unavailable.

## Review Process


negin513 (approved):
LGTM

dallasfoster:
/ok to test [f9c8bf5](https://github.com/NVIDIA/physicsnemo/commit/f9c8bf5d62eef9310de7f325bed4af3a8659c209)

physicsnemo/domain_parallel/shard_utils/halo_scatter.py
```diff
@@ -0,0 +1,714 @@
...
+        _torch_function_fallback_via_dtensor,
+    )
+
+    def _local(x):
+        # ``index`` / ``src`` may arrive promoted to a Replicate DTensor (the base
+        # ``_promote_plain_handler_args`` runs before this handler). Unwrap it with the
+        # differentiable ``to_local()`` -- the raw ``_local_tensor`` would sever the
+        # ``from_local`` bridge and drop the gradient back to the plain source.
+        if isinstance(x, ShardTensor):
+            return x._local_tensor
```
coreyjadams, line 655: why use `to_local()` on DTensor() but `_local_tensor` on ShardTensor?  One is a straight attributed, one is a differentiable method.

physicsnemo/domain_parallel/shard_utils/halo_scatter.py
```diff
@@ -0,0 +1,714 @@
...
+                grad_out._local_tensor
+                if isinstance(grad_out, ShardTensor)
+                else grad_out
+            )
+            return g, None, None, None
+
+    def _wrap_like(src, local, routing, requires_grad):
+        if requires_grad:
+            return _WrapLocalAsShard.apply(local, type(src), src._spec, routing)
+        return _build(type(src), local, src._spec, routing, False)
```
coreyjadams, line 698: If I'm reading this correctly, and I'm not sure I am, this looks really similar to `FromLocalTensor` except more targeted to subclasses (so don't force as ShardTensor) and with more flexibility in the backward pass routing.  What do you think about making this abstraction in `shard_tensor.py` so that other subclasses can consume it?

physicsnemo/domain_parallel/shard_utils/halo_scatter.py
```diff
@@ -0,0 +1,714 @@
...
+        local_result = _inplace_to_oop[f](_local(self), dim, _local(index), _local(src))
+        corrected = halo_scatter_correct(local_result, routing, group=self._spec.mesh)
+        out = _wrap_like(self, corrected, routing, _needs_grad(self, index, src))
+        # Write corrected values into local storage so a caller reusing the accumulator sees
+        # them; the detached no_grad copy avoids recording an in-place op on a leaf.
+        with torch.no_grad():
+            self._local_tensor.detach().copy_(corrected)
+        return out
+
+    for func in (torch.Tensor.scatter_add, torch.Tensor.index_add):
```
coreyjadams, line 745: index_add has an extra kwarg here, `alpha`, that is ignored in both of these routings.  I think you can pretty easily capture it in the handlers above, if f is index_add instead of scatter_add.

dallasfoster:
/ok to test [fb8fd98](https://github.com/NVIDIA/physicsnemo/pull/1914/commits/fb8fd98d791df626a7b4017663fd171c8da70427)

physicsnemo/domain_parallel/shard_utils/halo_scatter.py
```diff
@@ -0,0 +1,714 @@
...
+
+    # In-place forms map to out-of-place so the correction's backward chains cleanly (an
+    # in-place op on the inner tensor is dropped from the wrapper-subclass backward).
+    _inplace_to_oop = {
+        torch.Tensor.scatter_add_: torch.Tensor.scatter_add,
+        torch.Tensor.index_add_: torch.Tensor.index_add,
+    }
+
+    def _scatter_inplace_handler(f, types, args, kwargs):
+        self = args[0]
```
coreyjadams, line 728: I suspect, for this function and for `scatter_handler`, you should assert spec._mesh.ndim==1 else raise MissingShardWrapper, etc.  Meaning, this seems to only support sharding along a single mesh rank, which is fine, but it should raise loudly instead of fail weirdly.

physicsnemo/domain_parallel/shard_utils/halo_scatter.py
```diff
@@ -0,0 +1,714 @@
...
+        try:
+            world = dist.get_world_size(pg)
+            gathered: list[object] = [None] * world
+            dist.all_gather_object(gathered, socket.gethostname(), group=pg)
+            cached = len({str(h) for h in gathered}) > 1
+        except Exception:  # pragma: no cover - fall back to a local heuristic
+            try:
+                cached = dist.get_world_size(pg) > torch.cuda.device_count()
+            except Exception:
+                cached = False
```
coreyjadams: I have some reservations about bare Exception catching in this code: it decides which path to take and is doing collectives inside a try/except.  If one rank fails, or hits something unexpected, this will produce very difficult to debug crashes and hangs.  Can we remove the try/except?  I don't really think it's needed on this logic.

physicsnemo/domain_parallel/shard_utils/halo_scatter.py
```diff
@@ -0,0 +1,714 @@
...
+        )
+        send_counts = [int(send_sizes[r][rank]) for r in range(world_size)]
+        recv_counts = [int(send_sizes[rank][j]) for j in range(world_size)]
+        received_back = funcol_all_to_all_v_rows(
+            send_rows, send_counts, recv_counts, group
+        )
+
+        # Fold returned contributions into the lent owned rows; float64 accumulator
+        # for float32 inputs.
+        acc_dtype = torch.float64 if padded.dtype == torch.float32 else padded.dtype
```
coreyjadams: should this upcast accumulators for half precisions too?

physicsnemo/domain_parallel/shard_utils/halo_scatter.py
```diff
@@ -0,0 +1,714 @@
...
+        dtype = padded.dtype
+        group_name = _symm_group_name(group)
+        max_rows = _global_max_staged_rows(send_sizes, world_size)
+
+        # Reverse sends ghost contributions back to owners: readers pull from my buffer
+        # (peers I borrowed from); sources are the peers I pull from (peers I lent to).
+        readers = [s for s in range(world_size) if int(send_sizes[s][rank]) > 0]
+        sources = [j for j in range(world_size) if int(send_sizes[rank][j]) > 0]
+
+        acc_dtype = torch.float64 if dtype == torch.float32 else dtype
```
coreyjadams: Same comment about half precision accumulators.

coreyjadams (commented):
I've left a number of comments on `halo_scatter.py`.  Posting them right away while I go through `shard_tensor.py`.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -954,39 +975,130 @@ def placements(self) -> tuple[Placement, ...]:
...
+            if getattr(t, sentinel, None) is None:
+                for attr in attrs:
+                    setattr(t, attr, getattr(source, attr))
+
+        if isinstance(result, ShardTensor):
+            _apply(result)
+        elif isinstance(result, (tuple, list)):
+            for r in result:
+                _apply(r)
+        return result
```
coreyjadams, line 1069: Claude raised this error and I spent some time understanding it, fully, to see what it was going on about.  I think in `halo_scatter.py` you have a custom wrapper to make sure that when an auto-promoted object goes through the wrapper mechanations of shard tensor, it comes back up as the right class.  But, according to Dr. Claude, the subclass and metadata containers carried along can get lost in some of that machinery:

```python
agg.scatter_add(0, i, s)            # corrected
(agg * 1.0).scatter_add(0, i, s)    # base ShardTensor, no routing, no correction, no error
```

I'm not quite sure how to fix address this yet, just that it would sure be nice if we could subclass for you in a way that didn't require a million hoops.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -1368,11 +1535,24 @@ def __torch_function__(cls, func, types, args=(), kwargs=None):
...
+            # Under torch.compile tracing, route unpatched ops straight to
+            # ``__torch_dispatch__`` (like DTensor) so the aten-level path -- which AOT
+            # handles correctly -- owns the op and keeps the graph differentiable.
+            with torch._C.DisableTorchFunctionSubclass():
+                return func(*args, **kwargs)
+        else:
+            res = _torch_function_fallback_via_dtensor(func, args, kwargs)
+        # Ride subclass routing metadata onto op outputs (eager only; under compile it
+        # travels via the flatten context). No-op for base ShardTensor.
+        if cls._subclass_propagated_attrs and not torch.compiler.is_compiling():
```
coreyjadams: Senior Developer Claude thinks `is_compiling` will always read False inside `__torch_function__`, so `_is_tracing` should be used here instead.

physicsnemo/domain_parallel/shard_tensor.py
```diff
@@ -1754,3 +1934,62 @@ def scatter_tensor(
...
+def install_aot_plain_tangent_coercion() -> None:
+    r"""Patch ``AOTDispatchAutograd.process_runtime_tangent`` so a plain runtime backward
+    tangent is rebuilt into a :class:`ShardTensor` when the graph traced a ShardTensor
+    tangent there.
+
+    AOT can coerce a runtime *subclass* tangent but has no hook for a *plain* tensor where
+    the graph traced a subclass, so it raises "guessed its metadata incorrectly" when a
+    ShardTensor boundary crosses a graph break. The plain cotangent is value-correct at a
+    ``Replicate`` boundary, so rebuilding via ``__tensor_unflatten__`` is lossless. Scoped to
+    ShardTensor, idempotent, degrades gracefully; a general AOT gap pending an upstream hook."""
```
coreyjadams, line 1949: I admit to not understanding what is happening here, like at all.

coreyjadams (approved):
I am approving this, since he have to merge for code freeze.  I have one more comment that I want to raise but couldn't flag it in the code easily.

compiling uses hashing on the tensor spec to check for recompilation.  In subclasses, if you have extra tensor data - depending on the tensors - those might or might not need to trigger recompilation.  Without them in the hash they'll be ignored.

Otherwise, I think some of the IPC / NCCL differences would be nice - in a future PR, not here - to consider consolidating in to a cleaner more general utility we can use for the convolution and natten halos too, if they are more performant.  And the index select ops could benefit too, those are terrible for performance today, don't even look at them 🙈 .

Thanks for opening this @dallasfoster and it looks like good work.

physicsnemo/domain_parallel/shard_utils/halo_scatter.py
dallasfoster: Intentional, but the comment was in the wrong spot — fixed. For `self` the backward already comes from the custom op + wrapper, so `to_local()` there would just add a redundant autograd node. `index`/`src` come in as Replicate DTensors, and those *do* need `to_local()` or the gradient doesn't make it back to the plain source. Comment now covers both.

physicsnemo/domain_parallel/shard_utils/halo_scatter.py
dallasfoster: Agreed — this is basically a subclass-flavored `FromLocalTensor`, and it'd be nice to have it in `shard_tensor.py` for other subclasses to reuse. It's tangled up with the metadata-loss thing you flagged, so I'd rather do both in a follow-up than squeeze it in here. Tracking it.

physicsnemo/domain_parallel/shard_utils/halo_scatter.py
dallasfoster: Fixed — it passes `alpha` through for `index_add` now instead of dropping it.

physicsnemo/domain_parallel/shard_utils/halo_scatter.py
dallasfoster: Good call — both handlers now check for a 1-D mesh and raise if it isn't, instead of quietly doing the wrong thing.

physicsnemo/domain_parallel/shard_utils/halo_scatter.py
dallasfoster: Removed it. Now if the all-gather fails it raises on every rank instead of some ranks silently falling back to a different answer — which was the deadlock risk you're pointing at.

physicsnemo/domain_parallel/shard_utils/halo_scatter.py
dallasfoster: See above.

physicsnemo/domain_parallel/shard_tensor.py
dallasfoster: Yeah, same thing as the `_build` comment above. The propagation only re-attaches metadata to ShardTensors that come out of an op, so something like `agg * 1.0` drops to a plain ShardTensor with no routing, and then `scatter_add` correctly sees nothing to correct. Fixing it properly means carrying the extra inner tensors through the auto-wrap path — bigger than I want to touch under freeze, so it's on the follow-up list. For now `agg.scatter_add(...)` is the path that works.

physicsnemo/domain_parallel/shard_tensor.py
dallasfoster: Fixed — using `_is_tracing` now, like the check just above. You're right that `is_compiling()` is always False in here, so that propagation was actually running during tracing when it shouldn't have been.

physicsnemo/domain_parallel/shard_tensor.py
dallasfoster: Fair, it's dense — I'll expand the docstring. Short version: across a graph break the backward tangent can come back as a plain tensor even though the graph expected a ShardTensor, and torch has no hook to re-wrap it, so it errors. At a Replicate boundary that plain tensor is the same values as the local shard, so we just rebuild the ShardTensor before handing it back. Only touches ShardTensor, and no-ops if torch moves things around. I'll add an example.

dallasfoster:
/ok to test [fc2b1b4](https://github.com/NVIDIA/physicsnemo/pull/1914/commits/fc2b1b4a3ea8e286f18a7726f469f8b88862bf7c)

dallasfoster:
/ok to test [fc2b1b4](https://github.com/NVIDIA/physicsnemo/pull/1914/commits/fc2b1b4a3ea8e286f18a7726f469f8b88862bf7c)

dallasfoster:
/ok to test [055d13b](https://github.com/NVIDIA/physicsnemo/pull/1914/commits/055d13b6f989688d8a177cfe836f9f8ff84718c1)

PR #1921, opened 2026-08-13
Optimized HealDA data loaders to avoid GIL

negin513:
## Description

Migrates the HealDA `ZarrLoader` remote zarr reads and `s3://` downloads in `physicsnemo.core.filesystem` from fsspec/s3fs to [obstore](https://github.com/developmentseed/obstore). obstore's Rust-native client performs object fetches **off the GIL** with real connection pooling, so concurrent chunk reads no longer serialize on the Python interpreter.

### Benchmark

Concurrent zarr chunk reads against public HRRR data on NOAA S3 (4 surface fields × 96 chunks = 384 GETs of ~45 KB, cold cache, 3 trials per backend), using the exact `ZarrLoader` code path:

| Backend | Median wall clock |
|---|---|
| fsspec/s3fs (before) | 12.4 s |
| obstore (after) | **1.6 s (7.7× faster)** |

benchmark

This is the many-small-GET regime that HealDA's per-timestep 2D/3D variable reads hit; for large-chunk stores both backends converge toward bandwidth-bound parity.

### Changes

- **`ZarrLoader`**: remote URLs now open through `zarr.storage.ObjectStore` backed by `obstore.store.from_url`. fsspec-style `storage_options` keys (`anon`, `key`, `secret`, `token`, `endpoint_url`, `region_name`) are translated to obstore config names so existing call sites keep working; unknown keys pass through. Falls back to the fsspec path if obstore is not installed.
- **`core/filesystem.py`**: `s3://` single-file and recursive downloads in `_download_cached` stream through obstore (8 MB chunks, layout-preserving recursive fetch, `FileNotFoundError` on empty prefixes). `msc://` stays on fsspec; the checkpoint *write* path (`_get_fs`) is deliberately unchanged.
- Adds `obstore>=0.6.0` dependency.
- New tests: `test/datapipes/healda/test_zarr_loader_obstore.py` (local baseline, remote-via-obstore with data verification, storage-options translation, read-only store) and four obstore download tests in `test/core/test_filesystem.py` against an in-memory store.

### Notes for reviewers

- Credentials now resolve through obstore's chain (AWS env vars, profiles, IMDS) instead of s3fs/botocore — equivalent for standard setups, flagging in case anyone relies on exotic botocore config.
- The benchmark figure is committed under `docs/img/` to embed here; happy to drop it from the tree if preferred.

## Checklist

- [x] The CHANGELOG.md is up to date with these changes.


docs/img/healda_obstore_benchmark.png
pzharrington, line 1: Is this used in docs somewhere?

pzharrington (approved):
HealDA changes look good to me, nice speedup

coreyjadams (approved):

PR #1931, opened 2026-08-17
feat(diffusion): Added Flow Matching

Dibyajyoti-Chakraborty:
## Description

Adds flow matching throughout `physicsnemo.diffusion` with `RectifiedFlowNoiseScheduler`, flow-prediction conversions on `LinearGaussianNoiseScheduler`, and weighted and unweighted `FlowMatchingLoss` APIs.
Extends existing DSM losses, schedulers, predictors, and sampling documentation to support `"flow"` alongside `"x0"`, `"score"`, and `"epsilon"`.
Adds `MultiDiffusionFlowMatchingLoss` and its weighted counterpart for patch-based flow-matching training on large spatial domains.
Clarifies that `MultiDiffusionModel2D` and `MultiDiffusionPredictor` support both patch-based diffusion and flow matching during training and inference.
Adds parameterized constructor, non-regression, gradient-flow, and compile coverage with reference data, API documentation, and CHANGELOG updates.

## Checklist

## Review Process


physicsnemo/diffusion/noise_schedulers/flow_matching.py
```diff
@@ -0,0 +1,459 @@
...
+            ) -> Float[Tensor, " B *dims"]:
+                x0 = velocity_to_x0(velocity_predictor(x, t), x, t)
+                return x0_to_score_v(x0, x, t)
+
+        def sde_denoiser(
+            x: Float[Tensor, " B *dims"],
+            t: Float[Tensor, " B"],
+        ) -> Float[Tensor, " B *dims"]:
+            t_bc = _bc(t, x.ndim)
+            return velocity_fn(x, t) - (t_bc / (1 - t_bc)) * score_fn(x, t)
```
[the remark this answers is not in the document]
Dibyajyoti-Chakraborty: Updated!

physicsnemo/diffusion/metrics/losses.py
```diff
@@ -644,3 +644,311 @@ def __call__(
...
+            case _:
+                raise ValueError(
+                    f"prediction_type must be 'velocity', 'x0', 'epsilon', "
+                    f"or 'score', got '{prediction_type}'."
+                )
+
+        # Define the reduction callbacks
+        _reductions = {
+            "none": lambda x: x,
+            "mean": lambda x: x.mean(),
```
laserkelvin, line 874: Do we not need to specify dimension, or let users specify it?

My main reason for this is potentially support things that are not on a regular grid, e.g. graphs.

Perhaps `reduction` could also be an arbitrary user-provided Callable?

physicsnemo/diffusion/noise_schedulers/rectified_flow.py
```diff
@@ -0,0 +1,467 @@
...
+    """
+
+    def __init__(
+        self,
+        t_min: float = 0.0,
+        t_max: float = 1.0,
+    ) -> None:
+        if not 0.0 <= t_min < t_max <= 1.0:
+            raise ValueError(
+                f"t_min and t_max must satisfy 0 <= t_min < t_max <= 1, "
```
laserkelvin, line 127: I don't know if it's better to actually do this check at the device and precision we will actually run with, to make sure that they aren't rounded incorrectly at bf16. I can't think of a case where this is actually going to be an issue, but just wanted to bring it up

laserkelvin (commented):
Just two minor general comments.

I think the broader one that I have, which affects the interface more is if we wanted to mask loss contributions before the reduction. For example in a chemistry case where we have a padded batch of atoms, we would want to apply a padding mask

megnvidia (approved):
LGTM

CharlelieLrt:
> Just two minor general comments.
> 
> I think the broader one that I have, which affects the interface more is if we wanted to mask loss contributions before the reduction. For example in a chemistry case where we have a padded batch of atoms, we would want to apply a padding mask

For parity with the denoising score matching loss, we should have a `WeightedFlowMatchingLoss`, which supports arbitrary weights (including a binary mask).

@laserkelvin would [this](https://github.com/NVIDIA/physicsnemo/blob/94b355108cfd10e253d3d41cacdc0c0112b078e4/physicsnemo/diffusion/metrics/losses.py#L449) work for the chemistry case you mentioned?

docs/api/diffusion/noise_schedulers.rst
```diff
@@ -111,6 +111,8 @@ available at three levels:
   - :class:`IDDPMNoiseScheduler` --- Improved DDPM schedule.
   - :class:`StudentTEDMNoiseScheduler` --- EDM variant with Student-t noise
     for heavy-tailed data.
+  - :class:`FlowMatchingNoiseScheduler` --- Rectified flow / conditional
```
CharlelieLrt: Maybe it should be called `RectifiedFlowNoiseScheduler` or `OptimalTransportNoiseScheduler` instead? There are multiple noise schedules that could be used with flow matching (e.g. TrigFlow)

physicsnemo/diffusion/metrics/losses.py
```diff
@@ -644,3 +644,311 @@ def __call__(
...
         loss = w * loss
         return self._reduce(loss)
+
+
+class FlowMatchingLoss:
+    r"""
+    Flow matching loss for training continuous normalizing flows or
+    diffusion models with a velocity objective.
+
+    Implements the conditional flow matching objective. Given clean data
```
CharlelieLrt: Just curious, why the "conditional"?

physicsnemo/diffusion/metrics/losses.py
```diff
@@ -644,3 +644,311 @@ def __call__(
...
+    :math:`\mathbf{x}_0` and a linear-Gaussian path
+    :math:`\mathbf{x}_t = \alpha(t)\mathbf{x}_0
+    + \sigma(t)\boldsymbol{\epsilon}`, the loss is:
```
CharlelieLrt: For me there is a logic problem there. The noise path is part of the noise schedule. The loss should not have to know anything about the noise path. It would be best if we could remove this constarint of linear-gaussian noise schedule from the loss itself.

Maybe I am thinking too far ahead, but this would prevent the use of this loss for noise schedules that are not of the form `x_t = alpha_t * x_0 + sigma_t * epsilon`.

@laserkelvin any thoughts about noise schedules typically used in chemistry and material science?

physicsnemo/diffusion/metrics/losses.py
```diff
@@ -644,3 +644,311 @@ def __call__(
...
+        ``prediction_type``. Must satisfy the
+        :class:`~physicsnemo.diffusion.DiffusionModel` protocol.
+    noise_scheduler : LinearGaussianNoiseScheduler
+        Noise scheduler defining the interpolation path. Must be a
+        :class:`~physicsnemo.diffusion.noise_schedulers.LinearGaussianNoiseScheduler`
+        subclass (or a
+        :class:`~physicsnemo.diffusion.noise_schedulers.DomainParallelNoiseScheduler`
+        wrapping one). Typically a
+        :class:`~physicsnemo.diffusion.noise_schedulers.FlowMatchingNoiseScheduler`.
+    prediction_type : {"velocity", "x0", "epsilon", "score"}, default="velocity"
```
CharlelieLrt: Two remarks:

1. `prediction_type` has to be a `PredictorType`. See there: https://github.com/NVIDIA/physicsnemo/blob/94b355108cfd10e253d3d41cacdc0c0112b078e4/physicsnemo/diffusion/base.py#L25
So adding a new predictor should be done there (and propagated everywhere different types of predictions are expected to work).
2. Which name is better? "velocity"/"v" or "flow"? "Velocity"/"v" brings confusion with the other common parameterization `v_t = sigma_t * x_0 - alpha_t * epsilon`.

physicsnemo/diffusion/metrics/losses.py
Dibyajyoti-Chakraborty: Ya this was the paper's idea. The thing is it does not optimize for the actual flow (which is intractable), but a conditional flow. The minima of one is same as the minima of the other, so it gets to the same point. https://arxiv.org/pdf/2210.02747 (Sec 3.2)

physicsnemo/diffusion/metrics/losses.py
Dibyajyoti-Chakraborty: I feel like this can be done by the user. Didn't wanna make things too complex here. Something like:
```
loss = flow_matching_loss_fn(x0, condition=cond, reduction="none") 
loss_final = (loss * padding_mask).sum() / padding_mask.sum() # some complex reduction
```

physicsnemo/diffusion/metrics/losses.py
CharlelieLrt: Ok, got it. Might be worth clarifying this a bit in the docs (or not use the term at all), because the term of conditional is already used extensively in the diffusion module, but it usually refer to conditional in the sense of supervised on observation ``y``

Dibyajyoti-Chakraborty: Maybe I'll remove the conditional term entirely. Noone, calls that in practice anyway .

physicsnemo/diffusion/metrics/losses.py
Dibyajyoti-Chakraborty: Good point. Will update.

docs/api/diffusion/noise_schedulers.rst
Dibyajyoti-Chakraborty: Will update.

physicsnemo/diffusion/metrics/losses.py
Dibyajyoti-Chakraborty: Will add a WeightedFlowMatchingLoss for this.

CharlelieLrt: @Dibyajyoti-Chakraborty is your above comment for this thread? I am not sure  how a `WeightedFlowMatchingLoss` would enable different types of reduction?

physicsnemo/diffusion/metrics/losses.py
Dibyajyoti-Chakraborty: Will update.

Dibyajyoti-Chakraborty (commented):
Updated done based on the comments:
- Named rectified flow matching: `RectifiedFlowNoiseScheduler`, (@CharlelieLrt)
- Name changes - `"flow"` (named to avoid clashing with v-prediction), `flow_predictor`/`x0_to_flow`/`flow_to_x0` support. (@CharlelieLrt)
- `PredictorType` (name changed @CharlelieLrt ) .
- Added `WeightedFlowMatchingLoss` (support masking etc. @laserkelvin)
- Simplified the loss to delegate to the scheduler's `add_noise`/`x0_to_flow` instead of duplicating the noise-path formula. (@CharlelieLrt )

physicsnemo/diffusion/noise_schedulers/rectified_flow.py
Dibyajyoti-Chakraborty: Crazy catch. This can be caused when someone has timesteps like 0.9999 (which will be cast to 1) - unlikely. Added a warning message for now. I am still thinking about better ways.

physicsnemo/diffusion/metrics/losses.py
Dibyajyoti-Chakraborty: Okay, I see what you mean. The irregular grid/ reduction along a dim is still an issue. I was only thinking about different reductions as weighted/masked etc. 

Shall I add the custom user-provided `reduction` function capability? Or just let the user do it by setting `reduction=None` and use their own reduction?

physicsnemo/diffusion/noise_schedulers/rectified_flow.py
laserkelvin: Actually it just needs to be 0.999, which is why I pointed it out. I agree that it's unlikely, but it's a corner case to be aware of

laserkelvin:
> > Just two minor general comments.
> > I think the broader one that I have, which affects the interface more is if we wanted to mask loss contributions before the reduction. For example in a chemistry case where we have a padded batch of atoms, we would want to apply a padding mask
> 
> For parity with the denoising score matching loss, we should have a `WeightedFlowMatchingLoss`, which supports arbitrary weights (including a binary mask).
> 
> @laserkelvin would [this](https://github.com/NVIDIA/physicsnemo/blob/94b355108cfd10e253d3d41cacdc0c0112b078e4/physicsnemo/diffusion/metrics/losses.py#L449) work for the chemistry case you mentioned?

Yeah `weight` argument would work for this, thanks!

physicsnemo/diffusion/metrics/losses.py
laserkelvin: I'm satisfied with @Dibyajyoti-Chakraborty 's solution; if we can document this as a pattern somewhere that would be good.

physicsnemo/diffusion/metrics/losses.py
laserkelvin: > noise schedules typically used in chemistry and material science?

Not right now, no - noise schedules are pretty universal so... I think most flow matching models are going to be using linear interpolation anyway. I don't know off the top of my head if any use weird curved paths, etc. and in either case probably out of scope for this PR IMO

CharlelieLrt: Cool, got it - I can think of FM on discrete spaces (https://arxiv.org/abs/2407.15595), which is a a little exotic and out of scope for this PR, but still good to keep in mind that there are these things out there.

physicsnemo/diffusion/metrics/losses.py
Dibyajyoti-Chakraborty: Done.

physicsnemo/diffusion/noise_schedulers/noise_schedulers.py
```diff
@@ -27,6 +27,7 @@
 from .edm_log_uniform import EDMLogUniformNoiseScheduler  # noqa: F401
 from .iddpm import IDDPMNoiseScheduler  # noqa: F401
 from .linear_gaussian import LinearGaussianNoiseScheduler  # noqa: F401
+from .rectified_flow import RectifiedFlowNoiseScheduler  # noqa: F401
```
CharlelieLrt: Remove from there. Those are only shims for back-compatibility import. This is a new noise scheduler , so no reason to have backward-compatible import

physicsnemo/diffusion/noise_schedulers/rectified_flow.py
```diff
@@ -0,0 +1,481 @@
...
+
+        :meth:`get_denoiser` is overridden with closed-form expressions for
+        the probability-flow ODE right-hand side (the flow field), which
+        avoids the :math:`\dot{\alpha}(t)/\alpha(t)` singularity of the
+        generic formulation at :math:`t = 1`. Sampling can therefore start
+        exactly at :math:`t = 1` (pure noise) with ``x0_predictor`` or
+        ``flow_predictor``. For ``epsilon_predictor``,
+        ``score_predictor``, or ``denoising_type="sde"``, the reverse process
+        is inherently singular at :math:`t = 1` and ``t_max`` must be set
+        strictly below 1 (e.g. ``t_max=0.999``).
```
CharlelieLrt: Overriding the `get_denoiser` is really not ideal. I understand there is a singularity, but would it be possible to do handle it either in the `alpha(t)` method (prefrered), or even the `drift` method? This would work too since the default `get_denoiser` calls `drift` to compute `alpha_dot/alpha`.

Overriding `drift` is in any case required, because it has the same singularity at t=1.

physicsnemo/diffusion/noise_schedulers/linear_gaussian.py
```diff
@@ -725,21 +725,150 @@ def x0_to_epsilon(
         sigma_t_bc = self.sigma(t_bc)
         return (x_t - alpha_t_bc * x0) / sigma_t_bc
 
+    def x0_to_flow(
```
CharlelieLrt, line 805: We need conversions method score to flow (and reverse) as well. The ones with epsilon are not really necessary

physicsnemo/diffusion/metrics/losses.py
```diff
@@ -644,3 +648,545 @@ def __call__(
         w = _maybe_promote_to_mesh(w, loss)
         loss = w * loss
         return self._reduce(loss)
+
+
+class FlowMatchingLoss:
+    r"""
```
CharlelieLrt, line 688: Generally speaking, these docs are way too verbose and disconnected from real users. It would be good to tighten up everything.

physicsnemo/diffusion/noise_schedulers/noise_schedulers.py
Dibyajyoti-Chakraborty: Done

physicsnemo/diffusion/noise_schedulers/rectified_flow.py
Dibyajyoti-Chakraborty: Done

physicsnemo/diffusion/metrics/losses.py
Dibyajyoti-Chakraborty: Done.

physicsnemo/diffusion/noise_schedulers/linear_gaussian.py
Dibyajyoti-Chakraborty: Done.

physicsnemo/diffusion/utils/model_wrappers.py
```diff
@@ -82,6 +82,12 @@ class ConcatConditionWrapper(Module):
     vector_cond_key : str, optional
         TensorDict key for the vector conditioning tensor to pass through to
         the backbone, by default ``"cond_vec"``.
+    time_scale : float, optional
+        Multiplier applied to ``t`` before the backbone's time embedding
+        receives it, by default ``1.0``. Set to e.g. ``999.0`` for flow
+        matching, whose ``[0, 1]``-valued times under-drive DDPM++/DiT-style
+        embedders built for ``[0, 1000]``. Part of the wrapped module, so it
+        applies identically at training and sampling time.
```
pzharrington: IMO this isn't the right place for this kwarg. The `ConcatConditionWrapper` should act purely as a helper util that pipes various conditioning into the model in the expected format.

pzharrington: I think it would be more clear, if an extra wrapper is needed to handle this time rescaling, to create a separate wrapper in this file rather than blur the functionality of the existing `ConcatConditionWrapper`; the separate wrapper could compose with this one to get both functionalities if needed

CharlelieLrt: I agree with @pzharrington . Also, this rescaling is dependent on what is the model being wrapped and the type of time embedding it is using. For one type of time embedding, maybe no rescaling is required, for another it might be with a rescaling factor of 1000, for another 10000, etc...
IMO it's too much model dependent, while the wrapper is kind of general purpose.
I think the responsibility of rescaling should for the caller. A note in the documentation can emphasize this caveat.

CharlelieLrt: > I think it would be more clear, if an extra wrapper is needed to handle this time rescaling, to create a separate wrapper in this file rather than blur the functionality of the existing ConcatConditionWrapper; the separate wrapper could compose with this one to get both functionalities if needed

This makes me think, @Dibyajyoti-Chakraborty would it make sense to consider this additional wrapper as a preconditioner (maybe `FlowMatchingPreconditioner`?). That's a little different in purpose from the other EDM-like preconditioners, but maybe that's acceptable?

Dibyajyoti-Chakraborty: I'm not sure about the `FlowMatchingPreconditioner`. It is a bit of an overkill for the time scaling. I think it is best to remove it and just put a node in the documentation.

Dibyajyoti-Chakraborty: Done.

test/diffusion/test_noise_schedulers.py
CharlelieLrt, line 71: The right pattern for the new rectified flow noise scheduler should not be to add its own `test_rectified_flow.py`. Instead it should just be added to the `SCHEDULER_CONFIGS` here, with all the other scheduler and it should go through the same test suite with minimal changes to the tests themselves.

CharlelieLrt, line 83: The rectified flow should have its own constructor test class here though (these constructor tests cannot be handled cleanly by the pytest parameterization)

test/diffusion/test_noise_schedulers.py
```diff
@@ -272,6 +272,76 @@ def test_concrete_x0_to_score_to_x0(self, device):
...
+
+    def test_flow_to_score_matches_x0_composition(self, device):
+        """flow_to_score equals the flow_to_x0 -> x0_to_score composition."""
+        s = _MinimalScheduler()
+        v = make_input((2, 4), seed=29, device=device)
+        x_t = make_input((2, 4), seed=30, device=device)
+        t = torch.tensor([0.5, 0.5], device=device)
+        direct = s.flow_to_score(v, x_t, t)
+        composed = s.x0_to_score(s.flow_to_x0(v, x_t, t), x_t, t)
+        torch.testing.assert_close(direct, composed)
```
CharlelieLrt: This is a large number of repeated test functions. I'm not sure if it's possible, but if you find a way to instead use pytest parameterization to avoid code duplication, then that would be great. I am thinking something like a `test_source_prediction_to_target_to_source` and parameterize over the `source` prediction and `target` prediction. Ideally, we should have one single test method for all these roundtrip conversions.

test/diffusion/test_noise_schedulers.py
```diff
@@ -272,6 +272,76 @@ def test_concrete_x0_to_score_to_x0(self, device):
...
+
+        def flow_pred(x, t):
+            x0 = x0_pred(x, t)
+            return s.x0_to_flow(x0, x, t)
+
+        denoiser_x0 = s.get_denoiser(x0_predictor=x0_pred)
+        denoiser_v = s.get_denoiser(flow_predictor=flow_pred)
+        torch.testing.assert_close(denoiser_x0(x, t), denoiser_v(x, t))
+
+    def test_get_denoiser_validates_flow_predictor(self, device):
```
CharlelieLrt: Similar comment for these two new tests.

1. For `test_get_denoiser_with_flow_predictor`, it would be better to add pytest parameterization to the existing `test_concrete_denoiser` rather than duplicating the test for each type of prediction.
2. For `test_get_denoiser_validates_flow_predictor` it seems these types of validation tests were totally missing from the existing tests. Would be great if you could add 1 validation tests with the same parameterization pattern as for point 1 above.

test/diffusion/test_rectified_flow.py
```diff
@@ -0,0 +1,349 @@
...
+            warnings.simplefilter("error")
+            RectifiedFlowNoiseScheduler()
+
+
+# =============================================================================
+# Analytic Coefficient Tests
+# =============================================================================
+
+
+class TestRectifiedFlowCoefficients:
```
CharlelieLrt: This test class should not exist. The pattern for required tests is the same for all type of schedulers:
- a class for testing the constructor (test values of public attributes after constructor call with both default and user-provided attributes)
- A class for non-regression tests of all public scalar methods against golden files (scalar methods = methods that operates on time only and not the state x)
- A similar class for non-regression tests against golden files for spatial metjods (operates on state x)
- A class for compilation tests: has to full graph compilation
 
All other tests are not required and should be discarded unless really necessary.

All of this being said, you should not have to implement any test, and adding the new scheduler in the `SCHEDULER_CONFIGS` of `test_noise_schedulers.py` should be sufficient to cover all of this.

physicsnemo/diffusion/base.py
CharlelieLrt, line 1: Enabling support for flow prediction should be done on the entire module surface API (to avoid having some parts supporting flow prediction and some that do not). There are still places where it is missing, which is not ideal. For example in `multi_diffusion/predictor.py`, the class `MultiDiffusionPredictor` should support flow prediction through a `flow_to_x0_fn` callback. And I think there might be some other places where support for flow prediction with these user-provided callbacks is missing. It would be good to double-check everywhere.

docs/api/diffusion/metrics.rst
```diff
@@ -43,6 +49,25 @@ regions or channels (for example, land versus ocean in weather applications).
...
+per-element loss and apply your own reduction: a masked mean,
+``torch_scatter.scatter_mean`` keyed by a graph index, and so on.
+
+.. code-block:: python
+
+    loss_fn = MSEDSMLoss(model, scheduler, reduction="none")
+    loss = loss_fn(x0)                      # per-element, same shape as x0
+    masked_loss = (loss * mask).sum() / mask.sum()
+
+
```
CharlelieLrt: This should not exist there. Docstrings are for such detailed description of parameters variations. These APIs docs are fore higher-level explanations and examples that bind everything together.

physicsnemo/diffusion/metrics/losses.py
```diff
@@ -644,3 +648,471 @@ def __call__(
...
+        mismatched scale silently shifts the embedding out of its trained
+        range.
+
+    Parameters
+    ----------
+    model : DiffusionModel
+        Diffusion model to train. ``prediction_type`` sets how the loss
+        interprets the output. Must follow the
+        :class:`~physicsnemo.diffusion.DiffusionModel` protocol.
+    noise_scheduler : LinearGaussianNoiseScheduler
```
CharlelieLrt: There are some inconsistencies there: the docstring says `noise_scheduler` is `LinearGaussianNoiseScheduler`, while the type hint is `NoiseScheduler`. I think the latter is correct

physicsnemo/diffusion/metrics/losses.py
```diff
@@ -644,3 +648,471 @@ def __call__(
...
+        Noise scheduler defining the interpolation path. Must be a
+        :class:`~physicsnemo.diffusion.noise_schedulers.LinearGaussianNoiseScheduler`
```
CharlelieLrt: I think this hasn't been updated following my previous comments. The loss should not know anything about the type of noise schedule, as long as it implements the required `NoiseScheduler` protocol

physicsnemo/diffusion/metrics/losses.py
```diff
@@ -644,3 +648,471 @@ def __call__(
...
+        # DomainParallelNoiseScheduler, the conversion methods live on the
+        # wrapped inner scheduler.
+        coeff = getattr(noise_scheduler, "inner_scheduler", noise_scheduler)
+        if not callable(getattr(coeff, "x0_to_flow", None)):
+            raise ValueError(
+                "FlowMatchingLoss requires a noise scheduler (or a "
+                "DomainParallelNoiseScheduler wrapping one) that exposes "
+                f"x0_to_flow; {type(coeff).__name__} does not. "
+                "LinearGaussianNoiseScheduler subclasses provide this."
+            )
```
CharlelieLrt: IMO that's not a good pattern because it deviates too much from the existing `MSEDSMLoss`. Nothing wrong in the implementation in itself, but it would be preferrable to have API parity with the existing loss.
For example, the [optional] conversion methods should be provided as callbacks rather than inferred from the scheduler object. This can be useful when one needs to change just the prediction type without defining a new scheduler class. So, there is redundancy, but it improves API flexibility

CharlelieLrt: One detail though: the denoising score matching losses expose conversion to x0 (score-to-x0 and noise-to-x0). That is because the loss formulation is based on x0, and any conversion is just an adaptation of this. IMO, the basis prediction for the FlowMatching loss should be the flow prediction, so it should expose callbacks to convert to x0 (x0-to-flow, epsilon-to-flow, score-to-flow)

physicsnemo/diffusion/metrics/losses.py
```diff
@@ -199,7 +199,9 @@ class MSEDSMLoss:
     reduction : Literal["none", "mean", "sum"], default="mean"
         Reduction to apply to the output: ``"none"`` returns the
         per-element loss, ``"mean"`` returns the mean over all elements,
-        ``"sum"`` returns the sum over all elements.
+        ``"sum"`` returns the sum over all elements. See
+        :ref:`diffusion_metrics_reductions` for irregular/graph-structured
+        data.
```
CharlelieLrt: Remove this. I doubt we can reference API docs from python docstrings

physicsnemo/diffusion/metrics/losses.py
```diff
@@ -199,7 +199,9 @@ class MSEDSMLoss:
     reduction : Literal["none", "mean", "sum"], default="mean"
         Reduction to apply to the output: ``"none"`` returns the
         per-element loss, ``"mean"`` returns the mean over all elements,
```
CharlelieLrt, line 208: This goes back to my comment about adding support for flow-prediction over the entire API surface. Here, we have `prediction_type` that _**could**_ be `flow`. However, there is no exposed callback `flow_to_x0` that would be necessary to support a flow-prediction here. I think it should be added there and also in a few other places where it might be missing. Tests parameterization for the losses need to be adapted accordingly to cover this.

physicsnemo/diffusion/metrics/losses.py
```diff
@@ -644,3 +648,471 @@ def __call__(
...
+
+    **Example 2:** Training an x0-predictor with the flow matching loss.
+    The loss converts the prediction internally to a flow estimate; set
+    ``t_min`` above 0 to avoid the conversion singularity:
+
+    >>> scheduler = RectifiedFlowNoiseScheduler(t_min=1e-3)
+    >>> loss_fn = FlowMatchingLoss(model, scheduler, prediction_type="x0")
+    >>> loss = loss_fn(x0)
+    >>> loss.shape
+    torch.Size([])
```
CharlelieLrt: Remove this, the examples should be focusing on how to use the loss. There shouldn't be anything about inference here

physicsnemo/diffusion/metrics/losses.py
```diff
@@ -644,3 +648,471 @@ def __call__(
...
+        If ``prediction_type`` is not ``"flow"``, ``"x0"``,
+        ``"epsilon"``, or ``"score"``.
+    ValueError
+        If ``noise_scheduler`` does not expose ``x0_to_flow`` (and
+        ``epsilon_to_x0``/``score_to_flow`` when required by
+        ``prediction_type``).
+
+    Examples
+    --------
+    **Example 1:** Standard flow-predictor training with the flow
```
CharlelieLrt: IMO a single example would be sufficient in this docstring. The docstring for the `MSEDSMLoss` already has multi examples that show how to train a conditional vs unconditional model and also how to switch the type of predictor being trained. I don't think we need to repeat all of this here. We can refer to the `MSEDSMLoss` for more advanced usage patterns.

test/diffusion/test_losses.py
```diff
@@ -637,3 +644,507 @@ def test_non_regression(
...
+# docstrings).
+FM_SCHEDULER_KWARGS = {
+    "flow": {},
+    "x0": {"t_min": 1e-3},
+    "epsilon": {},
+    "score": {},
+}
+
+
+def _make_flow_matching_loss(model, scheduler, prediction_type):
```
CharlelieLrt: There is so much code duplication with these helpers. Would it be not possible to integrate/unify them with the existing helper for the diffusion losses instead?

test/diffusion/test_losses.py
CharlelieLrt, line 1: The new tests should be reorganized for clarity:
1. All pytest configs at the top
2. Constructor tests together
3. Non regression tests for the non-weighted losses (both DSM and flow mathcing)
4. Same for the weighted version

test/diffusion/test_losses.py
```diff
@@ -637,3 +644,507 @@ def test_non_regression(
...
+# =============================================================================
+
+
+class TestFlowMatchingLossGradientFlow:
+    """Tests that gradients flow through FlowMatchingLoss."""
+
+    @pytest.mark.parametrize(
+        "prediction_type", FM_PREDICTION_TYPES, ids=FM_PREDICTION_TYPES
+    )
+    def test_gradient_flow(self, device, prediction_type):
```
CharlelieLrt: The weighted version is not tested for gradient flow?

test/diffusion/test_losses.py
```diff
@@ -637,3 +644,507 @@ def test_non_regression(
             compare_outputs(losses[1], ref["loss_1"], **tolerances)
             compare_outputs(params[0], ref["param_0"], **tolerances)
             compare_outputs(params[1], ref["param_1"], **tolerances)
+
+
+# =============================================================================
+# FlowMatchingLoss — Constants and Helpers
+# =============================================================================
+
+FM_PREDICTION_TYPES = ["flow", "x0", "epsilon", "score"]
```
CharlelieLrt: Just add "flow" to the existing PREDICTION_TYPES. Why creating an almost identical `FM_PREDICTION_TYPES`

CharlelieLrt (changes requested):
Two major concerns:
- The PR claims to add support for "flow" prediction, but this support is only very partial with missing conversion callbacks at multiple place in the module
- New tests are a not great, with massive amounts of code duplication, new class-specific configs and tests, while it would be much better to extend the existing tests, etc...

test/diffusion/test_noise_schedulers.py
Dibyajyoti-Chakraborty: Done. Deleted `test_rectified_flow.py` and added `RectifiedFlowNoiseScheduler` to `SCHEDULER_CONFIGS` (default and custom `t_min`/`t_max` entries) plus `COMPILE_SCHEDULER_CONFIGS`, so it runs through the shared constructor/scalar/spatial/compile suites with no changes to the tests themselves.

Dibyajyoti-Chakraborty: Done. Added `TestRectifiedFlowNoiseSchedulerConstructor` with default/custom attribute checks and the `NoiseScheduler`/`LinearGaussianNoiseScheduler` isinstance checks.

test/diffusion/test_noise_schedulers.py
Dibyajyoti-Chakraborty: Done. Collapsed into a single `test_conversion_source_to_target_to_source(source, target)` parametrized over all six pairs (x0↔score, x0↔epsilon, x0↔flow, score↔flow), also absorbing the pre-existing x0→score→x0 test.

test/diffusion/test_noise_schedulers.py
Dibyajyoti-Chakraborty: Removed both flow-specific tests. `test_concrete_get_denoiser` is now parametrized over `["x0", "score", "epsilon", "flow"]` and asserts each predictor type reproduces the x0-predictor RHS. 
Added `test_get_denoiser_validates_predictor_count` with the same parametrization.

test/diffusion/test_rectified_flow.py
Dibyajyoti-Chakraborty: The file is deleted; no RF-specific tests remain beyond the constructor class. RF coverage now comes entirely from the `SCHEDULER_CONFIGS` entries in `test_noise_schedulers.py`.

physicsnemo/diffusion/base.py
Dibyajyoti-Chakraborty: Added the `flow_to_x0_fn` for "flow" prediction type case to every prediction-type dispatch: `MultiDiffusionPredictor`, `MSEDSMLoss`, `WeightedMSEDSMLoss`, `MultiDiffusionMSEDSMLoss`, and `MultiDiffusionWeightedMSEDSMLoss`. The rest of the places, like DPS guidance, have all been checked.

docs/api/diffusion/metrics.rst
Dibyajyoti-Chakraborty: Removed these from the API docs.

physicsnemo/diffusion/metrics/losses.py
Dibyajyoti-Chakraborty: Done. The docstrings now say `NoiseScheduler`, matching the type.

physicsnemo/diffusion/metrics/losses.py
Dibyajyoti-Chakraborty: Done, both points. The `noise_scheduler` parameter is documented as the `NoiseScheduler` protocol only, and the implementation no longer inspects the scheduler type.

 The flow matching losses now expose `x0_to_flow_fn`, `score_to_flow_fn`, and `epsilon_to_flow_fn` callbacks with the same pattern as `MSEDSMLoss`. Flow is the base prediction: every prediction type converts *to* flow. One note: `x0_to_flow_fn` is required for all prediction types because it also computes the regression target from clean data.

physicsnemo/diffusion/metrics/losses.py
Dibyajyoti-Chakraborty: Removed; the docstrings state the irregular-data guidance inline.

physicsnemo/diffusion/metrics/losses.py
Dibyajyoti-Chakraborty: Done. `MSEDSMLoss`/`WeightedMSEDSMLoss` (and the multi-diffusion losses and predictor) are added with `flow_to_x0_fn`.

physicsnemo/diffusion/metrics/losses.py
Dibyajyoti-Chakraborty: Removed the sampling example; the docstring shows training only.

test/diffusion/test_losses.py
Dibyajyoti-Chakraborty: Done. All FM-specific helpers are removed: a single `_make_loss(loss_cls, model, scheduler, prediction_type)` builds all four losses, and a single `_run_training_loop(..., weight=None)` plus a shared `_check_non_regression` serve both the DSM and flow matching tests.

physicsnemo/diffusion/metrics/losses.py
Dibyajyoti-Chakraborty: Done. One flow-predictor training example remains, ending with a pointer to `MSEDSMLoss` for conditional training, prediction-type switching, etc..

test/diffusion/test_losses.py
Dibyajyoti-Chakraborty: Done. The file now follows this layout, and the FM losses run through the *same* classes as the DSM losses via `LOSS_CONFIGS`/`WEIGHTED_LOSS_CONFIGS` rows `(loss_cls, sched_cls, kwargs, name)` — analogous to `SCHEDULER_CONFIGS`. Same for the multi-diffusion loss/predictor test files.

test/diffusion/test_losses.py
Dibyajyoti-Chakraborty: Added `test_weighted_gradient_flow` parametrized over all weighted loss configs (including `WeightedFlowMatchingLoss`) and all four prediction types.

test/diffusion/test_losses.py
Dibyajyoti-Chakraborty: Done. `FM_PREDICTION_TYPES` is removed; a single `PREDICTION_TYPES = ["x0", "score", "epsilon", "flow"]` drives every parametrization.

physicsnemo/diffusion/metrics/losses.py
Dibyajyoti-Chakraborty: Done. The docstrings now say `NoiseScheduler`, matching the type hints, in all four loss classes. The loss no longer assumes anything beyond the `NoiseScheduler` protocol. `RectifiedFlowNoiseScheduler` is only mentioned there as the typical choice.

CharlelieLrt (approved):
I have committed my own additions on top of this PR (patch-based support for flow matching, docs correction and improvements, extended test coverage for flow matching, etc.)

Everything looks good to me now.
Thanks @Dibyajyoti-Chakraborty for the nice addition to the diffusion module!

CharlelieLrt:
/ok to test [ca19e52](https://github.com/NVIDIA/physicsnemo/pull/1931/commits/ca19e522ba95da9738b0eba4f5e6ea620951aa59)

laserkelvin (approved):

PR #1932, opened 2026-08-18
Add exponential Euler, stochastic exponential Euler, DPM-Solver++(2M), and UniC-2 diffusion samplers

CharlelieLrt:
# PhysicsNeMo Pull Request

## Description

This PR adds four composable diffusion solvers to `physicsnemo.diffusion.samplers`. Each solver can be instantiated directly or selected by passing its string key to `sample()`.

### Solvers

- **`ExponentialEulerSolver`** (`"exponential_euler"`) is a first-order solver for semi-linear ODEs. It supports DDIM-like sampling and distilled few-step models.
- **`EDMStochasticExponentialEulerSolver`** (`"edm_stochastic_exponential_euler"`) adds EDM-style churn and controllable re-noising to exponential Euler. It supports stochastic, ancestral, and full re-noising configurations for distilled and consistency models.
- **`DPMPlusPlus2M`** (`"dpmpp_2m"`) is a configurable second-order multistep solver with one denoiser evaluation per step. Its defaults recover classical two-step Adams-Bashforth; configuring it with the schedule's semi-linear structure recovers the published DPM-Solver++(2M) formulation.
- **`DPMPlusPlus2MUniC2`** (`"dpmpp_2m_unic2"`) adds a UniC-2 corrector stage on top of `DPMPlusPlus2M`, raising the update to third order while keeping one denoiser evaluation per step, at the cost of a slightly larger memory footprint.

### API integration

- Adds `LinearGaussianNoiseScheduler.get_linear_denoiser()` to expose the semi-linear structure of the ODE/SDE right-hand side that the new exponential and multistep solvers rely on. It complements `get_denoiser()`, which returns the full right-hand side.
- Adds `LinearGaussianNoiseScheduler.snr()` to expose the schedule's signal-to-noise ratio, used as the multistep extrapolation coordinate by `DPMPlusPlus2M` and `DPMPlusPlus2MUniC2`.
- Extends `sample()` with string-key selection and solver-specific options for all four solvers.
- Documents the new solvers, their intended use cases, and their integration with linear-Gaussian noise schedules.

### Relationship to #1896

This PR supersedes [#1896](https://github.com/NVIDIA/physicsnemo/pull/1896). It builds on the DPM-Solver++(2M) implementation from that PR through a leaner and more compositional API. See #1896 for quantitative evidence of the sampler's benefits.

## Checklist

## Dependencies

None.

## Review Process


gertln:
Thanks for integrating the DPM++2M work into this PR.

I tested the public `sample(..., solver="dpmpp_2m")` path at commit `7f40b6a`.
Its implementation changed substantially in commit `35c7a1d8`, so I want to
confirm which numerical method the `"dpmpp_2m"` key is intended to run.

`87ae4004` introduced `DPMPlusPlus2M` using `λ = log(α/σ)` as its integration
variable. It used `alpha_fn` and `sigma_fn`, and handled the zero-noise endpoint
separately because λ diverges there. `35c7a1d8` replaced that implementation
with one that advances directly in the scheduler's time variable and accepts
one optional `linear_fn`. The file no longer references the log-SNR clock.

**The default path.** The class docstring says that omitting `linear_fn` gives
the "classical two-step Adams-Bashforth method", so that class behavior appears
intentional. What I could not tell is whether the same is intended for the
`"dpmpp_2m"` string key passed to `sample()`. Because `linear_fn` is optional,
`sample(..., solver="dpmpp_2m")` with no `solver_options` constructs the solver
with `linear_fn=None` and therefore selects that method.
I confirmed that its output matches an independent variable-step AB2
implementation to at most 4.4e-16, both with explicit Karras time steps and
with scheduler-generated ones. However, the `sample()` docstring and
`docs/api/diffusion/samplers.rst` describe the key as DPM-Solver++(2M) without
mentioning `linear_fn`.

On the analytic EDM problem in the attached reproduction, the public path has
relative error 2.39e-1 compared with 1.57e-1 for DPM++2M at 8 NFE, and 7.67e-3
compared with 3.69e-3 at 64 NFE. Both methods are second order, but their errors
at the same step count differ.

fig_pr1932_single

**Standard DPM++2M is still available**, which I want to be clear about. If I
pass the λ ladder as `time_steps`, use a denoiser that accepts λ as its time
input, and set `solver_options={"linear_fn": lambda t: -ones}`, then `sample()`
matches a direct implementation of the paper to at most 4.5e-15.

This recipe is not documented, and its time steps do not follow the documented
ordering. The λ values increase, while `time_steps` are documented as
decreasing. The reproduction stops at `σ_min`, so its final λ is finite.
Reaching zero noise needs a final `λ = +inf` rather than the documented
`t_0 = 0`. That is accepted, and looks correct — the `isfinite` guard drops
the step to first order and it returns the data prediction exactly — but is
passing `+inf` the intended way to take the terminal step under this recipe?

**The scheduler-supplied `linear_fn`.** The class `Examples` block uses
`linear_fn = get_linear_denoiser(prediction_type="x0")`. For EDM, this gives
`A(t) = 1/t`. The solver holds this value fixed over each step, so
`exp(h·A(t_cur))` is only an approximation to the linear update. The example
also steps to `t = 0.0`, where `z = −1`. Its final step therefore returns
`e⁻¹·x + (1 − e⁻¹)·D` instead of the data prediction `D`. Running the example
with `x0_pred = lambda x, t: x * 0.1` gives `x_0 = 0.566086`, while the data
prediction is `0.129176`. On a ladder ending at zero the error also falls far
more slowly than second order: 5.5e0 at 8 steps, 9.6e-1 at 128, and still
3.6e-1 at 2048.

This is also the configuration used by the solver test matrix
(`test_solvers.py:117`). For score or epsilon prediction on EDM, `A = 0`, so
this issue is specific to `x0` prediction.

**Test coverage.** As a check, I replaced the DPM++ multistep coefficient
`h_phi1/2 · r` with the plain AB2 coefficient `h/2 · r`. All 300 collected
`dpmpp_2m` tests still passed. The `dpmpp_2m` reference files are not in the
tree, so `load_or_create_reference` generates them on the first run rather than
comparing against a stored baseline.

Could you clarify:

1. Was the switch to the scheduler's time variable in `35c7a1d8` intended to
   change what `dpmpp_2m` selects, or was the λ-based update meant to remain?
2. If callers are expected to provide the λ transformation themselves, would
   it be worth documenting the recipe and explaining its different time-step
   ordering?
3. Is the `A(t) = 1/t` pairing shown in `Examples` intended for ladders that
   terminate at zero noise, given `timesteps()` always appends `0`?

A short CPU reproduction pinned to `7f40b6a` is below. It uses `sample()` and
`solver_options` throughout, although the λ ladder ascends rather than
decreases, unlike the documented `time_steps` behavior.

[repro_pr1932_public.py](https://github.com/user-attachments/files/31216619/repro_pr1932_public.py)

CharlelieLrt:
@gertln 

I reintroduced the extrapolation variable `lambda` as an additional callback passed to the solver. To reproduce DPM-Solver++(2M), lambda needs to be set to the log-SNR of the associated noise schedule. This is shown in the docstring example.

The time steps grid does not need to be in lambda coordinates. The solver always operates in (x, t) coordinates. The lambda conversion is purely internal (only used to approximate the moments/integrals).

CharlelieLrt:
@megnvidia Could I get your review on this PR for docs modifications? Thanks!

gertln:
Thanks @CharlelieLrt, `lambda_fn` makes the multistep extrapolation use the log-SNR coordinate. I believe the documented native-time config still differs from DMP++2M in the linear update.
The first step is the simplest check because there is no history, so `lambda_fn `does not affect the result. The current code propagates the linear term with `exp((t_next - t_cur) * A(t_cur))`, while DPM++2M uses the linear coefficient integrated over the interval. For EDM that exact multiplier is `sigma_next/sigma_cur`.
For the first 5.0 -> 2.5 step in the documented `x0` config with `x=2` and `x0_pred=0.1*x` we have:

current solver multiplier: `exp(-0.5) = 0.6065`
DPM++2M multiplier: `2.5/5.0 = 0.5`

current solver output: `1.291755`
DPM++2M output: `1.1`

To be clear the kernel can express canonical DPM2++2M but what I believe we can not reproduce is DPM++2M using the documented `x0` config on the scheduler's native time grid, where `A(t)` varies. Is this interpretation correct or am I missing something?

CharlelieLrt:
@gertln this AI slop is really not helpful. Can you please actually look at the code and let me know where you think there's an error?

physicsnemo/diffusion/noise_schedulers/linear_gaussian.py
```diff
@@ -761,6 +763,10 @@ def get_denoiser(
         stochastic term :math:`g(t) d\mathbf{W}` is handled by the solver,
         not returned by the denoiser itself.
 
+        Exponential integrators also need the coefficient of the linear part
+        of this right-hand side. Build that callback with
```
laserkelvin, line 788: I think "this right-hand side" is ambiguous, maybe just use the exact symbol

physicsnemo/diffusion/samplers/exponential_euler.py
```diff
@@ -0,0 +1,182 @@
...
+import torch
+from jaxtyping import Float
+from torch import Tensor
+
+from physicsnemo.diffusion.base import Denoiser
+
+from .base import Solver
+
+
+class ExponentialEulerSolver(Solver):
```
laserkelvin, line 70: Could you confirm that this MWE is a regression, and if so, address and add a unit test to cover it?

My understanding is that `t=0`, using `ExponentialEulerSolver` ends up having residual noise as opposed to the purely denoised expectation

```python
import torch

from physicsnemo.diffusion.samplers import (
    EulerSolver,
    ExponentialEulerSolver,
)

dtype = torch.float64

# Perfect clean-data prediction q=0.
q = torch.tensor([[0.0]], dtype=dtype)

# Valid EDM state: x(t) = q + t*epsilon, with t=1 and epsilon=1.
x = torch.tensor([[1.0]], dtype=dtype)
t_cur = torch.tensor([1.0], dtype=dtype)
t_next = torch.tensor([0.0], dtype=dtype)


def denoiser(x, t):
    """Exact EDM probability-flow ODE for the perfect prediction q."""
    t_bc = t.reshape(-1, 1)
    return (x - q) / t_bc


def linear_fn(t):
    """EDM x0 linear split returned by get_linear_denoiser()."""
    return 1.0 / t


exp_euler = ExponentialEulerSolver(denoiser, linear_fn=linear_fn)
euler = EulerSolver(denoiser)

actual = exp_euler.step(x, t_cur, t_next)
explicit_euler = euler.step(x, t_cur, t_next)
expected = q

# Literal closed form of the implementation.
implemented_formula = q + torch.exp(torch.tensor(-1.0, dtype=dtype)) * (x - q)

print("Expected endpoint:       ", expected.item())
print("Exponential Euler:       ", actual.item())
print("Implemented closed form: ", implemented_formula.item())
print("Explicit Euler:          ", explicit_euler.item())

# Confirms that the MWE faithfully represents the implementation.
torch.testing.assert_close(actual, implemented_formula)

# Even ordinary Euler reaches the exact endpoint for this example.
torch.testing.assert_close(explicit_euler, expected)

# This assertion intentionally fails.
torch.testing.assert_close(actual, expected)
```

megnvidia (approved):
LGTM

laserkelvin (changes requested):
I think more generally, adding regression tests that check the solver value at t=0, t=0.5, and t=1 against simple cases would be good for these three solvers (ideally all solvers). If you can resolve that and make sure all tests pass, it'll give me some more time to review as well

physicsnemo/diffusion/noise_schedulers/linear_gaussian.py
CharlelieLrt: Rewrote the docstring, now clearly uses the variable names.

physicsnemo/diffusion/samplers/exponential_euler.py
CharlelieLrt: I verified all math and there was a error in the solver implementation, that should now be fixed. It's still a good suggestion to add this type of consistency test in addition of the non-regression ones.

CharlelieLrt: Consistency test added.

physicsnemo/diffusion/samplers/samplers.py
```diff
@@ -360,7 +413,18 @@ def denoiser(
...
+        options = dict(solver_options)
+        required_args = {}
+        for name in _REQUIRED_SOLVER_ARGS[solver]:
+            if name not in options:
+                raise ValueError(
+                    f"Missing required solver option '{name}' for solver "
+                    f"'{solver_cls.__name__}'."
+                )
+            required_args[name] = options.pop(name)
+        solver_ = solver_cls(denoiser, **required_args, **options)
```
laserkelvin, line 427: Here the trap live. When a user write `sample(..., solver="dpmpp_2m")` with no `solver_options`, this line build the solver with empty pockets — no `bias_fn`/`bias_int_fn`/`slope_fn`/`lambda_fn`. Defaults then decide what actually run:

- `"exponential_euler"` -> plain explicit Euler (same as `"euler"`, plus quadrature overhead)
- `"edm_stochastic_exponential_euler"` -> explicit Euler too (defaults S_churn=0, renoise=0), plus one wasted RNG draw per step
- `"dpmpp_2m"` -> classical Adams-Bashforth-2 in t — not DPM-Solver++(2M)
- `"dpmpp_2m_unic2"` -> AB2 predictor + UniC-2 corrector in t coordinates

Nothing crash, samples come out — but from a different, weaker method than the key name promise, and neither `sample()`'s docstring nor `docs/api/diffusion/samplers.rst` say the callbacks are required. Me know the defaults are an intentional feature (class docstrings showcase the AB2 config, and the `dpmpp_2m_ab2` test pins it) — so me NOT ask to remove them or to raise. Ask instead:

1. `warnings.warn` when one of these four keys is used with no callback options — key-specific text saying what will actually run and what to pass;
2. per-key honesty in the `sample()` docstring + RST, e.g. "with default options this reduces to explicit Euler / AB2; pass the scheduler's `get_linear_denoiser(...)` callables (and `lambda_fn = lambda t: torch.log(scheduler.snr(t))` for the DPM keys) via `solver_options` to enable the named method."

Auto-wiring is not safe (`sample()` cannot introspect the denoiser's `prediction_type`), so docs + warning is the honest fix.

test/diffusion/test_solvers.py
```diff
@@ -274,26 +591,108 @@ def test_zero_churn_matches_deterministic(
...
+    "solver_cls,solver_kwargs,solver_name,uses_rng,time_scale",
+    SOLVER_CONFIGS,
+    ids=[c[2] for c in SOLVER_CONFIGS],
+)
+@pytest.mark.parametrize(
+    "spatial_name,shape,predictor_cls,predictor_kwargs",
+    SPATIAL_CONFIGS,
+    ids=[c[0] for c in SPATIAL_CONFIGS],
+)
+class TestStepConsistency:
```
laserkelvin, line 623: This oracle is elegant and me verify it is exact for every config — but it is blind in one eye. With `slope=0` the nonlinear term is N≡0, so the J quadrature, the DPM++2M extrapolation, and the UniC-2 corrector weights all get multiplied by zero — any bug in them passes 100%. The single-step design also never reaches UniC-2 steady state (corrector never runs). So the "second-order"/"third-order" claims today rest only on self-generated goldens — which catch *changes*, not *wrong-from-the-start*.

Suggest one golden-free order-of-accuracy test. Mock-up (CPU, no scheduler, no goldens):

```python
import numpy as np
import torch
import pytest
from physicsnemo.diffusion.samplers import (
    DPMPlusPlus2M, DPMPlusPlus2MUniC2, ExponentialEulerSolver,
)

# dx/dt = -x + cos(t): semi-linear with a=-1, A=-t, b=1, N=cos(t) (nonzero!).
# Exact: x(t) = e^{-(t-t0)} x0 + e^{-t} (F(t) - F(t0)),  F(t) = e^t (sin t + cos t)/2
def exact(t, t0=1.0, x0=1.0):
    F = lambda u: torch.exp(u) * (torch.sin(u) + torch.cos(u)) / 2
    return torch.exp(-(t - t0)) * x0 + torch.exp(-t) * (F(t) - F(t0))

@pytest.mark.parametrize("solver_cls,order", [
    (ExponentialEulerSolver, 1.0),
    (DPMPlusPlus2M, 2.0),
    (DPMPlusPlus2MUniC2, 3.0),
])
def test_empirical_order(solver_cls, order):
    denoiser = lambda x, t: -x + torch.cos(t).reshape(-1, *([1] * (x.ndim - 1)))
    errs, hs = [], []
    for n in (8, 16, 32, 64):
        solver = solver_cls(  # fresh instance per grid: multistep solvers are stateful
            denoiser,
            bias_fn=lambda t: -torch.ones_like(t),
            bias_int_fn=lambda t: -t,
            slope_fn=lambda t: torch.ones_like(t),
        )
        x = torch.ones(1, 1)
        ts = torch.linspace(1.0, 0.5, n + 1)
        for i in range(n):
            x = solver.step(x, ts[i].expand(1), ts[i + 1].expand(1))
        errs.append((x - exact(ts[-1])).abs().max().item())
        hs.append(0.5 / n)
    p = np.polyfit(np.log(hs), np.log(errs), 1)[0]
    assert p > order - 0.2, f"measured order {p:.2f}, expected ~{order}"
```

This measure the actual convergence rate — a wrong corrector weight or wrong J shows up immediately as a wrong slope. Cheap, fast, and directly proves the headline claims of the PR.

physicsnemo/diffusion/samplers/exponential_euler.py
```diff
@@ -0,0 +1,265 @@
...
+# Number of Gauss-Legendre points used by the internal quadrature of this
+# solver
+_NUM_QUADRATURE_POINTS = 4
+
+# Fractional offset from t_next toward t_cur of the fallback probe for the
+# slope-to-bias ratio; matches the innermost node of the 4-point rule
+_RATIO_PROBE_OFFSET = 0.0694318442029737
+
+
+def _nonlinear_weight(
```
laserkelvin, line 39: This helper (+ the `_NUM_QUADRATURE_POINTS` / `_RATIO_PROBE_OFFSET` constants) live byte-identical in all four solver files — the only difference anywhere is one comment word in `dpmpp_2m_unic2.py`. `_utils.py` exists for exactly this. Suggest moving `_nonlinear_weight`, the constants (and `_nonlinear_moment` too) into `_utils.py` and importing. Zero behavior change; existing tests cover all four paths. Copy-paste is fine only when a real customization need exists — here all four compute the identical J = ∫ e^{A(t')-A(s)} b(s) ds; if a solver ever needs a special kernel weight, that variant can earn its own name then.

physicsnemo/diffusion/noise_schedulers/linear_gaussian.py
```diff
@@ -905,6 +932,149 @@ def sde_denoiser(
...
+        # Capture methods as local variables to avoid referencing self
+        alpha = self.alpha
+        alpha_dot = self.alpha_dot
+        sigma = self.sigma
+        sigma_dot = self.sigma_dot
+
+        if prediction_type in ("score", "epsilon"):
+            # score- and noise-parameterizations share the bias coefficient;
+            # the reverse SDE doubles the slope of the noise-prediction term
+            factor = 1.0 if denoising_type == "ode" else 2.0
```
laserkelvin, line 1019: Small but sneaky: `1.0 if denoising_type == "ode" else 2.0` sends ANY invalid string to the SDE (2x) branch — `"ODE"`, `"sde "`, `"foo"` all silently get SDE coefficients. `get_denoiser` raises `ValueError` for the same parameter, so the two methods that must agree have different contracts:

```python
scheduler.get_linear_denoiser("x0", denoising_type="typo")      # silently SDE slope
scheduler.get_denoiser(x0_predictor=f, denoising_type="typo")   # ValueError (good)
```

Ask: `denoising_type = denoising_type.lower()` then raise unless in `{"ode", "sde"}` (lowercasing alone fixes case typos, but invalid garbage still falls through), plus one test (`test_get_denoiser_validation` already covers `prediction_type="bad"`; `denoising_type="bad"` is the missing twin). 

Longer-term we should consider an `is_ode_denoiser: bool` switch as a later API refactor — not this PR.

physicsnemo/diffusion/samplers/dpmpp_2m_unic2.py
```diff
@@ -0,0 +1,459 @@
...
+
+        x_next = (
+            e_bc * x
+            + j0_bc * n_cur
+            + c_old_bc * (n_old - n_cur)
+            + c_new_bc * (n_pred - n_cur)
+        )
+        self._n_old.copy_(n_cur)
+        self._lam_old.copy_(lam_cur_bc)
+        self._n_cur.copy_(n_pred)
```
laserkelvin, line 457: Question, not bug report. Me trace autograd carefully: in all shipped configurations the callbacks/timesteps are grad-free, so the caches never become saved tensors (mul saves an operand only when the OTHER operand requires grad), and backward through `sample()` works fine — the gradient-flow tests should pass.

BUT if timesteps ever carry grad (someone differentiates through the time grid), the coefficients require grad too, `j0_bc * n_cur` then saves the cache, and this `copy_` trips the SavedVariable version check -> RuntimeError at backward. Loud, not silent — but confusing and far from the cause:

```python
solver = DPMPlusPlus2MUniC2(denoiser, bias_fn=..., bias_int_fn=..., slope_fn=...)
x = torch.randn(4, 3)
for ta, tb in [(5.0, 2.5), (2.5, 1.25), (1.25, 0.625)]:
    x = solver.step(x, torch.full((4,), ta, requires_grad=True),   # t with grad
                       torch.full((4,), tb, requires_grad=True))
x.sum().backward()   # <- expect "modified by an inplace operation" here
```

Can you confirm grad-through-time is out of scope?

physicsnemo/diffusion/samplers/dpmpp_2m_unic2.py
```diff
@@ -0,0 +1,459 @@
...
+        lam_cur = self.lambda_fn(t_cur)
+        lam_cur_bc = lam_cur.reshape(expected_shape)
+        lam_next_bc = self.lambda_fn(t_next).reshape(expected_shape)
+
+        # The denoiser is generally singular at the terminal time t = 0, and
+        # the final step cannot reuse its endpoint evaluation anyway, so
+        # steps that land exactly at t = 0 skip the corrector and take the
+        # endpoint value at the current time instead. This keeps the update
+        # and its gradients finite on the last step of a trajectory.
+        t_eval = torch.where(t_next == 0, t_cur, t_next)
```
laserkelvin, line 373: Small pitfall: on the terminal step (`t_next == 0`) the corrector is masked off (good), but `n_pred = self._predictor_value(x_pred, t_cur, ...)` — a state/time-mismatched value — still gets cached into `_n_cur` below. Harmless when the trajectory really ends at 0; but if a caller keeps stepping past 0 without `reset()`, the history is quietly poisoned. Consider skipping the cache write when `at_zero`, or one comment marking the edge.

physicsnemo/diffusion/samplers/dpmpp_2m.py
```diff
@@ -0,0 +1,356 @@
...
+        expected_shape = (-1,) + (1,) * (x.ndim - 1)
+        lam_cur_bc = self.lambda_fn(t_cur).reshape(expected_shape)
+
+        # Recover the predictor-like term of the RHS: N = (D - a x) / b; the
+        # zero-slope guard drops the nonlinear term instead of dividing by
+        # zero
+        d_cur = self.denoiser(x, t_cur)
+        a_bc = self.bias_fn(t_cur).reshape(expected_shape)
+        b_bc = self.slope_fn(t_cur).reshape(expected_shape)
+        b_safe = torch.where(b_bc == 0, torch.ones_like(b_bc), b_bc)
```
laserkelvin, line 317: This guard silently drops the whole nonlinear term when `b(t_cur) == 0`. Fine for genuinely degenerate schedules — but it also fires silently when a user's callbacks don't match their denoiser's parameterization (e.g. x0 cheat-sheet with an epsilon denoiser), which is exactly the misconfiguration this callback API makes easy. Samples still come out, just wrong. Consider a warning when the guard actually fires (or at least a docstring note that a firing guard is a smell).

CHANGELOG.md
```diff
@@ -259,6 +270,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Extends `LinearGaussianNoiseScheduler` with two new methods:
+  `get_linear_denoiser` exposes the extended semi-linear structure used by
+  the new exponential and multistep solvers, returning a tuple of bias,
+  antiderivative, and slope callables (`bias`, `bias_int`, `slope`) and
+  taking a single `prediction_type` string; `snr` exposes the schedule's
```
laserkelvin, line 277: Two CHANGELOG fixes: (1) this line says `get_linear_denoiser` takes "a single `prediction_type` string" — the signature also takes `denoising_type`; (2) the VP `sigma_to_t` rationalization (`vp.py`) got no entry at all, but it changes VP numerics at extreme sigma — the old formula returns a slightly *negative* t at sigma=0. That is a user-visible fix worth its own line under Fixed.

test/diffusion/test_samplers.py
```diff
@@ -948,7 +1054,13 @@ def test_backward_through_sampling(
             num_steps=NUM_STEPS_SHORT,
```
laserkelvin, line 1054: This test's assertion is weak: `any(p.grad is not None and not isnan(...))` passes if ONE parameter has any gradient — a dead gradient path through the corrector or the history cache sails through, and inf grads pass too (only NaN is checked). Suggest `all(...)` + `torch.isfinite` + a nonzero-norm check. Also `NUM_STEPS_SHORT=2` means the `copy_`-chained history graph (steady state reads the cache written by the previous step's `copy_`) only actually participates in backward from step 3 onward — 3 steps would cover it.

test/diffusion/test_solvers.py
```diff
@@ -53,60 +60,149 @@
...
+    if kwargs.pop("_use_vp_scheduler", False):
+        scheduler = VPNoiseScheduler()
+    else:
+        scheduler = EDMNoiseScheduler()
+    model = instantiate_model_deterministic(
+        predictor_cls,
+        seed=0,
+        **predictor_kwargs,
+    ).to(device)
+    denoiser = scheduler.get_denoiser(x0_predictor=model, denoising_type="ode")
```
laserkelvin: Every new-solver test builds ODE denoisers, so the SDE coefficient branches (the factor-2 slopes from `get_linear_denoiser`) never run end-to-end — they exist only as callback goldens. One SDE config (any solver, any schedule) would close the hole, and pairs naturally with the `sde_x0` linear-denoiser goldens you already have.

laserkelvin (changes requested):
Ooga! Me look long at this PR. Short version: **me like**. Math check out — derived by hand twice (independent reviewers): DDIM reduction, DPM-Solver++(2M) update, UniC-2 corrector coefficients, EDM churn — all correct. VP `sigma_to_t` rewrite is a real fix (old code return slightly *negative* t at sigma=0). Golden files clean (334 new, all referenced, zero orphans). Compile design with masked updates + stable cache storage is smart.

Me leave comments inline. Two big asks, rest small.

**BIG ASK 1 — string keys not do what name say.** `sample(..., solver="dpmpp_2m")` with no `solver_options` build solver with empty pockets (no scheduler callbacks) → quietly run classical AB2, not DPM-Solver++(2M). Same story for the other three new keys. Defaults are a fine feature (me see the `dpmpp_2m_ab2` test use them on purpose) — but a user who not read the class docstrings get a different method than the name promise, and nothing tell them. Ask: warn when the new keys are used without callbacks + per-key honesty in the `sample()` docstring and docs page. Details inline at `samplers.py`.

**BIG ASK 2 — the new consistency test cannot see the fancy parts.** `TestStepConsistency` use slope=0, which make the nonlinear term N≡0 — so the J quadrature, the DPM extrapolation, and the UniC-2 corrector all get multiplied by zero, and the single-step design never reach UniC-2 steady state. The "second-order" / "third-order" claims today rest only on self-generated goldens, which catch changes, not wrong-from-the-start. Ask: one empirical order-of-accuracy test — golden-free mock-up inline at `test_solvers.py`, ~50 lines, CPU.

Small asks inline: dedup `_nonlinear_weight` ×4 into `_utils.py`; validate `denoising_type`; CHANGELOG fixes; strengthen the gradient-flow assertion; one SDE end-to-end config.

One *question* inline (`dpmpp_2m_unic2.py`): please confirm grads-through-time are out of scope — the `copy_` cache updates would crash backward loudly in that config; if out of scope, worth a one-line invariant comment.

And noting here because the file is not in this diff: `DomainParallelNoiseScheduler` forwards `get_denoiser` but not the new `get_linear_denoiser`/`snr` — domain-parallel users must find the `.inner_scheduler` trick themselves. Two small pass-through methods would keep parity.

Nice work overall — the numerics are solid; the asks above are about what users and tests can *see*.

physicsnemo/diffusion/samplers/samplers.py
CharlelieLrt: Yeah I agree, that's a good catch and a good Suggestion. In [ec1b5e5](https://github.com/NVIDIA/physicsnemo/pull/1932/commits/ec1b5e536a6bfc5cf21c0618fe00ed50c8b61557) I added some warnings in `sample()` when the solver is constructed without any optional arguments. I also added moire detailed explanations in the docs of `sample()` to clearly explain what needs to be passed to `sample()` to construct the desired solver.

test/diffusion/test_solvers.py
CharlelieLrt: Yeah that's a nice test to have, I didn't think of that. I added the new test in [a8f6ad7](https://github.com/NVIDIA/physicsnemo/pull/1932/commits/a8f6ad7fc8f0701b8be5759959f24788f108f558) (and kept the small 1-step test as a sanity check). I went a bit beyond the comment and applied this empirical order test to all solvers in the module not only the new ones introduced in this PR

physicsnemo/diffusion/samplers/exponential_euler.py
CharlelieLrt: I initially duplicated code because I like to have self-contained code even if it's at the cost of a bit of code duplication. All code in one place makes math much easier to follow usually. But that's just my personal preference, and not a strong one, so I moved the helpers to `_utils.py` in [24301d8](https://github.com/NVIDIA/physicsnemo/pull/1932/commits/24301d81e77b6c351c20f234f940d0edd03a678c)

physicsnemo/diffusion/noise_schedulers/linear_gaussian.py
CharlelieLrt: Good catch. Fixed in [537eb1e](https://github.com/NVIDIA/physicsnemo/pull/1932/commits/537eb1e7eed29f53a17d32c88e2986af07ebf1c0)

test/diffusion/test_samplers.py
CharlelieLrt: Good catch, I hardened the solver and sampler tests based on your suggestions in [61c144b](https://github.com/NVIDIA/physicsnemo/pull/1932/commits/61c144bb841d51af3021dad380a7a1c2bb58abd5)

physicsnemo/diffusion/samplers/dpmpp_2m_unic2.py
CharlelieLrt: That's a really good point, because we precisely **want to** be able to differentiate `solver.step()` and `sample()` wrt to time steps (and also other inputs, like initial condition, solver hyperparameters, etc.) So, two things in one:
- Your comment shows that there was a bling spot not covered by existing tests. There were tests for gradient flow through `solver.step()`, but they were not checking differentiability wrt timesteps, initial condition and such.
- The implementation of `solver.step()` was **not** differentiable

I fixed both in [61c144b](https://github.com/NVIDIA/physicsnemo/pull/1932/commits/61c144bb841d51af3021dad380a7a1c2bb58abd5) (tests expanded and hardened and implementation refactored to enable differentiability)

physicsnemo/diffusion/samplers/dpmpp_2m_unic2.py
CharlelieLrt: Yeah it's worth being safe there, there are definitely cases where the denoising trajectory does not end at 0. Fixed in [d24461f](https://github.com/NVIDIA/physicsnemo/pull/1932/commits/d24461f24788bb751900c4ce774351ffd2261678)

physicsnemo/diffusion/samplers/dpmpp_2m.py
CharlelieLrt: I added a docstring mention of this in all solver classes that requires this optional argument. Implemented in [d24461f](https://github.com/NVIDIA/physicsnemo/pull/1932/commits/d24461f24788bb751900c4ce774351ffd2261678)

CHANGELOG.md
CharlelieLrt: Fixed in [f64647e](https://github.com/NVIDIA/physicsnemo/pull/1932/commits/f64647efd4e65e25ce92bf07b64e14f6fcb4d085)

CharlelieLrt:
/ok to test [f64647e](https://github.com/NVIDIA/physicsnemo/pull/1932/commits/f64647efd4e65e25ce92bf07b64e14f6fcb4d085)

CharlelieLrt:
/ok to test [eb84621](https://github.com/NVIDIA/physicsnemo/pull/1932/commits/eb8462188144e35b86d1764e7187dac1a08aaa9a)

PR #1940, opened 2026-08-19
feat: add standalone FLARE++ and GeoTransolver backend

mnabian:
# PhysicsNeMo Pull Request

## Description

Implements [FLARE++](https://arxiv.org/abs/2608.11519) as two public, reusable APIs:

- `physicsnemo.nn.FLAREPlusPlus`: the paper-exact attention layer, using learned seeds, four input projections, and three scaled dot-product attention calls.
- `physicsnemo.models.flare.FLAREPlusPlus`: a standalone FLARE++ surrogate model with the same residual backbone and activation-checkpointing controls as FLARE.

The same mixer is also available inside GeoTransolver through `attention_type="GALE_FPP"`. The GeoTransolver-specific class is a thin adapter that inherits the standalone attention implementation and adds only optional geometry/global-context cross-attention; it does not contain a second FLARE++ implementation.

The branch is rebased onto current `main`; no FLARE++ code remains under `physicsnemo.experimental`. Input projections are fused into one linear operation followed by chunking, while retaining four independent parameter blocks. Transformer Engine and token-sharded context parallelism are rejected explicitly; replicated multi-GPU data parallelism remains supported.

## Review comments addressed

- Kept standalone FLARE++ public and moved the GeoTransolver context adapter into a private model-specific module.
- Replaced broad module-class annotations with explicit `Literal` backend selections.
- Removed the previous one-use helper chain and implicit `getattr`-based FLARE++ plumbing.
- Added standalone-model and GeoTransolver-backend documentation.
- Added constructor coverage, paper-equation parity, two committed forward goldens, a committed v1 checkpoint, save/load round trips, gradients, multi-input context tests, and `torch.compile(fullgraph=True)` coverage.
- Corrected GALEBlock's documentation to describe configurable attention.

## Validation

- `123 passed, 10 skipped` across the affected FLARE, FLARE++, GALE, and GeoTransolver suites.
- Domain-parallel regression suite: `11 skipped` because the local environment has no required multi-GPU runtime.
- Ruff lint/format and `git diff --check` pass.

## Checklist

## Dependencies

No new dependencies.

### DIFF
--- a/physicsnemo/experimental/models/flare/flare.py
+++ b/physicsnemo/experimental/models/flare/flare.py
@@ -1,4 +1,3 @@
-
 # SPDX-FileCopyrightText: Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES.
 # SPDX-FileCopyrightText: All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
@@ -15,11 +14,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-r"""FLARE Transolver: Transolver with FLARE attention.
+r"""FLARE-family Transolver models.
 
-Transolver variant that uses FLARE (Fast Low-rank Attention Routing Engine)
-attention instead of physics attention. Inherits from the core Transolver
-and replaces all attention blocks with FLARE blocks.
+These Transolver variants replace physics attention with either fixed-query
+FLARE or input-conditioned FLARE++ attention.
 """
 
 from __future__ import annotations
@@ -28,17 +26,17 @@ import torch
 import torch.nn as nn
 from jaxtyping import Float
 
+from physicsnemo.experimental.nn import FLARE as FLAREAttention
+from physicsnemo.experimental.nn import FLAREPlusPlus as FLAREPlusPlusAttention
 from physicsnemo.models.transolver import Transolver as CoreTransolver
 from physicsnemo.models.transolver.transolver import _TransolverMlp
 
-from physicsnemo.experimental.nn import FLARE as FLAREAttention
-
 
 class _FLAREBlock(nn.Module):
-    r"""Transformer block with FLARE attention instead of physics attention.
+    r"""Transformer block with FLARE-family attention.
 
-    Mirrors TransolverBlock structure but uses FLARE for the attention layer.
-    FLARE does not support Transformer Engine.
+    Mirrors ``TransolverBlock`` while replacing physics attention with a
+    supplied FLARE-family layer. Transformer Engine is not supported.
     """
 
     def __init__(
@@ -51,19 +49,25 @@ class _FLAREBlock(nn.Module):
         last_layer: bool = False,
         out_dim: int = 1,
         n_global_queries: int = 32,
+        attention_cls: type[nn.Module] = FLAREAttention,
+        attn_scale: float | None = None,
     ) -> None:
         super().__init__()
         self.last_layer = last_layer
         dim_head = hidden_dim // num_heads
 
         self.ln_1 = nn.LayerNorm(hidden_dim)
-        self.Attn = FLAREAttention(
+        attention_kwargs = {}
+        if attn_scale is not None:
+            attention_kwargs["attn_scale"] = attn_scale
+        self.Attn = attention_cls(
             dim=hidden_dim,
             heads=num_heads,
             dim_head=dim_head,
             dropout=dropout,
             n_global_queries=n_global_queries,
             use_te=False,
+            **attention_kwargs,
         )
         self.ln_mlp1 = nn.Sequential(
             nn.LayerNorm(hidden_dim),
@@ -89,6 +93,39 @@ class _FLAREBlock(nn.Module):
         if self.last_layer:
             return self.ln_mlp2(fx)
         return fx
+
+
+def _make_flare_blocks(
+    *,
+    attention_cls: type[nn.Module],
+    n_layers: int,
+    n_head: int,
+    n_hidden: int,
+    dropout: float,
+    act: str,
+    mlp_ratio: int,
+    out_dim: int,
+    n_global_queries: int,
+    attn_scale: float | None = None,
+) -> nn.ModuleList:
+    """Build a matched residual stack for a FLARE-family attention layer."""
+    return nn.ModuleList(
+        [
+            _FLAREBlock(
+                num_heads=n_head,
+                hidden_dim=n_hidden,
+                dropout=dropout,
+                act=act,
+                mlp_ratio=mlp_ratio,
+                last_layer=(i == n_layers - 1),
+                out_dim=out_dim,
+                n_global_queries=n_global_queries,
+                attention_cls=attention_cls,
+                attn_scale=attn_scale,
+            )
+            for i in range(n_layers)
+        ]
+    )
 
 
 class FLARE(CoreTransolver):
@@ -128,6 +165,9 @@ class FLARE(CoreTransolver):
         Shape of structured data. ``None`` for unstructured. Default is ``None``.
     time_input : bool, optional
         Whether to include time embeddings. Default is ``False``.
+    attn_scale : float, optional
+        Attention-logit scale. Default is ``1.0`` for backward compatibility
+        with the original FLARE implementation.
 
     Forward
     -------
@@ -159,6 +199,7 @@ class FLARE(CoreTransolver):
         ref: int = 8,
         structured_shape: None | tuple[int, ...] = None,
         time_input: bool = False,
+        attn_scale: float = 1.0,
     ) -> None:
         super().__init__(
             functional_dim=functional_dim,
@@ -180,19 +221,133 @@ class FLARE(CoreTransolver):
         )
 
         # Replace physics attention blocks with FLARE blocks
-        self.blocks = nn.ModuleList(
-            [
-                _FLAREBlock(
-                    num_heads=n_head,
-                    hidden_dim=n_hidden,
-                    dropout=dropout,
-                    act=act,
-                    mlp_ratio=mlp_ratio,
-                    last_layer=(i == n_layers - 1),
-                    out_dim=out_dim,
-                    n_global_queries=slice_num,
-                )
-                for i in range(n_layers)
-            ]
+        self.blocks = _make_flare_blocks(
+            attention_cls=FLAREAttention,
+            n_layers=n_layers,
+            n_head=n_head,
+            n_hidden=n_hidden,
+            dropout=dropout,
+            act=act,
+            mlp_ratio=mlp_ratio,
+            out_dim=out_dim,
+            n_global_queries=slice_num,
+            attn_scale=attn_scale,
+        )
+        self.initialize_weights()
+
+
+class FLAREPlusPlus(CoreTransolver):
+    r"""Transolver with FLARE++ dynamic-routing attention.
+
+    FLARE++ synthesizes its routing queries from each block's current input
+    before applying FLARE's low-rank gather-scatter operator. This changes only
+    the token mixer: the Transolver residual stream, normalization, and
+    feed-forward blocks are identical to :class:`FLARE`.
+
+    For architecture details, see:
+
+    - `FLARE++: Low-rank attention with dynamic attention routing
+      <https://arxiv.org/abs/2608.11519>`_
+
+    Parameters
+    ----------
+    functional_dim : int
+        Dimension of input values, not including embeddings.
+    out_dim : int
+        Dimension of model output.
+    embedding_dim : int | None, optional
+        Dimension of input embeddings. Required if ``unified_pos=False``.
+    n_layers : int, optional
+        Number of transformer blocks. Default is 4.
+    n_hidden : int, optional
+        Hidden dimension. Default is 256.
+    dropout : float, optional
+        Dropout rate. Default is 0.0.
+    n_head : int, optional
+        Number of attention heads. Default is 8.
+    act : str, optional
+        Activation function name. Default is ``"gelu"``.
+    mlp_ratio : int, optional
+        MLP hidden ratio. Default is 4.
+    slice_num : int, optional
+        Number of dynamic routing queries. Default is 32.
+    unified_pos : bool, optional
+        Whether to use unified positional embeddings. Default is ``False``.
+    ref : int, optional
+        Reference grid size for unified position. Default is 8.
+    structured_shape : None | tuple[int, ...], optional
+        Shape of structured data. ``None`` for unstructured data. Default is
+        ``None``.
+    time_input : bool, optional
+        Whether to include time embeddings. Default is ``False``.
+    attn_scale : float | None, optional
+        Attention-logit scale. ``None`` uses ``(n_hidden / n_head) ** -0.5``
+        as specified by FLARE++. Pass ``1.0`` for historical FLARE scaling.
+
+    Forward
+    -------
+    Same as :class:`~physicsnemo.models.transolver.Transolver`.
+
+    Outputs
+    -------
+    Same as :class:`~physicsnemo.models.transolver.Transolver`.
+
+    See Also
+    --------
+    :class:`FLARE` : Fixed-query FLARE model.
+    :class:`~physicsnemo.experimental.nn.flare_attention.FLAREPlusPlus` :
+        FLARE++ attention layer.
+    """
+
+    def __init__(
+        self,
+        functional_dim: int,
+        out_dim: int,
+        embedding_dim: int | None = None,
+        n_layers: int = 4,
+        n_hidden: int = 256,
+        dropout: float = 0.0,
+        n_head: int = 8,
+        act: str = "gelu",
+        mlp_ratio: int = 4,
+        slice_num: int = 32,
+        unified_pos: bool = False,
+        ref: int = 8,
+        structured_shape: None | tuple[int, ...] = None,
+        time_input: bool = False,
+        attn_scale: float | None = None,
+    ) -> None:
+        super().__init__(
+            functional_dim=functional_dim,
+            out_dim=out_dim,
+            embedding_dim=embedding_dim,
+            n_layers=n_layers,
+            n_hidden=n_hidden,
+            dropout=dropout,
+            n_head=n_head,
+            act=act,
+            mlp_ratio=mlp_ratio,
+            slice_num=slice_num,
+            unified_pos=unified_pos,
+            ref=ref,
+            structured_shape=structured_shape,
+            use_te=False,
+            time_input=time_input,
+            plus=False,
+        )
+
+        self.blocks = _make_flare_blocks(
+            attention_cls=FLAREPlusPlusAttention,
+            n_layers=n_layers,
+            n_head=n_head,
+            n_hidden=n_hidden,
+            dropout=dropout,
+            act=act,
+            mlp_ratio=mlp_ratio,
+            out_dim=out_dim,
+            n_global_queries=slice_num,
+            attn_scale=(n_hidden // n_head) ** -0.5
+            if attn_scale is None
+            else attn_scale,
         )
         self.initialize_weights()
--- a/physicsnemo/experimental/models/geotransolver/gale.py
+++ b/physicsnemo/experimental/models/geotransolver/gale.py
@@ -23,6 +23,8 @@ geometry and global context embeddings.
 
 from __future__ import annotations
 
+import math
+
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
@@ -30,17 +32,18 @@ from einops import rearrange
 from jaxtyping import Float
 
 import physicsnemo  # noqa: F401 for docs
-from physicsnemo.core.version_check import check_version_spec, OptionalImport
-from physicsnemo.nn import Mlp
+from physicsnemo.core.version_check import OptionalImport, check_version_spec
+from physicsnemo.experimental.nn.flare_attention import (
+    _flare_plus_plus_self_attention,
+    _flare_self_attention,
+)
+from physicsnemo.nn import ConcreteDropout, Mlp
 from physicsnemo.nn.module.physics_attention import (
     PhysicsAttentionIrregularMesh,
     PhysicsAttentionStructuredMesh2D,
     PhysicsAttentionStructuredMesh3D,
     _project_input,
 )
-from physicsnemo.experimental.nn.flare_attention import _flare_self_attention
-
-from physicsnemo.nn import ConcreteDropout
 
 # Check optional dependency availability
 TE_AVAILABLE = check_version_spec("transformer_engine", "0.1.0", hard_fail=False)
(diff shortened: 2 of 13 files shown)


physicsnemo/experimental/models/flare/flare.py
CharlelieLrt, line 1: Just a heads-up: FLARE moved out of experimental since the PR was opened, so everything in this PR should also move out of it.

physicsnemo/experimental/models/flare/flare.py
```diff
@@ -14,8 +14,340 @@
...
+        self,
+        num_heads: int,
+        hidden_dim: int,
+        dropout: float,
+        act: str = "gelu",
+        mlp_ratio: int = 4,
+        last_layer: bool = False,
+        out_dim: int = 1,
+        n_global_queries: int = 32,
+        attention_cls: type[nn.Module] = FLAREAttention,
```
CharlelieLrt: This type hint a little dangerous, because the `attention_cls` is actually supposed to have a specific constructor signature. We can't pass *any* `nn.Module`. A `Literal` with the accepted classes would be safer

physicsnemo/experimental/models/geotransolver/gale.py
```diff
@@ -84,6 +88,56 @@ def _mix_self_and_cross(
             raise ValueError(f"Invalid state_mixing_mode: {mode!r}")
 
 
+def _validate_attention_scale(scale: float) -> float:
```
CharlelieLrt: We really need a dedicated helper function for something so small? It's just called once.

physicsnemo/experimental/models/geotransolver/gale.py
```diff
@@ -84,6 +88,56 @@ def _mix_self_and_cross(
...
+    """Return a validated, finite positive attention-logit scale."""
+    scale = float(scale)
+    if not math.isfinite(scale) or scale <= 0.0:
+        raise ValueError(
+            f"attn_scale must be a finite value greater than 0, got {scale}"
+        )
+    return scale
+
+
+def _gale_flare_mix_context_and_project(
```
CharlelieLrt: I am very confused with these names: is this helper function for GALE? Or is it for FLARE/FLARE++? Or for both? Or for *any* kind of attention operation that we may support in the future?

Also, I do not totally understand the point to have two helper functions `_gale_flare_mix_context_and_project` and `_mix_self_and_cross`. This is extreme granularity in terms of helper function, as there is basically a helper function for every 5 liners of operations, and that make it very difficult tro follow the chain of operations done by the model.
 It would be great if you could reduce the number of helper function unless really necessary. For example, could it be possible to merge `_gale_flare_mix_context_and_project` and `_mix_self_and_cross` into a single function?

Can you also explain how it is needed/differs from `_gale_compute_slice_attention_cross`, which is also doing cross attention between slices and context?

physicsnemo/experimental/models/geotransolver/gale.py
```diff
@@ -84,6 +88,56 @@ def _mix_self_and_cross(
...
+    module: nn.Module,
+    x_mid: list[Float[torch.Tensor, "batch heads tokens dim"]],
```
CharlelieLrt: Just curious, what are the elements in `x_mid` here?

physicsnemo/experimental/models/geotransolver/gale.py
```diff
@@ -84,6 +88,56 @@ def _mix_self_and_cross(
...
+
+    Fixed-query FLARE and FLARE++ differ only in how their self-attention
+    outputs are produced. Their geometry/global cross-attention and token
+    projection paths are identical, so keeping that logic here prevents the
+    two GeoTransolver backends from drifting apart.
+    """
+    if context is not None:
+        q = [module.cross_q(_x_mid) for _x_mid in x_mid]
+        k = module.cross_k(context)
+        v = module.cross_v(context)
```
CharlelieLrt: Just double-checking whether this is optimal vs. one single pass and chuncking into q, k, v?

physicsnemo/experimental/models/geotransolver/geotransolver.py
```diff
@@ -14,8 +14,785 @@
...
+        Pass a plain ``dict`` whose keys match the fields of
+        :class:`~physicsnemo.experimental.guardrails.embedded.OODGuardConfig`
+        (``buffer_size`` required; ``knn_k`` and ``sensitivity`` optional), or
+        ``None`` to disable the guard entirely. A ``dict`` is required (rather
+        than the dataclass directly) so the model kwargs remain
+        JSON-serialisable for ``.mdlus`` checkpointing. When set, the guard
+        accumulates global-parameter bounds and pooled geometry latents during
+        training, and emits warnings on out-of-distribution inputs during
+        inference. Default is ``None``.
+    attention_type : str, optional
```
CharlelieLrt: Tighten up the type hint with a `Literal` instead

physicsnemo/experimental/models/geotransolver/geotransolver.py
```diff
@@ -14,8 +14,785 @@
...
+        input-conditioned FLARE++ routing. Default is ``"GALE"``.
+    state_mixing_mode : str, optional
+        How to blend self-attention and cross-attention outputs in GALE layers.
+        ``"weighted"`` uses a learnable sigmoid-gated weighted sum.
+        ``"concat_project"`` concatenates the two along the head dimension and
+        projects back with a linear layer. Default is ``"weighted"``.
+    attn_scale : float | None, optional
+        Attention-logit scale for FLARE-family backends. ``None`` selects the
+        backend default: ``1.0`` for ``"GALE_FA"`` and
+        the inverse square root of the effective per-head dimension for
```
CharlelieLrt: > the inverse square root of the effective per-head dimension

This type of sentence is just impossible to understand for a human being, IMO. There are a few of these kind of sentences in the docstrings in this PR. I think they should be made either more human readable, or just totally removed.

physicsnemo/experimental/models/geotransolver/gale.py
```diff
@@ -84,6 +88,56 @@ def _mix_self_and_cross(
...
+        cross_attention = [
+            F.scaled_dot_product_attention(_q, k, v, scale=module.scale) for _q in q
+        ]
+        outputs = [
+            _mix_self_and_cross(
+                self_attn,
+                cross_attn,
+                module.state_mixing_mode,
+                state_mixing=getattr(module, "state_mixing", None),
+                concat_project=getattr(module, "concat_project", None),
```
CharlelieLrt: These `getattr` are very implicit and are opening the door for lots of silent failures. It would be better to have something more explicit.

physicsnemo/experimental/models/geotransolver/gale.py
```diff
@@ -84,6 +88,56 @@ def _mix_self_and_cross(
...
+    scale = float(scale)
+    if not math.isfinite(scale) or scale <= 0.0:
+        raise ValueError(
+            f"attn_scale must be a finite value greater than 0, got {scale}"
+        )
+    return scale
+
+
+def _gale_flare_mix_context_and_project(
+    module: nn.Module,
```
CharlelieLrt: Same comment as in other places for the `module` argument: it would be cleaner if you could find a way to have a class that clearly defines which attributes and methods (with their signatures) the `module` should have.

test/experimental/models/flare/test_flare.py
```diff
@@ -95,32 +84,94 @@ def test_flare_constructor(config):
     assert hasattr(model, "meta"), "Model should have metadata"
 
 
-@requires_module("transformer_engine>=2.14.0")
-def test_flare_te_basic(device):
-    """Test the full FLARE model with Transformer Engine enabled."""
-    if device == "cpu":
-        pytest.skip("Transformer Engine requires CUDA")
+def test_flare_plus_plus_constructor():
```
CharlelieLrt: IMO test coverage is lacking. We should at least have tests for:
- Constructor with default vs custom parameters (checks the value of public attributes in both cases)
- Non-regression test of the forward pass (and all public methods operating on tensors) against golden files, in at least a few common configurations
- Test to load some saved reference checkpoints to make sure we can always load previous versions of the model
- Test that backward prop through the model (those are there it seems)
- If there is any compilation gain to expect, then text that the model can be compiled with fullgraph=True.

physicsnemo/experimental/models/geotransolver/gale.py
```diff
@@ -692,76 +725,173 @@ def forward(
...
+        query_v = [
+            _project_input(
+                _x,
+                self.query_synthesis_v,
+                self.heads,
+                self.dim_head,
+                "B N (H D) -> B N H D",
+            ).permute(0, 2, 1, 3)
+            for _x in x
+        ]
```
CharlelieLrt: Can you confirm this is optimal in comparison to 1 projection followed by chucnking into q/k/v?

physicsnemo/experimental/models/geotransolver/gale.py
```diff
@@ -692,76 +725,173 @@ def forward(
...
+                self.scale,
+            )
+            for _x_mid, _query_k, _query_v in zip(x_mid, query_k, query_v, strict=True)
+        ]
+        return _gale_flare_mix_context_and_project(self, x_mid, self_attention, context)
 
 
-class GALEBlock(nn.Module):
+class GALE_block(nn.Module):
     r"""Transformer encoder block using GALE attention.
```
CharlelieLrt: It's not part of the PR but can you correct the docstring? It's not using GALE attention but a configurable attention

physicsnemo/experimental/nn/flare_attention.py
```diff
@@ -0,0 +1,411 @@
...
+    G = q_global.to(dtype=x_mid.dtype).expand(x_mid.shape[0], -1, -1, -1)
+    k = self_k(x_mid)
+    v = self_v(x_mid)
+    z = F.scaled_dot_product_attention(G, k, v, scale=scale)
+    return F.scaled_dot_product_attention(k, G, z, scale=scale)
+
+
+def _flare_plus_plus_self_attention(
+    x_mid: Float[torch.Tensor, "B H N D"],
+    q_seed: nn.Parameter,
```
CharlelieLrt: Not: it is true that in the broader context of GeoT, `q_seed` is indeed a `nn.Parameter`, but strictly speaking in this self-contained function it could be any tensor.

physicsnemo/experimental/nn/flare_attention.py
```diff
@@ -0,0 +1,411 @@
...
+        Self-attended output of shape :math:`(B, H, N, D)`.
+    """
+    G = q_global.to(dtype=x_mid.dtype).expand(x_mid.shape[0], -1, -1, -1)
+    k = self_k(x_mid)
+    v = self_v(x_mid)
+    z = F.scaled_dot_product_attention(G, k, v, scale=scale)
+    return F.scaled_dot_product_attention(k, G, z, scale=scale)
+
+
+def _flare_plus_plus_self_attention(
```
CharlelieLrt: Do we really need yet another private helper function for 6 lines of math? It's only called once

physicsnemo/experimental/nn/flare_attention.py
```diff
@@ -0,0 +1,411 @@
...
+
+        y = _flare_plus_plus_self_attention(
+            x_mid,
+            self.q_seed,
+            query_k,
+            query_v,
+            self.self_k,
+            self.self_v,
+            self.scale,
+        )
```
CharlelieLrt: It's probably not because of this PR, but it is at least the 3rd time I read the same block of code in multiple different modules, including some that have inter-dependencies.
I'm not sure if you can do anything about it, as it's a bit outside the scope of this PR, but it would be awesome if you could find a better way to cleanup and streamline this 😄 

Note: I re-reading everything I realized maybe it's because of `models/flare/flare.py` which is kind of separated and independent form the others

physicsnemo/experimental/nn/flare_attention.py
```diff
@@ -0,0 +1,411 @@
...
+
+        out_x = y.permute(0, 2, 1, 3)  # (B, H, N, D) -> (B, N, H, D)
+        out_x = rearrange(out_x, "b n h d -> b n (h d)")
+        out_x = self.out_linear(out_x)
+        return self.out_dropout(out_x)
+
+
+class FLAREPlusPlus(FLARE):
```
CharlelieLrt: Why is this needed? The GeoTransolver is not even using it, right?

CharlelieLrt (changes requested):
Mostly minor comments, but maybe two that are more important than the others:
1. Test coverage is insuficient for the new flare++, in particular non-regresison tests of public methods and loading from existing checkpoints
2. Everything has moved out of experimental and `gale.py` is now in `physicsnemo/nn`, along side `flare_attention.py`. The problem is that in terms of organization, it's not clear what goes in `gale.py` and what goes in `flare_attention.py`. We end up in the worst possible situation with bunch of stuff that is specific  to flare/flare++ but ends up in `gale.py`. Then what is the point of having a separate `flare_attention.py`. The PR would gain from a better organization after rebasing on main.

physicsnemo/experimental/models/flare/flare.py
mnabian: Rebased onto current main. The attention layer, standalone model, GeoTransolver adapter, tests, and docs now live under the public `physicsnemo.nn` / `physicsnemo.models` paths; this PR no longer adds anything under `experimental`.

physicsnemo/experimental/models/flare/flare.py
mnabian: Removed the broad `attention_cls: type[nn.Module]` hook. `_FLAREBlock` now accepts `Literal["FLARE", "FLARE++"]` and constructs each supported class explicitly.

physicsnemo/experimental/models/geotransolver/gale.py
mnabian: Removed this one-use helper as part of the reorganization.

physicsnemo/experimental/models/geotransolver/gale.py
mnabian: Removed the FLARE-specific helper chain. FLARE++ now owns its mixer in `flare_attention.py`; `GALE_FPP` inherits that implementation and performs its optional context attention/mixing directly. `_gale_compute_slice_attention_cross` remains the existing GALE slice-token path and is not used by FLARE++.

physicsnemo/experimental/models/geotransolver/gale.py
mnabian: The old `x_mid` helper path is gone. The new implementation uses explicit `query_k`, `query_v`, `physical_k`, and `physical_v` names throughout.

physicsnemo/experimental/models/geotransolver/gale.py
mnabian: Changed this to a single fused input projection followed by chunking. It still represents four independent paper-defined projections, but executes them in one matrix multiplication.

physicsnemo/experimental/models/geotransolver/geotransolver.py
mnabian: Done. Both `GeoTransolver.attention_type` and `GALEBlock.attention_type` now use explicit `Literal` types including `GALE_FPP`.

physicsnemo/experimental/models/geotransolver/geotransolver.py
mnabian: Reworded the scale documentation in plain language: `None` uses the standard `1 / sqrt(dim_head)` attention scale. The other dense wording from the old patch was removed.

physicsnemo/experimental/models/geotransolver/gale.py
mnabian: The new FLARE++/GeoTransolver path has no implicit `getattr` dispatch. Weighted and concat-project mixing use explicit attributes in explicit branches.

physicsnemo/experimental/models/geotransolver/gale.py
mnabian: Removed the generic `module` helper entirely. `GALE_FPP` now inherits the concrete standalone `FLAREPlusPlus` class, so the available attributes and method signatures are defined by the class hierarchy.

test/experimental/models/flare/test_flare.py
mnabian: Expanded coverage substantially: default/custom constructor assertions, two committed forward goldens, a committed v1 `.mdlus` reference checkpoint plus save/load round trips, paper-equation parity, backward tests, multi-input/context tests, and `torch.compile(..., fullgraph=True)` for the layer, standalone model, and GeoTransolver adapter.

physicsnemo/experimental/models/geotransolver/gale.py
mnabian: Confirmed and changed: FLARE++ now uses one `C -> 4HD` projection and chunks it into query-synthesis K/V and physical K/V. Context K/V are likewise fused and chunked.

physicsnemo/experimental/models/geotransolver/gale.py
mnabian: Corrected: `GALEBlock` is now documented as a transformer block using configurable attention, with the three supported backends listed explicitly.

physicsnemo/experimental/nn/flare_attention.py
mnabian: The old self-contained helper and its overly narrow `nn.Parameter` annotation were removed. `q_seed` is now the actual parameter owned by the concrete `FLAREPlusPlus` module.

physicsnemo/experimental/nn/flare_attention.py
mnabian: Removed that one-use function. The remaining `_compute_attention` method is the shared implementation used by both standalone FLARE++ and the GeoTransolver adapter.

physicsnemo/experimental/nn/flare_attention.py
mnabian: Streamlined this so the FLARE++ mixer is implemented once in `flare_attention.py`. The standalone model composes that layer, and `GALE_FPP` inherits it and adds only context mixing; the core routing math is no longer duplicated.

physicsnemo/experimental/nn/flare_attention.py
mnabian: This is intentionally a public standalone FLARE++ attention/model API in addition to the GeoTransolver backend. `GALE_FPP` now directly inherits and reuses it, so the same implementation serves both use cases.

mnabian:
Follow-up pushed in `071c05d9`:

- Rejects `attention_type="GALE_FPP"` with `plus=True` eagerly at both the public `GALEBlock` and `GeoTransolver` entry points, since `plus` selects the separate Transolver++ Gumbel-slice path.
- Adds exact deterministic-evaluation and RNG-state regression coverage for standalone `physicsnemo.nn.FLAREPlusPlus`, standalone `physicsnemo.models.flare.FLAREPlusPlus`, `physicsnemo.nn.GALE_FPP`, and the GeoTransolver `GALE_FPP` backend.
- Exercises nonzero standard dropout and Concrete Dropout where supported.

Local validation: `121 passed, 10 skipped`; all configured pre-commit hooks pass.

docs/api/nn/layers/flare_attention.rst
CharlelieLrt, line 1: Just to confirm: are you sure this doc goes in `nn/layers`? Because the code itself is in `nn/module` and not `layers`. I'm not sure how the API ref for the physicsnemo.mm module is organized but worth double-checking

physicsnemo/nn/module/gale.py
```diff
@@ -761,8 +767,177 @@ def forward(
         return [self.out_dropout(_out) for _out in outputs]
 
 
+class GALE_FPP(FLAREPlusPlus):
```
CharlelieLrt: I think the overall organization is improved in comparison to the first round, but I am still slitghly confused as to why this `GALE_FPP` is in this module `nn.module/gale.py`
- It is for `FLARE`, not `GALE` attention, so what justifies putting it here rather than `nn.module.flare_attention.py`?
- It is also an adapter specific to `GeoTransolver`, so why putting it here instead of `models/geotransolver`? 

Actually looking at it, it seems it is only going to be useful for GeoTransolver (or other closely related models). In this case the second option might make more sense

test/models/flare/data/_generate_flare_plus_plus_references.py
CharlelieLrt, line 1: The purpose of this script is to generate the golden files, right? Instead it would be better if the **same test code** is used to:
- generate the reference golden files when they are not yet present in `data`
- load the reference golden files when they already exist in `data`
See the pattern in `test/diffusion/test_losses.py` for an example

test/models/flare/test_flare.py
```diff
@@ -139,6 +141,199 @@ def test_flare_constructor(config):
...
+    checkpoint = Path(__file__).parent / "data/flare_plus_plus_v1.mdlus"
+    model = Module.from_checkpoint(checkpoint).to(device)
+    functional_input = torch.randn(1, 13, 2, device=device)
+    embedding = torch.randn(1, 13, 3, device=device)
+
+    assert isinstance(model, FLAREPlusPlus)
+    assert model(functional_input, embedding).shape == (1, 13, 1)
+
+
+def test_flare_plus_plus_torch_compile_fullgraph(device):
```
CharlelieLrt, line 355: I think you need to set this too at the beginning of the test: `torch._dynamo.config.error_on_recompile = True`, and make sure it's automatically disabled by pytest when exiting the test (either completion or failure).

test/models/geotransolver/test_geotransolver.py
CharlelieLrt, line 1: Why does the test pattern for geotransolver does not follow the same as FLARE? There should be a similar set of tests with: constructor test, non-regression of a freshly initialized model against reference golden files, non-regression of a model loaded from a reference checkpoint, compile test and gradient test.

test/nn/module/test_flare_attention.py
CharlelieLrt, line 1: Sama here: there are IMO a few tests that are missing to have full coverage.

CharlelieLrt (approved):
I had a few more minor comments, but overall looks good

docs/api/nn/layers/flare_attention.rst
mnabian: Confirmed. The API reference organizes public physicsnemo.nn objects by concept under docs/api/nn/layers, independently of the source package directory. docs/api/physicsnemo.nn.layers.rst includes nn/layers/flare_attention in its toctree, alongside the other public layer categories, so this is the intended location.

physicsnemo/nn/module/gale.py
mnabian: Agreed. I moved the context adapter into the private model-specific module physicsnemo/models/geotransolver/flare_plus_plus.py. GALE_FPP is no longer a physicsnemo.nn class or a GALEBlock backend; only the GeoTransolver configuration string remains. The public standalone implementation is still physicsnemo.nn.FLAREPlusPlus and the private adapter subclasses it.

test/models/flare/data/_generate_flare_plus_plus_references.py
mnabian: Done. The standalone generator script has been removed. The forward and reference-checkpoint tests now create their golden artifacts from the same test code when absent and load/compare them when present, following the load-or-create pattern used by the diffusion tests.

test/models/flare/test_flare.py
mnabian: Done. The compile test now sets torch._dynamo.config.error_on_recompile = True before compilation and invokes the compiled graph twice. The repository-wide autouse reset_dynamo_state fixture resets Dynamo and restores error_on_recompile=False after every test, including failure paths. The attention and GeoTransolver FLARE++ compile tests use the same contract.

test/models/geotransolver/test_geotransolver.py
mnabian: Added a dedicated test/models/geotransolver/test_flare_plus_plus.py suite with constructor/backend selection, fresh-model golden output, committed reference checkpoint plus golden output, fullgraph compile with recompile detection, end-to-end gradients, exact standalone equivalence without context, context-equation parity, multi-stream gradients, validation, and deterministic eval with Concrete Dropout.

test/nn/module/test_flare_attention.py
mnabian: Expanded the standalone attention coverage with a fresh-initialization golden-output regression and stronger compile coverage that detects recompilation and executes the graph twice. Together with the existing constructor, paper-equation parity, input-conditioned routing, gradient, option-validation, deterministic-eval, and fullgraph tests, this now covers the expected attention-layer contract.

docs/api/models/flare.rst
```diff
@@ -9,6 +9,13 @@ a low-rank attention mechanism, and supports structured and unstructured data.
 For details of the attention mechanism, see the `FLARE paper
 <https://arxiv.org/abs/2508.12594>`__.
 
+The standalone :class:`~physicsnemo.models.flare.FLAREPlusPlus` model keeps the
```
ktangsali, line 12: Is this correct? The role targets `physicsnemo.models.flare.FLAREPlusPlus`, while autoclass registers `physicsnemo.models.flare.flare.FLAREPlusPlus`.

ktangsali (approved):
Approving to unblock, but please see my previous comment

docs/api/models/flare.rst
mnabian: Good catch—the two targets were inconsistent. I changed the `autoclass` directive to the public `physicsnemo.models.flare.FLAREPlusPlus` path, so it now matches the role and documents the intended user-facing import while leaving the implementation module unchanged.

loliverhennigh (approved):
LGTM

PR #1947, opened 2026-08-24
Fix ShardTensor on Cross Products

### LINKED ISSUE #1944: 🐛[BUG]: ShardTensor cross handlers mishandle replicated operands and valid API cases
## Summary

The registered `ShardTensor` handlers for `torch.cross`, `Tensor.cross`, and `torch.linalg.cross` have several confirmed correctness gaps for valid operand combinations. Most importantly, a replicated operand that requires gradients receives only its rank-local contribution instead of the sum across sharded ranks, silently producing a wrong training gradient.

The other confirmed cases are:

- A full-size `Replicate()` operand is accepted but not sliced to the reference tensor's local shard, causing a local shape error.
- A mixed `ShardTensor`/`DTensor` call can select the `DTensor` as its output reference and then access ShardTensor-only metadata.
- `torch.cross(dim=None)` chooses its default dimension from the broadcast output shape instead of the first input's shape.
- Under `torch.compile`, a mixed plain-tensor/ShardTensor call bypasses plain-tensor promotion and fails during tracing.

## Minimal reproducer

The silent gradient error requires two ranks. Run with:

```bash
torchrun --standalone --nproc-per-node=2 repro.py
```

```python
import torch
import torch.distributed as dist
from torch.distributed.device_mesh import DeviceMesh
from torch.distributed.tensor.placement_types import Shard

from physicsnemo.domain_parallel import ShardTensor


dist.init_process_group("gloo")
rank = dist.get_rank()
world_size = dist.get_world_size()
mesh = DeviceMesh("cpu", list(range(world_size)))

full_a = torch.tensor(
    [
        [1.0, 2.0, 3.0],
        [4.0, 5.0, 6.0],
        [7.0, 8.0, 9.0],
        [10.0, 11.0, 12.0],
    ]
)
local_a = full_a.chunk(world_size, dim=0)[rank].clone().requires_grad_(True)
sharded_a = ShardTensor.from_local(
    local_a,
    mesh,
    (Shard(0),),
    sharding_shapes="chunk",
    global_shape=tuple(full_a.shape),
)

axis = torch.tensor([[0.25, -0.5, 2.0]], requires_grad=True)
torch.linalg.cross(sharded_a, axis).full_tensor().sum().backward()

ref_a = full_a.clone().requires_grad_(True)
ref_axis = axis.detach().clone().requires_grad_(True)
torch.linalg.cross(ref_a, ref_axis).sum().backward()

print("observed:", axis.grad, "expected:", ref_axis.grad)
torch.testing.assert_close(axis.grad, ref_axis.grad)
```

Each rank observes:

```text
observed = [[2, -4, 2]]
expected = [[4, -8, 4]]
```

Expected: supported cross-product calls match native PyTorch in forward and backward. A replicated operand should be localized without a forward collective, and its rank-local gradient contributions should be reduced before reaching the original tensor.

Additional confirmed cases:

- `(Shard(0),)` shape `(4, 3)` crossed with `Replicate()` shape `(4, 3)` fails locally on `(2, 3)` versus `(4, 3)`.
- A replicated `ShardTensor` crossed with a sharded `DTensor` raises `AttributeError: 'DTensorSpec' object has no attribute 'sharding_shapes'`.
- With first-input shape `(1, 4, 3)` and second-input shape `(3, 4, 3)`, `torch.cross(dim=None)` chooses dimension 0 instead of native PyTorch's dimension 2 and fails.
- `torch.compile(fullgraph=True, backend="aot_eager")` fails for a ShardTensor plus plain tensor, while the ShardTensor/ShardTensor control passes.

Controls:

- The sharded operand's gradient is correct in the primary reproducer; only the replicated operand's gradient is missing the cross-rank reduction.
- On PyTorch 2.12, unregistering these handlers restores correct behavior for the primary reproducer, mixed-DTensor case, and `dim=None` case, although the fallback may replicate the output.
- PyTorch 2.10 has no native `linalg_cross` sharding strategy, so the handlers need to be fixed rather than removed.

Related:

- Discovered during post-hoc review of #1908.
- #1927 fixes scalar-with-gradient promotion, but not this missing cross-rank reduction or the explicit `Replicate()` case.
- #1796 tracks a separate `torch.compile` failure during ShardTensor construction.
- #1943 tracks the broader uneven-shard-boundary invariant exposed by binary operations; a cross-specific guard should not close that issue.

Observed with PyTorch 2.12 on one- and two-rank CPU/Gloo device meshes.

coreyjadams:
# PhysicsNeMo Pull Request

This PR fixes some bugs in the handling of sharded cross products.  Closes #1944 .

~We shouldn't merge this until after release.  Right now it's built against 2.2.0-rc, it needs that branch to merge back to main, then it can rebase onto main and merge.  I have started the PR against the RC so we can review it now, but the merge must wait until after release!~

## Checklist

## Review Process

### DIFF
--- a/physicsnemo/distributed/manager.py
+++ b/physicsnemo/distributed/manager.py
@@ -585,6 +585,11 @@ class DistributedManager(object):
         )
 
         if manager._distributed:
+            DistributedManager._isolate_torch_compile_cache(
+                manager._local_rank, manager._world_size
+            )
+
+        if manager._distributed:
             # Setup distributed process group
             try:
                 dist.init_process_group(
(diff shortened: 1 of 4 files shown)


physicsnemo/domain_parallel/custom_ops/_tensor_ops.py
```diff
@@ -273,74 +284,356 @@ def _normalize_cross_dim(
     return dim - ndim if dim >= 0 else dim
 
 
-def _cross_output_ref(
-    input_tensor: Any,
-    other_tensor: Any,
+# ---------------------------------------------------------------------------
```
peterdsharpe, line 287: Probably don't need this super long comment, but it's fine

peterdsharpe (approved):
LGTM! Thanks!

physicsnemo/distributed/manager.py
```diff
@@ -609,6 +614,73 @@ def setup(
 
         manager._initialization_method = method
 
+    @staticmethod
+    def _isolate_torch_compile_cache(local_rank: int, world_size: int) -> None:
+        """Give each rank its own ``torch.compile`` (inductor) cache directory.
```
coreyjadams, line 619: This is a new method I set up to fire whenever the distributed manager is used to initialize with pytorch - it will put the inductor cache in a per-rank local space, which is good for shard tensor.

physicsnemo/distributed/manager.py
```diff
@@ -741,6 +813,22 @@ def create_group_from_node(
...
+        parent : Optional[str], optional
+            Name of the parent process group to subdivide. Default None.
+        verbose : bool, default=False
+            Print group creation details.
+
+        Returns
+        -------
+        str
+            Name of the orthogonal group, usable as the parent for siblings.
+        """
```
coreyjadams, line 831: This is just a docstring that has been missing forever, adding it since it's been a grandfathered exception.  Same as below.

physicsnemo/domain_parallel/custom_ops/_tensor_ops.py
coreyjadams: I think this is outdated now, I cleaned up the comments.

negin513 (approved):
LGTM!

coreyjadams:
/ok to test [4acc910](https://github.com/NVIDIA/physicsnemo/pull/1947/commits/4acc910d55b94464d423b6bda48ff6269edbc91f)

PR #1948, opened 2026-08-24
Tighten import-linter dependency contracts

peterdsharpe:
## Summary

Tighten PhysicsNeMo's import boundaries by making complete contracts exhaustive, assigning previously untracked modules to their intended layers, and making currently independent siblings explicitly non-shareable. The partial `nn` contract remains non-exhaustive, while `compat` and `conftest` remain explicit top-level exceptions.

### Before

```ini
active_learning : diffusion
sym

samplers : metrics
multi_diffusion : preconditioners : guidance
utils
```

### After

```ini
exhaustive = True

deploy | active_learning | diffusion | sym

samplers | metrics
multi_diffusion | preconditioners | guidance
utils | base
```


.importlinter
```diff
@@ -86,28 +99,31 @@ layers =
...
 type = layers
+exhaustive = True
 containers=
    physicsnemo.models
 layers = 
-   mesh_reduced
+   mesh_reduced | flare | geotransolver
    afno | dlwp | dlwp_healpix | dit | domino | dpot | fengwu | figconvnet | fno | graphcast | meshgraphnet | pangu | pix2pix | rnn |  srrn | swinvrnn | topodiff | transolver  | vfgn | diffusion_unets
-   unet
+   unet | mlp
```
coreyjadams: perhaps mlp should be one layer below?

.importlinter
```diff
@@ -86,28 +99,31 @@ layers =
 [importlinter:contract:physicsnemo-models]
 name = Prevent Imports between physicsnemo models
 type = layers
+exhaustive = True
 containers=
    physicsnemo.models
 layers = 
-   mesh_reduced
+   mesh_reduced | flare | geotransolver
```
coreyjadams, line 106: @mnabian has an open PR from activation checkpointing that will bring these in too, let's compare and only merge one.

coreyjadams (approved):
Thanks, @peterdsharpe , for doing some maintenance here.  Good to clean this up periodically.  Left a few minor comments but overall its looking good.

.importlinter
peterdsharpe: Technically unet doesn't need mlp, so for now I'd set these apart - we can always do this stack later if we want them to share code!

PR #1961, opened 2026-09-02
Make config-driven field names reach nested TensorDict leaves in datapipes, mesh IO, and the aero recipe

peterdsharpe:
# PhysicsNeMo Pull Request

### The problem

A `TensorDict` is a tree, not a flat dict. `Mesh.point_data`, `cell_data`, and `global_data` can therefore hold nested groups of fields, e.g. `point_data = {"solution": {"gauge_pressure": ..., "wall_shear_stress": ...}, "sdf": ...}`, and our readers happily produce such meshes.

The catch is that TensorDict's dict-like surface hides the tree by default: `td.keys()` / `td.items()` / `td.values()` return only the top level (`include_nested=False`), yielding sub-TensorDicts as values, and a string key like `"solution/gauge_pressure"` is never parsed, it is just a literal top-level name. Code that treats a TensorDict as a plain dict is correct for flat data and wrong for nested data, usually without raising.

That pattern had spread across the datapipes and the unified external aero recipe. Every YAML-driven field name (`RenameMeshFields`, `NormalizeMeshFields`, `DropMeshFields`, `MeshToDomainMesh` targets, the TensorDict transforms, both collators, the recipe's `targets:`, nondim, loss, metrics, and output handling) was checked against `td.keys()` at the top level. The consequences, as a user reported, were that nested fields could not be addressed from YAML at all, and where they were present they were silently skipped: `RenameMeshFields` did nothing, `NormalizeMeshFields` left the field un-normalized, and `NormalizeMeshFields.inverse_td` (via `named_apply` without `nested_keys=True`) un-normalized every leaf whose *last name component* matched, including ones the forward pass never touched. `ConcatCollator` and `Reader(pin_memory=True)` crashed on nested samples, the PyVista round trip lost nesting, and GLOBE rejected any nested `global_data_ranks`.

The user's proposed workaround was a recipe-local flatten transform. That treats the symptom: the user should never have to think about flattening their mesh data.

### The fix, and why this shape

TensorDict already does the right thing when given a nested key: `key in td`, `td[key]`, `select`, `exclude`, `rename_key_`, `set`, and `named_apply(nested_keys=True)` all accept `("solution", "gauge_pressure")`. Nothing about the data needs to change. What is missing is a single, consistent way to turn a config string into that key, and the discipline of using the nested-aware TensorDict calls.

So this PR adds one small module, `physicsnemo.datapipes.keys`, and routes every field-name argument through it:

- `as_nested_key("solution.gauge_pressure")` returns `("solution", "gauge_pressure")`; a plain name stays a string; a list of strings is taken verbatim so a leaf whose own name contains a `"."` is still reachable. The `"."` separator is TensorDict's own `flatten_keys` default and the one our existing `Rename`, `Purge`, `RestructureTensorDict`, and GLOBE rank specs already used; a test pins it to the tensordict default.
- `get_leaf`, `rename_keys`, `with_leaf_name`, and `format_leaf_keys` cover the recurring lookup, rename, derived-output-name, and error-message needs, so each transform's fix is a one- or two-line change rather than a re-implementation. Missing fields are reported with the full leaf listing, and naming a group instead of a leaf gives an actionable `TypeError`.
- Inside the transforms, `key in td.keys()` becomes `key in td`, top-level iteration becomes `items(include_nested=True, leaves_only=True)`, and both `inverse_td` implementations use `named_apply(nested_keys=True)` so forward and inverse key on the same full path.

Two on-disk conventions are deliberately separate from the config convention. PyVista IO writes nested keys as `"/"`-joined VTK array names (mirroring HDF5 / zarr group paths; dots occur in real VTK names) and splits them back on read. The HDF5 / zarr readers keep `fields=["group/array"]` as a flat literal key, which is consistent with `"."` being the only parsed separator.

`physicsnemo/datapipes/datapipes.md` gains a short section stating the convention and the do / don't rules for iterating TensorDicts, so the pattern does not creep back in.

### Tests

New tests cover the helper module, every transform and collator on nested inputs (including the forward/inverse asymmetry), a nested PyVista round trip, and the recipe's nested-target flow end to end. `test/datapipes`, `test/mesh/io/io_pyvista`, `test/models/globe`, and the recipe tests pass locally (1548 passed). One unrelated local failure, `test_sdf_transform.py::test_sdf_normals_near_wall_use_face_normal`, is an SDF sign-tolerance check on code this PR does not touch.

### Reviewer notes

- `Reader.field_names` now reports nested leaves as tuple paths (previously top-level names only).
- `NormalizeMeshFields.stats` returns `"."`-joined string keys; the on-disk `.pt` format is unchanged for flat names.

## Checklist

## Dependencies

None.


physicsnemo/datapipes/multi_dataset.py
```diff
@@ -69,7 +70,9 @@ def _validate_strict_outputs(datasets: Sequence[DatasetBase]) -> list[str]:
         if len(ds) == 0:
             continue
         data, _ = ds[0]
-        keys = sorted(data.keys())
+        ### Compare leaf keys (nested included) so two datasets whose groups
+        ### hold different leaves are caught here, not mid-epoch in collate.
+        keys = format_leaf_keys(data)
```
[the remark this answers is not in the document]
peterdsharpe: Agreed, the rendered-name comparison was lossy. Now compares the actual leaf-key sets (`set(leaf_keys(data))`) and uses the dotted rendering only in the error message.

examples/cfd/external_aerodynamics/unified_external_aero_recipe/src/merge_global_data.py
```diff
@@ -119,7 +120,12 @@ def _load_sample(self, index: int) -> Mesh:
 
         ext_td = TensorDict.load_memmap(ext_path)
         merged = mesh.global_data.clone()
-        collisions = sorted(set(ext_td.keys()) & set(merged.keys()))
+        ### Compare leaves (nested included): two groups sharing a name but
+        ### holding different leaves are not a collision, and ``update``
+        ### merges them recursively.
+        collisions = sorted(
+            set(format_leaf_keys(ext_td)) & set(format_leaf_keys(merged))
```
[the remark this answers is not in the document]
peterdsharpe: Agreed. The collision check now intersects the actual leaf-key sets, so a literal `"a.b"` and a nested `("a", "b")` are distinct; `key_to_str` is used only to render the message.

physicsnemo/mesh/io/io_pyvista.py
```diff
@@ -61,7 +69,10 @@ def _vtk_data_to_tensor_dict(
             array = array[indices]
         elif force_copy:
             array = array.copy()
-        tensor_data[str(key)] = torch.as_tensor(array)
+        name = str(key)
+        parts = tuple(name.split(_VTK_KEY_SEPARATOR))
+        nested_key = parts if len(parts) > 1 and all(parts) else name
```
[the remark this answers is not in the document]
peterdsharpe: Leaving this as designed. `/` in a VTK array name is read the same way HDF5 and zarr read it: as a path into a group. An externally authored `a/b` array therefore loads as the nested key `("a", "b")`; no data is lost, and since every config-facing lookup goes through `as_nested_key`, a YAML name of `a.b` still reaches it. An escape mechanism would introduce a PhysicsNeMo-specific mini-format that no other VTK writer honors, for a naming pattern I have not seen in real datasets. The trade-off is documented on `_VTK_KEY_SEPARATOR`.

sylvesterkaczmarek (commented):
The conflict check explicitly permits a destination that is also being renamed, but the moves are then applied sequentially with `TensorDict.rename_key_(..., safe=False)`. For a mapping like `a -> b, b -> c`, the first move overwrites the original `b` before the second runs, so its value is silently lost (and swaps/cycles have the same problem). Please stage source values/use temporary keys, or reject dependent mappings, and add a chained-rename regression test.

physicsnemo/datapipes/keys.py
```diff
@@ -0,0 +1,300 @@
...
+    td: TensorDict,
+    mapping: Mapping[NestedKey, NestedKey],
+    *,
+    strict: bool,
+) -> TensorDict:
+    r"""Return a copy of ``td`` with keys renamed, nested keys included.
+
+    Both sides of ``mapping`` may be top-level or nested keys; a leaf can be
+    moved between groups (``("raw", "p") -> "pressure"``) and a whole group
+    can be renamed (``"raw" -> "solution"``). All moves are simultaneous, so
```
coreyjadams: I don't think this is strictly true that all moves are simultaneous, the are ordered, below, right?  I was able to have claude reduce it to this simple reproducer:

```

Order A: group first, then leaf. Raises.

import torch
from tensordict import TensorDict
from physicsnemo.datapipes.keys import rename_keys, leaf_keys

td = TensorDict(
    {"a": torch.zeros(2), "raw": {"p": torch.zeros(2), "q": torch.zeros(2)}},
    batch_size=[],
)

out = rename_keys(td, {"raw": "sol", ("raw", "p"): "pp"}, strict=True)

Output:

KeyError: 'No sub-tensordict with key raw.'

Order B: leaf first, then group. Succeeds.

import torch
from tensordict import TensorDict
from physicsnemo.datapipes.keys import rename_keys, leaf_keys

td = TensorDict(
    {"a": torch.zeros(2), "raw": {"p": torch.zeros(2), "q": torch.zeros(2)}},
    batch_size=[],
)

out = rename_keys(td, {("raw", "p"): "pp", "raw": "sol"}, strict=True)
print(sorted(map(str, leaf_keys(out))))
print(sorted(map(str, leaf_keys(td))))

Output:

['a', "('sol', 'q')", 'pp']
['a', "('raw', 'p')", "('raw', 'q')"]
```

CHANGELOG.md
```diff
@@ -18,6 +32,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- GLOBE no longer raises on a nested `global_data_ranks` declaration: the
+  `global_data.select(...)` pre-filter now uses nested keys instead of the
+  `"."`-joined names TensorDict does not parse.
+- `unified_external_aero_recipe`: `targets:`, `NonDimensionalizeByMetadata`
+  fields, `extract_targets`, loss / metric lookups, output normalization,
+  `attach_and_save`, and the global-data merge collision check all resolve
+  `"group.field"` names to nested leaves; `NonDimensionalizeByMetadata.inverse_td`
+  matches by full key rather than leaf name; TensorDict-valued forward kwargs
+  are batch-wrapped leaf by leaf in `input_type: tensors` mode.
+
```
coreyjadams: Nit: Can we simplify the changelog?

physicsnemo/datapipes/transforms/geometric.py
```diff
@@ -168,10 +172,16 @@ def __call__(self, data: TensorDict) -> TensorDict:
...
                     updates[self.closest_points_key] = closest_points
             else:
-                suffix = f"_{key}"
-                updates[f"{self.output_key}{suffix}"] = sdf.reshape(-1, 1)
+                ### Suffix the output leaf name with the input's leaf name,
+                ### keeping the output key's own nesting.
+                leaf = key if isinstance(key, str) else key[-1]
+                updates[with_leaf_name(self.output_key, lambda n: f"{n}_{leaf}")] = (
+                    sdf.reshape(-1, 1)
+                )
```
coreyjadams: Claude helped me find this one.  I thought it was BS at first, but I think with some of the ongoing dataset work this could actually matter.

Since this routes just the nested input leaf key name to the output name, something like:
```yaml
input_keys: [interior.coords, inlet.coords, slip.coords]
output_key: sdf
```
will all collide with `sdf_coords` names in the output.  I think `slip_coords` will actually win, but, it isn't good to silently discard data here.

physicsnemo/datapipes/keys.py
peterdsharpe: Yeah good point

peterdsharpe: https://github.com/NVIDIA/physicsnemo/pull/1961/commits/19b04d425ec704afc5f2321853cd70e2f6b5c747

coreyjadams: This comment got outdated so I couldn't update it in the code.  You can also have silent deletion.

With {("g","x"): "z", "g": "h", "other": "g"} the tensor moved onto g is deleted.

I don't know if it's still true after the updates, though!

coreyjadams (approved):
Approving to unblock.  Overall this looks good.  It cleans up many footguns, there are a couple subtle ones that I think are still here, and we should see if we can patch them, but they are harder to find than the previous way was for sure.  Even if we can't address it, it's better, but it'd be nice to address these!

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1961/commits/1f0f66be4e04802bc73163a4f91b63f009ea4228

physicsnemo/datapipes/transforms/geometric.py
peterdsharpe: fixed

PR #1966, opened 2026-09-08
Mesh.slice_points: avoid full-mesh index allocations for sparse cell blocks

peterdsharpe:
## Description

`Mesh.slice_points` no longer allocates point-count-sized index tensors when remapping a small cell block from a much larger mesh. It retains the lookup-table algorithm for ordinary meshes and switches to sorting the kept point IDs plus binary search when `n_points > 64 * cells.numel()` and fewer than `n_points / 64` points are kept. Meshes without cells skip the remap, and remapped connectivity is always int64 whatever the input dtype. This targets the reader's compaction of large meshes after cell sampling.

Index normalization allocates only the selected range/IDs. Empty forward slices remain empty; int32/int64 indices, negative indices, duplicates, unsorted selections, boolean masks and legacy uint8 masks retain their selection semantics. Invalid tensor dtypes, malformed masks, multidimensional selections and out-of-range indices are rejected. Negative-step slices remain unsupported, matching PyTorch indexing.

Points and nested point data use ordinary indexed gathers, including when loaded from a memory-mapped mesh. The change does not claim sequential range reads. Duplicate point IDs retain the previous last-position cell remapping behavior.

## Validation

- 96 focused tests passed across `test_slice_points_lookup_equivalence.py` and `test_slicing.py`, with both remapping algorithms forced in the equivalence/regression suite.
- Coverage includes memory-mapped meshes, nested fields, empty/clipped slices, mask/index dtype validation and positive/negative bounds errors.
- All changed-file pre-commit hooks passed, including import-linter.
- Local validation used Python 3.14, PyTorch 2.12 and TensorDict 0.14. Cluster (HiLiftAeroML) and synthetic benchmarks were rerun at f75eeead6; results and figures are in the comment thread.


peterdsharpe:
## Measurements: `main` vs this PR

Two implementations of `Mesh.slice_points` are compared throughout:

- **main: full-mesh lookup table.** `arange(n_points)` plus an `n_points`-long old-to-new map, then `old_to_new[cells]`; fancy-index gathers for points and point data.
- **this PR (f75eeead6): lookup table or binary search, chosen by mesh shape.** The same lookup table when `n_points ≤ 64 × cells.numel()` or when at least 1/64 of the points are kept; otherwise binary search of each cell vertex in the sorted kept ids. Points and point data are gathered with ordinary indexing.

Every measurement runs in a fresh process. Measured at the current head (f75eeead6), after the review changes. Figures, raw results and scripts are on [an assets branch of my fork](https://github.com/peterdsharpe/physicsnemo/tree/e5aabced37a65000d3e579616f0ae333ef2b51e2).

### 1. The real workload: HiLiftAeroML boundaries on a cluster (Lustre memmaps)

Each HiLiftAeroML boundary mesh has ~285M cells and ~142M vertices, stored as memory-mapped files on Lustre. The recipe's reader takes a contiguous block of cells per training sample and then calls `slice_points` to drop the vertices the block does not reference (a 10k-cell block keeps ~30k of the 142M vertices, so these meshes take the binary-search branch). Twelve cold cases per implementation per block size, 24 distinct cases in total, run in two rounds with the package order swapped and the two implementations alternating so file-system drift cancels. Median with min–max in parentheses.

| block cells | metric | main: lookup table | this PR: table or search, by shape | speed-up |
|---|---|---|---|---|
| 10,000 | cold sample (load + block + compaction) | 2.22 s (1.94 s–8.36 s) | 2.60 s (1.94 s–62.97 s) | 0.9x |
| 10,000 | second block, same case (partly cached) | 935 ms (680 ms–5.67 s) | 976 ms (621 ms–25.14 s) | 1.0x |
| 10,000 | slice_points alone, same block (point pages just read) | 48 ms (38 ms–6.03 s) | 2 ms (2 ms–2.01 s) | 27.9x |
| 10,000 | recipe reader `__getitem__` | 469 ms (314 ms–1.34 s) | 341 ms (266 ms–996 ms) | 1.4x |
| 10,000 | peak RSS increase during the sample | 3,081 (2,826–3,186) MiB | 963 (702–1,044) MiB | 3.2x |
| 100,000 | cold sample (load + block + compaction) | 3.25 s (2.20 s–8.78 s) | 2.92 s (2.24 s–6.43 s) | 1.1x |
| 100,000 | second block, same case (partly cached) | 95 ms (75 ms–6.82 s) | 46 ms (36 ms–10.60 s) | 2.1x |
| 100,000 | slice_points alone, same block (point pages just read) | 52 ms (44 ms–2.31 s) | 10 ms (9 ms–9.62 s) | 5.1x |
| 100,000 | recipe reader `__getitem__` | 147 ms (133 ms–2.41 s) | 119 ms (94 ms–5.49 s) | 1.2x |
| 100,000 | peak RSS increase during the sample | 3,622 (3,510–3,942) MiB | 1,476 (1,314–1,584) MiB | 2.5x |

HiLiftAeroML on the cluster

What this says:

- **Memory is the headline.** The per-sample peak RSS increase drops from ~3.1 GB to ~0.96 GB at 10k cells and from ~3.6 GB to ~1.5 GB at 100k cells. The ~2.1 GB saved is the two `n_points`-long int64 tensors (2 × 142M × 8 B) that every dataloader worker used to allocate per sample.
- **`slice_points` itself** goes from 48 ms to 2 ms on a 10k-cell block (28x at the median) and from 52 ms to 10 ms at 100k cells (5x).
- **Cold wall time is unchanged within noise (medians 2.2–2.6 s / 2.9–3.3 s, fastest cases identical at 1.94 s).** On Lustre the cold sample is dominated by reading 30k–300k scattered vertex rows out of a 1.7 GB point file, which no in-process algorithm changes. The outliers in both implementations (63 s, 25 s, 9 s, 8 s) are storage stalls on a loaded file system that hit whichever one was running; they also account for the min–max spread in the `slice_points` rows, whose point gather touches the same memory-mapped pages.
- The recipe reader's end-to-end `__getitem__` on a partly cached case improves modestly (469 → 341 ms, 147 → 119 ms).

### 2. Synthetic: the reader's path as a function of mesh size

`slice_cells(block)` then `slice_points(unique(cells))` on an N-vertex triangle mesh with 2N random cells, one scalar and one vector point field, saved once to a node-local memmap ("memory" rows clone everything into RAM first). The block is small; what matters is how compaction scales with the size of the mesh it came from. Median of 3 runs.

| N vertices | block cells | kept vertices | backing | main: lookup table | this PR: table or search, by shape | speed-up |
|---|---|---|---|---|---|---|
| 1,000,000 | 10,000 | 29,539 | memmap | 3 ms | 3 ms | 1.1x |
| 1,000,000 | 10,000 | 29,539 | memory | 2 ms | 2 ms | 1.0x |
| 1,000,000 | 100,000 | 259,232 | memmap | 3 ms | 3 ms | 1.0x |
| 1,000,000 | 100,000 | 259,232 | memory | 2 ms | 2 ms | 0.9x |
| 10,000,000 | 10,000 | 29,955 | memmap | 8 ms | 5 ms | 1.6x |
| 10,000,000 | 10,000 | 29,955 | memory | 4 ms | 2 ms | 1.8x |
| 10,000,000 | 100,000 | 295,678 | memmap | 8 ms | 6 ms | 1.3x |
| 10,000,000 | 100,000 | 295,678 | memory | 7 ms | 5 ms | 1.3x |
| 100,000,000 | 10,000 | 29,997 | memmap | 62 ms | 13 ms | 4.8x |
| 100,000,000 | 10,000 | 29,997 | memory | 49 ms | 2 ms | 20.8x |
| 100,000,000 | 100,000 | 299,571 | memmap | 65 ms | 17 ms | 3.8x |
| 100,000,000 | 100,000 | 299,571 | memory | 53 ms | 8 ms | 6.5x |

Synthetic reader path

- With the lookup table, the cost of compacting a fixed-size block grows linearly with the size of the source mesh (2 ms → 49 ms from 1M to 100M vertices for a 10k-cell block in memory). With this PR it is flat at ~2 ms: the work depends on the block, not on the mesh.
- The 1M-vertex rows are on the lookup-table side of the threshold (33 points per cell-vertex entry) and match `main`. The memmap rows at 100M vertices keep ~13 ms because the kept rows are scattered through the file and have to be read individually.

### 3. Synthetic: `slice_points` on the full mesh (keeping k random vertices)

Same meshes, now slicing the whole mesh to k random vertices, which is the usual use of `slice_points` (mesh transforms, cropping). Cell connectivity is 6N entries, so every row here takes the lookup-table branch of this PR. Median of 3 runs.

| N vertices | k kept | backing | main: lookup table | this PR: table or search, by shape | speed-up |
|---|---|---|---|---|---|
| 1,000,000 | 1,000 | memmap | 6 ms | 6 ms | 1.1x |
| 1,000,000 | 1,000 | memory | 15 ms | 5 ms | 3.0x |
| 1,000,000 | 30,000 | memmap | 6 ms | 6 ms | 1.0x |
| 1,000,000 | 30,000 | memory | 15 ms | 6 ms | 2.6x |
| 10,000,000 | 1,000 | memmap | 78 ms | 69 ms | 1.1x |
| 10,000,000 | 1,000 | memory | 76 ms | 66 ms | 1.1x |
| 10,000,000 | 30,000 | memmap | 80 ms | 73 ms | 1.1x |
| 10,000,000 | 30,000 | memory | 73 ms | 77 ms | 0.9x |
| 10,000,000 | 1,000,000 | memmap | 83 ms | 82 ms | 1.0x |
| 10,000,000 | 1,000,000 | memory | 78 ms | 81 ms | 1.0x |
| 100,000,000 | 1,000 | memmap | 1.01 s | 929 ms | 1.1x |
| 100,000,000 | 1,000 | memory | 961 ms | 923 ms | 1.0x |
| 100,000,000 | 30,000 | memmap | 995 ms | 947 ms | 1.1x |
| 100,000,000 | 30,000 | memory | 992 ms | 944 ms | 1.1x |
| 100,000,000 | 1,000,000 | memmap | 1.02 s | 992 ms | 1.0x |
| 100,000,000 | 1,000,000 | memory | 988 ms | 995 ms | 1.0x |

Synthetic full mesh

This PR matches `main` on full-mesh slices (0.9x–1.1x at 10M and 100M vertices; the 1M-vertex rows are all under 20 ms and within run-to-run noise): the common use of `slice_points` is unaffected, and the gains above come only where the mesh is far larger than the cells being kept.

<details>
<summary>Method and raw data</summary>

- **HiLift:** [`bench_hilift_slice.py`](https://github.com/peterdsharpe/physicsnemo/blob/e5aabced37a65000d3e579616f0ae333ef2b51e2/scripts/bench_hilift_slice.py) on one GB300 node of an internal Slurm cluster (CPU work only), with `PYTHONPATH` switched between a snapshot of `main` and this branch; both use the unchanged upstream reader (`_subsample_mesh_cells` → `slice_cells` → `slice_points`). Two dozen cases taken at a fixed stride through the dataset (12 per round, the second round on a disjoint dozen with the package order swapped), block seed 0 for the cold sample, seed 1 for the second block. Peak RSS from `ru_maxrss`. Raw: [`hilift_aga.jsonl`](https://github.com/peterdsharpe/physicsnemo/blob/e5aabced37a65000d3e579616f0ae333ef2b51e2/data/hilift_aga.jsonl).
- **Synthetic:** [`bench_slice_points.py`](https://github.com/peterdsharpe/physicsnemo/blob/e5aabced37a65000d3e579616f0ae333ef2b51e2/scripts/bench_slice_points.py) and [`bench_slice_points_block.py`](https://github.com/peterdsharpe/physicsnemo/blob/e5aabced37a65000d3e579616f0ae333ef2b51e2/scripts/bench_slice_points_block.py), one subprocess per (implementation, case), on the same node. The lookup-table algorithm from `main` is reproduced verbatim inside the script; the other column runs the branch's `Mesh.slice_points`. Raw: [`synthetic_full_mesh.json`](https://github.com/peterdsharpe/physicsnemo/blob/e5aabced37a65000d3e579616f0ae333ef2b51e2/data/synthetic_full_mesh.json), [`synthetic_reader_path.json`](https://github.com/peterdsharpe/physicsnemo/blob/e5aabced37a65000d3e579616f0ae333ef2b51e2/data/synthetic_reader_path.json).
- Tables and figures: [`analyze_bench1966.py`](https://github.com/peterdsharpe/physicsnemo/blob/e5aabced37a65000d3e579616f0ae333ef2b51e2/scripts/analyze_bench1966.py).

</details>

physicsnemo/mesh/mesh.py
```diff
@@ -1386,39 +1392,89 @@ def slice_points(
...
+        ###    where the table's allocation and fill dominated everything.
+        ### Measured crossover on synthetic meshes: the search wins from about
+        ### n_points ~ 300 x cells.numel(); the table is faster below ~ 30 x.
+        n_kept = kept_indices.numel()
+        cells = self.cells
+        if n_kept == 0:
+            valid_cells_mask = torch.zeros(
+                cells.shape[0], dtype=torch.bool, device=device
+            )
+            new_cells = cells[valid_cells_mask]
```
negin513: Small dtype leak here: this path indexes the original `cells`, so if the mesh was built with int32 connectivity the empty result comes back int32, while the other two branches (and `main`) always hand back int64. Ends up being `slice_points(slice(100, 200))` → int32 but `slice_points([0])` → int64 on the same mesh.

Probably just `cells.new_empty((0, cells.shape[1]), dtype=torch.long)`. Might be worth one equivalence case with `cells.to(torch.int32)` too — the fixtures all go through `torch.randint` so they're int64 and wouldn't have caught this.

physicsnemo/mesh/mesh.py
```diff
@@ -1386,39 +1392,89 @@ def slice_points(
...
+        elif n_points <= _SEARCH_REMAP_RATIO * cells.numel():
```
negin513: I think the gate needs to look at `n_kept` as well. The expensive part of the search branch is the sort, which scales with how many points you keep, not with `cells.numel()`. So a mesh with lots of points and few cells (point-cloud-ish, or literally zero cells) where you keep most of the points will land in the search branch and sort ~everything, which is quite a bit slower than just building the table.

Not a problem for the reader, since it keeps a tiny fraction — but `slice_points` is public. Since `n_kept` is already computed a few lines up, something like `and n_kept * _SEARCH_REMAP_RATIO < n_points` would cover it.

physicsnemo/mesh/mesh.py
```diff
@@ -1386,39 +1392,89 @@ def slice_points(
         if indices is None or indices is ...:
             return self
 
-        ### Normalize indices to a 1D tensor of point indices to keep
-        all_indices = torch.arange(self.n_points, device=self.points.device)
+        ### Normalize indices to a 1D tensor of point indices to keep. Nothing
+        ### here is sized by n_points: a boolean mask becomes its nonzero
```
negin513: Nit on the comment: this doesn't quite hold for the bool/uint8 case below, where the mask has to be `n_points` long and we `nonzero()` the whole thing. The tensor is the caller's, but the scan is ours. Maybe just say this for slices and integer indices, so nobody reads it as "bool masks are free on huge memmap meshes".

negin513 (changes requested):
Reviewed the two remap paths and the index normalization carefully — the sort+searchsorted branch is correct (stable sort + `right=True` reproduces last-occurrence semantics; the clamp + equality check filters dropped points safely), the dense path is unchanged in complexity, and the tests genuinely force both algorithms. Nice.

One real regression in the empty-selection branch, one gap in the algorithm gate, and one comment that overstates the cost model — inline below.

Two broader notes. (1) The 28x / RSS numbers in the description were measured at 1e605e0, before the row-gather helper was removed; worth a rerun on head before they're cited. (2) Since #1894, the zarr reader path (`_zarr_mesh_subsampled` in `datapipes/readers/mesh.py`) compacts via `torch.unique(return_inverse=True)` and never calls `slice_points`, so this optimization only serves the `.pmsh`/`.pdmsh` memmap reader and external callers. Still worthwhile, but it argues for keeping the gate simple rather than tuning it further.

physicsnemo/mesh/mesh.py
peterdsharpe: I think that we *want* to preserve the original dtype, right? Let me look into this more, just to check

physicsnemo/mesh/mesh.py
peterdsharpe: Fair, will fix

physicsnemo/mesh/mesh.py
peterdsharpe: Ok - I do think that long-term we want to preserve dtype, but since this behavior differs outside this file, I'll set that aside for a separate PR. Going with the `.new_empty()` fix here

physicsnemo/mesh/mesh.py
peterdsharpe: Good idea, will find a setting that works here

peterdsharpe:
^ fixes pushed up, benchmarks re-running

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1966/commits/f75eeead65d8c6222f2b5f8b900fb900ca5d6201

negin513 (approved):
LGTM!

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1966/commits/9b90b80723ea680100605baf573433bb46b51528

PR #1968, opened 2026-09-08
Configure the distributed process-group timeout in code or environment

peterdsharpe:
# PhysicsNeMo Pull Request

## Description

Add a configurable timeout for the default distributed process group:

```python
DistributedManager.initialize(timeout=6000)
```

The argument accepts numeric seconds (including fractional seconds) or `datetime.timedelta`. Explicit values override `PHYSICSNEMO_DIST_TIMEOUT_S`; with no argument, the environment variable supplies seconds. Unset or empty configuration preserves PyTorch's backend default.

Resolve and validate the selected timeout before initialization-state changes or launcher autodetection. Invalid types, nonfinite/nonpositive values, overflow, and durations that round to zero fail clearly, and corrected configuration can be retried. Explicit settings override even a malformed environment value.

The effective timeout reaches ENV, SLURM and OPENMPI initialization, direct setup, and both current and compatibility `init_process_group` calls. Use the same setting on all ranks. Separately created process groups retain their own timeout policies.

## Validation

- 120 new CPU-capable tests passed for precedence, fractional seconds/timedelta, all launcher routes, compatibility forwarding, direct setup, invalid configuration and corrected retries.
- Existing manager suite: 14 passed, 22 skipped.
- Three new `multigpu_dynamic` tests initialize two real CUDA/NCCL ranks, verify the configured argument/environment/precedence choice, perform an all-reduce and clean up. These were skipped locally because this machine has one GPU; the existing `ci:multi-gpu` label enables both CI streams on the updated trusted mirror.
- All changed-file pre-commit checks passed. The implementation received an independent code review.

## Checklist

- [x] Contribution guidelines followed; new commits carry sign-offs.
- [x] Regression tests, API documentation, and CHANGELOG updated.
- [x] No model implementation changes.
- [x] Fresh multi-GPU CI results pending.

### DIFF
--- a/physicsnemo/distributed/manager.py
+++ b/physicsnemo/distributed/manager.py
@@ -18,6 +18,7 @@ import atexit
 import os
 import queue
 import warnings
+from datetime import timedelta
 from typing import Optional, Tuple
 from warnings import warn
 
@@ -387,6 +388,10 @@ class DistributedManager(object):
         listed above. Initialization method can also be explicitly controlled using the
         `PHYSICSNEMO_DISTRIBUTED_INITIALIZATION_METHOD` environment variable and setting it
         to one of the options above.
+
+        The process-group timeout can be overridden by setting the
+        `PHYSICSNEMO_DIST_TIMEOUT_S` environment variable to a number of seconds.
+        When unset, the backend default (10 minutes for NCCL) is used.
         """
         if DistributedManager.is_initialized():
             warn("Distributed manager is already initialized")
@@ -590,6 +595,14 @@ class DistributedManager(object):
             )
 
         if manager._distributed:
+            # Optional process-group timeout override (seconds). The backend
+            # default (10 minutes for NCCL) is too long when a rank dies
+            # between collectives, since its peers hang for the full window,
+            # and too short when a long step or a stalled file system delays
+            # one rank. Unset leaves the backend default in place.
+            timeout_env = os.environ.get("PHYSICSNEMO_DIST_TIMEOUT_S")
+            timeout = timedelta(seconds=float(timeout_env)) if timeout_env else None
+
             # Setup distributed process group
             try:
                 dist.init_process_group(
@@ -597,6 +610,7 @@ class DistributedManager(object):
                     rank=manager.rank,
                     world_size=manager.world_size,
                     device_id=manager.device,
+                    timeout=timeout,
                 )
             except TypeError:
                 # device_id only introduced in PyTorch 2.3
@@ -604,6 +618,7 @@ class DistributedManager(object):
                     backend,
                     rank=manager.rank,
                     world_size=manager.world_size,
+                    timeout=timeout,
                 )
 
         if torch.cuda.is_available():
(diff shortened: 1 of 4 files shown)


coreyjadams:
I think, if you'd like to add this, it'd be good to support a programmatic interface too.  `initialize` could take a `timeout=6000`.  And we definitely need multigpu-ci on this PR :)

peterdsharpe:
Will do! Adding multigpu tests, and will set up something like `DistributedManager.initialize(timeout=3600)` or so

coreyjadams (approved):

physicsnemo/distributed/manager.py
```diff
@@ -387,11 +429,37 @@ def initialize():
...
+        TypeError
+            If an explicit timeout is not numeric seconds or a ``timedelta``.
+        ValueError
+            If the selected timeout is invalid, nonpositive, or unrepresentable.
+            Validation precedes initialization state changes, so callers can
+            correct the configuration and retry.
         """
         if DistributedManager.is_initialized():
             warn("Distributed manager is already initialized")
             return
```
negin513, line 457: The new `Raises` section above promises `TypeError`/`ValueError` for a bad `timeout`, but this early return runs before `_resolve_timeout()`, so on an already-initialized manager any `timeout` value is silently dropped with only the generic "already initialized" warning. That applies to invalid values (`timeout=-1` won't raise) and to valid-but-different ones (`initialize(timeout=6000)` after a default init leaves the original timeout in effect with no signal that the request was ignored).

The idempotent early return is pre-existing and fine to keep. Two cheap options:
- Move `_resolve_timeout(timeout)` above the `is_initialized()` check so the documented validation contract holds unconditionally (it's pure and has no side effects), and/or
- Extend the warning when `timeout is not None` to say the value was ignored because the manager is already initialized, and add a one-line caveat to the docstring.

physicsnemo/distributed/manager.py
```diff
@@ -597,13 +673,15 @@ def setup(
                     rank=manager.rank,
                     world_size=manager.world_size,
                     device_id=manager.device,
+                    timeout=timeout,
```
negin513, line 676: Heads-up, not a defect in this PR: open PR #1938 ("Fix DistributedManager initialization on CPU-only hosts") rewrites this same `try/except TypeError` block, because unconditionally passing `device_id=manager.device` raises `ValueError: init_process_group device_id parameter must be an accelerator with an index` on CPU-only hosts with recent PyTorch. Both PRs touch these lines plus `CHANGELOG.md` and `test/distributed/test_manager.py`, so whichever lands second will need a rebase. Worth coordinating so the `timeout=` forwarding survives into both `init_process_group` call sites after #1938's restructuring.

negin513 (commented):
Looked at this end to end — the plumbing is solid: `timeout` reaches both `init_process_group` call sites, `_resolve_timeout` runs before any shared-state mutation in `initialize()` and `setup()`, explicit arg correctly shadows `PHYSICSNEMO_DIST_TIMEOUT_S`, and nan/inf/overflow/bool/sub-microsecond inputs are all rejected with the retry path covered by tests.

Two inline notes: one small docstring/behavior mismatch on the already-initialized early return, and a heads-up about overlap with open #1938. Neither is blocking.

negin513 (approved):
LGTM! But a minor nit.

PR #1969, opened 2026-09-08
Mesh: FieldLayout for packing named scalar and vector fields

peterdsharpe:
## Description

Add `FieldLayout(rank_spec, spatial_dim)` to pack named scalar and Cartesian-vector fields from a point `TensorDict` into separate channel tensors and unpack them into their named hierarchy. Channel order is sorted by flattened name, independent of dictionary insertion order. Scalars have shape `(N, C_s)` and vectors `(N, C_v, D)`.

Packing validates only the selected tuple paths. Unselected tensors or metadata cannot interfere with validation, and literal separators remain distinct from nested paths when the configured flattened names do not collide. Selected leaves must have the declared scalar/vector shape and share dtype/device. `ScalarVectorFields` represents the packed pair; layout properties expose channel names, counts and rank schemas.

The new `validate_rank_spec` validates schema keys and non-negative integer ranks. `validate_data_contains_ranks` now rejects malformed schemas up front; valid existing GLOBE schemas retain their behavior. New public symbols are exported from `physicsnemo.mesh`.

## Validation

- 18 tests passed across `test/mesh/test_fields.py` and `test/models/globe/test_validation.py`.
- Regression coverage includes ignored metadata, unrelated literal-dot keys, custom separators with reversed input insertion order, and selected non-tensor leaves.
- All changed-file pre-commit hooks passed, including import-linter.
- Local validation used Python 3.14, PyTorch 2.12 and TensorDict 0.14.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   in `physicsnemo.datapipes` through it, so a `"."` in a YAML field name
   (`"solution.pressure"`) addresses a leaf inside a nested `TensorDict`.
   Nested `Mesh` data no longer needs to be flattened before use.
+- Adds `physicsnemo.mesh.FieldLayout`, which packs the named scalar and vector
+  fields of a point `TensorDict` into two dense channel tensors in a fixed,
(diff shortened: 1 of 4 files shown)


peterdsharpe:
No longer needed, retired upstream

PR #1746, opened 2026-06-23
Relative L2 loss

mnabian:
# PhysicsNeMo Pull Request

## Description

Adds relative MES and relative L2 to `physicsnemo.metrics.general`, and gives mse/rmse an optional weights argument for a masked/weighted mean. Both relative metrics accept element weights (subsuming a validity mask and per-point weights) and a dim reduction, matching general.mse.

## Checklist

- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.
- [ ] If I am implementing a new model or modifying any existing model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation).

## Review Process

**All PRs are reviewed by the PhysicsNeMo team before merging.**

Depending on which files are changed, GitHub may automatically assign a maintainer for review.

We are also testing AI-based code review tools (e.g., Greptile), which may add automated comments with a confidence score.
This score reflects the AI’s assessment of merge readiness and is **not** a qualitative judgment of your work, nor is
it an indication that the PR will be accepted / rejected.

AI-generated feedback should be reviewed critically for usefulness.
You are not required to respond to every AI comment, but they are intended to help both authors and reviewers.
Please react to Greptile comments with 👍 or 👎 to provide feedback on their accuracy.

PR #1974, opened 2026-09-08
Unified aero recipe: rename the misnamed "rmse" loss to relative_mse (deprecated alias kept)

peterdsharpe:
## Summary

Fixes a misleading loss name in the unified external aero recipe's `src/loss.py`.

**What is wrong today.** `training.loss_type: rmse` does not compute a root-mean-square error. For each field it computes

```
mean((pred - target)^2) / mean(target^2)
```

(per component and summed, for vector fields) with no square root anywhere. That is the target-normalized *relative MSE*. Anyone comparing a training curve labelled "rmse" against reported RMSE numbers is comparing different quantities.

**What this does.**

- Adds `relative_mse` as the `loss_type` name for the quantity the old name computed. It delegates to `physicsnemo.metrics.general.relative_error.relative_mse` (from #1746) instead of a hand-rolled ratio.
- Keeps `rmse` as a deprecated alias: `LossCalculator` rewrites it to `relative_mse` at construction and emits a `FutureWarning` naming the replacement. (FutureWarning rather than DeprecationWarning because it is a user-facing config value and DeprecationWarning is hidden by default outside `__main__`.)
- Updates the README's loss-type list and adds a CHANGELOG entry under Deprecated.
- No shipped config uses `rmse`; `base.yaml` defaults to `huber`, which is untouched.
- No new loss types are added. Relative L2 remains available as a reported metric only; it is not offered as a training loss (its gradient is undefined at exactly zero error).

**Numerical change for existing configs.** `huber` and `mse` are untouched. For `rmse` -> `relative_mse` the ratio is identical for any target with nonzero energy (`mean/mean == sum/sum`). The only difference is the `1e-8` guard on the denominator: previously added to `mean(target^2)`, now a floor on `sum(target^2)` inside the shared implementation. For a field whose target is identically zero, the new value is larger by the element count; both values are meaningless in that case (a finite error divided by `1e-8`), and the warning text states the difference.

## Tests

`tests/test_loss_equivalence.py` (16 passed):

- `relative_mse` per-field values match the explicit target-normalized formula for a scalar field and a per-component-summed vector field.
- `rmse` warns at construction, reports `loss_type == "relative_mse"`, and produces bit-identical totals and per-field values to `relative_mse`.
- Unknown loss types are rejected with the accepted names in the message; per-field key tests cover `relative_mse`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,9 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Unified external aerodynamics recipe: `training.loss_type: rmse` is
   deprecated in favour of `relative_mse`, which names what it always
-  computed (target-normalized relative MSE, no square root); `relative_l2`
-  is added for the square root. Both delegate to
-  `physicsnemo.metrics.general.relative_error`. `rmse` still works and
+  computed (target-normalized relative MSE, no square root) and delegates
+  to `physicsnemo.metrics.general.relative_error`. `rmse` still works and
   warns.
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Deprecated
 
+- Unified external aerodynamics recipe: `training.loss_type: rmse` is
+  deprecated in favour of `relative_mse`, which names what it always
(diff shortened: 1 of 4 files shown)


coreyjadams (approved):
OK, this is just name changing, I don't mind :)

peterdsharpe:
/ok to test https://github.com/NVIDIA/physicsnemo/pull/1974/commits/dc7103ebe9459d82a6db76de355af256963734d9

PR #1957, opened 2026-09-01
Update to cuda13

coreyjadams:
# PhysicsNeMo Pull Request

v2 of #1950, opened fresh after a rebase, copy of description:

This PR updates our CI from cuda12 to cuda13.

~It also fixes a bug in how we check yaml files for license headers. So it looks a lot bigger than it is, because I also fixed all those license headers.~ This was split out and merged separately.

Additionally, the nightly container and pre-commit workflows are deleted:

nightly container was not used.
pre-commit is superceded by the app version of the same tool, which is faster and more efficient.

## Checklist

- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.
- [ ] If I am implementing a new model or modifying any existing model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation).

## Review Process

**All PRs are reviewed by the PhysicsNeMo team before merging.**

Depending on which files are changed, GitHub may automatically assign a maintainer for review.

We are also testing AI-based code review tools (e.g., Greptile), which may add automated comments with a confidence score.
This score reflects the AI’s assessment of merge readiness and is **not** a qualitative judgment of your work, nor is
it an indication that the PR will be accepted / rejected.

AI-generated feedback should be reviewed critically for usefulness.
You are not required to respond to every AI comment, but they are intended to help both authors and reviewers.
Please react to Greptile comments with 👍 or 👎 to provide feedback on their accuracy.

PR #1975, opened 2026-09-08
Build transformer-engine-torch against the runtime torch (fixes nightly TE ImportError)

coreyjadams:
## Summary

Main's nightly has failed every night since 2026-09-05 with

```
ImportError: transformer_engine_torch.cpython-312-x86_64-linux-gnu.so:
  undefined symbol: _ZN3c104cuda19CUDAErrorLogCaptureC1Ev
```

`transformer-engine-torch` is built from its sdist in an isolated uv build environment. With no `extra-build-dependencies` entry it pulls whatever torch PyPI serves today (2.14), so the extension is compiled against a newer c10 ABI than the locked runtime torch (2.11.0+cu128). `c10::cuda::CUDAErrorLogCapture` exists in torch 2.14 and not in 2.11.

This was latent until the 09-05 nightly, when the cu12 uv download cache was evicted (repo cache went over the 10 GB cap) and TE was rebuilt cold. Every nightly since has hit the cache and reused the poisoned wheel.

## Change

One line in `pyproject.toml`: declare `torch` as an extra build dependency for `transformer-engine-torch` with `match-runtime = true`, the same pattern the PyG packages already use. This pins the build env's torch to the runtime version. No `uv.lock` change.

The same line is already on #1957 for the cu13 stack, where the branch nightly verified it end to end (frozen sync, lockfile guard, TE tests passing).

## Follow-up after merge

The cached cu12 uv slot still holds the bad wheel and must be deleted once so the next nightly rebuilds TE cleanly:

```
gh cache delete "uv-cache-nightly-cuda12.8.1-cudnn-devel-ubuntu24.04-py3.12-uv0.11.7-fullextras-latest" --repo NVIDIA/physicsnemo
```

Then dispatch the nightly. Note that last night's failed run published an empty `testmon-nightly-latest`, so PRs on main are currently running the full suite until a good nightly republishes the DB.

Until the slot is deleted, this PR's own GPU jobs will hit the poisoned cache and fail on the same ImportError.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,6 +84,7 @@ conflicts = [
 torch-sparse = ["torch"]
 torch-cluster = ["torch"]
 torch-scatter = ["torch"]
+transformer-engine-torch = [{ requirement = "torch", match-runtime = true }]
 
 [[tool.uv.index]]
 name = "nvidia"


pyproject.toml
ktangsali, line 83: Does it make sense to add the `match-runtime` to the other dependencies in this list as well?

ktangsali (approved):

pyproject.toml
coreyjadams: yes probably.  It's not biting us because we're more deliberate with those packages in the build ... but it might be nice to do this for the downstream folks.

coreyjadams:
[Deleting this message that claude slop-posted WITHOUT PERMISSION.]

pyproject.toml
coreyjadams: Actually, no, the `torch_[cluster|scatter|sparse]` packages don't seem to support this.  Nevermind!

coreyjadams:
/ok to test [2ca26f2](https://github.com/NVIDIA/physicsnemo/pull/1975/commits/2ca26f2b2de82cca3cfb15e18059bfeb9776d0b7)

PR #1978, opened 2026-09-09
Fix mixed Q/K/V dtypes in NATTEN attention when qk_norm is used under autocast

linneawolniewicz:
# PhysicsNeMo Pull Request

## Description
`Natten2DSelfAttention` and `RopeNatten2DSelfAttention` fail with `qk_norm=True`
under `torch.autocast` in a reduced-precision dtype. NATTEN rejects the call:
```
ValueError: Query, key, and value must assume the same data type, got
query.dtype=torch.float32, key.dtype=torch.float32, value.dtype=torch.bfloat16
```
`layer_norm` is on PyTorch's list of [CUDA ops that autocast to `float32`](https://docs.pytorch.org/docs/stable/amp.html#cuda-ops-that-can-autocast-to-float32), so `self.q_norm(q)` and `self.k_norm(k)` return `float32` while `v` stays in the autocast dtype. Which combinations fail depends on the norm backend and on whether the block is compiled; see the reproduction note below.

Nothing between the norm and the attention operator reconciles the dtypes:
- `physicsnemo.nn.functional.na2d` is a `__torch_function__` dispatch shim that
  forwards straight to `natten.functional.na2d`.
- NATTEN validates dtype equality as the first statement of `na2d`
  (`na_tensor_checks`, whose `raise_error` defaults to `True`) and repeats the
  check in every backend, so there is no permissive fallback.
- NATTEN's AMP integration is `functools.partial(custom_fwd, device_type="cuda")`
  with no `cast_inputs`, which propagates autocast state into the backward pass
  but performs no casting.
This casts the normalized queries and keys back to the dtype of the QKV
projection — what the values already carry and what the attention operator
requires. Where the dtypes already agree, the cast is a no-op.

Note on reproduction: with `layernorm_backend="torch"` this fails in both eager
and compiled execution. With `layernorm_backend="apex"` it fails only under
`torch.compile`, because apex's `FusedLayerNorm` is a custom `autograd.Function`
that autocast's op policy does not see, so eager returns the autocast dtype
while Dynamo lowers it to aten `layer_norm` and the float32 policy applies.
Verified on H100, torch 2.13, natten 0.21.6:

| case | eager | compiled |
| --- | --- | --- |
| torch LayerNorm output | float32 | float32 |
| apex FusedLayerNorm output | bfloat16 | float32 |
| `na2d(fp32 Q/K, bf16 V)` | ValueError | ValueError |
| attention block, with cast | bfloat16 | bfloat16 |
| attention block, without cast | ValueError | ValueError |

The cast fixes the failure, makes the two norm backends agree with each
other, and makes eager and compiled execution agree for apex.

### Scope
The timm-based `TimmSelfAttention` path is unaffected and left unchanged. Its
qk-norm modules also return float32 under autocast (both `LayerNorm` and timm's
`RmsNorm`, eager and compiled), but `scaled_dot_product_attention` accepts
float32 Q/K against a bfloat16 V and returns the value dtype rather than
raising, so the block runs correctly. Only NATTEN requires uniform dtypes.

## Testing
- Added a CPU unit test that verifies normalized Q and K are cast back to the
  QKV projection dtype before NATTEN.
- Added dependency-gated CUDA integration coverage using real NATTEN 0.21.6,
  bf16 autocast, and both PyTorch and Apex LayerNorm backends.
- Added compiled coverage with `torch.compile`, including the Apex-specific
  regression path.
- Verified on an NVIDIA H100:
  - With the fix: all eager and compiled variants pass.
  - Without the fix: PyTorch eager/compiled and Apex compiled fail due to mixed
    Q/K/V dtypes; Apex eager passes, reproducing the reported compile-specific
    behavior.
- `pre-commit` passes for all changed files.

## Checklist

## Dependencies
N/A


linneawolniewicz:
@pzharrington

test/nn/module/test_dit_layers.py
```diff
@@ -399,6 +401,49 @@ def test_mask_token_arithmetic_matches_where():
...
+        rope_cos, rope_sin = build_axial_rope_cos_sin_2d(h, w, head_dim)
+        kwargs |= {"rope_cos": rope_cos, "rope_sin": rope_sin}
+
+    observed = {}
+
+    def fake_na2d(q, k, v, **_kwargs):
+        observed["dtypes"] = (q.dtype, k.dtype, v.dtype)
+        return v
+
+    monkeypatch.setattr(dit_layers, "_na2d_func", fake_na2d)
```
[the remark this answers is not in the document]
linneawolniewicz: Addressed, I've added dependency-gated CUDA integration tests that exercise real PyTorch/Apex LayerNorm, bf16 autocast, the real NATTEN kernel, and torch.compile.

CHANGELOG.md
```diff
@@ -34,6 +34,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `Natten2DSelfAttention` and `RopeNatten2DSelfAttention` with `qk_norm=True` no
+  longer fail under `torch.autocast`. `LayerNorm` is an autocast-to-float32
+  operation, so the normalized queries and keys were returned as float32 while
+  the values stayed in the autocast dtype, and NATTEN requires queries, keys,
+  and values to share a dtype. The normalized queries and keys are now cast
+  back to the dtype of the QKV projection.
```
pzharrington: Can shorten this down to something like
```
- Fixed an issue in `Natten2DSelfAttention` and `RopeNatten2DSelfAttention` with `qk_norm=True` mixing `LayerNorm` fp32 Q/K outputs with autocasted V dtypes
```

pzharrington (approved):

CHANGELOG.md
linneawolniewicz: Shortened!

loliverhennigh (approved):

coreyjadams:
/ok to test [409dd82](https://github.com/NVIDIA/physicsnemo/pull/1978/commits/409dd8264b17395c461e4dc12a4e71422d99d7a3)

coreyjadams:
/ok to test [581fece](https://github.com/NVIDIA/physicsnemo/pull/1978/commits/581fece31722ff53bd13c76a97c887f2bacc1a02)

PR #1979, opened 2026-09-10
Standalone datapipe and ShardTensor fixes ahead of domain-parallel reading

coreyjadams:
# PhysicsNeMo Pull Request

This PR brings standalone fixes and reorganizations that are necessary to start work for the domain-parallel data loading and unified recipe domain parallel support.

Datapipes:
- `MeshReader` / `DomainMeshReader` gain `close()` (fixes `MeshDataset.close()`
  raising `AttributeError`); the extra-boundary loader is split out.
- Zarr readers: extract-method refactor shared with the tensorstore reader,
  a shared `Reader._window_indices`, and a deterministic subsampling window
  key (configured list order instead of set iteration order).
- `record_consumer_stream` unwraps `ShardTensor` / `DTensor` leaves.

ShardTensor:
- `scatter_tensor` returns a shard that owns its storage (dynamo could not
  fakeify the nonzero-offset ranks' views).
- Redistribute planner key is immutable and the spec hash is invalidated on
  mutation (planner cache no longer grows per step).
- Shard-shape gather is device agnostic; three small `torch.compile` fixes
  (`_conversion_scope`, plain-tensor promotion under `no_grad`, `__new__`).

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/physicsnemo/blob/main/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [x] The [CHANGELOG.md](https://github.com/NVIDIA/physicsnemo/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/physicsnemo/issues) is linked to this pull request.
- [ ] If I am implementing a new model or modifying any existing model, I have followed the [Models Implementation Coding Standards](https://github.com/NVIDIA/physicsnemo/wiki/Coding-standards-for-models-implementation).

## Review Process

**All PRs are reviewed by the PhysicsNeMo team before merging.**

Depending on which files are changed, GitHub may automatically assign a maintainer for review.

We are also testing AI-based code review tools (e.g., Greptile), which may add automated comments with a confidence score.
This score reflects the AI’s assessment of merge readiness and is **not** a qualitative judgment of your work, nor is
it an indication that the PR will be accepted / rejected.

AI-generated feedback should be reviewed critically for usefulness.
You are not required to respond to every AI comment, but they are intended to help both authors and reviewers.
[...]

PR #1988, opened 2026-09-14
CI: stop testmon from displacing coverage in the PR Coverage job

coreyjadams:
## Problem

The PR `Coverage` job runs `coverage run ... -m pytest --testmon ...`. pytest-testmon tracks test dependencies with its **own** coverage.py tracer, installed per test, which displaces the outer `coverage run` tracer. Result: nothing executed *inside* a test is recorded. The merged report is just the restored nightly baseline (recorded against `main`'s line numbers) plus import-time lines from the PR.

Any changed file whose lines shifted therefore appears to lose coverage in proportion to the shift. On #1979, for example, `physicsnemo/datapipes/readers/mesh.py:365` (`if self.include_index_in_metadata:`) was reported missing in the same job where 28 mesh-reader tests passed through it, and Codecov showed -471 hits for a +19 line diff.

Reproduced locally on `test/datapipes/readers/test_mesh_readers.py` (28 tests):

| invocation | mesh.py coverage |
|---|---|
| `coverage run -m pytest` | 60% |
| `coverage run -m pytest --testmon` | 15% (import-time lines only) |
| `coverage run -m pytest --testmon --testmon-nocollect` | 60% |

testmon has an automatic "coverage.py detected, disabling collection" guard, but it did not trigger in this configuration (the CI header only printed `testmon: changed files: 35, ...`).

## Fix

Pass `--testmon-nocollect` in the Coverage job so testmon only *selects* tests from the restored nightly DB and does not install its own tracer. The Coverage job restores the DB but never publishes it (the `Testmon` job and the nightly maintain it), so skipping collection here loses nothing.

## Verification

The next PR Coverage run after this lands should show the codecov patch coverage of changed files reflecting the tests that actually ran, rather than dropping whenever lines shift.


coreyjadams:
/ok to test [d121a9b](https://github.com/NVIDIA/physicsnemo/pull/1988/commits/d121a9ba80bd1486053b48ebba12d6596407e786)

PR #1995, opened 2026-09-15
Unified External Aero Recipe: scale geometry once per NonDimensionalizeByMetadata chain

melo-gonzo:
**Summary**

`NonDimensionalizeByMetadata` divides the mesh coordinates by `L_ref` on each call. A dataset
chain with two instances, one for interior `point_data` fields and one for boundary `cell_data`
fields (inlet velocity, wall shear) read by a boundary model such as GLOBE, scales the geometry
by `L_ref` twice. With `L_ref = 0.4 m` the coordinates were 2.5x too large in body units while
the fields were correct.

This adds `scale_geometry: bool = True`. The default is unchanged. A second instance sets
`scale_geometry: false` and scales fields only. `inverse()` uses the same flag, so a chain of
two instances inverts to the original geometry.

A related bug in inference: `build_redim_field_types` kept only the last
`NonDimensionalizeByMetadata` block's field map, so with two instances the interior predictions
would not have been re-dimensionalized. The maps are now merged.

**Changes**
- `src/nondim.py`: `scale_geometry` argument, applied in forward and inverse; docstring and
  YAML example.
- `src/infer.py`: merge the field maps of all `NonDimensionalizeByMetadata` blocks.
- `tests/test_nondim.py` (new): default scales once; second instance does not rescale geometry;
  inverse of both instances restores the input.
- `tests/test_infer.py`: two-instance field-map merge.
- `README.md`, `CHANGELOG.md`.

**Testing**
`tests/test_nondim.py`, `tests/test_infer.py`, `tests/test_nested_targets.py`,
`tests/test_synthetic_configs.py`: 70 passed. Pre-commit hooks pass. No dataset YAML in the
repo uses two instances, so existing recipes are unchanged.


CHANGELOG.md
```diff
@@ -44,6 +44,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Unified external aero recipe: `NonDimensionalizeByMetadata` gains
```
peterdsharpe: I would probably categorize this as "new functionality" rather than "fixed"? "Fixed" has the unintended side effect of implying "this may change behaviors in existing workflows", when that is not the case

peterdsharpe (approved):
Looks good, one docs comment

CHANGELOG.md
melo-gonzo: Fair point, moved changelog comment to `added` section (c5577046).

melo-gonzo:
/ok to test [c557704](https://github.com/NVIDIA/physicsnemo/pull/1995/commits/c55770460b5c626c8613115b9b11b9d0059d2009)

melo-gonzo:
/ok to test [ebf0967](https://github.com/NVIDIA/physicsnemo/pull/1995/commits/ebf0967eeac8a50a855ce065f3f2d1925b79d0a6)

PR #1999, opened 2026-09-17
Fix AeroJEPA unbatched mask handling and validate chunked-decode arguments

### LINKED ISSUE #1998: 🐛[BUG]: AeroJEPA ignores unbatched TokenSet masks and decode_field_chunked accepts invalid arguments
### Version

2.3.0a0 (`main` at 3bbbe86a)

Source

### Describe the issue

Two groups of input-handling problems in the experimental AeroJEPA model (`physicsnemo.experimental.models.aerojepa`). None of them raise an error; they give wrong results or cryptic failures.

**1. The `mask` of an unbatched `TokenSet` is ignored.** `QueryTokenDecoder.forward` drops masked-out target tokens, but three other components treat every token of an unbatched `TokenSet` as valid:

- `PrototypeTokenJEPAHead` builds an all-`True` context mask for unbatched input, so masked-out context tokens still take part in cross-attention. The batched path uses `context_tokens.mask` correctly.
- `pad_token_sets` sets `mask[i, :count] = True` and computes the fallback global token with `masked_mean(features, None)`. Tokens that the input set marks invalid become valid after packing, and they also enter the global token.
- `flatten_valid_token_features` returns rank-2 features unchanged even when a mask is passed.

Expected: masked-out tokens are excluded in every case, as they already are in the batched path.

**2. `AeroJEPA.decode_field_chunked` does not check its arguments.**

- An unknown `precision` (for example `"bfloat16"` or `"amp"`) falls back to fp32 without any warning. The docstring says "Anything other than `"fp32"` enables `torch.autocast`", but the code only enables autocast for `"fp16"` and `"bf16"`.
- `chunk_size <= 0` fails inside the decoder with `torch.cat(): expected a non-empty list of Tensors`.
- `query_sdf` is required, even though `decode_field` and the decoder accept `None` when `use_sdf=False`. Omitting it raises `TypeError: 'NoneType' object is not subscriptable`.
- The same goes for `QueryTokenDecoder(query_chunk_size=0)`: the constructor accepts it, and `forward` then fails with `range() arg 3 must not be zero`.

Expected: invalid arguments raise a clear `ValueError`, and `query_sdf` is optional when the decoder does not use it.

```shell
import torch
from physicsnemo.experimental.models.aerojepa import (
    PrototypeTokenJEPAHead, QueryTokenDecoder, TokenSet,
)
from physicsnemo.experimental.models.aerojepa.layers import (
    flatten_valid_token_features, pad_token_sets,
)

head = PrototypeTokenJEPAHead(
    token_dim=16, cond_dim=0, depth=1, num_heads=2, neighbor_k=4, query_pe_bands=2
).eval()
feats, coords, targets = torch.randn(10, 16), torch.randn(10, 3), torch.randn(5, 3)
mask = torch.arange(10) < 6

with torch.no_grad():
    trimmed = head(context_tokens=TokenSet(feats[:6], coords[:6]), target_positions=targets)
    masked = head(context_tokens=TokenSet(feats, coords, mask=mask), target_positions=targets)
print(torch.allclose(trimmed, masked, atol=1e-5))                        # False

packed = pad_token_sets([TokenSet(feats, coords, mask=mask)])
print(int(packed.mask.sum()))                                             # 10, expected 6
print(flatten_valid_token_features(feats, mask).shape[0])                 # 10, expected 6

QueryTokenDecoder(token_dim=16, use_sdf=False, query_chunk_size=0)        # accepted
```

For `decode_field_chunked`, run `test_decode_field_chunked_*` from the linked PR against `main`.

### Relevant log output

```shell
precision='bfloat16' : accepted, autocast enabled=False
chunk_size=0         : ValueError: torch.cat(): expected a non-empty list of Tensors
query_sdf=None       : TypeError: 'NoneType' object is not subscriptable
query_chunk_size=0   : ValueError: range() arg 3 must not be zero
```

### Environment details

```shell
Bare-metal, source install (uv, editable), Python 3.12.3, torch 2.14.0+cpu, Linux aarch64
```

harshaa765:
# PhysicsNeMo Pull Request

## Description

Closes #1998.

This PR fixes two groups of input-handling bugs in the experimental AeroJEPA model. The PR has one commit per group.

**1. Honor unbatched `TokenSet` masks.** `QueryTokenDecoder` already drops masked-out tokens. These three helpers now do the same:

- `PrototypeTokenJEPAHead` uses `context_tokens.mask` for unbatched context. Before this fix it built an all-`True` mask, so masked-out tokens were still used in cross-attention.
- `pad_token_sets` copies each set's mask into the packed mask and uses it for the fallback global token. It also checks the mask shape.
- `flatten_valid_token_features` applies a rank-2 mask. Before this fix it ignored the mask. The docstring of the example's `TokenLatentSIGReg` is updated to match.

**2. Validate `decode_field_chunked` arguments.**

- An unknown `precision` now raises `ValueError`. Before this fix it silently fell back to fp32. Matching is case-insensitive, the same as the recipe's `get_autocast_context`. The docstring now describes what the code does.
- A non-positive `chunk_size` now raises `ValueError` up front. Before this fix it failed later with an unrelated `torch.cat` error.
- `query_sdf` is optional, as it already is in `decode_field` and the decoder. The `AeroJEPATrunk.decode_queries` annotation and docstring are updated to match.
- `QueryTokenDecoder` rejects a non-positive `query_chunk_size` in its constructor. Before this fix it failed on the first `forward` call.

Only invalid inputs change behavior: valid calls give the same results as before, and the SuperWing recipe's configs (`precision: bf16`) keep working.

### Tests

- New tests in `test/experimental/models/aerojepa/`:
  - Masked unbatched context gives the same prediction as the trimmed context.
  - `pad_token_sets` keeps per-set masks and averages only valid features into the global token.
  - `flatten_valid_token_features` applies rank-2 masks and rejects masks with the wrong shape.
  - `decode_field_chunked` rejects invalid arguments, matches `decode_field` in fp32 (for both `"fp32"` and `"FP32"`), and works without `query_sdf`.
  - `QueryTokenDecoder` rejects a non-positive `query_chunk_size`.
- 13 of the new tests fail on `main` and pass with this change.
- `pytest test/experimental/models/aerojepa examples/cfd/external_aerodynamics/aerojepa/tests`: 169 passed, 36 skipped. The skipped tests need CUDA. Each commit passes the suite on its own.
- `ruff check`/`ruff format` (v0.12.5, `--force-exclude`), the docstring-coverage hook, and the doctests in the touched modules all pass.
- I ran these on CPU only (aarch64, torch 2.14). The autocast tests on CUDA were not run locally.

## Checklist

## Dependencies

None.


harshaa765:
@greptileai please re-review. a0fef0da addresses both findings: the predictor now skips the cross-neighbor search when the context is fully masked (with a regression test), and the touched docstrings follow MOD-003d/e/h.

harshaa765:
@mnabian can you please review.

PR #2011, opened 2026-09-19
Aero recipe: prevent near-wall SDF sign noise from reversing normals

peterdsharpe:
## Description

Near-wall volume queries can receive a tiny negative signed distance from float32 roundoff. `ComputeSDFFromBoundary` previously used that sign to reverse the fallback face normal, turning an outward normal inward for points on the wall.

Use separate coordinate-scaled tolerances for the SDF sign and closest-point direction. The direction fallback retains its broad `128 * eps * coordinate_scale` band; only negative distances within the smaller `8 * eps * coordinate_scale` sign-noise band default to the outward face normal. Interior points beyond the sign-noise band retain inward normals, including when the direction fallback is active. Stored SDF values and closest-point directions outside the fallback band are preserved.

For example, translating a box by 10,000 units gives a direction-fallback tolerance of about 0.153. Points 0.01 and 0.1 units inside still receive inward normals; a point on the wall receives the outward face normal. This addresses the interior-normal regression identified in review.

## Validation

- 21 SDF-transform cases passed. Coverage includes real near-wall geometry, controlled sign noise, and real interior/on-wall/exterior queries at original coordinates and after translation along one or all three axes. Near-wall and translation cases run on CPU and CUDA; the translated-geometry regression exercises both winding-number and pseudo-normal SDF signs.
- Eight translated-interior cases failed on the previous PR head and passed after separating the tolerances. The original near-wall correction remains covered, along with degenerate faces, unchanged SDF values, and cache preservation.
- 19 synthetic recipe configuration tests passed.
- Ruff checks and all applicable changed-file pre-commit hooks passed.

## Checklist

- [x] I am familiar with the contributing guidelines.
- [x] The changelog is up to date with these changes.
- [ ] An issue is linked to this pull request.

## Dependencies

None. Based directly on main.


peterdsharpe:
@greptileai

